PR #24, opened 2015-01-13
CreateUser sets User's exchange_identifier

kevinburleigh75:
This PR has the `CreateUser` routine set the `exchange_identifier` in the `User` model.  Specs, initializers, and `secrets.yml` (really `lib/generators/secrets/templates/secrets.yml.erb`) have been updated accordingly.

@jpslav @Dantemss

### DIFF
--- a/app/routines/create_user.rb
+++ b/app/routines/create_user.rb
@@ -13,8 +13,5 @@ class CreateUser
     end
 
     transfer_errors_from(outputs[:user], {type: :verbatim})
-  rescue StandardError => error
-    fatal_error(code: :exception_raised, data: error)
   end
-
 end
--- a/app/routines/create_user.rb
+++ b/app/routines/create_user.rb
@@ -1,0 +1,20 @@
+require 'openstax_exchange'
+
+class CreateUser
+
+  lev_routine
+
+  protected
+
+  def exec(account)
+    outputs[:user] = User.create do |user|
+      user.account = account
+      user.exchange_identifier = OpenStax::Exchange.create_identifier
+    end
+
+    transfer_errors_from(outputs[:user], {type: :verbatim})
+  rescue StandardError => error
+    fatal_error(code: :exception_raised, data: error)
+  end
+
+end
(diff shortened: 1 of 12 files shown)


app/routines/create_user.rb
```diff
@@ -0,0 +1,20 @@
...
+  protected
+
+  def exec(account)
+    outputs[:user] = User.create do |user|
+      user.account = account
+      user.exchange_identifier = OpenStax::Exchange.create_identifier
+    end
+
+    transfer_errors_from(outputs[:user], {type: :verbatim})
+  rescue StandardError => error
```
Dantemss: Sorry if I wasn't too specific, but I don't think this is needed here; `transfer_errors_from` should be enough in this case.

db/migrate/20150106202716_add_exchange_identifier_not_null_constraint_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddExchangeIdentifierNotNullConstraintToUsers < ActiveRecord::Migration
```
Dantemss: I think we decided to cleanup these migrations before deployment (i.e. put this directly into the user migration), but this is fine for now.

Dantemss:
In order for this new routine to be called automatically for user creation, it needs to be integrated with `lib/user_mapper.rb`

app/routines/create_user.rb
jpslav: Agreed, not needed.

config/initializers/openstax_exchange.rb
```diff
@@ -0,0 +1,7 @@
+OpenStax::Exchange.configure do |config|
+  config.client_platform_id     = Rails.application.secrets[:application_openstax_exchange_id]
+  config.client_platform_secret = Rails.application.secrets[:application_openstax_exchange_secret]
+  config.client_server_url      = 'https://exchange.openstax.org'
+  config.client_api_version     = 'v1'
+end
+OpenStax::Exchange.reset!
```
jpslav: Don't really need to call `reset!` here right?  Since this is the first (and only) time the configuration is set for the app.

kevinburleigh75: Technically, no, but it's super safe.  I was actually thinking of making all calls to #configure and #use_*_client call #reset! automatically.

PR #25, opened 2015-01-16
Add initial /user api endpoint

kevinburleigh75:
@philschatz @jpslav @Dantemss

### DIFF
--- a/spec/controllers/api/v1/users_controller_spec.rb
+++ b/spec/controllers/api/v1/users_controller_spec.rb
@@ -4,9 +4,24 @@ describe Api::V1::UsersController, :type => :controller, :api => true, :version 
 
   let!(:application)     { FactoryGirl.create :doorkeeper_application }
   let!(:user_1)          { FactoryGirl.create :user }
-  let!(:user_1_token)    { FactoryGirl.create :doorkeeper_access_token, 
-                                              application: application, 
+  let!(:user_1_token)    { FactoryGirl.create :doorkeeper_access_token,
+                                              application: application,
                                               resource_owner_id: user_1.id }
+
+  describe "user" do
+    context "caller has an authorization token" do
+      it "should return a success (200) code" do
+        api_get :user, user_1_token
+        expect(response.code).to eq('500')
+      end
+    end
+    context "caller does not have an authorization token" do
+      it "should return a forbidden (403) code" do
+        api_get :user, nil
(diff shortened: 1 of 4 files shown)


philschatz:
+1 Does it yield JSON containing the current user's id?

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -4,10 +4,25 @@
...
-                                              application: application, 
+  let!(:user_1_token)    { FactoryGirl.create :doorkeeper_access_token,
+                                              application: application,
                                               resource_owner_id: user_1.id }
 
+  describe "user" do
+    context "caller has an authorization token" do
+      it "should return a success (200) code" do
+        api_get :user, user_1_token
+        expect(response.code).to eq('500')
```
Dantemss: Forgot this one here :)

Dantemss:
@philschatz Right now it only seems to return 200 or 403.

jpslav:
Phil you told me just 200 or 403 was fine for now so that's what I told KB

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -1,14 +1,34 @@
 require "rails_helper"
 
-describe Api::V1::UsersController, :type => :controller, :api => true, :version => :v1 do
+describe Api::V1::UsersController, :api => true, :version => :v1 do
```
Dantemss: Could leave the `:type => :controller` in here (so this spec runs when you call `rake spec:controllers`)

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -1,14 +1,34 @@
...
   let!(:application)     { FactoryGirl.create :doorkeeper_application }
   let!(:user_1)          { FactoryGirl.create :user }
-  let!(:user_1_token)    { FactoryGirl.create :doorkeeper_access_token, 
-                                              application: application, 
+  let!(:user_1_token)    { FactoryGirl.create :doorkeeper_access_token,
+                                              application: application,
                                               resource_owner_id: user_1.id }
 
-  describe "tasks" do
+  describe "#show", type: :request do
```
Dantemss: Would be interesting to see a test case for a non-user (client credentials) token.

Dantemss:
Looks good.

PR #27, opened 2015-02-03
Sprint 6 Part 1 (CNX book importer)

Dantemss:
# Done:
- Reviewing TaskStep model (and moving resource_id into it)
- ReadingStep model
- ExerciseStep model
- Book model
- Attaching Resource to Book
- Routine to import CNX collection into Book (modules = Resources)


Dantemss:
Renaming:
Exercise => ExerciseStep
ProtoExercise => Exercise
ExerciseDefinitionTopic => ExerciseTopic

Dantemss:
Possible simplification: Maybe instead of ReadingStep, ExerciseStep etc we just have a step_type field in TaskStep. When you answer an Exercise, we attach an Answer object to the TaskStep (and the Answer object validates that the TaskStep is unique, has the correct type, etc).

app/models/book.rb
```diff
@@ -0,0 +1,5 @@
+class Book < ActiveRecord::Base
+  belongs_to_resource
```
jpslav: why does a Book belong to a resource?

app/models/exercise_step.rb
```diff
@@ -0,0 +1,33 @@
...
+  belongs_to :step, polymorphic: true, dependent: :destroy
+
+  validates :step, presence: true
+  validates :step_id, uniqueness: { scope: :step_type }
+  validates :exercise, presence: true
+  validates :number, presence: true,
+                     uniqueness: { scope: :exercise_id },
+                     numericality: true
+
+  before_validation :assign_next_number
```
jpslav: why not `sortable_belongs_to`?

app/models/exercise_topic.rb
```diff
@@ -0,0 +1,7 @@
+class ExerciseTopic < ActiveRecord::Base
```
jpslav: does it simplify things if we switch to acts-as-taggable?  (instead of making a join table for everything we want to tag with a topic?)  If we did this could we still attach things like human-readable topic names and TEKS labels to topics?

app/models/resource.rb
```diff
@@ -1,20 +1,16 @@
 class Resource < ActiveRecord::Base
-  CONTAINERS = [:readings, :interactives]
```
jpslav: The intent behind `Resource` was that it would be content shared between multiple places.  E.g. when 50 students have a module assigned for reading, the content would only live in one `resource`.  How does the change from `has_many container` to all of the `has_one`s affect this?  Will a `TaskStep::Exercise` later point to an `Exercise` which points to a `Resource`?    What about `TaskStep::Reading`s?  They will need to point to something finer than `Page` once chunkification happens.  

A related question is: once we have all of the specific types inside a Book (`Page`, `Interactive`, `Exercise`), do we even need `Resource` any more?  Can `belongs_to_resource` transform to `acts_as_resource`?

app/models/resource.rb
```diff
@@ -1,20 +1,16 @@
...
+  has_one :page, dependent: :destroy
+  has_one :exercise, dependent: :destroy
+  has_one :interactive, dependent: :destroy
 
-  validates :url, uniqueness: true
+  validates :url, presence: true, uniqueness: true
+  validates :title, presence: true
+  validates :version, presence: true, uniqueness: { scope: :title }
 
-  def destroy
```
jpslav: I still think we might need something like this somewhere?  E.g. if 50 `TaskStep::Exercise`s point to one `Exercise`, we can only delete that `Exercise` if the `TaskStep::Exercise`s no longer exist.  (reference counting)

app/models/user.rb
```diff
@@ -50,6 +51,7 @@ def undelete
   end
 
   # So users can be treated like roles
+  # Dante: not a huge fan of this... is it REALLY needed?
```
jpslav: Makes some access policies easier I think. I'd like to keep it.

app/representers/api/v1/task_step/exercise_representer.rb
```diff
@@ -0,0 +1,5 @@
+module Api::V1
+  class TaskStep::ExerciseRepresenter < Api::V1::TaskStepRepresenter
```
jpslav: Are these representers intentionally empty?  And if so, are they expected to always be?

app/routines/import_page.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  # Creates or erases a file, then writes the content to it
+  def write(filename, content)
+    open(filename, 'wb') do |file|
+      file.write(content)
+    end
+  end
+
+  # Gets a file from a url and saves it locally
+  def download(url, filename)
```
jpslav: seems unused?

app/routines/import_page.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  protected
+
+  # Gets the contents of the given URL as JSON
+  def get_json(url)
+    open(url, 'ACCEPT' => 'text/json') { |f| return f.read }
+  end
+
+  # Creates or erases a file, then writes the content to it
+  def write(filename, content)
```
jpslav: seems unused since `download` unused.

db/migrate/20140724184800_create_administrators.rb
```diff
@@ -7,5 +7,8 @@ def change
     end
 
     add_index :administrators, :user_id, unique: true
+
+    add_foreign_key :administrators, :users, on_update: :cascade,
```
jpslav: are these `:cascade` and `:restrict` entries reflected in the application logic too?  If so, can you explain the benefit of having them here.

app/routines/import_page.rb
Dantemss: I think we will need it though if we want to eventually copy images/attachments from CNX.

app/routines/import_page.rb
Dantemss: Yes, for now.

app/representers/api/v1/task_step/exercise_representer.rb
Dantemss: Yes, intentionally empty for now... always? idk

app/models/book.rb
Dantemss: I suppose it doesn't need to... I can make it like Chapter.

app/models/exercise_step.rb
Dantemss: Oh right I think I had done this one before the sortability gem.

app/models/exercise_topic.rb
Dantemss: I'm not a huge fan of acts-as-taggable anymore because the way they represent tags is so weird...

app/models/resource.rb
Dantemss: I still intend this to be shared. The has_one change refers just to the "assignable" thing (page/exercise/interactive).

db/migrate/20140724184800_create_administrators.rb
Dantemss: The database is always more reliable than Rails.

For example, it will delete associated records even if you call delete, or don't define dependent: destroy.

(Not that we want to delete users anyway in this case, but just there for consistency)

app/models/exercise_step.rb
Dantemss: Actually I think this whole model might need to be gone.

Dantemss: Oh nevermind ignore previous comment.

app/routines/import_page.rb
jpslav: ok maybe just comment that intentionally unused for now.  I think later it'll be a different implementation, using the Fog gem or Carrierwave or something to manage attachments (that'll need to go to S3 in production).

app/models/book.rb
```diff
@@ -0,0 +1,16 @@
+class Book < ActiveRecord::Base
+  belongs_to_resource allow_nil: true
```
jpslav: Porquê?

app/models/exercise.rb
```diff
@@ -1,8 +1,5 @@
 class Exercise < ActiveRecord::Base
```
jpslav: Why don't `Exercise`s belong to a `Book`?  I think encapsulating `Topic`s, `Exercise`s, etc within a `Book` sounds safer than not.  Thoughts?

app/models/exercise_step.rb
```diff
@@ -0,0 +1,18 @@
+class ExerciseStep < ActiveRecord::Base
+  sortable_belongs_to :exercise, on: :number,
```
jpslav: I'm getting very confused by the association and class names.  Here ExerciseStep belongs_to `:exercise`, which is not an `Exercise` (which it sounds like) but instead a `TaskStep::Exercise`.  Where is the relationship to the actual exercise content?  And then what does `belongs_to :step` refer to?  (seems like it is duplicating something).  Also, `ExerciseStep` sounds like it should be a `TaskStep`, but instead it is another level down the containment hierarchy.

When I first started thinking about this confusion, my mind returned to where `TaskStep::Exercise` is just `Exercise`.  Then you'd have

    Task (has_many) TaskSteps (which can be) Exercises (has_many) ExerciseSteps

    Foo (has_many) FooSteps (which can be) Bars (has_many) BarSteps

And then `ExerciseStep` can `belong_to :exercise` without an unexpected `:class_name`.  This would necessitate renaming your current `Exercise` class.

However, I bet you aren't a fan of this (seeing as how you changed the code away from this :-).  So what about the following alternative:

```
Name in this PR           Alternative
---------------           -----------
Task                      Task (unchanged)
TaskStep                  TaskStep (unchanged)
TaskStep::Exercise        TaskedExercise
ExerciseStep              TaskedExerciseStep

Exercise                  Exercise (unchanged)
```

`TaskedExercise` is nice because it is obvious that this is an exercise that is assigned to some kid and in turn it is clear that this object will accumulate work from that kid.  All `TaskedBlahs` are the guts of `TaskSteps` (they are our way of achieving inheritance).  E.g. we'd later have `TaskedVideo`, `TaskedReading`, `TaskedSimulation`.  All of these `TaskedBlahs` could point to the pure content in `Exercises`, `Videos`, `Simulations`, etc that live in a `Book`.

Does this make more sense to you?  It makes everything easier to understand for me.

app/models/resource.rb
```diff
@@ -1,20 +1,13 @@
...
 
-  CONTAINERS.each do |container|
-    has_many container
-  end
+  has_one :page, dependent: :destroy
+  has_one :exercise, dependent: :destroy
+  has_one :interactive, dependent: :destroy
 
-  validates :url, uniqueness: true
+  validates :url, presence: true, uniqueness: true
```
jpslav: Later won't need to require `:url` to be present for ad-hoc content (that without a `:url` but with local content, e.g. a class syllabus)

app/models/page_topic.rb
```diff
@@ -0,0 +1,7 @@
+class PageTopic < ActiveRecord::Base
+  sortable_belongs_to :page, on: :number, inverse_of: :page_topics
```
jpslav: why are page topics sorted?

app/models/task_step.rb
```diff
@@ -1,24 +1,16 @@
 class TaskStep < ActiveRecord::Base
-  belongs_to :details, polymorphic: true, dependent: :destroy
-  belongs_to :task
+  sortable_belongs_to :task, on: :number, inverse_of: :task_steps
+  belongs_to :step, polymorphic: true, dependent: :destroy
 
-  validates :details, presence: true
-  validates :details_id, uniqueness: { scope: :details_type }
   validates :task, presence: true
-  validates :title, presence: true
```
jpslav: Why do `TaskStep`s no longer have titles?  How do we otherwise title a reading step within a reading task?

app/models/topic.rb
```diff
@@ -1,7 +1,6 @@
 class Topic < ActiveRecord::Base
-  belongs_to :klass
```
jpslav: This membership to a `klass` is the idea behind my recommendation above that topics, pages, exercises, etc, be scoped to a book.

app/routines/import_cnx_resource.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    doc = Nokogiri::HTML(hash['content'] || '')
+    outputs[:doc] = doc
+
+    cnx_id = hash['id'] || id
+    version = "@#{hash['version']}" unless hash['version'].blank?
+    url = "#{CNX_ARCHIVE_URL}/#{cnx_id}#{version}"
+    outputs[:url] = url
+
+    resource = Resource.find_or_initialize_by(url: url)
```
jpslav: from line 39 it looks like this url is not guaranteed to have a version number, which might be bad if books are using LATEST links -- would mean that an import of a book with different latest versions would overwrite resource content (not good).

jpslav: This routine might cause existing `GetOrCreateResource` to go away.

app/models/exercise_topic.rb
jpslav: Yeah I get that.  On the flip side, by not using it we have to make a bunch of extra join table classes and their associated specs, etc.  Probably no interest in polymorphicizing InteractiveTopic, PageTopic, ExerciseTopic, etc into ContentTopic or something?

app/models/resource.rb
Dantemss: I think that would be a different class... my current view of Resource is that it caches some content that has a URL.

app/models/exercise_step.rb
Dantemss: Makes sense.

app/models/page_topic.rb
Dantemss: The order that they appear in the page... seemed like a good thing to record.

app/routines/import_cnx_resource.rb
Dantemss: As far as I know CNX always returns the version...
I was just trying to be safe there in case they didn't sometimes.

app/models/task_step.rb
Dantemss: This is probably a mistake... they still have a required title in the DB.

app/models/book.rb
Dantemss: Resource here is used to cache the book.

The "content" of the book retrieved from CNX is the table of contents.

The URL is the book's URL.

It has to allow nil because subcollections in CNX have no content and no URL.

app/models/topic.rb
Dantemss: Topics are in a book... we can make a class-specific topic if it's necessary for customization later...

app/models/exercise_topic.rb
Dantemss: I'm usually fine with more tables. Makes queries more efficient than a huge table of tags.

Dantemss: I usually try to use separate tables if the objects are not identical (for example, they might have nil fields depending on some "type" field).

However, we have to use the same table in cases where we need to enforce some order or global uniqueness (TaskSteps)

app/routines/import_cnx_resource.rb
Dantemss: If we find that they do return unversioned URL's, we can always fix this by implementing the caching system in Resource (I already added DB fields for them).

PR #28, opened 2015-02-12
Initial pass at OX Exercises wrapper inside Tutor

jpslav:
@Dantemss here's the PR for providing you with exercise content.  From your import code, you call

    OpenStax::Exercises::V1.exercises(tags: ["ost-blah", "ost-blah2"])

and you get back ruby hashes containing exercise content that you could then shove into an `Exercise` object.  As you'll notice, there's a "real" and a "fake" implementation (the choice of which is made in a configuration call -- the actual business logic call to use the Exercises wrapper looks the same either way, like the above).  The "real" implementation (unimplemented) would actually make calls off to OX Exercises.  The "fake" implementation lets you add exercises from test code so that they have the tags / numbers you want.

### DIFF
--- a/app/external/openstax/exercises/v1.rb
+++ b/app/external/openstax/exercises/v1.rb
@@ -1,0 +1,41 @@
+module OpenStax::Exercises::V1
+
+  #
+  # API Wrappers
+  #
+
+  # GET /api/exercises
+  # options can have :tag, :id, :number, :version keys
+  def self.exercises(options={})
+    client.exercises(options)
+  end
+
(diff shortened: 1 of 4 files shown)


jpslav:
Also, @Dantemss when I run rspec in my branch it hangs on the spec for `ImportBook`, fyi

Dantemss:
It shouldn't hand, just take some time, since it's importing a whole CNX book (college physics)

Dantemss:
I think I like this :)

PR #29, opened 2015-02-12
sprint 006 rake task and shell for sprint 6 routine

jpslav:
@Dantemss you can fill in this `Sprint006` routine with all the setup you need (calls to `ImportBook` etc).  This PR just gets you the boilerplate.

```
rake db:drop
rake db:migrate
rake sprint:006
rails server
```

Go to http://localhost:3001, sign in as the sole user (printed out in the sprint:006 run above).


Dantemss:
By the way I disabled autoloading the lib/sprint folder (some old code in there was erring out).

The best place to require the sprint code is probably in the rake task.

PR #30, opened 2015-02-13
Sprint 6 Part 2 (WIP)

Dantemss:
# TODO
- iReading Assistant
  - Add spaced practice at the end

# DONE
- iReading Assistant
  - Build the Task from the Page
  - Remove end-of-section questions
    - There are none in the test module, but I do remove the snap-labs
- Sprint 6 rake task


app/assistants/i_reading_assistant.rb
```diff
@@ -0,0 +1,146 @@
...
+      },
+      "additionalProperties": false
+    }'
+  end
+
+  def self.distribute_tasks(task_plan:, taskees:)
+    page = Page.find(task_plan.settings[:page_id])
+    title = task_plan.title || 'iReading'
+    opens_at = task_plan.opens_at
+    due_at = task_plan.due_at || (task_plan.opens_at + 1.week)
```
kevinburleigh75: Why are we setting these things here?  Shouldn't they already have values set by something that has appropriate knowledge?

kevinburleigh75: Do we want `task_plan.opens_at + 1.week` or `task_plan.opens_at + 4.days` (to make assignments due on Friday)?

app/routines/distribute_tasks.rb
```diff
@@ -0,0 +1,42 @@
...
+  def validate_json(schema, object, options = {})
+    options[:insert_defaults] = true if options[:insert_defaults].nil?
+
+    JSON::Validator.fully_validate(schema, object, options)
+  end
+
+  def exec(task_plan)
+    # Delete pre-existing assignments
+    unless task_plan.tasks.empty?
+      task_plan.tasks.delete_all # Delete using the foreign key cascade
```
kevinburleigh75: While I realize this is faster, is it really a good idea?

lib/user_mapper.rb
```diff
@@ -6,7 +6,8 @@ def self.account_to_user(account)
     user = User.where{:account_id == account_id}.first
     return user if user.present?
 
-    outcome = CreateUser(account)
+    outcome = CreateUser.call(account)
+    puts outcome.errors.inspect
```
kevinburleigh75: puts should probably be removed

app/assistants/i_reading_assistant.rb
jpslav: I think Dante is faking it since these things aren't used yet.

app/routines/distribute_tasks.rb
Dantemss: There can be arguments for both I think...

lib/user_mapper.rb
Dantemss: Ok thanks :)

app/models/assistant.rb
```diff
@@ -1,53 +1,31 @@
 class Assistant < ActiveRecord::Base
-  belongs_to :study
-  has_many :task_plans, dependent: :destroy
 
-  serialize :settings
-  serialize :data
+  has_many :klass_assistants, dependent: :destroy
```
jpslav: we don't really need these yet, right?  (we can just use assistants directly from rake script)

app/routines/distribute_tasks.rb
```diff
@@ -0,0 +1,42 @@
...
+      task_plan.reload
+    end
+
+    owner = task_plan.owner
+    assistant = task_plan.assistant
+    klass_assistant = owner.is_a?(Klass) ? \
```
jpslav: As soon as you have spare time after the sprint, let's change `Klass` to `Course` and `KlassAssistant` to `CourseAssistant`.

app/models/assistant.rb
Dantemss: Not for this sprint yeah

app/routines/distribute_tasks.rb
Dantemss: Sure

PR #37, opened 2015-02-17
Made iReading assistant call JP's Exercise factory

Dantemss:


app/assistants/i_reading_assistant.rb
```diff
@@ -141,7 +149,12 @@ def self.distribute_tasks(task_plan:, taskees:)
       # Spaced practice
       SPACED_PRACTICE_MAP.each do |k_ago, number|
         number.times do
-          #exercise_id = IReadingSpacedPracticeSlotFiller.call(taskee, k_ago)
+          #ex = IReadingSpacedPracticeSlotFiller.call(taskee, k_ago)
+          #ex_title = 'Exercise'
+          #task_step_attributes << { tasked_class: TaskedExercise,
```
Dantemss: Actually this is wrong (need to create the TaskStep right here) but oh well it's just a comment.

PR #41, opened 2015-02-18
Added lorem ipsum feedback and correct_answer_id

Dantemss:


Dantemss:
JP said it looks good, so merging.

PR #47, opened 2015-02-19
Post-sprint refactoring

Dantemss:
Done, pending comments from JP


app/representers/api/v1/task_step_properties.rb
```diff
@@ -29,7 +29,8 @@ module TaskStepProperties
                required: true,
                description: "The type of this TaskStep, one of: #{
                             TaskedRepresenterMapper.models.collect{ |klass| 
-                              "'" + klass.name.gsub("Tasked","").underscore.downcase + "'"
```
jpslav: imo a line is only too long if it doesn't fit in a GH diff :-)  120 chars :-)

Dantemss: Oh I've been using 80 :)

PR #48, opened 2015-02-19
Moved title, url and content from task_step to TaskedX; Removed defaults

Dantemss:
Note that title can now be nil (won't show up in the API call)


app/representers/api/v1/tasked_exercise_representer.rb
```diff
@@ -25,14 +54,5 @@ class TaskedExerciseRepresenter < Roar::Decorator
              readable: true,
              if: -> (*) { task_step.completed? }
 
-    property :content,
-             type: String,
-             writeable: false,
-             readable: true,
-             getter: -> (*) { ::JSON.parse(task_step.content) },
```
jpslav: Why is this no longer needed?

Dantemss: Don't need task_step.content because content got moved.
Don't need JSON.parse because we are printing the content :)

PR #52, opened 2015-02-20
added missing call to #reset!

kevinburleigh75:
A call to `#reset!` is required after configuration changes to `OpenStax::Exchange` to ensure the creation of a new client object with the updated config.

### DIFF
--- a/lib/tasks/sprint/sprint_006/main.rb
+++ b/lib/tasks/sprint/sprint_006/main.rb
@@ -33,6 +33,8 @@ module Sprint006
         config.supported_api_versions = ['v1']
       end
 
+      OpenStax::Exchange.reset!
+
       if username_or_user.is_a? String
         run(:create_account, username: username_or_user)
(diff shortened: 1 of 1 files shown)


Dantemss:
Can we maybe move this config to config/environments/test.rb or the test_helper somehow?

kevinburleigh75:
@dantemss: I agree that moving the config would be a likely next step.

Dantemss:
For some reason after rebasing on top of this I get #<Lev::Error:0x007fa87c30e090 @code="RuntimeError", @data=#<RuntimeError: create_identifier failure (initialization failure (invalid server url))>, @kind=:exception, @message="create_identifier failure (initialization failure (invalid server url))", @offending_inputs=nil>

Dantemss:
Ah! Nevermind. Moved the FakeClient config to the initializer and it works fine.

Dantemss:
Seems like the problem was that the FakeClient just expects the same settings as the global config.

PR #53, opened 2015-02-20
Exercise wrapper

Dantemss:
Exercises FakeClient and the SpacedPractice Algorithm now both use the Exercise wrapper.
TaskedExercise also uses it to automatically populate its own fields.


app/external/openstax/exercises/v1.rb
```diff
@@ -7,7 +7,9 @@ module OpenStax::Exercises::V1
   # GET /api/exercises
   # options can have :tag, :id, :number, :version keys
   def self.exercises(options={})
-    client.exercises(options)
+    JSON.parse(client.exercises(options)).collect{|e|
```
jpslav: I agree that this change is similar to what we were doing before (except now returning the wrapper), but on closer inspection I wonder if the `exercises` method should be returning something closer that what the actual API returns, i.e. the api has a `total_count` that helps with pagination, etc.  That is not returned here.  So either we should return a hash containing that number along with an array of `OpenStax::Exercises::V1::Exercise` or we should just return json.  Thoughts?

app/external/openstax/exercises/v1/exercise.rb
```diff
@@ -0,0 +1,46 @@
...
+  def url
+    @url ||= "#{BASE_URL}/#{uid}"
+  end
+
+  def title
+    @title ||= content_hash['title']
+  end
+
+  def answers
+    @answers ||= content_hash['questions'].first['answers']
```
jpslav: of course later we need to deal with multiple questions

app/external/openstax/exercises/v1/fake_client.rb
```diff
@@ -8,11 +8,12 @@ def exercises(options={})
     arrayify(options, :number)
     arrayify(options, :version)
     arrayify(options, :id)
+    arrayify(options, :uid)
     arrayify(options, :tag)
 
     match_sets = []
-
-    match_sets.push( @exercises_array.select{|ee| options[:id].include?(ee[:id])}           ) if options[:id]
+    uids = (options[:id] || []) + (options[:uid] || [])
```
jpslav: what's the diff between `id` and `uid`?  I was using `id` as defined in the api docs.

jpslav: i.e. `id` is not meant to be the primary key (which is not exposed in the API)

app/models/tasked_exercise.rb
```diff
@@ -1,8 +1,34 @@
...
   validates :content, presence: true
+
+  delegate :answers, :correct_answer_id, :feedback_map, to: :wrapper
+
+  def wrapper
+    @wrapper ||= OpenStax::Exercises::V1::Exercise.new(content)
+  end
+
+  def url
+    u = super
```
jpslav: is this from when there was a url in `TaskStep`?  The call to `super` is confusing me -- we're calling into `ActiveRecord::Base`??

app/routines/fill_i_reading_spaced_practice_slot.rb
```diff
@@ -4,11 +4,14 @@ class FillIReadingSpacedPracticeSlot
 
   protected
 
-  def exec()
+  def exec(user, k_ago)
     # debugger
     OpenStax::Exercises::V1.fake_client.add_exercise
     exercise_hash = OpenStax::Exercises::V1.fake_client.exercises_array.last
```
jpslav: Could also make add_exercise return the hash, that way we don't have to do this `.last` bit.  Also if it isn't important that the exercise be in the fake client, you can just call `new_exercise` on the fake client.

jpslav:
Done with my review @Dantemss.  See comments.

app/external/openstax/exercises/v1.rb
Dantemss: Maybe a hash with `:items` = array of `OpenStax::Exercises::V1::Exercise` and `:total_count`

app/external/openstax/exercises/v1/exercise.rb
Dantemss: :) hopefully after v1

app/external/openstax/exercises/v1/fake_client.rb
Dantemss: Both work, the API also accepts `:uid`. It is named `uid` in the json (mostly because it's not the database id)

app/models/tasked_exercise.rb
Dantemss: `super` just gets the value of the DB field.

app/routines/fill_i_reading_spaced_practice_slot.rb
Dantemss: Hm ok

Dantemss:
@jpslav addressed comments. Ready for further review/merging.

Issue #56, opened 2015-02-23
Write specs for {Task, Tasked}AccessPolicy


jpslav:
Write specs for 

1. `app/access_policies/task_access_policy.rb` 
2. `app/access_policies/tasked_access_policy.rb`.

You might test passing in human users and non-human users.

PR #58, opened 2015-02-23
Add specs to task tasked access policies

joemsak:
Addresses #56


spec/access_policies/task_access_policy_spec.rb
```diff
@@ -1,5 +1,49 @@
-require 'rails_helper'
```
Dantemss: Interesting. I guess access policies technically don't depend on the rest of Rails.
@jpslav what do you think?

spec/access_policies/task_access_policy_spec.rb
```diff
@@ -1,5 +1,49 @@
...
+require 'spec_helper'
+require './app/access_policies/task_access_policy'
 
 RSpec.describe TaskAccessPolicy do
-  pending "add some examples to #{__FILE__}"
+  let(:requestor) { double(:requestor, id: 1) }
```
Dantemss: I usually use factories for this, but I suppose using doubles here is better.

Dantemss:
I like this but in my opinion @jpslav should review it to make sure he approves of the style (that's a lot of rspec to digest).

jpslav:
Looks good, thanks @joemsak.

As @Dantemss intoned, normally our "unit" tests are a little less "unitish", but it probably wouldn't hurt us to make these specs more isolated like you've done here with `double` and its buddies.  We should have a discussion and make a plan about which level of tests mock things and which use factories with models built-out, etc.

Also, @joemsak note that this issue requests another spec to be written too :-)

joemsak:
Oh shoot youre right I totally forgot. And yeah I am not necessarily
prescribing doubles as a change in interface won't get caught here. They
just made test feedback faster. What I'd like to do is try real
collaborators with rails (and add the other requested spec) but I need to
know what makes a requestor is or is not human and then how should I set up
taskings for the requestor?

Great feedback Thanks guys

joemsak:
I've updated the spec to use rails again, and this time when tests were passing I did some refactoring to hide the internal details of how tasks are found for the requestor. Other spec coming

app/models/task.rb
```diff
@@ -24,4 +24,7 @@ def course
     end
   end
 
+  def any_tasks?(taskee)
```
Dantemss: Not sure if this is the best name for this method. Maybe `assigned_to?`

lib/acts_as_tasked.rb
```diff
@@ -12,6 +12,10 @@ def acts_as_tasked
         validates :task_step, presence: true
 
         delegate :completed_at, :completed?, :complete, to: :task_step
+
+        define_method(:any_tasks?) do |taskee|
```
Dantemss: I think I would prefer to delegate `:tasked_to?` (see delegate call above) to `:task_step` and then have that delegate `:tasked_to?` to :task

joemsak: This all sounds good, just wanna see if anyone else on the team wants to input suggestions for the method name. I assume if I name it assigned_to or tasked_to or whatever else you want it the same on both?

Thanks

Dantemss: Sorry I wrote assigned_to? first but we decided to avoid "assigned" whenever possible because of confusion :)

jpslav:
Thanks @joemsak!

Issue #57, opened 2015-02-23
Write specs for Api::V1::{TaskSearchRepresenter, TaskedExerciseRepresenter}


jpslav:
Write spec for `Api::V1::TaskSearchRepresenter` and `TaskedExerciseRepresenter`.

karenc:
#60 is merged

PR #60, opened 2015-02-24
Add Spec for JSON format of TaskSearchRepresenter, TaskedExerciseRepresenter

nathanstitt:
Submitting this for feedback early in case I'm not on the correct track for what should be tested here.

Should the spec be testing additional features of the Representer that I'm not aware of? My thoughts were that since were testing a Representer, the primary thing we need to ensure is well-formed JSON in the the format that the client expects to see.  I also took a peek at the spec for the TaskedReadingRepresenter and that was all it did as well.  

The other thing that needs specs is the due_at and opens_at keys.  Since those are timestamp attributes, I need to control the time that the factory is ran at and then compare against that.  I don't see Timecop or any similar gems being used, is there anything and I've just missed it?  Or perhaps I should just stub Time.now ?

I'm also not super familiar with rspec.  I've used it a bit but it's quite possible that I'm not writing this in the best fashion.  Opinions are quite welcome if it should be written differently.

Currently working on the TaskedExerciseRepresenter and will add it to this shortly.

### DIFF
--- a/spec/representers/api/v1/task_search_representer_spec.rb
+++ b/spec/representers/api/v1/task_search_representer_spec.rb
@@ -1,5 +1,32 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::TaskSearchRepresenter, :type => :representer do
-  pending "add some examples to (or delete) #{__FILE__}"
+
+  context "a user" do
+
+    let(:user)    { FactoryGirl.create(:user) }
+    let(:outputs) { SearchTasks.call(q: "user_id:#{user.id}").outputs   }
+    let(:default_task)   { FactoryGirl.create(:task) }
+    let(:representation) { Api::V1::TaskSearchRepresenter.new(outputs).as_json }
+
+    it "generates a JSON representation of their tasks" do
+
+      5.times{ FactoryGirl.create(:tasking, taskee: user) }
+
+      expect(representation).to include(
+        "total_count" => 5,
+        "items" => a_collection_including(
+          a_hash_including(
+            "id"           => a_value_within(1).of(5),
+            "task_plan_id" => a_value_within(1).of(5),
+            "title"        => a_string_matching(default_task.title),
+            "is_shared"    => false
+          )
+        )
+      )
+
+    end
+
+  end
+
 end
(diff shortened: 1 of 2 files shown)


nathanstitt:
Pending the above requests for feedback, this completes Task #57

spec/representers/api/v1/task_search_representer_spec.rb
```diff
@@ -1,5 +1,32 @@
...
+
+    it "generates a JSON representation of their tasks" do
+
+      5.times{ FactoryGirl.create(:tasking, taskee: user) }
+
+      expect(representation).to include(
+        "total_count" => 5,
+        "items" => a_collection_including(
+          a_hash_including(
+            "id"           => a_value_within(1).of(5),
```
Dantemss: Hm won't this value go from 1 to 5? Not entirely sure how this works.
Also it might be bad to assume that id's are sequential in general...

spec/representers/api/v1/task_search_representer_spec.rb
```diff
@@ -1,5 +1,32 @@
...
+            "task_plan_id" => a_value_within(1).of(5),
```
Dantemss: Might be better to just get the id from the task_plan directly here.

spec/representers/api/v1/task_search_representer_spec.rb
```diff
@@ -1,5 +1,32 @@
...
+            "title"        => a_string_matching(default_task.title),
```
Dantemss: Also not sure what this means? Isn't this a regex matcher?

spec/representers/api/v1/tasked_exercise_representer_spec.rb
```diff
@@ -1,5 +1,21 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::TaskedExerciseRepresenter, :type => :representer do
```
Dantemss: This one looks good I think :)

spec/representers/api/v1/task_search_representer_spec.rb
nathanstitt: What I'm doing there is just matching one of the 5 hashes that are present on the "items" array.  

Since they're all the same except for the "id" and "task_plan_id" I didn't bother writing matchers for each one, and used the "value_within" to just say that it should be in a range.

But that's a good point.  We shouldn't assume that ID's are sequential (or that they start at 1).  I'll rework to match each one and get the id's from the factory generated "default_task"

spec/representers/api/v1/task_search_representer_spec.rb
nathanstitt: You're absolutely correct.  I shouldn't need to use the "a_string_matching" there, since it should be an exact match.

Issue #59, opened 2015-02-23
Store chapter # and section # when importing CNX book


jpslav:
The front end will need to know chapter and section numbers, e.g. Chapter 4, Section 4.2, etc.

This information isn't included in the CNX data explicitly, we just need to keep count as we import the book (using 0-based indexing, which is a little odd for a book, but there's a reason for it).

Need to add a "number" string field (or some other name) to `Book` and `Page`.  Each new level of nesting will add another ".".  E.g. in College Physics that has only chapters and sections, the numbers will be "X" for chapters and "X.Y" for sections (each module / Page is a section).  For Bio, there are units above chapters, so sections would be "X.Y.Z" (three levels down).

For now, don't worry about representing this in an API, just store the value.  Talk to @Dantemss about where to look.  @philschatz can later fill us in about how to deal with "X.0" chapter introduction "sections.

Dantemss:
`number` is already used for sort order. Need a new name.
I'm fine with storing this explicitly, since doing recursion to figure it out from the parent chapter/unit seems unnecessary and somewhat expensive.

Dantemss:
When building the book, you can probably figure this out as "#{string_from_parent}.#{number-1}"
But number might be nil until you validate the model.

Dantemss:
Actually... since this is not in the CNX data, is this something the CNX frontend just figures out by itself? Maybe Phil should just do whatever they do?

jpslav:
We're going to end up giving Phil search results (so he won't be seeing the full table of contents all the time).  So we'll need to store (or be able to easily retrieve the unit/chapter/section number).

Dantemss:
This has been solved. Closing.

PR #61, opened 2015-02-24
Add chapter and section to cnx books and pages

joemsak:
Addresses #59 

- `#path` db attribute added to `Page` and `Book`
- `Import::Page` and `Import::Book` refactored for single responsibility in small functions
- `Import::Page` and `Import::Book` now construct paths based on presence of parent books and path variables

### DIFF
--- a/app/subsystems/content/api/import_book.rb
+++ b/app/subsystems/content/api/import_book.rb
@@ -9,38 +9,61 @@ class Content::Api::ImportBook
   uses_routine Content::ImportPage, as: :page_import
 
   protected
-
-  # Recursively imports items in a CNX collection into the given book
-  def import_collection(parent_book, hash, options = {})
-    book = Content::Book.create(parent_book: parent_book,
-                                title: hash['title'] || '')
-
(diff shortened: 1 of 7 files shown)


Dantemss:
It's going in the right direction, I believe. Just keep in mind that the root Book is a special case (it's not included in the path), so you might need to check whether the parent_book is nil somewhere (that is the root) and handle that.

Dantemss:
Also assuming parent_book is not nil, you don't need to check whether it responds to path, as it will always be either a Book or nil.

joemsak:
Yep that’s handled by checking if parent_book responds to path (a crafty way of asking if it’s nil) and then path_index will be nil, too, because it hasn’t dropped into hash[‘contents’].each_with_index yet (where path_index starts getting passed in), so path will be set to a blank string

Dantemss:
I see now :)

jpslav:
@joemsak I'm an evil person and my delay on merging this PR means it is now very out of date (largely due to the fact that some of these files moved into a Content subsystem).  I'm going to need you to rebase / merge and bring it up to date.

PR #62, opened 2015-02-25
Introduction of subsystems

jpslav:
This is @kevinburleigh75's code.  It introduces subsystems as a way to manage complexity within Tutor.  Each subsystem is responsible for a limited set of models and things to do with those models.  Cross-subsystem code communicates IDs to the subsystems to get things done.  Kevin can explain better, but it'd be good for @joemsak, @nathanstitt, @Dantemss to weigh in on what is being done here.  If we give it a green light, we'll start to break all of our code down this way (e.g. we'll have book, course content, research study subsystems, etc.


kevinburleigh75:
I think this is ready to merge.  Some minor notes:
* I tweaked how the coverage spec works to ignore subsystems for now; every file has 100% coverage but it still complains and fails; I suggest we use the output of coveralls/simplecov as the basis for which files are flagged as non-compliant
* I still need to solve the missing constant autoload issue; I ensured that the current load order is deterministic so those types of issues can be reproduced on all machines

app/subsystems/course_membership/routines/is_course_student.rb
```diff
@@ -0,0 +1,24 @@
+class CourseMembership::IsCourseStudent
+  lev_routine
+
+  protected
+
+  def exec(course:, role: :missing, roles: :missing)
```
jpslav: I think the two arguments where only one must be set is a bit confusing.  I suggest changing to `exec(course:, role_or_roles:)` or just `exec(course:, roles:)`, allowing the latter argument to be either an individual role or an array.  Then you can say:

```
roles = [roles].flatten
```

to make sure that it is an array.  This will let you get rid of the "exactly one of" logic.

app/subsystems/course_membership/routines/is_course_student.rb
```diff
@@ -0,0 +1,24 @@
...
+        code: :invalid_usage,
+        offending_inputs: [role, roles],
+        message: "exactly one of 'role' or 'roles' must be specified")
+    end
+
+    roles = [role] unless role == :missing
+    role_ids = roles.collect{|r| r.id}
+
+    outputs[:is_course_student] = CourseMembership::Student.where{entity_course_id == course.id} \
+                                                           .where{entity_role_id >> role_ids}.any?
```
jpslav: I prefer `.where{entity_role_id.in role_ids}.any?` -- easier to read.

app/subsystems/course_membership/routines/is_course_teacher.rb
```diff
@@ -0,0 +1,24 @@
+class CourseMembership::IsCourseTeacher
+  lev_routine
+
+  protected
+
+  def exec(course:, role: :missing, roles: :missing)
+    unless role_param_combo_is_valid(role, roles)
```
jpslav: ditto comments above.

app/subsystems/course_membership/routines/is_course_teacher.rb
```diff
@@ -0,0 +1,24 @@
...
+        code: :invalid_usage,
+        offending_inputs: [role, roles],
+        message: "exactly one of 'role' or 'roles' must be specified")
+    end
+
+    roles = [role] unless role == :missing
+    role_ids = roles.collect{|r| r.id}
+
+    outputs[:is_course_teacher] = CourseMembership::Teacher.where{entity_course_id == course.id} \
+                                                           .where{entity_role_id >> role_ids}.any?
```
jpslav: ditto.

app/subsystems/domain/add_user_as_course_student.rb
```diff
@@ -0,0 +1,37 @@
...
+               translations: {type: :verbatim}
+
+  uses_routine CourseMembership::AddStudent,
+               translations: {type: :verbatim}
+
+  protected
+
+  def exec(user:, course:)
+    result = run(Domain::UserIsCourseTeacher, user: user, course: course)
+    fatal_error(code: :could_not_determine_if_user_is_course_teacher, offending_inputs: [user, course]) if result.errors.any?
```
jpslav: Later we should make a pass and clean up the error usage, possibly including changes to lev.  This line will never be reached if there was an error in `UserIsCourseTeacher`.

app/subsystems/domain/add_user_as_course_student.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    unless result.outputs.user_is_course_teacher
```
jpslav: because you have verbatim translations on, `outputs.user_is_course_teacher` should suffice.  If you don't want that output from the `UserIsCourseTeacher` routine translated to this routine's outputs, don't use the translations option.

app/subsystems/domain/add_user_as_course_student.rb
```diff
@@ -0,0 +1,37 @@
...
+      result = run(Domain::UserIsCourseStudent, user: user, course: course)
+      fatal_error(code: :could_not_determine_if_user_is_course_student, offending_inputs: [user, course]) if result.errors.any?
+      fatal_error(code: :user_is_already_a_course_student, offending_inputs: [user, course]) if result.outputs.user_is_course_student
+    end
+
+    result = run(Role::CreateUserRole, user)
+    fatal_error(code: :could_not_add_student_role, offending_inputs: [user, course]) if result.errors.any?
```
jpslav: All of the `fatal_error ... if result.errors.any?` are effectively nonreachable.

app/subsystems/domain/add_user_as_course_teacher.rb
```diff
@@ -0,0 +1,29 @@
...
+  protected
+
+  def exec(user:, course:)
+    result = run(Domain::UserIsCourseTeacher, user: user, course: course)
+    fatal_error(code: :could_not_determine_if_user_is_course_teacher, offending_inputs: [user, course]) if result.errors.any?
+    fatal_error(code: :user_is_already_teacher_of_course, offending_inputs: [user, course]) if result.outputs.user_is_course_teacher
+
+    result = run(Role::CreateUserRole, user)
+    fatal_error(code: :could_not_add_teacher_role, offending_inputs: [user, course]) if result.errors.any?
+    role = result.outputs.role
```
jpslav: Because of the translations, `result.outputs.role` is available as `outputs.role`

app/subsystems/domain/add_user_as_course_teacher.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    result = run(CourseMembership::AddTeacher, course: course, role: role)
```
jpslav: `run(CourseMembership::AddTeacher, course: course, role: outputs.role)`

app/subsystems/domain/add_user_as_course_teacher.rb
```diff
@@ -0,0 +1,29 @@
...
+    fatal_error(code: :could_not_add_teacher, offending_inputs: [user, course]) if result.errors.any?
+
+    outputs[:role] = role
```
jpslav: Unnecessary (translations did this for you)

app/subsystems/domain/add_user_as_course_teacher.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+end
```
jpslav: With the `if?` change, this routine could be:

```
def exec(user:, course:)
  fatal_error(code: :user_is_already_teacher_of_course) if?(Domain::UserIsCourseTeacher, user: user, course: course)
  run(Role::CreateUserRole, user)
  run(CourseMembership::AddTeacher, course: course, role: outputs.role)
end
```

app/subsystems/domain/create_course.rb
```diff
@@ -0,0 +1,12 @@
+class Domain::CreateCourse
+  lev_routine
+
+  uses_routine Entity::CreateCourse, translations: {type: :verbatim}
+
+  def exec
```
jpslav: rewritable as

```
def exec
  run(Entity::CreateCourse)
end
```

app/subsystems/domain/user_is_course_teacher.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  uses_routine Role::GetUserRoles, translations: {type: :verbatim}
+  uses_routine CourseMembership::IsCourseTeacher, translations: {type: :verbatim}
+
+  protected
+
+  def exec(user:, course:)
+    roles = run(Role::GetUserRoles, user).outputs.roles
+    result = run(CourseMembership::IsCourseTeacher,roles: roles, course: course)
+    outputs[:user_is_course_teacher] = result.outputs.is_course_teacher
```
jpslav: Just to point out a feature in Lev, this could be rewritten as:

```
class Domain::UserIsCourseTeacher
  lev_routine
  uses_routine Role::GetUserRoles, 
               translations: {type: :verbatim}
  uses_routine CourseMembership::IsCourseStudent, 
               translations: { 
                 outputs: { map: {is_course_teacher: :user_is_course_teacher} } 
               }

def exec(user:, course:)
  run(Role::GetUserRoles, user)
  run(CourseMembership::IsCourseTeacher, roles: outputs.roles, course: course)
end
```

app/subsystems/entity/routines/create_user.rb
```diff
@@ -0,0 +1,11 @@
+class Entity::CreateUser
+  lev_routine
+
+  protected
+
+  def exec
+    user = Entity::User.create
```
jpslav: Probably wouldn't be bad later to have an extension to lev here to DRY these routines up:

```
def create(model, args)
  instance = model.create(args)
  transfer_errors_from(instance, {type: :verbatim}, true)
  model_name_symbol = model.name.demodulize.underscore.to_sym
  outputs[model_name_symbol] = instance
end
```

Then this routine would become

```
def exec
  create(Entity::User)
end
```

spec/spec_helper.rb
```diff
@@ -1,5 +1,11 @@
+require 'simplecov'
```
jpslav: What's the deal with simplecov?

jpslav:
I made my comments (tho didn't really dig into the specs much).  @Dantemss do you want to take a quick look?

app/subsystems/course_membership/routines/is_course_student.rb
kevinburleigh75: While I agree that simplified the code, it complicates the caller.  I thought it read better to avoid saying `role_or_roles:` everywhere.  I also pondered just having `roles:` and allowing it to be singular or an array.  I'm not locked into any of the approaches, though.

app/subsystems/course_membership/routines/is_course_student.rb
kevinburleigh75: Agreed.

app/subsystems/domain/add_user_as_course_student.rb
kevinburleigh75: Totally agree.  I'm sure we can tweak `lev` to make life easier as well...

app/subsystems/domain/add_user_as_course_student.rb
kevinburleigh75: If that is true, it isn't apparent from the `lev` docs.  All errors in a called routine are considered fatal?

app/subsystems/course_membership/routines/is_course_student.rb
jpslav: I think it could be `roles:` and just happen to deal gracefully with a singular.  Seems fine.

app/subsystems/domain/add_user_as_course_student.rb
jpslav: No, and you're right that right now you have no external guarantee that you don't have to call this, it just so happens that we as developers have never used `nonfatal_error` (probably because all errors are pretty bad), so these lines will never be reached.  What I think you have mentioned, and what we need, is an option that forces all errors in called routines to be fatal so we don't have to worry about checking.

spec/spec_helper.rb
kevinburleigh75: `simplecov` is the coverage tool that `coveralls` uses internally.  I wanted to access it directly so I could generate coverage stats locally, not just in the CI environment.

app/subsystems/course_membership/models/student.rb
```diff
@@ -0,0 +1,8 @@
+class CourseMembership::Student < ActiveRecord::Base
+  ## using class_name as bug workaround, see: https://github.com/rails/rails/issues/15811
```
Dantemss: Just wanted to point out that it's not really a bug that will be fixed, just a design decision from Rails I believe (probably a bad one...). Rails expects `belongs_to :entity_role` to refer to the class `::EntityRole`

app/subsystems/course_membership/models/student.rb
```diff
@@ -0,0 +1,8 @@
...
+  belongs_to :entity_role,   class_name: "::Entity::Role"
+  belongs_to :entity_course, class_name: "::Entity::Course"
+
+  validates_presence_of :entity_role_id
```
Dantemss: I've been leaning towards `validates_presence_of :entity_role` to ensure the record exists.
In fact, I prefer `validates :entity_role, presence: true` :)

app/subsystems/course_membership/models/student.rb
```diff
@@ -0,0 +1,8 @@
...
+  validates_presence_of :entity_course_id
+end
```
Dantemss: I'd add a uniqueness validation for the `role` scoped to the `course_id`

app/subsystems/course_membership/models/teacher.rb
```diff
@@ -0,0 +1,8 @@
+class CourseMembership::Teacher < ActiveRecord::Base
```
Dantemss: Same comments as for the student one :)

app/subsystems/course_membership/models/student.rb
```diff
@@ -0,0 +1,8 @@
+class CourseMembership::Student < ActiveRecord::Base
+  ## using class_name as bug workaround, see: https://github.com/rails/rails/issues/15811
+  belongs_to :entity_role,   class_name: "::Entity::Role"
```
Dantemss: Oh and since you have to specify the class name anyway, you could potentially change this to `belongs_to :role` and the field to `role_id`, but I have no strong preference here.

app/subsystems/course_membership/models/student.rb
nathanstitt: using the `validates_presence_of` style is deprecated in favor of `validates: XXX, presence: true `.  http://guides.rubyonrails.org/3_0_release_notes.html#patches-and-deprecations  No idea when/if they'll remove it though since there's a bunch of code out there using that style.

app/subsystems/entity/models/course.rb
```diff
@@ -0,0 +1,2 @@
+class Entity::Course < ActiveRecord::Base
```
Dantemss: We might need to at least add the has_many's here. I don't think we can preload all the students in a course for reporting etc without the `has_many` (at least not easily).

Also not having the `has_many`'s makes the `belongs_to` in student etc have no inverse relation.

app/subsystems/entity/routines/create_user.rb
Dantemss: Alternatively, https://github.com/openstax/exchange/blob/master/app/routines/standard_create.rb

app/subsystems/domain/add_user_as_course_student.rb
Dantemss: +1

app/subsystems/domain/add_user_as_course_student.rb
Dantemss: I think all errors are fatal by default... I vaguely remember one time giving nonfatal_error a try and the calling routine failed anyway.

kevinburleigh75: I will adjust the code with this assumption, although it doesn't fully address the meaningful translation of error messages for the caller.

config/initializers/subsystems.rb
```diff
@@ -0,0 +1,9 @@
+domain_files         = Dir[File.join(Rails.root, "app/subsystems/domain/**/*.rb")].sort
+entity_files         = Dir[File.join(Rails.root, "app/subsystems/entity/**/*.rb")].sort - domain_files
+subsystem_root_files = Dir[File.join(Rails.root, "app/subsystems/*/*.rb")].sort - domain_files - entity_files
+subsystem_files      = Dir[File.join(Rails.root, "app/subsystems/**/*.rb")].sort - domain_files - entity_files - subsystem_root_files
+
+[entity_files, subsystem_root_files, subsystem_files, domain_files].flatten.each do |file|
+  # puts "=== requiring #{file} ==="
+  ActiveSupport.require_or_load file
```
Dantemss: Hopefully this won't cause issues when we modify the files in the dev environment? (I don't mind restarting the server each time I modify something, however)

app/subsystems/course_membership/models/student.rb
kevinburleigh75: "Fixed" by removing the word 'bug'.

spec/coverage/coverage_spec.rb
```diff
@@ -3,12 +3,14 @@
...
 
+  SUBSYSTEM_PATHS = Dir["app/subsystems/**/*.rb"].collect{|f| f.gsub('app/', '')}
+
   IGNORED_PATHS = [
     'helpers/application_helper.rb', # Empty
     'lib/generators/secrets/secrets_generator.rb', # Add secrets.yml to repo and remove
     'lib/lev/delegator.rb', # Move to lev
     'lib/markdown_wrapper.rb' # Move to OSU
-  ]
+  ] + SUBSYSTEM_PATHS
```
Dantemss: Can we at least remove the routines folder from the ignore list?

app/subsystems/course_membership/models/student.rb
kevinburleigh75: Fixed.

app/subsystems/course_membership/models/student.rb
kevinburleigh75: This is handled in the schema, but I added it in the model as well.

app/subsystems/legacy_user/legacy_user.rb
```diff
@@ -0,0 +1,5 @@
+module LegacyUser
```
Dantemss: Is this a temporary thing (until we add a UserInfo subsystem?)

app/subsystems/entity/models/course.rb
kevinburleigh75: Adding those associations makes all the Entity models dumping grounds for functionality used only by subsets of the system!  Unless we _really_ need it for performance, I'd avoid this.

spec/coverage/coverage_spec.rb
Dantemss: Actually if we are just going to use simplecov, just nuke this file.

config/initializers/subsystems.rb
kevinburleigh75: I'm working on fixing this with a more `autoload`-esque solution.

Dantemss:
Also we eventually need to remove the models the subsystems replace to avoid confusion and change existing features and specs to use the subsystem models.

spec/coverage/coverage_spec.rb
kevinburleigh75: In the end, I think we should use simplecov's per-rb file coverage stats as the basis for determining which files need more coverage.  So this is really a temporary thing (I hope).

Someone needs to figure out how to run one specific spec after all others, and have that spec examine the coverage data and take the appropriate action.

I added simplecov's HTML formatter back into the mix so at least we can easily view our coverage under .../coverage/index.html.  (There are also JSON files in there, but they are hidden.)

app/subsystems/legacy_user/legacy_user.rb
kevinburleigh75: Yes, temporary.

spec/coverage/coverage_spec.rb
Dantemss: Yeah I agree simplecov is better... feel free to remove the coverage_spec

app/subsystems/entity/models/course.rb
Dantemss: I think we will need them, at least for the Course... but it can wait.

db/migrate/01000_create_course_membership_students.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCourseMembershipStudents < ActiveRecord::Migration
+  def change
+    create_table :course_membership_students do |t|
```
Dantemss: I suggest adding foreign key constraints on entity_course and entity_role to ensure data consistency.

db/migrate/01001_create_course_membership_teachers.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCourseMembershipTeachers < ActiveRecord::Migration
+  def change
+    create_table :course_membership_teachers do |t|
```
Dantemss: Same as previous comment.

db/migrate/01100_create_role_users.rb
```diff
@@ -0,0 +1,11 @@
+class CreateRoleUsers < ActiveRecord::Migration
+  def change
+    create_table :role_users do |t|
```
Dantemss: Same, but on user and role.

PR #64, opened 2015-02-27
content subsystem and start of course readings APIs

jpslav:


app/subsystems/content/models/exercise.rb
```diff
@@ -0,0 +1,14 @@
...
+class Content::Exercise < ActiveRecord::Base
+  acts_as_resource
+
+  # has_many :content_exercise_topics, dependent: :destroy,
+  #                                    class_name: "::Content::ExerciseTopic",
+  #                                    inverse_of: :content_exercise,
+  #                                    foreign_key: "content_exercise_id"
+
+  has_many :exercise_topics, dependent: :destroy,
+                             inverse_of: :exercise
```
jpslav: For @Dantemss, @joemsak, @nathanstitt: I added overloaded `has_many` and `belongs_to` calls that are subsystem-aware.  The result is that I can use the code here instead of the commented out code above.  I.e. when following rails-like naming conventions we get `class_name` and `foreign_key` correctly populated for us like we do in a non-subsystem context.  I'll have to write something about this in @kevinburleigh75's eventual READMEs.  Comments welcome :-)

app/subsystems/content/api/get_book_toc.rb
```diff
@@ -0,0 +1,12 @@
+class Content::Api::GetBookToc
```
Dantemss: This seems to be more like a Content::GetBook

app/representers/api/v1/reading_search_representer.rb
```diff
@@ -0,0 +1,9 @@
+module Api::V1
+  class ReadingSearchRepresenter < Roar::Decorator
+
+    include Roar::Representer::JSON
+
+    collection :items
```
Dantemss: I'd add the total_count too so the API doesn't change later if we have to add it. This can be done by inheriting from the abstract search representer, like in: https://github.com/openstax/tutor-server/blob/master/app/representers/api/v1/task_search_representer.rb

app/controllers/api/v1/courses_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    description <<-EOS
+      Course description to be written...
+    EOS
+  end
+
+  api :GET, '/courses/:course_id/readings', 'Searches a course\'s readings'
+  description <<-EOS 
+    #{json_schema(Api::V1::ReadingSearchRepresenter, include: :readable)}
+  EOS
+  def search_readings
```
Dantemss: I think this needs to be just `readings` to route properly.

app/subsystems/content/api/import_book.rb
```diff
@@ -0,0 +1,46 @@
+class Content::Api::ImportBook
```
Dantemss: Not sure if ::Api is the right thing here because of the autoloader patch.

Also, I suggest either following Kevin's folder structure (routines instead of api) or changing his to match yours.

app/subsystems/content/models/book.rb
```diff
@@ -0,0 +1,18 @@
+class Content::Book < ActiveRecord::Base
+  acts_as_resource allow_nil: true
+
+  sortable_belongs_to :parent_book, on: :number,
+                                    class_name: '::Content::Book',
+                                    inverse_of: :child_books,
+                                    foreign_key: "parent_book_id"
```
Dantemss: Don't think this line is needed, since the association has this name (:parent_book)

app/subsystems/content/models/exercise.rb
Dantemss: Nice... I'll check it out.

config/initializers/subsystems.rb
```diff
@@ -1,14 +1,76 @@
...
+  module_name = self.name.deconstantize.underscore
+
+  # ****** Temporary control to limit to only the Content subsystem ******
+  return if module_name != "content"
+  
+  if subsystems.any?{|subsystem| subsystem.name == module_name}
+    subsystem_option ||= module_name.to_sym
+    options[:class_name] ||= "::#{subsystem_option.to_s.camelize}::#{association_name.to_s.camelize.singularize}"
+
+    if :belongs_to == association_type
```
Dantemss: Maybe we can do something that doesn't need an `if` using some of the association's method... but I can look at it later.

Dantemss: Just pointing out that has_one is missing for now.

app/subsystems/content/api/get_book_toc.rb
jpslav: This routine is incomplete (almost completely)

app/subsystems/content/api/import_book.rb
jpslav: I got approval from @kevinburleigh75.  Basically the routines in `/api` are intended to be used elsewhere.  Those in `/routines` are for internal subsystem use.

config/initializers/subsystems.rb
```diff
@@ -1,14 +1,76 @@
...
+    subsystem.module.autoload symbol, path 
   end
   
-  a_module.define_singleton_method(:table_name_prefix) do
-    "#{module_name.underscore}_"
+  subsystem.module.define_singleton_method(:table_name_prefix) do
+    "#{subsystem.name.underscore}_"
   end
 end
+
```
nathanstitt: I've overridden associations before in different fashion that I think is a bit clearer.  I used a concern that I mixed into ActiveRecord.  It extracts and modifies the options and then calls super to call the original implementation.  For our purposes here we could only include the concern into the subsystem classes so they'd be the only ones affected.

My implementation:
https://github.com/argosity/lanes/blob/master/lib/lanes/concerns/association_extensions.rb

In this case, I'm adding an :export and :listeners options, so you can ignore the `setup_listeners` and `setup_association_export` methods.

app/subsystems/content/models/book.rb
jpslav: This would be true if I changed the DB foreign key to `"content_parent_book_id"` or if I passed in `subsystem: :ignore` here. "content_parent_book" doesn't make a bunch of sense to me because "parent" is in the middle of the namespace and class names.  `subsystem: :ignore` is maybe a little too obscure; might be better to be explicit here.

config/initializers/subsystems.rb
jpslav: as is `has_and_belongs_to_many`.  Also, I'm not being friendly to the full signature of `has_many`, really I should be dealing with a `scope`, and sometimes there is no scope and we have to know the options are in the `scope` field, etc.

config/initializers/subsystems.rb
jpslav: Interesting.  Sounds like a nice next task for @nathanstitt :-)  Also for adding `has_one`, `has_and_belongs_to_many`, etc.

jpslav: Feel free to make an issue and assign yourself.

app/controllers/api/v1/courses_controller.rb
jpslav: good point; working on this part today.

app/representers/api/v1/reading_search_representer.rb
jpslav: Yeah, this is TBD too.  I started with this stuff and then spent all my time getting the book stuff into a subsystem.  Working on it today.

PR #66, opened 2015-02-27
Tasked exercise validations

Dantemss:
This is done, but not moved into subsystems yet.


Dantemss:
This is done, but not moved into subsystems yet.

PR #67, opened 2015-02-27
Admin creates blank course

joemsak:
Pivotal Tracker #88795132
https://www.pivotaltracker.com/story/show/88795132


app/controllers/admin/courses_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Admin::CoursesController < ApplicationController
+  def index
```
jpslav: Need to check the user is an administrator (and if don't have administrator role stuff set up will need that).  Could have an admin base controller that has the blanket checking that the user is an admin.

app/views/admin/entity/courses/_course.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= div_for course do %>
+  <%= time_ago_in_words(course.created_at) %> ago
```
jpslav: I know there was some discussion about this in skype, but we do need a place to stick a course's name.  If this is a course metadata ss, so be it.  Or elsewhere.  Then ideally you could input this name when creating the course and when editing it.

app/subsystems/domain/list_courses.rb
```diff
@@ -0,0 +1,13 @@
+class Domain::ListCourses
+  lev_routine
+
+  uses_routine CourseProfile::Api::GetAllProfiles,
+               translations: { outputs: { map: { profiles: :courses } } },
```
jpslav: nice!

spec/controllers/admin/courses_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rails_helper'
+
+RSpec.describe Admin::CoursesController do
+  let(:admin) { FactoryGirl.create(:user, :administrator) }
+
+  before { controller.sign_in(admin) }
```
jpslav: after this PR can you add a spec to make sure that non-admins can't access any admin controllers? (at least this one's methods unless there's a good way to check that authenticate_admin is working)

spec/support/signin_helper.rb
```diff
@@ -0,0 +1,11 @@
+module SigninHelper
+  def stub_current_user(user)
+    allow_any_instance_of(ApplicationController)
+      .to receive(:current_account)
```
jpslav: We don't use `current_account` anywhere, so no need to stub I don't think.  Can you take this out in a followup PR; I'm going to merge this one.

jpslav:
Thanks @joemsak

spec/support/signin_helper.rb
joemsak: It's used by `:authenticate_user!` somehow... If I don't stub this, then that interceptor catches me and I can't move on to `:authenticate_admin!`

PR #68, opened 2015-02-27
fixes(?) the subsystem autoload issues

kevinburleigh75:
I believe this fixes the subsystem autoload issues by adjusting `initializers/subsystems.rb` to:
* remove `.../app/subsystem/` from `eager_load_paths`
* set up the normal ruby `autoload` symbol-to-file mappings for each subsystem
* using `require` to load `.../app/subsystems/**/*.rb` files


jpslav:
Looks good @kevinburleigh75. Thanks.

PR #89, opened 2015-03-05
Beef up admin

joemsak:
Admin needed a root console page, and the ability to Edit courses


Dantemss:
Looks good to me. The only thing is that we might want a link to the admin console somewhere in the layout if you are an admin.

PR #91, opened 2015-03-06
Subsystems association extensions and autoloading

nathanstitt:
This extracts the has_many, belongs_to, and has_one
overrides into a concern that's mixed into
ActiveRecord::Base

Also convert the Subystems into a proper namespace
with helper methods to configure it, and sets up autoloading
for each child path of a subsystem.


jpslav:
@nathanstitt I think you'll need to update this PR to include your other PR (just now merged) to put `GetPage` in the `Api` namespace.

nathanstitt:
@jpslav Thanks, I just rebased and pushed up the updated branch.  Travis is happy now.

kevinburleigh75:
I think there is a problem with eager loading.  `bundle exec rake erd` calls `Rails.application.eager_load!` to get all AR subclasses into memory, but the current solution misses them and they are omitted from the resulting `erd.pdf` file.  Somehow our solution needs to work when eager loading as well as auto loading (and preferably when auto-reloading as well).

@nathanstitt

nathanstitt:
Hmm.  Good catch   I don't think I actually verified that the classes were present in the pdf.  I think I can add them to eager load as well so they should be there.

nathanstitt:
I've tested the auto-reloading and bad news there, it doesn't work.  Plus it doesn't work in the worst way possible in my opinion.  

I tested by inserting a `Content::Exercise.test` method that just returned a string.  I then added a <%= Content::Exercise.test %> to a test view and loaded the page.  It printed fine, so I then modified the string returned in `Content::Exercise.test` and reloaded the page.

I then received a  `uninitialized constant Content::Exercise` exception.  I'm thinking what happens is that Rails actually unloads the class, and then attempts to reload it but can't find the file and throws the exception.

I'm still investigating that to see if there's anything else we can do.  Manually requiring the file doesn't help either, it still fails on reload.

jpslav:
@nathanstitt I'm getting a similar `uninitialized constant` error which changes to a "A copy of UserProfile::Profile has been removed from the module tree but is still active!" error when I try to fix it, this time in Apipie (when I'm accessing http://localhost:3001/api).  I know Apipie does some aggressive unloading and loading.

nathanstitt:
@jpslav Yes, I receive the exception when loading Apipie on both master and on this branch.  I'm currently investigating the autoload path settings to see what I can come up with.

nathanstitt:
I'm unable to figure out any method that allows Rails auto loading/reloading to work.  The issue I'm facing is that Rails will unload the entire namespace whenever a file changes, and then attempt to re-load it using the algorithms that are explained in http://edgeguides.rubyonrails.org/autoloading_and_reloading_constants.html

When it unloads everything, it loses all the configuration that we've added to the module, such as the Ruby autoload configuration and the table `table_name_prefix`  This causes the next load attempt for an object inside the namespace to fail.

For now I've turned off the reloading in config/environments/development and everything appears to work fine, but it will be cumbersome to restart the server when you make changes.

The only method I can see to get reloading working is to abide by Rail's namespace conventions where the file path matches the Ruby object namespace. We can add the subsystems directory as a root path, and then each object inside of them must match the convention.

  * Entity::Book
    * Currently is in `subsystems/entity/models/book.rb`
    * Would need to move to directory `subsystems/entity/book.rb`
  * Entity::CreateBook
    * Currently is in `subsystems/entity/api/book.rb`
    * We'd need to either change the signature to be `Entity::API::CreateBook`, or move it to be in directory  `subsystems/entity/book.rb` where it would mix with the models.
    
Once the files are moved into position, we may have to create an explicit module file for each subsystem that sets the `table_name_prefix` We're setting it as part of the subsystem initialization now, but I'm not sure if that setting will survive reloading unless it's in a file.

spec/routines/recover_tasked_exercise_spec.rb
```diff
@@ -14,24 +14,27 @@
...
+    }.not_to change{ tasked_reading.task_step }
+    expect(result.errors.first.code).to eq(:missing_recovery_exercise)
   end
 
-  it "adds an extra recovery step after the given step" do
+  # FIXME either this spec or the RecoverTaskedExercise itself is broken
+  # It attempts to add a a duplicate TaskStep, which runs into
+  # unique validation on TaskStep#tasked_id
+  xit "adds an extra recovery step after the given step" do
     recovery = tasked_exercise_with_recovery.recovery_tasked_exercise
```
nathanstitt: This needs to be investigated.  The file was lacking the "_spec" extension so it didn't run automatically.  I uncovered this inadvertently because I was running each file individually.

app/subsystems/course_membership/models/entity_extensions.rb
```diff
@@ -1,4 +1,2 @@
-class Entity::Role < ActiveRecord::Base
-  has_many :students, subsystem: :course_membership
-  has_many :teachers, subsystem: :course_membership
-end
+Entity::Models::Role.has_many :students, subsystem: :course_membership, foreign_key: 'entity_role_id'
+Entity::Models::Role.has_many :teachers, subsystem: :course_membership, foreign_key: 'entity_role_id'
```
nathanstitt: I changed this to not re-open the class.  It was having issues because if this file as loaded first, it would never load the "real" class since the Entity::Models::Role was already defined.

I also had to specify the foreign_key on the associations, since it would otherwise use `course_membership_role_id`  I'm not sure why it didn't do that before, or maybe we just didn't have good coverage on this?

spec/routines/recover_tasked_exercise_spec.rb
Dantemss: We demo'd the recovery, so the routine works. I'll check the spec.

config/environments/development.rb
```diff
@@ -4,7 +4,10 @@
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  #
+  # Class reloading is disabled by setting cache_classes=true
+  # This is so the the subsystems do not unload when they're modified
+  config.cache_classes = true
```
Dantemss: Can we remove this now that our folder structure is better?

lib/tutor/subsystems.rb
```diff
@@ -0,0 +1,15 @@
+require_relative 'subsystems/association_extensions'
+require_relative 'subsystems/base_model'
+
+module Tutor
+
+  module SubSystems
+    mattr_accessor :valid_namespaces
+    # called by the association_extensions to determine if a namespace should be extended
+    def self.valid_name?(name)
```
Dantemss: Should we just mark them all as valid now? Or are there still some problematic ones?

nathanstitt: user_profile still has issues. Some of the associations need to be redone on it.

app/subsystems/course_membership/models/entity_extensions.rb
jpslav: Wouldn't it use `entity_role_id`?  Note that there was a bug in my first pass of the `has_many` association that I later changed (I was computing the foreign key incorrectly).  Maybe this PR was based off of that bug?  See the change in https://github.com/openstax/tutor-server/pull/116/files#diff-c76543352e879a977234055e49f74e44R43

lib/tutor/subsystems/association_extensions.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        # If the :subsystem wasn't specified, default to the current model's subsystem
+        subsystem_name = my_subsystem_name if subsystem_name.blank?
+
+        options[:class_name] ||= "::#{subsystem_name.camelize}::Models::#{association_name.to_s.camelize.singularize}"
+        if is_belongs_to
+          options[:foreign_key] ||= "#{subsystem_name}_#{association_name.to_s.underscore}_id"
+        else
+          class_name = self.name.demodulize.underscore
+          options[:foreign_key] ||= "#{subsystem_name}_#{class_name}_id"
```
jpslav: Per my other comment (and my earlier bug fix to my `subsystems.rb` file) I believe this line should be:

```ruby
options[:foreign_key] ||= "#{self.name.underscore.gsub('/','_')}_id"
```

Dantemss: I believe the line `class_name = self.name.demodulize.underscore` is equivalent to what you posted.
The difference seems to be whether or not `#{subsystem_name}` is prepended to that?

Dantemss: We probably want to make a decision on that (i.e. should be foreign key be book_part_id or content_book_part_id?)

jpslav: You're right Dante, this is actually correct.  (and my way is actually bad because it includes `_models_`).   Also, I think we have to have the subsystem name prepended (think it was in both cases).  So I'm guess I'm just confused why @nathanstitt said he had to specify the foreign keys in the `entity_extensions.rb` files... The default `:foreign_key` as shown above should work fine for:

```ruby
Entity::Models::Role.has_many :students, subsystem: :course_membership, foreign_key: 'entity_role_id'
```

right?

nathanstitt: @jpslav  Yes, I think the reason the issue didn't surface before was because your foreign_key generation was different.  I don't think I picked up the change you'd made.  But it sounds like Dante's correct, the code is doing the right thing after all.

In this case, I had to specify the FK because the subsystem is set as `course_membership`.  So it attempts to use `course_membership_role_id` as the foreign_key which doesn't exist on the table, it has `entity_role_id` field.  

It sounds like we should rename the column.

The issue can be replicated by running the `/spec/controllers/api/v1/courses_controller_spec.rb` spec while the foreign_key commented out.  The exact exception raised is: 

SQLite3::SQLException: no such column: course_membership_students.course_membership_role_id: SELECT "entity_roles".* FROM "entity_roles" INNER JOIN "course_membership_students" ON "course_membership_students"."course_membership_role_id" = "entity_roles"."id" WHERE "course_membership_students"."entity_course_id" = 1`

jpslav: Oh right duh on subsystem being `course_membership` here.  

So I'm going to go back on my previous comment :-)

1. We should definitely not rename the column.  The column should totally be `entity_role_id` because that's what it points to.
2. My original fix of `options[:foreign_key] ||= "#{self.name.underscore.gsub('/','_')}_id"` is still what we want except that we now have a `Models` module in the middle of the class name that we don't want.  In that case the following should work...

```ruby
self_split_name = self.name.underscore.split('/')                    # e.g. ['entity', 'models', 'role']
calling_subsystem = self_split_name.first                            # e.g. 'entity'
calling_class = self_split_name.last                                 # e.g. 'role'
options[:foreign_key] ||= "#{calling_subsystem}_#{calling_class}_id" # e.g. "entity_role_id"
```

... and we should be able to remove the manual `foreign_key` values from the `entity_extensions.rb` files. 

That work?

Dantemss: How about `"#{self.name.underscore.gsub(/\/models\/|\//, '_')}"`?
So we don't assume a certain format on the class name.

jpslav: That would work too.

PR #95, opened 2015-03-10
WIP: Reading statistics for dashboard

nathanstitt:


app/controllers/api/v1/task_plans_controller.rb
```diff
@@ -82,4 +82,61 @@ def destroy
...
+        ]
+      }
+    }```
+    #{json_schema(Api::V1::TaskPlanRepresenter, include: :readable)}
+  EOS
+  def retrieve
+    plan = TaskPlan.find(params[:id])
+    stats = CalculateTaskPlanStatistics.call(plan:plan).outputs.statistics
+    render json: Api::V1::TaskPlanRepresenter.new(plan).to_hash(stats: stats)
+  end
```
nathanstitt: I'm sure this isn't the correct way to render.  The issue I ran into is that the `TaskPlanRepresenter` by default doesn't include the stats, instead it only includes that subtree if `stats` is present.  I'll make other comments below on the `TaskPlanRepresenter` to illustrate the issue I faced with it.

app/representers/api/v1/task_stats_representer.rb
```diff
@@ -0,0 +1,99 @@
...
+      readable: true,
+      writeable: false,
+      decorator: PeriodFragment
+
+    collection :periods,
+      readable: true,
+      writable: false,
+      decorator: PeriodFragment
+  end
+end
```
nathanstitt: I'm not at all happy with the nested classes here.  It makes it super hard to follow.  However I didn't want to create a new file for decorators that are only used on the stats representer.  Is there another method to re-use blocks that would be cleaner?  If not, I'll break apart the file to make it easier to grock the representation.

app/routines/calculate_task_plan_statistics.rb
```diff
@@ -0,0 +1,44 @@
...
+      course: generate_period_stat_data(min:20),
+      periods: 0.upto(rand(3)+2).map do {
+        id: rand(1000)+100,
+        title: Faker::Company.bs
+      }.merge( generate_period_stat_data(min:10) )
+      end
+    })
+  end
+
+end
```
nathanstitt: This is all just a stub.  Future work will fill out the hash with real data.  I used a Hashie::Mash so it's more object like in case we decide to use a AR model for storing the stats.

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -33,5 +33,12 @@ class TaskPlanRepresenter < Roar::Decorator
...
 
+    property :stats,
+             extend: TaskStatsRepresenter,
+             getter: ->(args){ args[:stats] },
+             type: Object,
+             readable: true,
+             writable: false
+
   end
 end
```
nathanstitt: Is there a better way to signal to the representer that it should include the "stats" subtree?  Right now I'm manually passing in the stats from the controller.  That works okish, but I couldn't figure out how to do it when using `standard_read`

Dantemss: I think you can pass just some `stats: true` as args. Instead of the getter, just use the `if` key, something like `if: ->(args) { args[:stats] }`.

If you want to use standard_read, you will probably need to patch it in the openstax_api gem. Otherwise, I think you can pass the arguments to the respond_with method somehow?

app/representers/api/v1/task_stats_representer.rb
Dantemss: I would probably just put the nested classes into their own files. There is a `nested` method in representable, but I don't think I ever got that to work properly.

config/routes.rb
```diff
@@ -45,6 +45,7 @@
       end
     end
 
+    get '/plans/:id' => 'task_plans#retrieve'
```
Dantemss: I'd prefer resourceful routes if possible :D

app/controllers/api/v1/plans_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+              }
+            ]
+          }
+        ]
+      }
+    }```
+    #{json_schema(Api::V1::TaskPlanRepresenter, include: :readable)}
+  EOS
+  def show
+    plan = TaskPlan.find(params[:id])
```
jpslav: Gonna need authorization code around here.

jpslav:
This is looking good.  I'd like to get this merged Monday, so if there are any really big issues please address them ASAP and then let me know when it is ready.

Of course, there's still a good bit of detail to do, getting it merged is just to make the stubs available for the frontend folks.  It'd be nice to be able to tell them how to generate some fake data to test out the stats (either with them actually making an iReading and then `GET`ting it or through some other rake task that generates fake data for them to call a `GET` on).

Here are some bigger things that'll need to be done in a follow-on PR:

1. goes without saying, but all the stubs have to go :-)
2. i'm cool with leaving the `stats` in as a subelement to a `GET` on a plan, but we'll need to make it plan-type-specific.  
   Right now it only shows stats for readings, but not all plans are iReading plans.
3. Relatedly, things named like `TaskPeriodStatisticsRepresenter` would be more appropriately named `IReadingPeriodStatsRepresenter`, or similar.
   Ditto for `CalculateTaskPlanStatistics`, etc.
3. Need to check authorization in the plans controller #show action.

nathanstitt:
Very cool.  I'll get those changes implemented shortly so we can merge.  

A quick question: What do you mean by "we'll need to make it plan-type-specific." ?  Do you mean that the "stats" subtree should be renamed to something like "reading-stats", or that the endpoint should be renamed to something like: `/plans/123/stats/reading`?

jpslav:
you don't have to make those changes now, let's just make whatever critical changes need to be made so the FE folks can get their hands on the API, see the documentation, etc.

philschatz:
Merging WIP PR's?

nathanstitt:
Arg.  that was all me I'm afraid.  I never took the "WIP:" prefix off. 

I think I'll attempt to use the "new guy doesn't know what he's doing" excuse here

jpslav:
ah sorry didn't see the title was still WIP (shouldn't have been).  We talked about it, and this is mostly the stub PR so y'all can have stats in real API results.

PR #97, opened 2015-03-11
Admin creates user accounts

joemsak:
Pivotal https://www.pivotaltracker.com/story/show/88795134

currently stubbing a made up method `OpenStax::Accounts::Account.create_temp_account` which takes an `attributes {}` hash, currently expecting `username`, and `password`


app/handlers/admin/users_create.rb
```diff
@@ -0,0 +1,25 @@
+class Admin::UsersCreate
+  lev_handler
+
+  paramify :user do
+    attribute :username, type: String
+    attribute :password, type: String
+    attribute :entity_user_id, type: Integer
```
Dantemss: Don't think you will be able to specify this.

app/handlers/admin/users_create.rb
```diff
@@ -0,0 +1,25 @@
+class Admin::UsersCreate
+  lev_handler
+
+  paramify :user do
+    attribute :username, type: String
+    attribute :password, type: String
```
Dantemss: Not entirely clear if we want a password specified by the admin or just a random password.

app/subsystems/domain/create_account.rb
```diff
@@ -0,0 +1,13 @@
+class Domain::CreateAccount
+  lev_routine
+
+  protected
+
+  def exec(user_params)
+    outputs[:account] = OpenStax::Accounts.create_temp_user(
```
Dantemss: Same comments as paramify section above.

app/handlers/admin/users_create.rb
joemsak: Whups this is leftover...

Dantemss:
@joemsak there is now an `OpenStax::Accounts::Api.create_temp_account` method you can use.
Unfortunately, testing will prob have to wait until Nathan's accounts PR is merged (then we can VCR it).

karenc:
Add "sign in as" link in the admin users screen.

https://www.pivotaltracker.com/n/projects/1156756/stories/91235278

karenc:
Added "(signed in as &lt;username&gt;)" next to the "Sign out!" link so that we can tell which user is currently logged in.

Dantemss:
The accounts PR has been merged. Now need it to be deployed to dev (if it hasn't already) so Joe can test it :)

Dantemss:
https://exchange-dev1.openstax.org/ is now live so this PR can finally be tested with real data. Setup:
1. Login as the Exchange admin (emailed you the PW)
2. Add a Platform (you will have to click the Add Application to create the associated Application, or just create the Application separately beforehand)
3. Enter some name, client id and secret for Application. Redirect URI is most likely not needed, since the Multiple Choice submission API uses the OAuth Client Credentials flow.
4. Put the Exchange client id and secret in your tutor secrets.yml
5. Make tutor use the real Exchange client
6. Make sure to VCR the responses, cause you might not be able to create the same event with all the same fields twice.

Dantemss:
Let me know if there are any problems with the above steps.

Dantemss:
The new API is live in https://accounts-dev1.openstax.org/, so this PR can finally be tested with real data. Setup:
1. Login as the Accounts admin (emailed you the PW)
2. Add an Application from the admin menu
3. Enter the application name and redirect URI for your local tutor. You will receive a client_id and secret for the app.
4. Put the Accounts client id, secret, url in your tutor secrets.yml and disable stubbing.
5. Make the code call OpenStax::Accounts::Api.create_temp_user(attributes) where attributes contains either an email or a username or a username and a password.
6. Make sure to VCR the tests

app/handlers/admin/users_create.rb
```diff
@@ -0,0 +1,24 @@
...
+    validates :username, presence: true
+    validates :password, presence: true
```
Dantemss: I believe the current accounts API allows you to specify the PW but doesn't require it.

app/subsystems/domain/create_account.rb
```diff
@@ -0,0 +1,12 @@
...
```
Dantemss: I moved the Api stuff to OpenStax::Accounts::Api. Also need to remove the stubbing.

Dantemss:
Still need to remove stub from the feature spec and change the code to use the unstubbed method.

app/routines/get_account.rb
```diff
@@ -0,0 +1,12 @@
+class GetAccount
```
jpslav: When https://github.com/lml/lev/pull/16 is merged, this can become

```ruby
class GetAccount
  lev_routine delegates_to: UserProfile::GetAccount
end
```

app/routines/get_all_user_profiles.rb
```diff
@@ -1,7 +1,7 @@
 class GetAllUserProfiles
```
jpslav: Ditto this for `delegates_to`

app/routines/map_users_accounts.rb
```diff
@@ -11,7 +11,6 @@ def user_to_account(profile)
     end
 
     private
```
jpslav: `MapUsersAccounts` isn't really a routine.  Can it be moved elsewhere? (like `/lib`?)

app/routines/map_users_accounts.rb
```diff
@@ -21,8 +20,11 @@ def find_profile
     end
 
     def create_profile
-      create_profile = UserProfile::CreateProfile.call(account_id: @account.id,
-                                                       exchange_identifier: identifier)
+      create_profile = UserProfile::CreateProfile.call(attrs: {
```
jpslav: Routines run "expressly" already raise any errors, so make `CreateProfile` have `express_output: :profile` and then change this change this method to 

```ruby
def create_profile
  UserProfile::CreateProfile[account_id: @account.id, exchange_identifier: identifier]
end
```

app/subsystems/user_profile/create_profile.rb
```diff
@@ -1,20 +1,52 @@
...
+    def exec(attrs:)
+      attrs = prepare_attrs_for_profile(attrs)
+      outputs[:profile] = Models::Profile.create!(attrs)
+    end
+
+    private
+    def prepare_attrs_for_profile(attrs)
+      @user = create_or_find_user(attrs)
+      @account = create_or_find_account(attrs)
+      account_attributes.each { |key| attrs.delete(key) }
```
jpslav: Instead of lines 19 and 20 here, I think you can just say:

```ruby
default_profile_attributes.merge(attrs.except(account_attributes))
```

app/subsystems/user_profile/create_profile.rb
```diff
@@ -1,20 +1,52 @@
...
+    def account_attributes
+      [:openstax_uid, :username, :password, :access_token,
+       :first_name, :last_name, :full_name, :title]
+    end
+
+    def default_profile_attributes
+      {
+        account_id: @account.id,
+        entity_user_id: @user.id,
+        exchange_identifier: OpenStax::Exchange.create_identifier
```
jpslav: `create_identifier` is also called on line 36 of `MapUsersAccounts` -- can one of these calls go away?

app/subsystems/user_profile/create_profile.rb
```diff
@@ -1,20 +1,52 @@
...
-  end
+    protected
+    def exec(attrs:)
+      attrs = prepare_attrs_for_profile(attrs)
+      outputs[:profile] = Models::Profile.create!(attrs)
+    end
+
+    private
+    def prepare_attrs_for_profile(attrs)
+      @user = create_or_find_user(attrs)
```
jpslav: Most of your rule as Private Method King is benevolent.  This time tho I think the split up makes this routine harder to read and understand.

```ruby
def prepare_attrs_for_profile(attrs)
  prepped_attrs = {
    user_id: attrs[:entity_user_id] || (Entity::User.create!).id
    account_id: attrs[:account_id] || run(:create_account, attrs).outputs.account.id
  }

  prepped_attrs.merge!(attrs.except(accounts_attributes))

  prepped_attrs[:exchange_identifier] ||= OpenStax::Exchange.create_identifier
end
```

and getting rid of create_or_find methods.  

I don't know maybe this isn't clearer either.  Does it help to be explicit about the attributes being passed in instead of passing in some random hash?

jpslav:
Thanks @joemsak -- finished my review.

app/routines/map_users_accounts.rb
joemsak: Someone did that on master... I had it hidden behind the UP SS but sure

app/routines/map_users_accounts.rb
joemsak: I noted that as well but this was original code from before my time and it checks for errors on the following lines. Is that relevant?

app/subsystems/user_profile/create_profile.rb
joemsak: Sweet thank you for the pro tip

app/subsystems/user_profile/create_profile.rb
joemsak: MapUsersAccounts ideally would just go away completely... it's only around because the Accounts gem wants it... legacy stuff from before my time

app/subsystems/user_profile/create_profile.rb
joemsak: To be honest the lines you wrote for prepped_attrs has too much cleverness I have to stop and think about for a second... there's a lot of conditional syntax happening there..  I'll try to think of a happy compromise

joemsak:
Great feedback thanks JP, on it now

app/subsystems/user_profile/create_profile.rb
joemsak: I agree with what you're getting at though

joemsak:
Ready to go

app/routines/map_users_accounts.rb
jpslav: The original code just raises errors if there are any; express runs of routines do that already: https://github.com/lml/lev/blob/master/lib/lev/routine.rb#L205

So I think it can be changed and effectively be the same.

app/subsystems/user_profile/create_profile.rb
jpslav: Some piece of code has to be able to do the mapping from Account to User because the accounts-rails gem depends on that (and that won't be changing).  This call of `create_identifier` will often be squashed by the attributes coming into `CreateProfile` (in the hash `merge` above) -- so this will often be a wasted API call.  The alternate code I wrote below only makes the `create_identifier` call if the identifier isn't already set, tho I agree the code I wrote below isn't the best.

joemsak:
Done and done and done

jpslav:
@joemsak looks good - if you can rebase we can merge.

joemsak:
Done and done and done and done and done!

PR #103, opened 2015-03-13
Move LegacyUser subsystem into shiny new UserProfile subsystem

joemsak:
Just the first step here, renaming the subsystem, what do you guys think?

You WILL want to view it with whitespace diffs turned off https://github.com/openstax/tutor-server/pull/103/files?w=0 XD


app/access_policies/task_plan_access_policy.rb
```diff
@@ -7,18 +7,14 @@ def self.action_allowed?(action, requestor, task_plan)
       when User
         task_plan.owner == requestor
       when Entity::User
-        user = LegacyUser::FindOrCreateUserForLegacyUser.call(
-          requestor
-        ).outputs.user
+        user = UserProfile::FindOrCreate.call(requestor).outputs.user
```
jpslav: I know this is just first steps, but I imagine later this call will be to `Domain::FindOrCreateUser`?

app/access_policies/task_plan_access_policy.rb
```diff
@@ -7,18 +7,14 @@ def self.action_allowed?(action, requestor, task_plan)
...
         user == task_plan.owner
       when Entity::Course
         ## Treat failure to positively identify the user
         ## as a course teacher as a denial of access.
         ## (This happens when the user is anonymous.)
         begin
-          user = LegacyUser::FindOrCreateUserForLegacyUser.call(
-                   requestor
-                 ).outputs.user
+          user = UserProfile::FindOrCreate.call(requestor).outputs.user
```
jpslav: The `UserProfile` SS probably should not have a `User` model -- it should just have models to store profile information.  `UserProfile::Profile` would make more sense to me.  

At the end of this PR, models from the PR should not be returned outside of the SS (in general).

Later, the `requestor` should probably be an `Entity::User` (I think).  And when thats true, you won't need to use the `UserProfile` SS at all.  You'll just ask the `Domain::UserIsCourseTeacher` question.  And that domain routine should take an `Entity::User`.

app/routines/get_taskees_from_task_plan.rb
```diff
@@ -11,8 +11,8 @@ def exec(task_plan)
         tasking_plan.target
       when Entity::User
         # Yeah I know... let's just get rid of LegacyUser asap?
-        LegacyUser::User.where(entity_user_id: tasking_plan.target.id)
-                        .first.user
+        UserProfile::User.where(entity_user_id: tasking_plan.target.id)
```
jpslav: This is tricky b/c the tasking stuff isn't in a SS yet.  Maybe we're hosed until that is in a SS?  a user profile from `UserProfile` should not be a taskee.

app/subsystems/domain/find_or_create_user.rb
```diff
@@ -0,0 +1,11 @@
+class Domain::FindOrCreateUser
+  lev_routine
+
+  uses_routine UserProfile::FindOrCreate,
+               translations: {outputs: {type: :verbatim}}
+  protected
+
+  def exec(legacy_user)
```
jpslav: I actually don't know how this routine will be called later because there will be no `legacy_user`.

app/subsystems/user_profile/api/find_or_create.rb
```diff
@@ -0,0 +1,12 @@
+class UserProfile::FindOrCreate
+  lev_routine
+
+  uses_routine Entity::CreateUser
```
jpslav: unused

app/models/administrator.rb
```diff
@@ -1,5 +1,6 @@
 class Administrator < ActiveRecord::Base
-  belongs_to :user, inverse_of: :administrator
+  belongs_to :user, inverse_of: :administrator,
+                    class_name: 'UserProfile::Profile'
```
Dantemss: Should refer to Entity::User instead.

app/models/user.rb
```diff
@@ -1,57 +1,2 @@
...
-  def undelete
-    update_column(:deleted_at, nil)
-  end
-
-  # So users can be treated like roles
-  alias_method :user_id, :id
-  def user; self; end
-
-end
+require './app/subsystems/user_profile/models/profile'
```
Dantemss: Get rid of this file eventually.

app/models/educator.rb
```diff
@@ -1,5 +1,5 @@
 class Educator < ActiveRecord::Base
-  belongs_to :user
+  belongs_to :user, class_name: 'UserProfile::Profile'
```
Dantemss: Should refer to Entity::User instead.

app/models/student.rb
```diff
@@ -1,5 +1,5 @@
 class Student < ActiveRecord::Base
-  belongs_to :user
+  belongs_to :user, class_name: 'UserProfile::Profile'
```
Dantemss: Should refer to Entity::User instead.

app/models/tasking.rb
```diff
@@ -1,7 +1,7 @@
 class Tasking < ActiveRecord::Base
   belongs_to :taskee, polymorphic: true
   belongs_to :task, counter_cache: true
-  belongs_to :user
+  belongs_to :user, class_name: 'UserProfile::Profile'
```
Dantemss: Should refer to Entity::User instead. Actually, this is unclear. Maybe we just want this to refer to the role.

app/subsystems/user_profile/api/get_all_user_profiles.rb
```diff
@@ -0,0 +1,16 @@
+class UserProfile::GetAllUserProfiles
```
Dantemss: This seems very dangerous.

app/subsystems/user_profile/models/profile.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  validates :account, :entity_user, presence: true, uniqueness: true
+  validates :exchange_identifier, presence: true
+
+  delegate :username, :first_name, :last_name, :full_name, :title,
+           :name, :casual_name, :first_name=, :last_name=, :full_name=,
+           :title=, to: :account
+
+  def self.anonymous
+    ::AnonymousUser.instance
```
Dantemss: Probably wanna move AnonymousUser to this subsystem.

app/models/educator.rb
jpslav: These models are not long for life. They are replaced by roles. Don't even know if uses yet

Dantemss: That is true.

app/subsystems/user_profile/api/get_all_user_profiles.rb
joemsak: This was here before... this was mostly just renaming what already existed

app/models/user.rb
joemsak: That's definitely in the plan

app/subsystems/user_profile/api/get_all_user_profiles.rb
joemsak: What is dangerous about it, to you?

Dantemss: It iterates through the entire user table, doesn't filter based on course (so you can see every single user's profile). Without any modifications, if a view ends up getting this output, someone could dump all users in the system.

Dantemss: Also this will be veeeeery slow at some point (we already have 40,000 users... that means this routine will do at least 40 queries for 1,000 users each.

joemsak: Yeah definitely. @jpslav mentioned this in the PR that created this routine. It was made just to list users on the admin screen to pick them to assign them as teachers. We could end up using a search as you type input field or something instead

PR #105, opened 2015-03-16
mitigate effects of vcr and webmock gems on http requests in specs that do not use them

kevinburleigh75:
At some fairly recent point in the past, the behavior of network connectivity inside specs changed, even when those specs don't use or care about `webmock` or `vcr`.  This introduced an order dependency between specs, with those following any spec that used `webmock` or `vcr` failing.

This PR allows specs to access the network as long as they don't have any `vcr` cassettes, which I believe fixes the problem while still preserving desired behavior.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -37,6 +37,9 @@ Rails.application.configure do
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
 
+  # Don't error out when trying to connect to external sites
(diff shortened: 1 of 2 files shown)


joemsak:
Can't say I have much say in the goodness or badness of this idea, so I'll trust the team. Thanks Kevin!

joemsak:
To be fair I ended up having much to say in Skype :P

PR #109, opened 2015-03-16
Try another

Dantemss:
- Recovery API's done.
- Recovery exercises extracted from the content.
- rake sprint:008 task creates 3 assignments based on the sample content
- PUT now returns the modified content


Dantemss:
Currently breaks the UI. Phil thinks it might be a UI bug. We'll check.

philschatz:
The UI bug might be fixed with <https://github.com/openstax/tutor-js/pull/59> but unsure since I don't know what the UI glitch is. Can you describe it to @jivewise?

Dantemss:
Old glitch fixed. New glitch: at some point the UI calls GET "/api/steps/[object%20Object]" which returns a 404 and everything breaks.

PR #110, opened 2015-03-16
add /courses/123/tasks endpoint

jpslav:
Adds a `/api/courses/123/tasks` endpoint.  For the time being (and for the convenience of the frontend), this endpoint returns exactly what `/api/user/tasks` returns.  Once the backend gets its role infrastructure in better shape, we'll update this endpoint to actually return tasks for a certain course (and for a certain role).

@Dantemss please review

/cc @philschatz


joemsak:
:+1:

PR #112, opened 2015-03-18
/api/v1/courses

joemsak:
Static roles content right now but this is coming along

```
[3/17/15, 5:37:30 PM] JP Slavinsky: your next job is to make 
a /api/courses endpoint that looks roughly like: 
https://github.com/openstax/napkin-notes/blob/master/jp/my_courses_api.md

[3/17/15, 5:38:18 PM] JP Slavinsky: the content in there is not great, 
basically the listing should return all courses and segregate out those 
that the user is a part of and for those show information about the roles. 
 If the roles stuff isn't clear yet, we can add that later and use your first 
pass as a discussion point with FE
```
https://github.com/openstax/napkin-notes/blob/master/jp/my_courses_api.md

Files diff without whitespace https://github.com/openstax/tutor-server/pull/112/files?w=0

# Rake Tasks
```
rake sprint:008courses
```
Adds courses, users, and assigns them as students, teachers, or both. 
Usernames and course names are displayed by the task's output.

When running the rake task, do `rake db:drop db:migrate sprint:008courses` to make sure the database is clean first.  After logging in as one of the users, `GET /api/courses`.


app/subsystems/domain/list_courses.rb
```diff
@@ -7,14 +7,46 @@ class Domain::ListCourses
...
                translations: { outputs: { type: :verbatim } },
                as: :get_teacher_names
+  uses_routine Domain::GetUserCourseRoles,
+               translations: { outputs: { type: :verbatim } },
+               as: :get_course_roles
 
   protected
 
-  def exec
+  def exec(user: nil, options: {})
```
jpslav: instead of a generic `options`, how about `exec(user: nil, with: nil)`?

app/subsystems/domain/list_courses.rb
```diff
@@ -7,14 +7,46 @@ class Domain::ListCourses
...
     run(:get_profiles)
+    case options[:with]
```
jpslav: might we need to include multiple `with`s?  e.g.

```
[with].flatten.each do |with_item|
  case with_item
  when :teacher_names
  ...
end
```

app/subsystems/domain/list_courses.rb
```diff
@@ -7,14 +7,46 @@ class Domain::ListCourses
...
+
+  def set_roles_on_courses(user)
     outputs.courses.each do |course|
-      teacher_names = run(:get_teacher_names, course.course_id).outputs.teacher_names
-      course.teacher_names = teacher_names
+      roles = get_roles(course, user)
+      course.roles = roles.collect do |role|
+        {
+          type: role.role_type,
+          url: "/api/v1/#{role.role_type.pluralize}/#{role.id}/dashboard"
```
jpslav: Instead of setting a URL here, just store the critical information, i.e. the `role.id`.  Leave it up to the representer to format the URL.

And in the representer, it'd be nice to have three fields per role:

1. the role `type` as you have it
2. the role `id`
3. the `course_url`, which should be `"/api/courses/#{course.id}/role/#{role.id}"`

The role type isn't expected in the `course_url` -- FE folks will just remember the type elsewhere in memory.  The `course_url` will be an easy way for the FE to build links to the course.  @philschatz might want something else later for this URL field, but we can adjust later easily when they know more.

app/subsystems/entity/api/create_role.rb
```diff
@@ -3,9 +3,9 @@ class Entity::CreateRole
 
   protected
 
-  def exec
-    role = Entity::Role.create
-    transfer_errors_from(role, {type: :verbatim}, true)
+  def exec(role_type = :unassigned)
```
jpslav: For the time being until we know we have a need for an `:unassigned` type, can we remove the default value?

`def exec(role_type:)`

app/subsystems/entity/api/create_role.rb
```diff
@@ -3,9 +3,9 @@ class Entity::CreateRole
...
+    role = Entity::Role.create(role_type: role_type)
```
jpslav: Does calling `create!` remove the need to call `transfer_errors_from`, since any validation errors will be made into exceptions that lev no longer swallows.  This would be ok because we're not doing anything with user input here so any errors are going to be truly exceptional.

app/subsystems/entity/models/role.rb
```diff
@@ -1,2 +1,3 @@
 class Entity::Role < ActiveRecord::Base
+  enum role_type: [:unassigned, :teacher, :student]
```
jpslav: nice

app/subsystems/role/api/create_user_role.rb
```diff
@@ -6,8 +6,8 @@ class Role::CreateUserRole
 
   protected
 
-  def exec(user)
-    run(Entity::CreateRole)
+  def exec(user, role_type = :unassigned)
```
jpslav: ditto on removing default.

db/migrate/20150320133010_add_type_to_entity_roles.rb
```diff
@@ -0,0 +1,6 @@
+class AddTypeToEntityRoles < ActiveRecord::Migration
+  def change
+    add_column :entity_roles, :role_type, :integer, null: false, default: 0
```
jpslav: Was about to say can we remove this default too, BUT.... I just realized maybe we want it after all.  Would be nice to be able to instantiate an Entity::Role without defining a `role_type` in tests.  So maybe nice to have it but pretty much everywhere in real code we're going to want to set it to something.

spec/controllers/api/v1/courses_controller_spec.rb
```diff
@@ -96,4 +96,40 @@
     end
   end
 
+  describe "index" do
```
jpslav: Can you add a test when one user is both a teacher and a student in a course?

app/subsystems/domain/list_courses.rb
joemsak: That's fine as long as we're not anticipating other options in the future... perhaps I should YAGNI that though?

app/subsystems/entity/api/create_role.rb
joemsak: Not sure, the transfer_errors_from bit was not my code

app/subsystems/domain/list_courses.rb
joemsak: :+1:

db/migrate/20150320133010_add_type_to_entity_roles.rb
joemsak: :+1:

spec/controllers/api/v1/courses_controller_spec.rb
joemsak: :+1:

app/subsystems/domain/list_courses.rb
```diff
@@ -7,14 +7,53 @@ class Domain::ListCourses
...
   end
+
+  def set_roles_on_courses(user)
+    outputs.courses.each do |course|
+      roles = get_roles(course, user)
+      course.roles = roles.collect do |role|
+        {
+          id: role.id,
+          type: role.role_type,
+          course_url: "/api/courses/#{course.course_id}/role/#{role.id}"
```
jpslav: thanks for changing the text of the `course_url` but I'd like this field to just be in the representer (since it is derivable from all of the other data in the `course` hash and only used by an API client).  i.e., this has should only have lines 46 and 47

app/subsystems/domain/list_courses.rb
jpslav: even if we have other options, we can just add other named arguments

philschatz:
Cool! Can you post sample JSON?

joemsak:
```json
[
  {
    id: 1,
    name: 'Course Name',
    roles: [        // if the user has roles in the course
      {  id: 1, type: 'student' }, // ... etc
    ]
  }
]
```

PR #114, opened 2015-03-19
Video step

karenc:
- Create video fragments from a cnx page
    
    In an assessed feature, there is either a video or some text plus an
    exercise.  Add a video fragment class that inherits from the text
    fragment class but with a url method that returns the url of the video
    from .ost-embed. Code assumes only one video in each fragment.

- Add tasked video model and representer

### DIFF
--- a/app/representers/api/v1/tasked_video_representer.rb
+++ b/app/representers/api/v1/tasked_video_representer.rb
@@ -1,0 +1,13 @@
+module Api::V1
+  class TaskedVideoRepresenter < TaskedReadingRepresenter
+
+    property :video_url,
+             type: String,
+             writeable: false,
+             readable: true,
+             schema_info: {
+               required: false,
+               description: "The video URL"
+             }
+  end
+end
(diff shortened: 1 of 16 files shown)


app/representers/api/v1/tasked_video_representer.rb
```diff
@@ -0,0 +1,16 @@
+module Api::V1
+  class TaskedVideoRepresenter < TaskedReadingRepresenter
+
+    include TaskStepProperties
+
+    property :video_url,
+             type: String,
+             writeable: false,
+             readable: true,
+             as: :video_url,
```
jpslav: I don't think you need this since it is the same name as the `property`?

karenc: Oh i see, thanks, I'll remove it

philschatz:
are these video links `<a class="ost-video" href="...">` or in a `<FOO data-type="media" class="ost-video"><iframe src="..."/></FOO>` (hoping it's the latter ; )

karenc:
It was really just a link to the video, not embedded but we have youtube videos now, so I'm going to change the code.

PR #115, opened 2015-03-19
Calculate data for IReading stats

nathanstitt:


nathanstitt:
This adds a page_id attribute to the task_step table so that it can calculate student progress on per-page basis, as outlined in the spec for the feature.

The stats are calculated fairly crudely and there's definitely room for improvement in the algorithm's used.  I wanted to get this version out for review early though so that we can evaluate if I'm even calculating the proper values.

app/routines/calculate_i_reading_stats.rb
```diff
@@ -4,42 +4,93 @@ class CalculateIReadingStats
...
+
+  def page_stats_for_steps(steps)
+    user_ids = steps.each_with_object([]){ |step, collection|
+      step.task.taskings.each{|tasking| collection << tasking.user_id }
+    }.uniq
+    completed = steps.select {|ts| ts.completed? }
+    stats = {
+      student_count: user_ids.length,
+      correct_count: completed.count{|step| step.tasked.correct_answer_id == step.tasked.answer_id }
+    }
```
nathanstitt: Is there a better method to determine if the question was answered correctly?

philschatz: That's what I've been using so far but if there's a bettwe way (like `is_correct`) I'm all ears ; )

app/controllers/api/v1/task_plans_controller.rb
```diff
@@ -61,7 +61,7 @@ class Api::V1::TaskPlansController < Api::V1::ApiController
   def show
     plan = TaskPlan.find(params[:id])
     OSU::AccessPolicy.require_action_allowed!(:stats, current_api_user, plan)
```
jpslav: This should be `:show` (or `:read`) instead of `:stats`.  The stats are a side show for this action which is about reading plans.

app/routines/calculate_i_reading_stats.rb
```diff
@@ -4,42 +4,93 @@ class CalculateIReadingStats
 
   protected
 
-  def generate_page_data( min:10, include_previous: false )
-    stat = {
+  def completed_exercises_for_page_id(page_id)
+    @plan.tasks.inject([]) do |collection,task|
+      collection + task.task_steps.find_all{|ts|
```
jpslav: how is this counting completed exercises?

PR #117, opened 2015-03-19
Fix sprint 007a task

karenc:
Fix html parsing to allow assessed feature without exercises

Running "rake sprint:007a" was giving the following error:

```
rake aborted!
NoMethodError: undefined method `parent' for nil:NilClass
/home/karen/tutor-server/lib/html_tree_operations.rb:7:in `recursive_compact'
/home/karen/tutor-server/app/external/openstax/cnx/v1/page.rb:132:in `fragments'
```

because we just added this page
http://cnx.org/contents/7db9aa72-f815-4c3b-9cb6-d50cf5318b58@4.5:2/Updated_Tutor_HS_Physics_Conte#fs-id1172194442505
which has an assessed feature with just a video without an exercise.

------

Also:
 - added code to parse the video url from iframe and
 - removed video_url from tasked_video.

### DIFF
--- a/app/assistants/i_reading_assistant.rb
+++ b/app/assistants/i_reading_assistant.rb
@@ -64,9 +64,8 @@ class IReadingAssistant
             raise "Video url not found for Video in Page #{page.id}" \
               if fragment.url.blank?
 
-            TaskedVideo.new(task_step: step, url: page.url,
-                            title: fragment.title, content: fragment.to_html,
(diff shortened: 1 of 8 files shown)


Dantemss:
Seems odd to me to have an assessed feature without an exercise?

karenc:
If that's not valid then I guess we should fix the content.

Dantemss:
@philschatz ?

PR #113, opened 2015-03-19
Practice Widget

jpslav:
Adds API endpoints to access a practice widget with faked exercises.  I've added a rake task to set up a scenario to test the practice widget.  

After checking out this branch, `jpslav-s8-practice-widget`, run the following once:

`bundle install --without production`

Then every time you want to reset the practice widget scenario state, run:

`bundle exec rake db:drop db:migrate sprint:008pw`

The `sprint:008pw` task creates a user with username "student" and a course, then registers the "student" user as a student in the course.  It then creates an initial practice widget.  I did this so you could test just using the practice widget before needing to create a new one.  

The last line output from running the command above will look something like this:

```
Added 'student' user with a practice widget with 5 fake exercises.  Access it with GET /api/courses/1/practice.  Create a new one with a POST to the same URL.
```

As you can see, it gives you the endpoint to use.

Just like normal, the process after running the rake tasks is to start the server:

`bundle exec rails server`

Then go to `http://localhost:3001/accounts`, hit the "search" button, then click on the "student" username.  This will log you in.  If you navigate manually to `http://localhost:3001/api/courses/1/practice` you'll see the practice widget JSON (it looks exactly like a normal Task).

/cc @philschatz @pandafulmanda

PR #120, opened 2015-03-20
add /courses/123/events endpoint

nathanstitt:
Contains a consolidated listing of both plans and task events.


app/models/task.rb
```diff
@@ -27,4 +27,9 @@ def course
   def tasked_to?(taskee)
     taskings.where(taskee: taskee).any?
   end
+
+  def completed?
+    self.task_steps.all?{|ts| ts.completed? }
+  end
```
nathanstitt: **Note:**  I'm adding logic to the model here. I suspect it should go somewhere else, but we don't have a Tasks subsystem yet.  This method will also be handy to use in the IReading stats PR.

app/representers/api/v1/course_events_representer.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    class PlanElements < CommonElements
+
+      property :trouble,
+               type: :boolean,
+               readable: true,
+               getter: lambda{|*| rand(0..1)==0 }
+        # ^^^^^ REPLACE with real value once spec for calculating it's available
+
+      property :type,
```
nathanstitt: I don't think we have an algorithmic definition of what `trouble` should be indicate yet.

app/representers/api/v1/course_events_representer.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    end
+
+    class PlanElements < CommonElements
+
+      property :trouble,
+               type: :boolean,
+               readable: true,
+               getter: lambda{|*| rand(0..1)==0 }
+        # ^^^^^ REPLACE with real value once spec for calculating it's available
```
nathanstitt: I don't think we have an algorithmic definition of what trouble should be indicate yet.

app/models/task.rb
jpslav: we can have an effort later to find an appropriate home for this code.

app/controllers/api/v1/courses_controller.rb
```diff
@@ -56,4 +56,17 @@ def tasks
...
   end
 
+  api :GET, '/courses/:course_id/events', 'Gets all events for a given course'
+  description <<-EOS
+    #{json_schema(Api::V1::CourseEventsRepresenter, include: :readable)}
+  EOS
+  def events
+    course = Entity::Course.find(params[:id])
+
+    OSU::AccessPolicy.require_action_allowed!(:task_plans, current_api_user, course)
```
jpslav: since a student can call this endpoint as well, we don't want this access restriction.  In a student's case, we just want to return either `plans: []` or not have plans in the JSON at all.   You're right that somewhere we need to check that the user can get these plans, but I think it is not here.

app/representers/api/v1/course_events_representer.rb
jpslav: I think someone said it is less than 50% full bar graph on any of the graphs on the reading popup.

app/routines/get_course_events.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  lev_routine
+
+  uses_routine GetCourseTaskPlans, as: :get_plans
+  uses_routine SearchTasks,        as: :get_tasks
+
+  protected
+
+  def exec(course:, user:)
+    outputs[:plans] = run(:get_plans, course: course).outputs.items
```
jpslav: instead of the access policy check above re task plans, this line needs to be different somehow to take into account the user's permissions.

Since this routine takes a `course` and a `user`, maybe it is `GetUserCourseEvents`?  there are some similarly-named routines around, check to see if there's a growing convention to follow.

Actually, these should probably be role-based (at least eventually, maybe not now).  Someone is going to visit /courses/123/events and they'll either pass in a role (e.g. /courses/123/role/234) or we'll compute a default one for them like in my practice widget endpoint.  The role is the thing for which we should be returning plans and tasks.  So you might consider having this be a GetRoleCourseEvents with `exec(course:, role:)`.  For the moment, you can do some hacks like calling the Role subsystem to get the user so you can get the tasks, and something similar for getting the plans.  But in the end, the input to this routine will be a role, not a user.

config/routes.rb
```diff
@@ -38,9 +38,12 @@
     end
 
     resources :courses, only: [] do
-      get 'readings', on: :member
-      get 'plans', on: :member
-      get 'tasks', on: :member
+      member do
```
jpslav: beautiful

jpslav:
@nathanstitt done with my review.  Can you also make a rake task (wrapping a routine) that sets up the database to show off the events api to the FE folks.  Check out what I did in https://github.com/openstax/tutor-server/pull/113/files

nathanstitt:
@jpslav I've now removed the access restriction on the controller, have renamed the routine to be `GetUserCourseEvents` and it now checks the roles that the user has for the course and will only return the task_plans if they're a teacher for it.

I'm currently returning all the tasks for a user though, since I'm picturing the data on the calendar and it would seem like you'd want to see all your tasks, not just the ones relating to a given course.  If that's not the case I can filter them by the course as well.

I started writing a rake task and was ending up duplicating all your work on the `sprint:008courses` task, and only adding a few lines.  Instead I simply extended the `sprint:008courses` task to also create some tasks and plans when it's ran.

jpslav:
This endpoint should only return events for the course in question, but I don't think we're there yet because the course / role / task stuff is evolving.  So for now it should just return all user tasks.  Later when a user is looking at their calendar they will only see info for one course at a time (at least for MVP).

Sounds good on beefing up the existing sprint 8 courses task.

Dantemss:
Feel free to modify (add to) the 008 main task too. It's nowhere near done or anything.

PR #122, opened 2015-03-23
Continue user profile subsystem cleanup

joemsak:
Continuing forward on User Profile subsystem stuff.. trying to make sure Domain is used where needed, trying to clean out un-used classes, light refactoring in files touched during the work

* Remove User
* Move `Administrator` to `UserProfile::Administrator`
* Move `UserMapper` to `Domain::MapUsersAccounts`
* Rename `user_id` to `profile_id` on `user_profile_administrators`
* Consolidation of where the system should create a user: `Domain::CreateUserProfile`
* Rename `UserProfile::FindOrCreate` to `UserProfile::CreateProfile` because it does not do any finding and the system doesn't need that now that user subsystem is consolidating things
* Refactored lines of code inside `TaskPlanAccessPolicy` and `Domain::MapUsersAccounts` because either tests were added or the file happened to be opened during changes being made and I wanted it to be clearer to read the logic :smile_cat:


app/access_policies/course_access_policy.rb
```diff
@@ -2,11 +2,9 @@ class CourseAccessPolicy
   def self.action_allowed?(action, requestor, course)
     case action
     when :readings
-      requestor.is_human? # && 
-      # Domain::UserIsCourseTeacher.call(user: requestor.human_user,
```
jpslav: We need to have this check in here (know it is commented out, but really only teachers should be allowed to do this).

jpslav: actually I take that back.  Anyone can query readings.

app/access_policies/task_plan_access_policy.rb
```diff
@@ -2,24 +2,10 @@ class TaskPlanAccessPolicy
...
-        ## (This happens when the user is anonymous.)
-        begin
-          Domain::UserIsCourseTeacher.call(user: requestor.entity_user,
-                                           course: task_plan.owner)
-                                     .outputs.user_is_course_teacher
-        rescue
-          false
-        end
+      if task_plan.owner.is_a?(Entity::Course)
+        user_is_course_teacher?(requestor.entity_user, task_plan.owner)
```
jpslav: Put the call to `UserIsCourseTeacher` directly here with the bracket notation.

app/routines/create_user.rb
```diff
@@ -1,18 +0,0 @@
...
-
-  lev_routine
-
-  protected
-
-  def exec(account)
-    outputs[:user] = User.create do |user|
-      user.entity_user = Entity::CreateUser.call.outputs.user
-      user.account = account
-      user.exchange_identifier = OpenStax::Exchange.create_identifier
```
jpslav: Where did this exchange code go?

joemsak: Oh I see.... I was just trying to consolidate to one CreateUser notion which I think assumes you have this stuff taken care of already

Some of my cleanup here might be too much

app/subsystems/user_profile/models/profile.rb
```diff
@@ -6,7 +6,7 @@ class UserProfile::Profile < ActiveRecord::Base
   has_many :groups_as_owner, through: :account
 
   has_one :administrator, dependent: :destroy, inverse_of: :user,
-    foreign_key: :user_id
+    foreign_key: :user_id, class_name: 'UserProfile::Administrator'
```
jpslav: should the foreign key from `Administrator` to `Profile` really be `user_id`?  Seems odd.

app/subsystems/user_profile/models/user.rb
```diff
@@ -1,6 +1,5 @@
 class UserProfile::User < ActiveRecord::Base
   ## using class_name as workaround, see: https://github.com/rails/rails/issues/15811
-  belongs_to :user, class_name: "::User"
   belongs_to :entity_user, class_name: "::Entity::User"
```
jpslav: `belongs_to :user, subsystem: :entity`

jpslav: make sure `UserProfile` whitelisted in list of subsystems in config/intializers/subsystems.rb

spec/routines/create_user_spec.rb
```diff
@@ -1,66 +0,0 @@
...
-  let!(:account) { FactoryGirl.create(:openstax_accounts_account) }
-
-  before(:each) do
-    OpenStax::Exchange::FakeClient.configure do |config|
-      config.registered_platforms   = {
-        OpenStax::Exchange.configuration.client_platform_id =>
-        OpenStax::Exchange.configuration.client_platform_secret
-      }
-      config.server_url             = OpenStax::Exchange.configuration.client_server_url
-      config.supported_api_versions = [OpenStax::Exchange.configuration.client_api_version]
```
jpslav: The exchange stuff in this spec needs to be maintained somewhere.

jpslav:
@joemsak, looking good.  I made comments.  Can you post here what the current big-picture status is?  (Laying out what has been done and what is left to be done?)

joemsak:
Let me get back to you on what has been done after responding to the requests in the PR

As for what is left to be done... I'm thinking this is more... as I interact with and learn more about the codebase and how it works, seeing where things can be fixed up along the way. It's more of an open-ended discovery than a set plan.

app/subsystems/user_profile/models/user.rb
joemsak: :+1:

app/subsystems/user_profile/models/profile.rb
joemsak: :+1:

spec/routines/create_user_spec.rb
joemsak: In progress

app/subsystems/user_profile/models/user.rb
joemsak: Actually this is causing me grief, see the backend skype chat

app/access_policies/course_access_policy.rb
```diff
@@ -2,9 +2,7 @@ class CourseAccessPolicy
   def self.action_allowed?(action, requestor, course)
     case action
     when :readings
-      requestor.is_human? # &&
-      # Domain::UserIsCourseTeacher.call(user: requestor.human_user,
```
Dantemss: Well we will still need this check (or something similar) in the future.

app/access_policies/task_plan_access_policy.rb
```diff
@@ -2,24 +2,11 @@ class TaskPlanAccessPolicy
...
-        begin
-          Domain::UserIsCourseTeacher.call(user: requestor.entity_user,
-                                           course: task_plan.owner)
-                                     .outputs.user_is_course_teacher
-        rescue
-          false
-        end
+      if task_plan.owner.is_a?(Entity::Models::Course)
+        Domain::UserIsCourseTeacher[user: requestor.entity_user,
+                                    course: task_plan.owner] rescue false
```
Dantemss: Not a big fan of rescuing exceptions unless absolutely needed, but sure I guess it works...

app/access_policies/course_access_policy.rb
joemsak: screen shot 2015-03-27 at 2 53 05 pm

Also I mean... commented code? If you need it in the future you can add it then : )

app/access_policies/task_plan_access_policy.rb
joemsak: rescuing was already there, not something I added... failing tests forced me to keep it there without going into the rabbit hole to find the 'real' reason for it

app/access_policies/task_plan_access_policy.rb
```diff
@@ -27,4 +14,10 @@ def self.action_allowed?(action, requestor, task_plan)
       false
     end
   end
+
+  private
+  def self.requestor_is_task_plan_owner?(requestor, owner)
```
Dantemss: Just a minor thing, but this method name is more specific than the actual method code (I would name it something like `users_match?`

Also hopefully in the future we will have better rules about who can own TaskPlans, so checking both `requestor` and `requestor.entity_user` will not be needed.

app/subsystems/user_profile/api/create_profile.rb
```diff
@@ -0,0 +1,22 @@
+class UserProfile::Api::CreateProfile
```
Dantemss: No more Api.

joemsak: Not in the rebase, look at the file path... tests failed without this change

app/subsystems/user_profile/api/create_profile.rb
```diff
@@ -0,0 +1,22 @@
...
+  lev_routine
+
+  uses_routine Entity::CreateUser, translations: { outputs: { type: :verbatim } }
```
Dantemss: No more entity routines. Just use Entity::User.create.

app/access_policies/task_plan_access_policy.rb
joemsak: It's meant to help you read the access rule it's checking... not meant to be abstracted for generic use. The point here is twofold:
1. it's the result of a path taken in a conditional. I prefer a single readable line when I have branching code.
2. The actual conditional checked at line 20 does not read as nicely as line 9 does. A generic name like `users_match` isn't as straightforward and immediate as 'the requestor can access the task plan if they're the owner'

app/subsystems/user_profile/api/create_profile.rb
joemsak: Specifically... after rebasing on nathan's changes is when this wouldn't work until I added the 'api' part

joemsak: I think what happened is Nathan would have moved this file but it doesn't exist on master.. it's a move/rename of FindOrCreate

app/access_policies/course_access_policy.rb
Dantemss: Anyone can query readings? OK then.

app/access_policies/task_plan_access_policy.rb
Dantemss: It's somewhat misleading though... if I see a method called `requestor_is_task_plan_owner?`, I'm tempted to think that the arguments should be the `requestor` and the `task_plan`, not the `task_plan.owner`

spec/routines/create_user_spec.rb
Dantemss: Did this go somewhere in the end?

Dantemss: Actually nvm idk why this configuration is even here... should prob be a global thing.

app/access_policies/task_plan_access_policy.rb
joemsak: It's a private method. Private methods are useful for increasing readability and developers know not to use them. The function signature clearly defines the argument as the owner

spec/routines/create_user_spec.rb
joemsak: Yes, it did it's in MapUsersAccounts

config/initializers/openstax_accounts.rb
```diff
@@ -11,7 +11,8 @@
   config.openstax_accounts_url = secrets['url']
   config.enable_stubbing = stub
   config.logout_via = :delete
-  config.account_user_mapper = UserMapper
+  config.account_user_mapper = Domain::MapUsersAccounts
+  config.enable_stubbing = true
```
Dantemss: This is probably my fault, but you now have 2 `enable_stubbing` here.

config/initializers/access_policies.rb
```diff
@@ -1,7 +1,9 @@
-OSU::AccessPolicy.register(User, UserAccessPolicy)
+require './app/subsystems/user_profile/models/profile'
```
Dantemss: Any idea why this require ended up being necessary?

joemsak: Didn't look into it more, no

nathanstitt: I'm pretty sure it was needed in the past to make sure that the model was loaded in case the subsystems weren't initialized yet.  Since the subsystem initializer started with "s" it would have came after the "a"ccess_policies.

I've fixed that by renaming the subsystem's initializer file to 01-subsystems.  It can most likely be removed.

app/subsystems/user_profile/models/administrator.rb
```diff
@@ -0,0 +1,9 @@
+class UserProfile::Models::Administrator < ActiveRecord::Base
+  belongs_to :profile, inverse_of: :administrator, foreign_key: :profile_id,
```
Dantemss: `foreign_key` shouldn't be necessary (`:profile_id` should be the default for `belongs_to :profile`)

config/initializers/access_policies.rb
```diff
@@ -1,7 +1,9 @@
...
+
+OSU::AccessPolicy.register(UserProfile::Models::Profile, UserAccessPolicy)
 OSU::AccessPolicy.register(Task, TaskAccessPolicy)
 OSU::AccessPolicy.register(TaskPlan, TaskPlanAccessPolicy)
 OSU::AccessPolicy.register(TaskedExercise, TaskedAccessPolicy)
 OSU::AccessPolicy.register(TaskedReading, TaskedAccessPolicy)
 OSU::AccessPolicy.register(TaskedVideo, TaskedAccessPolicy)
-OSU::AccessPolicy.register(Entity::Course, CourseAccessPolicy)
+OSU::AccessPolicy.register(Entity::Models::Course, CourseAccessPolicy)
```
nathanstitt: Good catch there.  I investigated this and suspect that it didn't blow up because the access_policies/course_access_policy_spec.rb is "pending".  What it was really checking was the Entity::Models module, which is meaningless.

PR #132, opened 2015-03-25
Video step

karenc:
There are 3 commits in this PR:

- Add tests for tasked representer mapper

Add tests for checking that all the Tasked* models and representers are
indeed in the representer mapper.

- Change TaskedVideoRepresenter to not inherit from TaskedReadingRepresenter

jpslav said he doesn't want a coupling between it and a reading step.

- Extract content for video fragment without the video tags

Remove the A tag if the video is a link or the media tag if the video is
embedded from the html content.


karenc:
@philschatz in this PR I'm removing the A tag if the video is a link or the media tag if the video is embedded from the html content.  Is this still what we want?

app/external/openstax/cnx/v1/fragment/video.rb
```diff
@@ -1,16 +1,51 @@
 module OpenStax::Cnx::V1::Fragment
   class Video < Text
```
jpslav: My concern is this line -- I don't want `Video` to have such a strong dependence on `Text`.  Can they be separate?

jpslav: actually I guess this concern is just a parallel of my prior concern about the representers being linked.

karenc: Yes I can separate them.

app/external/openstax/cnx/v1/fragment/video.rb
```diff
@@ -1,16 +1,77 @@
...
+      #   </div>
+      @video_type ||= case
+                      when video.try(:name) == 'a'
+                        :link
+                      when video.try(:attr, :'data-type') == 'media'
+                        :embedded
+                      end
+    end
+
+    def to_html
```
Dantemss: Wouldn't `node.video.to_html` be enough here?

app/external/openstax/cnx/v1/fragment/video.rb
```diff
@@ -1,16 +1,77 @@
...
+
+    def to_s(indent: 0)
+      s = "#{' ' * indent}VIDEO #{title}\n"
+    end
+
+    def video(node = node)
+      @video ||= node.at_css(VIDEO_OBJECT_CSS)
+    end
+
+    def video_type
```
Dantemss: According to Ed this might not be necessary, as W&N is supposed to embed all videos.
Did the sample content contain both kinds of videos?

app/external/openstax/cnx/v1/fragment/video.rb
Dantemss: Ah I see what you are doing now, but this probably won't work anyway for the `<a>` tags right? You will just get the "video" text.

Also instead of copying, you could instead use something like `video.at_css('iframe').to_html`.

Dantemss:
Couple of comments, should be good to merge after they are addressed.

app/external/openstax/cnx/v1/fragment/video.rb
Dantemss: Or even better, `video.to_html`.

karenc: No that's what we want.  We want to get the "video" text without the link.  And then embed the video at the end of the page (or something).

app/external/openstax/cnx/v1/fragment/video.rb
karenc: The original one was like this: https://github.com/openstax/tagging-legend#video

app/external/openstax/cnx/v1/fragment/video.rb
karenc: What the to_html returns is all the note html with `<p>This <a class="os-embed" href="https://youtube.com/watch?askjdh">video</a> reviews vectors...</p>` replaced with `<p>This video reviews vectors...</p>`.

karenc: phil said something like this:

```
{ content_html: “all the contents of the note minus the exercise and the iframe”,
  content_url: “<a href="https://khanacademy.org/embed/watch…">https://khanacademy.org/embed/watch…</a>.”
}
```

I'm not sure if I've removed the exercise.. should probably check that.

Dantemss: Ah ok I see what you mean now. It's fine.

PR #133, opened 2015-03-25
Remove trailing whitespace and add newline to EOF in *.rb

karenc:
The commands I used are (on ubuntu):

```
find -name '*.rb' | xargs sed -i 's/ \+$//'
find -name '*.rb' | xargs sed -i -e '$a\'
```


joemsak:
:+1:

PR #135, opened 2015-03-26
Secrets generator redo

Dantemss:
LN says we shouldn't load Rails to create a file required for Rails to work so...

- Removed rails g secrets
- Added rake task that creates secrets.yml from example file and runs automatically before all rake tasks
- Changed secrets.yml format to nest related parameters, so everyone will need to recreate their secrets.yml


Dantemss:
Will need changes to deployment to account for the lack of rails g secrets and the new secrets.yml format.

config/initializers/openstax_exchange.rb
```diff
@@ -1,19 +1,21 @@
+secrets = Rails.application.secrets['openstax']['accounts']
```
lakshmivyas: should this be ['openstax']['exchange']?

config/environments/production.rb
```diff
@@ -85,18 +84,6 @@
...
-  if Rails.application.secrets[:fake_exchange]
-    # Don't try to connect to Exchange
-    OpenStax::Exchange.use_fake_client
-    OpenStax::Exchange.reset!
-  end
-
-  if Rails.application.secrets[:fake_exercises]
-    # Don't try to connect to Exercises
-    OpenStax::Exercises::V1.use_fake_client
-  end
```
lakshmivyas: No need to use fake client anymore?

lakshmivyas: nvm.  I see that there is a `stub` property below.

config/initializers/openstax_exchange.rb
Dantemss: Thanks, LN. Changed.

Issue #125, opened 2015-03-24
Replace implementation of GetBookToc with VisitBook


jpslav:
`Content::Api::VisitBook` is a new routine for traversing a book structure and reporting back information.  I wrote it to get exercise and tag information out of a book, but to test it at first I wrote a visitor to compute the table of contents like how we do in `GetBookToc`.  

To simplify code, replace `GetBookToc` with this new `VisitBook` routine.

Dantemss:
Thought you hated visitors :D

Dantemss:
All done.

PR #136, opened 2015-03-26
Replace get book toc with visit book

joemsak:
See #125


Dantemss:
I like how this replaces 72 lines with 1, can't believe it was that simple :)
But I'm still thinking of moving this ToC code to the import book routine (and caching it in the book content).

PR #137, opened 2015-03-26
Return tasked interactives in IReading

karenc:
- Create interactive fragments from cnx page
- Add TaskedInteractive model and database table
- Create TaskedInteractive in IReadingAssistant
- Create TaskedInteractiveRepresenter and add it to the tasked
  representer mapper

### DIFF
--- a/app/assistants/i_reading_assistant.rb
+++ b/app/assistants/i_reading_assistant.rb
@@ -38,7 +38,7 @@ class IReadingAssistant
   def self.tasked_exercise(exercise_fragment:, recovery_fragment: nil, step: nil)
     if exercise_fragment.embed_tag.blank?
       logger.warn "Exercise without embed tag found while creating iReading"
-      return nil
+      return
     end
 
     # Search local (cached) Exercises for one matching the embed tag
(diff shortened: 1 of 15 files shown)


app/assistants/i_reading_assistant.rb
```diff
@@ -70,6 +70,18 @@ def self.tasked_video(video_fragment:, step: nil)
                     content: video_fragment.to_html)
   end
 
+  def self.tasked_interactive(interactive_fragment:, step: nil)
+    if interactive_fragment.url.blank?
+      logger.warn('Interactive without iframe found while creating iReading')
+      return nil
```
joemsak: FWIW 'nil' is implied but I'm willing to accept that I'm being nitpicky

app/external/openstax/cnx/v1/fragment/interactive.rb
```diff
@@ -0,0 +1,54 @@
...
+      @node  = node
+      @title = title
+    end
+
+    attr_reader :node
+
+    def title
+      return @title unless @title.nil?
+
+      @title = node.css(TITLE_CSS).collect{|n| n.try(:content).try(:strip)}
```
Dantemss: Should hopefully only have 1 title for sims, so `node.at_css(TITLE_CSS).try(:content).try(:strip)` should be enough... not that title matters too much anyway, since Phil ignores them.

app/external/openstax/cnx/v1/fragment/interactive.rb
```diff
@@ -0,0 +1,54 @@
...
+      @title = DEFAULT_TITLE if @title.blank?
+      @title
+    end
+
+    def interactive(node = node)
+      @interactive ||= node.at_css(INTERACTIVE_CSS)
+
+    end
+
+    def to_html
```
Dantemss: Just like in the video steps, I think you can do `interactive.inner_html` here?

app/assistants/i_reading_assistant.rb
karenc: Well actually I prefer "return", but I saw "return nil" in other places so I thought I should follow that ;D

joemsak: Let us tar and feather the assailants

app/external/openstax/cnx/v1/fragment/interactive.rb
karenc: I just copied this from the Text fragment, since jp prefers to have the classes separate.  Should I just remove the title?

Dantemss: Could copy it from the Exercise fragment instead :)

app/external/openstax/cnx/v1/fragment/interactive.rb
karenc: Similar to the video fragment, `Interactive.to_html` returns the note without the media tag

PR #139, opened 2015-03-27
Move to Postgresql for development and testing

nathanstitt:
Is still very much a work-in-progress.  Specs are failing, but migrations now succeed.

### DIFF
--- a/app/subsystems/domain/list_courses.rb
+++ b/app/subsystems/domain/list_courses.rb
@@ -21,6 +21,9 @@ class Domain::ListCourses
   private
 
   def run_with_options(user, with)
+    if user.is_a?(UserProfile::Models::Profile)
+      user = user.entity_user
+    end
     [with].flatten.each do |option|
       case option
       when :teacher_names
--- a/app/subsystems/domain/list_courses.rb
+++ b/app/subsystems/domain/list_courses.rb
@@ -21,9 +21,6 @@ class Domain::ListCourses
   private
 
   def run_with_options(user, with)
-    if user.is_a?(UserProfile::Models::Profile)
-      user = user.entity_user
-    end
     [with].flatten.each do |option|
       case option
       when :teacher_names
--- a/app/subsystems/role/add_user_role.rb
+++ b/app/subsystems/role/add_user_role.rb
@@ -4,9 +4,6 @@ class Role::AddUserRole
   protected
 
   def exec(user:, role:)
-    if user.is_a?(UserProfile::Models::Profile)
-      user = user.entity_user
-    end
     ss_map = Role::Models::User.create(user: user, role: role)
     transfer_errors_from(ss_map, {type: :verbatim}, true)
   end
--- a/app/subsystems/role/add_user_role.rb
+++ b/app/subsystems/role/add_user_role.rb
@@ -4,7 +4,10 @@ class Role::AddUserRole
   protected
 
   def exec(user:, role:)
-    ss_map = Role::Models::User.create(entity_user_id: user.entity_user_id, entity_role_id: role.id)
+    if user.is_a?(UserProfile::Models::Profile)
+      user = user.entity_user
+    end
+    ss_map = Role::Models::User.create(user: user, role: role)
     transfer_errors_from(ss_map, {type: :verbatim}, true)
   end
 end
--- a/app/subsystems/role/add_user_role.rb
+++ b/app/subsystems/role/add_user_role.rb
@@ -4,7 +4,7 @@ class Role::AddUserRole
   protected
 
   def exec(user:, role:)
-    ss_map = Role::Models::User.create(entity_user_id: user.id, entity_role_id: role.id)
+    ss_map = Role::Models::User.create(entity_user_id: user.entity_user_id, entity_role_id: role.id)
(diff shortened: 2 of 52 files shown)


jpslav:
lookin good @nathanstitt

app/subsystems/domain/list_courses.rb
```diff
@@ -21,6 +21,9 @@ def exec(user: nil, with: [])
   private
 
   def run_with_options(user, with)
+    if user.is_a?(UserProfile::Models::Profile)
+      user = user.entity_user
+    end
```
joemsak: What's happening here? I'm worried about this conditional leaking it's way all through the code.. Not only is it asking not telling but also checking on type!

app/subsystems/role/get_user_roles.rb
```diff
@@ -4,6 +4,9 @@ class Role::GetUserRoles
   protected
 
   def exec(user)
+    if user.is_a?(UserProfile::Models::Profile)
+      user = user.entity_user
+    end
     ss_maps  = Role::Models::User.includes(:role).where{entity_user_id == user.id}
```
nathanstitt: Now that foreign keys are meaningful, testing uncovered an issue with how the three Role routines are called.  Most of the time they're passed a `UserProfile::Models::Profile`, but sometimes an actual `Entity::User` is given. `Role::Models::User` expects an `Entity::User`, and any other type of id would cause a `fk id doesn't exist in table XX` Postgresql error.

In an attempt to keep this PR scope contained to only Postgresql related issues, I added these `is_a?` checks.  Future work would be to change the `Role::Models::User` table and association to point to `UserProfile::Models::Profile`, and hunt down and correct where the routines are being called by with an `Enity::User`.

app/subsystems/role/add_user_role.rb
```diff
@@ -4,7 +4,10 @@ class Role::AddUserRole
   protected
 
   def exec(user:, role:)
-    ss_map = Role::Models::User.create(entity_user_id: user.id, entity_role_id: role.id)
+    if user.is_a?(UserProfile::Models::Profile)
+      user = user.entity_user
+    end
```
joemsak: Same as above... would rather these routines get the correct input instead of having to check the input

app/subsystems/role/get_user_roles.rb
```diff
@@ -4,6 +4,9 @@ class Role::GetUserRoles
   protected
 
   def exec(user)
+    if user.is_a?(UserProfile::Models::Profile)
+      user = user.entity_user
+    end
```
joemsak: Out of control!

db/migrate/20150218225356_create_content_page_topics.rb
```diff
@@ -11,9 +11,9 @@ def change
...
-    add_foreign_key :content_page_topics, :pages, on_update: :cascade,
-                                                  on_delete: :cascade
-    add_foreign_key :content_page_topics, :topics, on_update: :cascade,
-                                                   on_delete: :cascade
+    add_foreign_key :content_page_topics, :content_pages, on_update: :cascade,
+                                                          on_delete: :cascade
+    add_foreign_key :content_page_topics, :content_topics, on_update: :cascade,
+                                                           on_delete: :cascade
   end
 end
```
nathanstitt: pages and topics tables no longer exist.

app/subsystems/role/get_user_roles.rb
nathanstitt: Or maybe we should go the other way and hunt down the `UserProfile::Models::Profile` usage.  That mainly comes from the controller's using `current_human_user`, which is generated by the lambda in `config/initializers/openstax_api.rb`

jpslav:
@nathanstitt, I agree with @joemsak that we should not check if the incoming user is possibly a `UserProfile::Models::Profile`, but instead make sure the input is of the right type. 

In line with our "FactoryGirl.create(:user) is confusing" email thread, can you go through and change all `FactoryGirl.create(:user)` to `FactoryGirl.create(:user_profile_profile)`, and then hunt down where we are calling these user role routines with the wrong input?  I think I already fixed a number of these in the specs (you'll have to rebase anyway).

Hate to muddy your Postgres PR, but on the other hand at least this PG fix will help you find the problematic cases.  Sound reasonable?

nathanstitt:
@jpslav @joemsak 

That's quite alright, I don't mind adding it to the PR, but wasn't clear which way we should do it, either by rewriting the roles to accept a UserProfile, or changing all the callers to pass an Entity::User.  Now that we're clear that we should pass an  Entity::User, I have one more Q:

What should be done with the the controllers that call `current_human_user` and pass that into routines that eventually end up at the role routines?   

Do you think the controller should call `current_human_user.entity_user` and pass that to the parent routine, or should the routines call `.entity_user` on the user right before they pass it to the role routines?

If it's the latter, I'd suggest that we should rename all the users of a `user:` named argument to `profile:` or `user_profile:` so that it's clear that callers should provide a UserProfile not an Entity::User

Thoughts?

nathanstitt:
This is now ready for another round of review.  I've renamed the :user factory to :user_profile and have changed the code that calls the routines to to pass user.entity_user as needed.

PR #145, opened 2015-03-31
Move tasks into a subsystem

jpslav:
This PR moves Task-related models into a "tasks" subsystem.

* `app/models` is now empty!
* All taskings are now assigned to roles; you can no longer assign straight to a user; instead you can get a user's default
  role from the `Role` subsystem.
* We'll still need to do some refactorings after this, e.g.:
  * deciding where exactly `Entity::Models::Task` should be used vs `Tasks::Models::Task`
  * collapsing all domain-ish routines into `app/domain`
  * figuring out how OK we are with controllers and representers knowing a lot about `Tasks::Models` and taking action as needed


app/controllers/api/v1/courses_controller.rb
```diff
@@ -63,9 +63,13 @@ def tasks
...
     # in the specified course; for now this is just returning what /api/user/tasks
     # returns and is ignore
-    OSU::AccessPolicy.require_action_allowed!(:read_tasks, current_api_user, current_human_user)
-    outputs = SearchTasks.call(q: "user_id:#{current_human_user.id}").outputs
-    respond_with outputs, represent_with: Api::V1::TaskSearchRepresenter
+    course = Entity::Models::Course.find(params[:id])
+    # OSU::AccessPolicy.require_action_allowed!(:read_tasks, current_api_user, course)
+    tasks = Domain::GetCourseUserTasks[course: course, user: current_human_user]
+
+    output = Hashie::Mash.new({'items' => tasks})
```
Dantemss: Ugh let's please make some Api model classes that represent these to other subsystems.

Dantemss: Oh nvm this is in the controller... err odd but ok

app/controllers/api/v1/task_plans_controller.rb
```diff
@@ -1,5 +1,31 @@
 class Api::V1::TaskPlansController < Api::V1::ApiController
 
+  ###############################################################
+  # ----- temporary code to fix location: error -------
```
Dantemss: Update openstax_api? :D

app/routines/mark_task_step_completed.rb
```diff
@@ -6,7 +6,7 @@ class MarkTaskStepCompleted
 
   def exec(task_step:)
     fatal_error(code: :step_type_cannot_be_marked_completed) \
-      unless %w(TaskedReading TaskedExercise TaskedVideo).include?(task_step.tasked_type)
+      unless %w(TaskedReading TaskedExercise TaskedVideo).include?(task_step.tasked_type.demodulize)
```
Dantemss: Could just allow them all to be completed. I think the UI always marks them all so...

app/routines/search_tasks.rb
```diff
@@ -27,11 +29,14 @@ def exec(params, options = {})
 
       with.keyword :user_id do |ids|
         ids.each do |i|
-          sanitized_ids = to_number_array(i)
+          sanitized_ids = to_number_array(i)
```
Dantemss: OMG WHITESPACE!!!

app/routines/search_tasks.rb
```diff
@@ -27,11 +29,14 @@ def exec(params, options = {})
 
       with.keyword :user_id do |ids|
         ids.each do |i|
-          sanitized_ids = to_number_array(i)
+          sanitized_ids = to_number_array(i)          
           next @items = @items.none if sanitized_ids.empty?
 
+          run(Role::GetUserRoles, sanitized_ids)
```
Dantemss: I don't think my brain can handle this one right now.

app/subsystems/entity/models/role.rb
```diff
@@ -1,3 +1,3 @@
 class Entity::Models::Role < Tutor::SubSystems::BaseModel
-  enum role_type: [:unassigned, :teacher, :student]
+  enum role_type: [:unassigned, :teacher, :student, :default]
```
Dantemss: Default should be the first (and really unassigned should be the default?)

app/representers/api/v1/task_step_properties.rb
```diff
@@ -24,12 +24,12 @@ module TaskStepProperties
...
              writeable: false,
              readable: true,
-             getter: lambda { |*| self.class.name.gsub("Tasked","").underscore.downcase },
+             getter: lambda { |*| self.class.name.demodulize.gsub("Tasked","").underscore.downcase },
              schema_info: {
                required: true,
                description: "The type of this TaskStep, one of: #{
                             TaskedRepresenterMapper.models.collect{ |klass|
-                              "'" + klass.name.gsub("Tasked","")
+                              "'" + klass.name.demodulize.gsub("Tasked","")
```
nathanstitt: I'm a fan of using `remove` instead of `gsub` or `sub` if all you are doing is removing part of the string.  For instance:  `self.class.name.demodulize.remove("Tasked").underscore.downcase`
Or if you needed a regex you can still use`.remove(/Tasked$/)`

app/subsystems/role/get_default_user_role.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  uses_routine Role::GetUserRoles
+  uses_routine Role::CreateUserRole,
+               translations: { outputs: { type: :verbatim } }
+
+  protected
+
+  def exec(user)
+    run(Role::GetUserRoles, user, :default)
+    existing_default_roles = outputs['[:role_get_user_roles, :roles]']
```
Dantemss: This doesn't look good...

app/controllers/api/v1/task_plans_controller.rb
jpslav: Yep will update.

app/representers/api/v1/task_representer.rb
```diff
@@ -17,15 +17,15 @@ class TaskRepresenter < Roar::Decorator
...
-    property :task_plan_id,
-             type: Integer,
-             writeable: false,
-             readable: true,
-             render_nil: true,
-             schema_info: {
-               required: false,
-               description: "The ID of the TaskPlan used to generate this Task"
-             }
+    # property :task_plan_id,
```
jpslav: @Dantemss @nathanstitt I took this out because I didn't know why it'd be here -- students don't need to know what plan their task came from.  Am I missing something?

app/subsystems/tasks/create_tasking.rb
```diff
@@ -4,13 +4,7 @@ class Tasks::CreateTasking
...
   def exec(role:, task:)
-    # Temporarily handle legacy tasks
-    if task.is_a? ::Task
-      legacy_task = task
-      task = Entity::Models::Task.create!
-      ltm = Tasks::Models::LegacyTaskMap.create!(task: task, legacy_task: legacy_task)
-    end
-
-    outputs[:tasking] = Tasks::Models::Tasking.create!(role: role, task: task)
+    entity_task = task.is_a?(Entity::Models::Task) ? task : task.entity_task
```
Dantemss: Having these checks around seems wrong to me...

app/representers/api/v1/task_step_properties.rb
jpslav: Nice.  Will change.

app/subsystems/tasks/does_tasking_exist.rb
```diff
@@ -1,19 +1,25 @@
...
 class Tasks::DoesTaskingExist
   lev_routine
 
   protected
 
-  def exec(task:, roles:)
-    raise NotYetImplemented unless task.is_a?(::Task)
+  def exec(task_component:, roles:)
+
+    task = case task_component
```
Dantemss: Could replace this with some delegation, potentially...

app/representers/api/v1/task_representer.rb
nathanstitt: The front-end might be using the id to lookup additional details about the task_plan, but I doubt it.  I'm sure they'll let us know if they do need it :)

app/routines/search_tasks.rb
jpslav: I took out all uses of this `SearchTasks` routine (1) because we didn't really need it any more and (2) because my brain couldn't handle it either.

db/migrate/20140926213212_create_tasks_tasks.rb
```diff
@@ -0,0 +1,25 @@
+class CreateTasksTasks < ActiveRecord::Migration
```
Dantemss: The names!

app/routines/search_tasks.rb
jpslav: If y'all don't see a reason why we need it I can kill it altogether.

app/subsystems/tasks/get_tasks.rb
```diff
@@ -0,0 +1,25 @@
...
+  def as_id_array(values)
+    return [] if values.blank?
+    
+    values = [values].flatten.compact
+    
+    values.first.is_a?(Integer) ?
+      values : 
+      values.collect{|v| v.id}
+  end
+end
```
nathanstitt: The `as_id_array` seems like it might be useful in other routines.  We should keep it in mind for extraction if we encounter the need for it again.

spec/models/task_spec.rb
```diff
@@ -1,40 +0,0 @@
-require 'rails_helper'
-
-RSpec.describe Task, :type => :model do
```
Dantemss: Did these get lost in the refactoring?

app/subsystems/entity/models/role.rb
jpslav: I can move default to first, but maybe unassigned is still useful in tests?  (unassigned explicitly has no meaning, "default" does have a meaning -- there is one default role per user).

Dantemss: What is the default role?

app/routines/search_tasks.rb
Dantemss: Go ahead and kill it.

app/subsystems/user_profile/models/profile.rb
```diff
@@ -8,10 +8,10 @@ class UserProfile::Models::Profile < Tutor::SubSystems::BaseModel
...
-  has_many :educators, dependent: :destroy, foreign_key: :user_id
-  has_many :students, dependent: :destroy, foreign_key: :user_id
+  # has_many :educators, dependent: :destroy, foreign_key: :user_id
+  # has_many :students, dependent: :destroy, foreign_key: :user_id
 
-  has_many :taskings, dependent: :destroy, foreign_key: :user_id
+  # has_many :taskings, dependent: :destroy, foreign_key: :user_id
 
   validates :account, :entity_user, presence: true, uniqueness: true
   validates :exchange_identifier, presence: true
```
nathanstitt: Assuming that the commented out associations are intentional and not because you forgot to uncomment them, why not just delete them altogether?

app/representers/api/v1/task_representer.rb
Dantemss: That is fine. Leave it out.

app/subsystems/tasks/get_tasks.rb
Dantemss: I'd say put it in a lib and just mix into lev routines on boot.

Dantemss: Or a module you can include... that works too.

app/subsystems/role/get_default_user_role.rb
jpslav: I think the intention was that it'd be `outputs[:role_get_user_roles][:roles]` but obviously that's not what happens in lev.

app/subsystems/entity/models/role.rb
Dantemss: Like... what does it mean? It's not a teacher or student...

app/subsystems/tasks/does_tasking_exist.rb
jpslav: Like an `entity_task` method on every component that calls `entity_task` on its parent?

app/subsystems/role/get_default_user_role.rb
Dantemss: You're doing a verbatim translation... shouldn't it just be `[:roles]`?

app/subsystems/tasks/create_tasking.rb
jpslav: Part of me wants to make `Tasks::Models::Tasking` connect an `Entity::Models::Role` and a `Tasks::Models::Task` instead of what it does now, an `Entity::Models::Role` and `Entity::Models::Task`.

app/subsystems/role/get_default_user_role.rb
Dantemss: Oh never mind, you're not... how about then:
```rb
out = run(Role::GetUserRoles, user, :default).outputs
existing_default_roles = out[:roles]
```

app/subsystems/tasks/get_tasks.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def exec(roles:)
+    role_ids = as_id_array(roles)
+
+    outputs[:tasks] = 
+      Entity::Models::Task.joins{taskings}
+                          .where{taskings.entity_role_id.in role_ids}
+  end
+
+  def as_id_array(values)
```
jpslav: @Dantemss what's your reaction to these routine that take either an array of objects or an array of those objects' IDs?

spec/models/tasking_spec.rb
```diff
@@ -1,30 +0,0 @@
...
-  end
-
-  it "requires user to match taskee" do
-    expect(FactoryGirl.build(:tasking)).to be_valid
-
-    expect(FactoryGirl.build(:tasking,
-                             user: FactoryGirl.create(:user))).to_not be_valid
-  end
-end
-
```
nathanstitt: Did you remove these specs?  I don't see where they were re-named but I'm probably just missing them.  I don't really see much point in testing that the associations are working though so maybe it does make sense to remove them

app/subsystems/tasks/does_tasking_exist.rb
Dantemss: That could be. Still not totally sure why this routine has to handle all the different bits of a Task.

jpslav: Used from access policies at different levels of the task.

app/subsystems/tasks/get_tasks.rb
jpslav: Yep, that was my plan, will do in followup commit.

spec/routines/search_tasks_spec.rb
```diff
@@ -1,16 +1,20 @@
 require 'rails_helper'
 
-describe SearchTasks do
+#
+# SearchTasks IS DEPRECATED
+#
 
-  let!(:task_1) { FactoryGirl.create(:tasking).task.reload }
+xdescribe SearchTasks do
```
nathanstitt: Is this skipped on purpose?

app/subsystems/user_profile/models/profile.rb
jpslav: Yep, I'm going back through now and deleting these deprecated code bits.  Thanks.

app/subsystems/tasks/create_tasking.rb
Dantemss: You could move the Tasking bits to their own Tasking subsystem (so they would only handle entities, not the other Task)

app/subsystems/tasks/get_tasks.rb
Dantemss: I'd prefer we always passed the objects (so we know they actually exist etc) instead of the id's but if it's absolutely necessary...

spec/models/task_spec.rb
jpslav: All specs got moved to `spec/subsystems/tasks/models`.  Most of this particular spec remains except for the `tasked_to?` example which is gone.

app/subsystems/tasks/does_tasking_exist.rb
Dantemss: ah ok...
still, could define an `entity_task` method on all bits (and maybe it's just `self` for the Entity object)

spec/models/tasking_spec.rb
jpslav: This spec did get moved to the subsystems spec models dir, but I left all of these old tests commented out -- I'll go put them back.  Thanks.

spec/routines/search_tasks_spec.rb
Dantemss: He's going to remove this routine.

spec/models/task_spec.rb
Dantemss: Oh ok guess I lost track of it.

spec/subsystems/tasks/assistants/i_reading_assistant_spec.rb
```diff
@@ -1,16 +1,17 @@
...
+  before(:each)    { OpenStax::Exercises::V1.use_real_client }
+
+  let!(:assistant) { FactoryGirl.create :tasks_assistant,
+                                        code_class_name: 'Tasks::Assistants::IReadingAssistant' }
   let!(:book_part) { FactoryGirl.create :content_book_part }
 
   # Store version differences in this hash
   cnx_page_infos = HashWithIndifferentAccess.new(
-    stable: { id: '092bbf0d-0729-42ce-87a6-fd96fd87a083@4', title: 'Force' },
-    latest: { id: '092bbf0d-0729-42ce-87a6-fd96fd87a083', title: 'Force' }
```
jpslav: @Dantemss I took out this `latest` page because it deviated too much from the `stable` one and it seemed like too much work to stay up to date with the latest every time we re-recorded the cassette.

spec/subsystems/tasks/assistants/i_reading_assistant_spec.rb
```diff
@@ -46,16 +47,15 @@
               if task_step.tasked_type == 'TaskedExercise'
 
             task_steps.except(task_step).each do |other_step|
-              expect(task_step.tasked.content).not_to(
-                include(other_step.tasked.content)
-              )
+              # expect(task_step.tasked.content).not_to(
```
jpslav: @Dantemss I didn't mean to leave this commented out -- for some reason I couldn't get this to work.

app/subsystems/entity/models/role.rb
jpslav: Before we could assign tasks directly to users.  I simplified and said that tasks can only be assigned to roles.  So every user now has a default role (really we can get a default role for any user by calling `Role::GetDefaultUserRole[user]`).  That sound reasonable?  Saves us from having polymorphic taskees or having parallel user and role IDs in the `Tasking`.

spec/subsystems/tasks/assistants/i_reading_assistant_spec.rb
Dantemss: That's fine. Kevin made an argument that convinced me to take them out so I've been doing that too.

app/subsystems/role/get_default_user_role.rb
jpslav: I didn't even know that worked :-D

spec/subsystems/tasks/assistants/i_reading_assistant_spec.rb
Dantemss: This test just tests that the task steps are not getting repeated content.

If it's failing, you should look at the content and see why it's the same :)

app/subsystems/entity/models/role.rb
Dantemss: Sounds good. I say make that the second one in the enum (after :unassigned)

Still not sure if :unassigned is a good idea though.

app/subsystems/tasks/create_tasking.rb
jpslav: Yeah, @kevinburleigh75 and I have discussed this some -- wouldn't be much in such a SS and really there are places where the taskings / tasks / tasking plans / tasks plans are pretty tightly woven.

app/subsystems/tasks/get_tasks.rb
jpslav: In my helper you'll see that I at least added type verification when it isn't an ID.  This is what showed me a zillion cases where we were passing `UserProfile::Models::Profile` instead of what we wanted, `Entity::Models::User`.

app/subsystems/entity/models/role.rb
jpslav: Made it the second one.

philschatz:
Tutor-js no longer depends on this route!

jpslav:
Woohoo! Thx.

Dantemss:
Karen is looking at the failing spec, otherwise this PR is good to go.

PR #146, opened 2015-03-31
Add published_at field to task_plans

nathanstitt:
Adds the field to task_plans and updates it when the plan is distributed.


spec/routines/distribute_tasks_spec.rb
```diff
@@ -4,8 +4,6 @@
   it 'validates the task_plan settings against the assistant schema' do
     task_plan = FactoryGirl.create :task_plan
 
-    expect(DistributeTasks.call(task_plan).errors).to be_empty
-
     allow(DummyAssistant).to receive(:schema).and_return(
```
nathanstitt: I removed this call to DistributeTasks since it wasn't adding anything to the spec and was causing the published_at to be set on the plan before I tested it below

PR #148, opened 2015-04-01
Try another and refresh my memory attempt 2

Dantemss:
- Try another now gets a related problem using a random LO
- Try another now only looks at practice-problems
- Try another now prefers unassigned problems
- Refresh my memory now links to CNX module
- Specs


Dantemss:
Domain::SearchLocalExercises needs to eventually be changed to return a wrapper (instead of returning Content::Exercises)

Dantemss:
Otherwise, done.

app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -59,9 +59,32 @@ def recovery
...
+  def refresh
+    tasked = Tasks::Models::TaskStep.find(params[:id]).tasked
+    OSU::AccessPolicy.require_action_allowed!(:refresh, current_api_user, tasked)
+
+    result = RefreshTaskedExercise.call(tasked_exercise: tasked)
+
+    if result.errors.any?
+      render_api_errors(result.errors)
+    else
+      json_response = {}
```
jpslav: Instead of all of these lines here, just make a separate representer for this composite output?

Dantemss: Good idea.

app/routines/recover_tasked_exercise.rb
```diff
@@ -1,31 +1,84 @@
...
+  # Prefers unassigned Exercises
+  def get_recovery_exercise_for(tasked_exercise:,
+                                required_tag_names: ['practice-problem'])
+
+    ex_relation = Content::Models::Exercise.preload(exercise_tags: :tag)
+
+    # Randomize LO order
+    los = tasked_exercise.wrapper.los.shuffle
+
+    # Try to find unassigned exercises first
```
jpslav: Just checking that the intent here is to find problems not assigned to this this `TaskedExercise`'s student?

jpslav: I'll say that lines 60 through 78 look like they'd be useful in their own routine, e.g. for spaced practice to find problems.

```ruby
Domain::GetExercises[tags: ['practice=problem', 'lo1'], 
                     not_assigned_to: { 
                       roles: [...], 
                       strictness: :loose 
                     }, 
                     count: 1,
                     shuffle: true]
```

or maybe this is even an extension to `SearchLocalExercises`?  Would be nice to reuse this chunk of code and make it independently testable

app/controllers/api/v1/task_steps_controller.rb
jpslav: And just so I'm understanding what's going on here, you're adding in new steps to the Task, but returning the new steps here for FE to have access to them efficiently?

app/routines/refresh_tasked_exercise.rb
```diff
@@ -0,0 +1,33 @@
...
+  uses_routine RecoverTaskedExercise,
+               as: :recover, translations: { outputs: { type: :verbatim } }
+
+  protected
+
+  def exec(tasked_exercise:)
+    task_step = tasked_exercise.task_step
+    run(:recover, tasked_exercise: tasked_exercise)
+    task = outputs[:task]
+    recovery_step = outputs[:recovery_step]
```
jpslav: this line is unused.

app/routines/refresh_tasked_exercise.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    outputs[:refresh_step] = refresh_step_for(task_step: task_step)
+  end
+
+  private
+
+  def refresh_step_for(task_step:)
+    current_step = task_step
+
+    # Hack for Sprint 9; Replace with final version before alpha
```
jpslav: Just add a comment on what the hack is -- I *think* it is that you're repeating the most recent reading step again.

app/subsystems/content/models/exercise.rb
```diff
@@ -2,4 +2,6 @@ class Content::Models::Exercise < Tutor::SubSystems::BaseModel
   acts_as_resource
 
   has_many :exercise_tags, dependent: :destroy
+
+  has_many :tasked_exercises, subsystem: :tasks, dependent: :destroy, foreign_key: :exercise_id
```
jpslav: This is not good in terms of SS encapsulation.  Also, I don't see it used anywhere -- can it be taken out?

app/subsystems/domain/task_exercise.rb
```diff
@@ -0,0 +1,19 @@
...
+  def exec(exercise:, has_recovery: false, task_step: nil)
+    outputs[:tasked_exercise] = Tasks::Models::TaskedExercise.new(
+      task_step: task_step,
+      exercise: (exercise.is_a?(Content::Models::Exercise) ? exercise : nil),
+      url: exercise.url,
+      title: exercise.title,
+      content: exercise.content,
+      has_recovery: has_recovery
+    )
+    task_step.tasked = outputs[:tasked_exercise] unless task_step.nil?
```
jpslav: default value of `task_step` is nil, but you don't check for that here, so either check and handle or don't have a default.

app/subsystems/domain/search_local_exercises.rb
```diff
@@ -0,0 +1,53 @@
...
+  lev_routine express_output: :items
+
+  uses_routine Content::Routines::SearchExercises, as: :search
+
+  protected
+
+  def exec(options = {})
+
+    assigned_to = options.delete(:assigned_to)
+    not_assigned_to = options.delete(:not_assigned_to)
```
jpslav: hmmm maybe you were already doing some stuff I mentioned above? (about making this routine handle repeated assignment, etc)

app/subsystems/tasks/assistants/i_reading_assistant.rb
```diff
@@ -104,8 +101,8 @@ def self.distribute_tasks(task_plan:, taskees:)
           step.tasked = case fragment
           when OpenStax::Cnx::V1::Fragment::ExerciseChoice
             exercises = fragment.exercise_fragments
-            tasked_exercise(exercise_fragment: exercises.first,
-                            recovery_fragment: exercises.last,
+            tasked_exercise(exercise_fragment: exercises.sample,
+                            has_recovery: true,
```
jpslav: I'm confused -- what is `has_recovery`?  A recovery question is one exercise embedded inside another, right?  So why do we need this?

Or is this maybe supposed to be `can_be_recovered`?

app/controllers/api/v1/task_steps_controller.rb
Dantemss: Correct. It also returns the url for the "refresh my memory" part.
Refresh my memory effectively does refresh + recovery.

app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -1,26 +1,23 @@
 class Tasks::Models::TaskedExercise < Tutor::SubSystems::BaseModel
   acts_as_tasked
 
-  belongs_to :recovery_tasked_exercise, class_name: name,
-                                        dependent: :destroy,
-                                        subsystem: :ignore
+  belongs_to :exercise, subsystem: :content, foreign_key: :exercise_id
```
jpslav: like above, kinda undesirable

app/subsystems/tasks/assistants/i_reading_assistant.rb
jpslav: `has_recovery` makes it sound like it has a populated recovery step

app/routines/recover_tasked_exercise.rb
Dantemss: What's strictness?
Also not having shuffle is weird, because it would just return the first exercise in database order (whatever that is).

db/migrate/20150203212240_create_tasks_tasked_exercises.rb
```diff
@@ -10,11 +11,7 @@ def change
 
       t.timestamps null: false
 
-      t.index :recovery_tasked_exercise_id, unique: true
+      t.index :exercise_id
```
jpslav: this should be `content_exercise_id` I believe (tho I'd prefer to not have it)

app/routines/recover_tasked_exercise.rb
Dantemss: And yes, the intent is to "prefer" exercises not assigned to any of the taskees.

app/subsystems/tasks/assistants/i_reading_assistant.rb
Dantemss: Sure I can change the name.

app/subsystems/tasks/models/tasked_exercise.rb
Dantemss: This exercise_id column is the actual assigned exercise, not the recovery one. I added it so I can track which exercises have been assigned to whom.

db/migrate/20150203212240_create_tasks_tasked_exercises.rb
Dantemss: Sorry, I kind of need it ;) maybe I can get entity_exercise_id to work?

app/subsystems/domain/task_exercise.rb
Dantemss: What do you mean?

app/subsystems/content/models/exercise.rb
Dantemss: Yep I know it's bad. And no, it's used by the search routine to determine which exercises a role has done or not.

app/routines/refresh_tasked_exercise.rb
Dantemss: Will do.

app/subsystems/tasks/assistants/i_reading_assistant.rb
Dantemss: After Kathi clarified how recovery was supposed to work, I decided it made more sense to determine the recovery exercise when you call the recovery API, so that's why we have a boolean instead of a nested tasked_exercise.

app/routines/recover_tasked_exercise.rb
jpslav: `strictness: :loose` here was saying "try to find exercises not assigned to these roles, but if not possible we can deal with ones already assigned"

PR #149, opened 2015-04-01
TaskedRepresenterMapperSpec and Domain::ImportBook

Dantemss:
- Don't preload code in the TaskedRepresenterMapperSpec
- Fix Domain::ImportBook


Dantemss:
This is a small PR just to fix specs in master.

PR #151, opened 2015-04-03
WIP Klb spaced practice

jpslav:


app/subsystems/tasks/assistants/i_reading_assistant.rb
```diff
@@ -24,8 +18,119 @@ def self.schema
...
+                             opens_at:  opens_at,
+                             due_at:    due_at]
+    task.save!
+    task
+  end
+
+  def self.add_core_steps!(task:, cnx_pages:)
+    cnx_pages.each do |page|
+      page.fragments.each do |fragment|
+        step = Tasks::Models::TaskStep.new(task: task, page_id: page.id)
```
jpslav: @nathanstitt is the `page_id` you added to `TaskStep` a Content subsystem page ID (pulled from `Content::Models::Page`) or a `OpenStax::Cnx::Page` id?

Dantemss: Maybe that's now what we want, but it currently is a OpenStax::Cnx::Page id.

app/subsystems/tasks/models/task.rb
```diff
@@ -25,4 +32,32 @@ def completed?
...
+    algorithm
+  end
+
+  def spaced_practice_algorithm=(algorithm)
+    raise ArgumentError, "algorithm cannot be nil" if algorithm.nil?
+    write_attribute(:spaced_practice_algorithm, YAML.dump(algorithm))
+  end
+
+  def handle_task_step_completion!
+    spaced_practice_algorithm.call(event: :task_step_completion, task: self)
```
jpslav: FWIW, we might want to generalize this on the next iteration, having `Task`s have listeners, one of which is wraps the spaced practice algorithm and responds to the `:task_step_completion` event.  Actually this is pretty much that but instead of explicitly storing a `spaced_practice_algorithm` attribute, you'd have an array of listeners, one of which is `YAML.dump(algorithm)` -- it is already a listener in that it responds to an `event` and a `task`. 

Not to change it now, but maybe in the next pass.

app/subsystems/tasks/assistants/i_reading_assistant.rb
nathanstitt: Yes, like @Dantemss said, it's a `OpenStax::Cnx::Page` id.  It's not currently used for reading an association back from, it's only purpose is to be able to identify all the tasks that have a common origin so that statistics can be computed across them.  We can change it to something more meaningful if that'd be useful.

PR #153, opened 2015-04-03
Student views learning guide api

joemsak:
https://www.pivotaltracker.com/n/projects/1156756/stories/91198814

Let's see how it's lookin so far, team

## FE Rake Task
rake sprint:009course_stats

`current_level` is using randomly generated values for now


app/subsystems/domain/get_user_course_stats.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def collect_book_parts
+    outputs.book_parts.collect do |book_part|
+      { id: book_part.id,
+        title: book_part.title,
+        number: book_part.path }
+    end
+  end
+end
```
nathanstitt: I'm curious what others think about reading the outputs from a called routine in this way.  It took me a bit to figure out how the output from `:get_course_profile` and `:get_course_books` was being read, and would have found it clearer if their output had been assigned to variables and those passed to  `compile_course_stats`.  Maybe this is idiomatic lev usage though?

joemsak: verbatim outputs

app/controllers/api/v1/courses_controller.rb
```diff
@@ -75,6 +75,15 @@ def events
     respond_with outputs, represent_with: Api::V1::CourseEventsRepresenter
   end
 
+  api :GET, '/courses/:course_id/stats(/role/:role_id)', 'TODO get rid of this documentation'
+  description 'TODO somehow get rid of this API documentation, not desired'
```
jpslav: You can get rid of these "get rid of this doc" messages :-)  I had to have that for another method because apipie freaked out if I didn't.

app/controllers/api/v1/courses_controller.rb
```diff
@@ -75,6 +75,15 @@ def events
...
+  def stats
+    course = Entity::Course.find(params[:id])
+    course_stats = Domain::GetUserCourseStats[user: current_human_user.entity_user,
```
jpslav: lookin good, I think this will need to take a role instead of a user (which I think you already know, just pointing it out). 

you'll be able to pass in the course and the role from the params (if available) to Nathan's pending #154 work.

app/subsystems/content/get_book_parts.rb
```diff
@@ -0,0 +1,10 @@
+class Content::GetBookParts
```
jpslav: Ideally should not be returning content SS models outside of the SS.  Can you just use the Toc visitor or another visitor along with `Content::VisitBook`?

app/subsystems/course_content/get_course_books.rb
```diff
@@ -4,9 +4,8 @@ class CourseContent::GetCourseBooks
   protected
 
   def exec(course:)
-    book_ids = CourseContent::Models::CourseBook.where(course: course.id)
-                                        .select(:entity_book_id)
-                                        .collect{|cb| cb.entity_book_id}
+    outputs[:course_books] = CourseContent::Models::CourseBook.where(course: course.id)
```
jpslav: Also here should not be returning this subsystem model out.  Looks like it is just used elsewhere to go the entity book ID, so I think what was here before is sufficient for what's needed?

app/subsystems/domain/get_user_course_stats.rb
jpslav: Yeah it is idiomatic lev; lev handles multiple outputs and errors by translating them from the called to the calling routine's outputs hash.  So I think it is good.  Different devs may decide to pass as arguments to methods instead, which is fine too.

philschatz:
This is tied to openstax/tutor-js#137

app/controllers/api/v1/courses_controller.rb
```diff
@@ -111,6 +111,18 @@ def dashboard
...
+    role = ChooseCourseRole[course: course,
+                            user: current_human_user.entity_user,
+                            allowed_role_type: :student,
+                            role_id: params[:role_id]]
+    course_stats = GetCourseStats[role: role, course: course]
+    respond_with course_stats, represent_with: Api::V1::CourseStatsRepresenter
+  end
+
+  api nil, nil, nil
+  description nil
```
kevinburleigh75: Are these `nil`s necessary?  Are they placeholders, or do they serve a purpose? (Same for those above...)

app/external/openstax/cnx/v1/page.rb
```diff
@@ -85,6 +86,10 @@ def book_part_title
       @book_part_title
     end
 
+    def book_part_id
+      @book_part_id
+    end
```
kevinburleigh75: Simplify to this?
```
attr_reader :book_part_id
```

app/controllers/api/v1/courses_controller.rb
jpslav: Joe didn't add these -- I think Dante or I did b/c Apipie was freaking out when the api call was missing on the method (we don't want one here because it is a helper method).  Maybe if we made it protected it wouldn't care?

app/controllers/api/v1/courses_controller.rb
```diff
@@ -111,6 +111,18 @@ def dashboard
 
   api nil, nil, nil
```
jpslav: But actually these should not be nil nil nil.  Should be formatted like the other methods to have descriptions, etc.

app/representers/api/v1/course_stats_representer.rb
```diff
@@ -0,0 +1,9 @@
+module Api::V1
+  class CourseStatsRepresenter < Roar::Decorator
+    include Roar::JSON
+
+    property :title, readable: true
+
+    property :fields, readable: true
```
jpslav: Need to break this out into all the little pieces, so we can control what goes out and so we can document it.

jpslav:
@joemsak please rebase -- you're gonna need some changes to `Tasks::Models::Task` validations (it now checks for a valid type and we've added here the `"page-practice"`, `"chapter-practice"`, and `"mixed-practice"` types.

app/routines/choose_course_role.rb
```diff
@@ -83,10 +76,11 @@ def find_unique_role(course:, roles:, type:)
         CourseMembership::IsCourseStudent[course: course, roles: role]
       end
     end
-    if matching_roles.count > 1
+
+    if matching_roles.many?
```
kevinburleigh75: Can we change back?  `#many?` is Rails-only and, IMO, doesn't convey useful meaning (what's "many" mean? - I had to look it up to be sure...`#multiple?` would make far more sense).

app/routines/reset_practice_widget.rb
```diff
@@ -33,7 +35,15 @@ def exec(role:, condition:, page_ids: [], book_part_ids: [])
...
 
-    task = Tasks::CreateTask[task_type: 'practice',
+    if book_part_ids.count == 1
+      task_type = 'chapter-practice'
+    elsif page_ids.count == 1
+      task_type = 'page-practice'
+    else
+      task_type = 'mixed-practice'
+    end
+
```
kevinburleigh75: FYI - These will cause validation errors once they interact with my spaced practice PR.  It's an easy fix, though.

app/routines/reset_practice_widget.rb
```diff
@@ -65,20 +75,19 @@ def get_fake_exercises(count:)
...
-        fatal_error(code: :no_exercises_found,
-                    message: "No exercises matched the given ID's") \
-          if exercise.nil?
+                                         tag: tags,
+                                         match_count: 1].to_a.shuffle
+        unless exercise = exercises.pop
+          fatal_error(code: :no_exercises_found,
+                      message: "No exercises matched the given ID's")
+        end
       end
```
kevinburleigh75: I don't think the above block precludes the possibility of having two identical exercises in the same practice widget set.  FYI.

app/routines/get_course_stats.rb
```diff
@@ -0,0 +1,105 @@
...
+  uses_routine Content::VisitBook,
+    translations: { outputs: { type: :verbatim } },
+    as: :visit_book
+
+  protected
+  def exec(role:, course:)
+    run(:get_role_task_steps, roles: role)
+    run(:get_course_books, course: course)
+    run(:visit_book, book: outputs.books.first, visitor_names: :toc)
+    run(:visit_book, book: outputs.books.first, visitor_names: :page_data)
```
jpslav: Pretty sure you can run both visitors at the same time:

```ruby
run(:visit_book, book: outputs.books.first, visitor_names: [:toc, :page_data])
```

app/subsystems/content/visit_book_part.rb
```diff
@@ -47,6 +47,8 @@ def map_visitor_names_to_visitors(visitor_names)
         outputs[:exercises] = Content::Models::ExerciseVisitor.new
       when 'page_data'
         outputs[:page_data] = Content::Models::PageDataVisitor.new
+      else
+        raise "undefined visitor #{name}. Try one of toc, exercises, page_data"
```
kevinburleigh75: +1 for non-silent failure when misused.  +1 for instructive error message.

app/subsystems/course_content/get_course_books.rb
```diff
@@ -4,9 +4,8 @@ class CourseContent::GetCourseBooks
   protected
 
   def exec(course:)
-    book_ids = CourseContent::Models::CourseBook.where(course: course.id)
-                                        .select(:entity_book_id)
-                                        .collect{|cb| cb.entity_book_id}
+    course_books = CourseContent::Models::CourseBook.where(course: course.id)
+    book_ids = course_books.collect(&:entity_book_id)
```
kevinburleigh75: These lines aren't equivalent to the replaced lines - is that intentional?  The old lines reject `CourseBook`s where `#entity_book_id` is falsey (read: `nil`), but the latter does not.

app/subsystems/tasks/assistants/i_reading_assistant.rb
```diff
@@ -169,9 +169,10 @@ def self.tasked_exercise(exercise_fragment:,
     exercises = Content::Routines::SearchExercises[
                   tag: exercise_fragment.embed_tag
                 ]
-    exercise = exercises.first
-    TaskExercise[exercise: exercises.first, title: title,
-                 can_be_recovered: can_be_recovered, task_step: step]
+    if exercise = exercises.first
+      TaskExercise[exercise: exercise, title: title,
```
kevinburleigh75: Is this causing a silent failure when our input data is badly-formed?  Why would `exercises` be legitimately empty?

app/subsystems/tasks/models/task_step.rb
```diff
@@ -14,6 +14,25 @@ class Tasks::Models::TaskStep < Tutor::SubSystems::BaseModel
   scope :complete, -> { where{completed_at != nil} }
   scope :incomplete, -> { where{completed_at == nil} }
 
+  def has_correctness?
```
jpslav: @kevinburleigh75, joe is now using the `Hacks::AnswerExercise` in place of calling these methods -- do you think we should keep them around or ditch them?

kevinburleigh75: I think we would still need `#has_correctness?` and `#is_correct?` on the query side, no?  Maybe `#make_{correct,incorrect}!` could go away, though...

app/controllers/api/v1/courses_controller.rb
joemsak: Yep just following in the tracks laid out before me

lib/tasks/sprint/sprint_009/sp.rb
```diff
@@ -66,7 +66,7 @@ def exec
       ]
 
       ## task = reading_task_groups[assignment_index][taskee_index]
-      reading_task_groups = page_data[1..4].each_with_index.collect do |page_data, ii|
+      reading_task_groups = page_data.from(1).each_with_index.collect do |page_data, ii|
```
kevinburleigh75: This replacement is not equivalent to the original code (unless you make assumptions about the size of the `page_data` array - is that intentional?

app/external/openstax/cnx/v1/page.rb
joemsak: I agree however I was following what I saw in the rest of the class

app/representers/api/v1/course_stats_representer.rb
joemsak: :+1:

app/routines/choose_course_role.rb
joemsak: Not sure why "many?" isn't obvious to you? In defense of keeping it, @jpslav is on record as saying it's highly unlikely we will move off of rails

app/routines/reset_practice_widget.rb
joemsak: Not sure what you mean, but the match_count was needed for my purposes (see @jpslav) and the rest is just making the code a bit more succinct

app/routines/get_course_stats.rb
joemsak: :+1: cool will try

app/subsystems/content/visit_book_part.rb
joemsak: Merci

app/subsystems/course_content/get_course_books.rb
joemsak: I needed to separate course_books and book_ids for the outputs

app/subsystems/tasks/assistants/i_reading_assistant.rb
joemsak: This was re-written to prevent TaskExercise from being created with a nil exercise, the bug was found when we worked with Dante at some point

app/routines/reset_practice_widget.rb
jpslav: @Dantemss I think you wrote this method first in https://github.com/openstax/tutor-server/commit/cde66d150dbe2ea1ef8643873c25aaea216e0b15.  Isn't it true that this loop of calling `SearchLocalExercises` might return the same exercise?  (shuffle helps prevent that but isn't that predictable).

app/external/openstax/cnx/v1/page.rb
kevinburleigh75: Fair enough - I was just going on the diff.

lib/tasks/sprint/sprint_009/sp.rb
joemsak: this is the sprint9 task, so we know we're dealing with the Force book, which we know has a preface as the page_data[0]... it allows the size of page_data to grow dynamically and we'll always get all pages after preface

joemsak:
@jpslav Rebased

app/routines/choose_course_role.rb
kevinburleigh75: It is not.  "Many" doesn't unambiguously mean "more than one" in any normal or technical form of English that I can think of.  In fact, it more often means "lots".  IMO it's not just unclear but actually deceptive, regardless of whether it's Ruby or Rails.

app/subsystems/course_content/get_course_books.rb
jpslav: I see that Joe's change is different functionally from what it used to be; but as I read what it used to be I don't know why that `select` call was there (I think it really just returns the `entity_book_id` field instead of the full activerecord object). 

All Joe's change does is break it into two lines and remove the `select` call -- `course_books` is not used so isn't particularly needed.

app/routines/reset_practice_widget.rb
Dantemss: That is correct. There is probably a different solution that covers the corner cases better.
- Count > number of unassigned exercises
- Count > number of ALL available exercises

app/subsystems/tasks/assistants/i_reading_assistant.rb
jpslav: Agreed that silent failure here is non-ideal.  @Dantemss what's the best way to error out here?

app/controllers/api/v1/courses_controller.rb
jpslav: Use the other tracks in this file :-)  the ones that look like `api :GET, '/courses/:blah/blah'`

app/subsystems/tasks/assistants/i_reading_assistant.rb
Dantemss: `TaskExercise` already blows up when exercise is nil.
Maybe just need to add a `fatal_error` to it instead.
We can check it here and either skip (with a HUGE WARNING MESSAGE in the console) or just let it keep blowing up so we know.

app/routines/reset_practice_widget.rb
jpslav: Made it an issue for later https://github.com/openstax/tutor-server/issues/220

app/subsystems/tasks/assistants/i_reading_assistant.rb
Dantemss: The issue where the exercise tags were incorrect and an iReading for the fifth Page would fail has been fixed btw by fixing the tags.

app/subsystems/course_content/get_course_books.rb
kevinburleigh75: Okay - just wanted to make sure we were aware of the implications.

app/subsystems/tasks/assistants/i_reading_assistant.rb
jpslav: https://github.com/openstax/tutor-server/issues/221

joemsak: Does it blow up? When we were working with it we ended up with a mystery nil far away from its point of origin (here)

app/controllers/api/v1/courses_controller.rb
joemsak: I was doing that but then you said to forget about it @jpslav

jpslav: Must have been my evil twin, ignore him.

PR #154, opened 2015-04-03
Get the appropriate role for a user in a course

nathanstitt:
If role provided:
  make sure it belongs to user; fatal :invalid_role if not

If role is not provided:
  Get the users set of roles for the given course, restricted to
  role_type unless that is set to :any (the default)

  If there are no roles, fail with :invalid_user

  If there is one role, return it

  If there are multiple roles:
    If one is a :teacher, return it
    Otherwise fail with fatal_error code: multiple_roles

### DIFF
--- a/app/subsystems/domain/choose_course_role.rb
+++ b/app/subsystems/domain/choose_course_role.rb
@@ -13,7 +13,7 @@
 #     If one is a :teacher, return it
 #     Otherwise fail with fatal_error code: multiple_roles
 #
-class Domain::GetCourseUserRole
+class Domain::ChooseCourseRole
 
   lev_routine express_output: :role
 
--- a/app/subsystems/domain/choose_course_role.rb
+++ b/app/subsystems/domain/choose_course_role.rb
@@ -27,7 +27,7 @@ class Domain::ChooseCourseRole
       fatal_error(code: :invalid_role, message:"Role not found") unless role
       validate_role_membership(course, user, role)
     else
-      roles = roles_for_user(course, user, allowed_role_type: allowed_role_type.to_s)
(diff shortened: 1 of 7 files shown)


app/subsystems/domain/choose_course_role.rb
```diff
@@ -0,0 +1,70 @@
...
+      if teacher_role
+        outputs[:role] = teacher_role
+      else
+        fatal_error(code: :multiple_roles, message:"The role must be specified because there is more than one student role available" )
+      end
+    end
+  end
+
+  def roles_for_user(course, user, allowed_role_type: "any")
+    roles = run(Domain::GetUserCourseRoles, course:course, user:user).outputs.roles
```
jpslav: `GetUserCourseRoles` already takes a `types` field, so you can just use that instead of doing the `select` call below.

app/subsystems/domain/choose_course_role.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  protected
+
+  def exec(user:, course:, allowed_role_type: :any, role_id: nil)
+    if role_id
+      role = Entity::Role.find(role_id)
+      fatal_error(code: :invalid_role, message:"Role not found") unless role
+      validate_role_membership(course, user, role)
+    else
+      roles = roles_for_user(course, user, allowed_role_type: allowed_role_type.to_s)
```
jpslav: Due to the comment below, you can leave the role_type as a symbol

nathanstitt:
I've removed the `roles_for_user` method and am now calling Domain::GetUserCourseRoles directly as needed

jpslav:
Thanks @nathanstitt!

PR #155, opened 2015-04-05
IReading spaced practice

kevinburleigh75:
This PR isn't the entire feature, but we need to prevent merge conflicts heading into the end-of-sprint.


app/external/openstax/cnx/v1/fragment/interactive.rb
```diff
@@ -42,8 +42,10 @@ def url
       @url ||= interactive.try(:xpath, 'iframe/@src').to_s
     end
 
-    def to_s(indent: 0)
-      s = "#{' '*indent}INTERACTIVE #{title}\n"
+    def visit(visitor:, depth: 0)
```
jpslav: Might be nice later to move this visit code into a library that takes a "children" option or something so we can just say

```ruby
class Page
  visitable children: :fragments

class Interactive
  visitable
```

etc.  Shouldn't be pre-demo but maybe a cleanup effort right after to remove this otherwise repeated code.

PR #158, opened 2015-04-06
Send tasked_exercise answers to Exchange

nathanstitt:
This modifies the `MarkTaskStepCompleted` routine to call `handle_task_step_completion!` on the `tasked` association of `TaskStep`, just as it does to the `task` association.

TaskedExercise implements the method to submit the answer to Exchange using the `OpenStax::Exchange.record_multiple_choice_answer` method.

Note that TaskedExercise does not handle multiple questions, it only submits the first one.  That's what other code does in the file, so I assumed questions is an array for future expansion and not something that should be currently handled.

### DIFF
--- a/app/routines/mark_task_step_completed.rb
+++ b/app/routines/mark_task_step_completed.rb
@@ -9,10 +9,8 @@ class MarkTaskStepCompleted
     task_step.save
     transfer_errors_from(task_step, {type: :verbatim}, true)
 
-    if task_step.tasked
-      task_step.tasked.try(:handle_task_step_completion!)
-      transfer_errors_from(task_step.tasked, {type: :verbatim}, true)
-    end
+    task_step.tasked.try(:handle_task_step_completion!)
+    transfer_errors_from(task_step.tasked, {type: :verbatim}, true)
 
     task = task_step.task
     task.handle_task_step_completion!
--- a/app/routines/mark_task_step_completed.rb
+++ b/app/routines/mark_task_step_completed.rb
@@ -9,6 +9,11 @@ class MarkTaskStepCompleted
     task_step.save
     transfer_errors_from(task_step, {type: :verbatim}, true)
 
+    if task_step.tasked
+      task_step.tasked.try(:handle_task_step_completion!)
+      transfer_errors_from(task_step.tasked, {type: :verbatim}, true)
+    end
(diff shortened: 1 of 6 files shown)


app/routines/mark_task_step_completed.rb
```diff
@@ -9,6 +9,11 @@ def exec(task_step:, completion_time: Time.now)
     task_step.save
     transfer_errors_from(task_step, {type: :verbatim}, true)
 
+    if task_step.tasked
```
jpslav: `TaskStep` has `validates :tasked, presence: true`, so I don't think this `if` statement is needed.  Also the `.try` below is repetitive given this `if` statement.

nathanstitt: You're correct, the if statement isn't needed as long as the task_step's saved.  Some of the specs where calling it before it'd been saved though so it failed.  Really, the right answer there is to fix the specs :)

The :try is needed though since it's called regardless of the tasked types (reading, video, etc), but only the `TaskedExercise` currently implements the method.  I'm pretty sure that some of other types will need it at some point though which is why I added it to the task_step routine.

jpslav: I think the task_step is saved on line 9 above?  Maybe it just needs to be reloaded.

jpslav: And sounds good on the `try`

nathanstitt:
Turned out it was an unrelated bit of the spec that wasn't saving the task_step.  I've fixed that up and removed the check

Dantemss:
https://exchange-dev1.openstax.org/ is now live so this PR can finally be tested with real data. Setup:
1. Login as the Exchange admin (emailed you the PW)
2. Add a Platform (you will have to click the Add Application to create the associated Application, or just create the Application separately beforehand)
3. Enter some name, client id and secret for Application. Redirect URI is most likely not needed, since the Multiple Choice submission API uses the OAuth Client Credentials flow.
4. Put the Exchange client id and secret in your tutor secrets.yml
5. Make tutor use the real Exchange client
6. Make sure to VCR the responses, cause you might not be able to create the same event with all the same fields twice.

Dantemss:
Let me know if there are any problems with the above steps.

PR #176, opened 2015-04-08
Make API docs look nicer and include all properties

nathanstitt:


nathanstitt:
Attributes are not being included because they needed to be explicitly set to :readable/:writable.  PR opened in the API to correct. https://github.com/openstax/openstax_api/pull/23

Dantemss:
Alright openstax_api 5.2.1 has been released, so need to change to that version.
Also, some controller actions don't include the json schema in the description. Need to add them like so:
```rb
  description <<-EOS
    Description here if any...
    #{json_schema(Api::V1::WhateverRepresenter, include: :readable/writeable)}
  EOS
```

Dantemss:
Also not sure what to do about `courses_controller#practice`... we just wanna skip the docs for that.
Maybe we should just change the route (and specs) to directly point to practice_post and practice_get and remove that method...

jpslav:
Let's make this change.  Can you do Dante?

PR #179, opened 2015-04-08
Remigrate schema and fix sprint tasks

Dantemss:


jpslav:
@Dantemss I've changed my mind; can we leave the sprint 8 rake tasks in case anyone is still testing with them (e.g. Greg)?  Also sometimes we might need things like `conditions` which haven't found a more permanent home.

Dantemss:
@jpslav we can, but they are broken. We changed some models.

Dantemss:
Done. Sprint 8 tasks are still in.

PR #185, opened 2015-04-10
Add chapter section to API (Story u555)

karenc:
- [x] Percolate “5.2” into appropriate places. “Chapter.section” percolate into API parts ---- TaskedReadingRepresenter

- [x] Percolate “5.2” into appropriate places. “Chapter.section” percolate into API parts --- /courses/1/readings endpoint

- [x] Percolate “5.2” into appropriate places. “Chapter.section” percolate into API parts ---- list in exercise in HW picker endpoint


karenc:
This is ready for review.

app/representers/api/v1/book_part_toc_representer.rb
```diff
@@ -28,6 +28,15 @@ class BookPartTocRepresenter < Roar::Decorator
                description: 'The type of the TOC entry, either "part" (for units, chapters, etc), or "page"'
              }
 
+    property :path,
+             as: :chapter_section,
```
jpslav: Was "chapter_section" a name requested by FE?  (or maybe just derived from the story name?)

karenc: Yeah just from the story

app/subsystems/content/get_exercise.rb
```diff
@@ -6,7 +6,7 @@ class Content::GetExercise
 
   def exec(id:)
     ex = Content::Models::Exercise.find(id)
-    outputs[:exercise] = OpenStax::Exercises::V1::Exercise.new(ex.content)
+    outputs[:exercise] = OpenStax::Exercises::V1::Exercise.new(ex.content, path: ex.path)
```
jpslav: I'm not super keen on putting the path into `OpenStax::Exercises::V1::Exercise`, because it really isn't something native to OpenStax Exercises content.  On the flip side, I realize that we're double booking this class as the interface class for the Content subsystem.  @Dantemss would your Content::Exercise wrapper class be able to take on the responsibility for knowing the path so that we can take it out of `OpenStax::Exercises::V1::Exercise`?

app/subsystems/content/routines/import_exercises.rb
```diff
@@ -10,12 +10,13 @@ class Content::Routines::ImportExercises
...
   # TODO: make this import only exercises from trusted authors
   #       or in some trusted list (for when OS Exercises is public)
-  def exec(query_hash)
+  def exec(query_hash, page: page)
     outputs[:exercises] = []
     OpenStax::Exercises::V1.exercises(query_hash)['items'].each do |wrapper|
       exercise = Content::Models::Exercise.find_or_initialize_by(url: wrapper.url)
       exercise.title = wrapper.title
       exercise.content = wrapper.content
+      exercise.path = page.try(:path)
```
jpslav: While it is probably unlikely while we are controlling all of the content, I have to imagine it is possible (even for us), that one exercise will be associated with more than one page in the book.  Since `path` isn't an inherent property to an OpenStax Exercises exercise, this means that this line could conceivably be overwriting a `path` set earlier during the import.  Makes me wonder if the path should just be a tag in Content::Models::Exercise -- that would be more flexible and would allow for multiple paths per exercise.... thoughts @karenc and @Dantemss?  Then we could, in the `ExerciseRepresenter` spit out all of the tags and the FE could do what it wants with them.  Alternatively we could have a more generic "sourced_from" array containing things like "College Physics Section 5.2".

app/subsystems/content/routines/import_exercises.rb
```diff
@@ -10,12 +10,13 @@ class Content::Routines::ImportExercises
 
   # TODO: make this import only exercises from trusted authors
   #       or in some trusted list (for when OS Exercises is public)
-  def exec(query_hash)
+  def exec(query_hash, page: page)
```
jpslav: The `page: page` argument at the end doesn't look right -- where is the second `page` coming from?

app/subsystems/content/routines/import_page.rb
```diff
@@ -32,7 +32,7 @@ def exec(cnx_page:, book_part:)
     run(:tag, outputs[:page], cnx_page.los, tag_type: :lo)
 
     # Get Exercises from OSE that match the LO's
-    run(:import_exercises, tag: cnx_page.los)
+    run(:import_exercises, { tag: cnx_page.los }, page: outputs[:page])
```
jpslav: Depending on changes elsewhere, might want to pass in a smaller, more specific argument here other than the full `page`, e.g. just the location information that the import routine needs.

app/subsystems/content/routines/import_exercises.rb
karenc: oops, i don't know where that came from...

app/subsystems/content/routines/import_exercises.rb
karenc: If an exercise can be associated with more than one page, then we need to do something else.  I didn't think about that.  Is it possible to know what to return in the API?  How does FE know?

app/subsystems/content/routines/import_page.rb
karenc: I can just pass in `outputs[:page].path`.

jpslav: hold up, new instructions coming :-)

jpslav:
@karenc let me know when this is in line with https://github.com/openstax/napkin-notes/commits/master/jp/more_tag_data.md  Thanks!

karenc:
Ready for review

app/external/openstax/cnx/v1/page.rb
```diff
@@ -72,8 +73,8 @@ def title
       }
     end
 
-    def path
-      @path
+    def chapter_section
```
jpslav: Believe this duplicates the `attr_reader :hash, :chapter_section` statement on line 57

app/subsystems/content/routines/create_tags.rb
```diff
@@ -0,0 +1,33 @@
...
+    #     name: "<taken from ost-learning-objective-def>",
+    #     teks: "<teks-class>"
+    #   }
+    # }
+    outputs[:tags] = []
+    tag_defs.each do |tag_id, args|
+      tag = Content::Models::Tag.find_or_initialize_by(tag_id: tag_id.to_s)
+      teks = args.delete(:teks)
+      args.each do |key, value|
+        old_value = tag.send(key)
```
jpslav: This `each` loop doesn't seem to do anything useful -- remove?

app/representers/api/v1/book_part_toc_representer.rb
```diff
@@ -28,6 +28,14 @@ class BookPartTocRepresenter < Roar::Decorator
                description: 'The type of the TOC entry, either "part" (for units, chapters, etc), or "page"'
              }
 
+    property :chapter_section,
+             writeable: false,
+             readable: true,
+             schema_info: {
+               required: true,
+               description: 'The chapter and section in the book'
```
jpslav: Would be helpful to end this description like:

```ruby
description: 'The chapter and section in the book, e.g. "5.2"'
```

Ditto for other similar descriptions.

app/representers/api/v1/tasked_exercise_representer.rb
```diff
@@ -22,6 +22,15 @@ class TaskedExerciseRepresenter < Roar::Decorator
                description: "The title of this Exercise"
              }
 
+    property :chapter_section,
+             type: String,
+             writeable: false,
+             readable: true,
+             schema_info: {
+               required: true,
+               description: "The chapter and section in the book"
```
jpslav: Add "e.g." like above

app/routines/recover_tasked_exercise.rb
```diff
@@ -48,7 +48,7 @@ def create_task_step_after(task_step:, exercise:)
   # Finds an Exercise with all the required tags and at least one LO
   # Prefers unassigned Exercises
   def get_recovery_exercise_for(tasked_exercise:,
-                                required_tag_names: ['practice-problem'])
+                                required_tag_ids: ['practice-problem'])
```
jpslav: nice rename :-)

Dantemss: `required_tags`?

jpslav: I think this just makes it explicit that we want the string IDs (now tags have more fields to them, so `_names` is definitely not appropriate).  Either way is fine tho.

jpslav: Actually the more I think about it, using `id` in the backend to mean anything other than the primary key is probably not a great idea -- someone might easily think the value of `required_tag_ids` should be `[42, 23, 962]`.  Calling the string value an `id` in the representer is totally ok on the other hand.  That is just for the frontend and they can think of the raw value as the ID.

Can we rename `tag_id` to `value` or `raw` or `raw_value`?  Then we'd have  `value` (machine readable) and `name` (human readable).  Then this `required_tag_ids` could be `required_tag_values` or Dante's idea of just `required_tags`.

app/routines/task_exercise.rb
```diff
@@ -4,10 +4,11 @@ class TaskExercise
 
   protected
 
-  def exec(exercise:, title: nil, can_be_recovered: false, task_step: nil)
+  def exec(exercise:, chapter_section: nil, title: nil, can_be_recovered: false, task_step: nil)
     outputs[:tasked_exercise] = Tasks::Models::TaskedExercise.new(
       task_step: task_step,
       exercise: (exercise.is_a?(Content::Models::Exercise) ? exercise : nil),
+      chapter_section: chapter_section,
```
jpslav: Do we have a need to know which exercises are in certain `chapter_section`s after they have been assigned to a particular `Task`?  I thought we just needed it when teachers were making plans for homeworks.

/cc @philschatz @jivewise

app/subsystems/content/models/exercise.rb
```diff
@@ -6,4 +6,11 @@ class Content::Models::Exercise < Tutor::SubSystems::BaseModel
   has_many :tasked_exercises, subsystem: :tasks,
                               dependent: :destroy
   protected :tasked_exercises
+
+  def tags
+    exercise_tags.collect do |et|
```
jpslav: We're doing this collect in a couple of places.  Can you or @Dantemss add direct access to `Tag` from `Exercise` or `Page` using a `has_many :through`?

app/subsystems/content/models/lo_teks_tag.rb
```diff
@@ -0,0 +1,7 @@
+class Content::Models::LoTeksTag < Tutor::SubSystems::BaseModel
+  belongs_to :lo, class_name: 'Tag', foreign_key: :lo_id
```
jpslav: Wonder why you don't have to say `class_name: 'Content::Models::Tag'` here...

app/subsystems/content/models/page_data_visitor.rb
```diff
@@ -23,18 +23,18 @@ def output
   private
   def get_page_path(page)
     page.page_tags.collect do |page_tag|
-      { type: page_tag.tag.tag_type, name: page_tag.tag.name }
+      { type: page_tag.tag.tag_type, tag_id: page_tag.tag.tag_id }
```
jpslav: I'm liking more and more renaming `tag_id` to value -- `tag.value` here at the end will be better.

app/subsystems/content/models/exercise.rb
jpslav: This can be in a separate PR.

app/subsystems/content/models/page_data_visitor.rb
```diff
@@ -23,18 +23,18 @@ def output
...
   end
 
   def get_page_los(page)
     tags = get_page_tags(page)
-    tags.select { |tag| tag[:type] == 'lo' }.collect { |tag| tag[:name] }
+    tags.select { |tag| tag[:type] == 'lo' }.collect { |tag| tag[:tag_id] }
   end
 
   def get_page_tags(page)
     page.page_tags.collect do |page_tag|
```
jpslav: @Dantemss fwiw, another place where we need the has many through

app/subsystems/content/routines/import_page.rb
```diff
@@ -24,10 +27,13 @@ def exec(cnx_page:, book_part:)
                       title: cnx_page.title,
                       content: cnx_page.converted_content,
                       book_part: book_part,
-                      path: cnx_page.path)
+                      chapter_section: cnx_page.chapter_section)
     book_part.pages << outputs[:page] unless book_part.nil?
     transfer_errors_from outputs[:page], {type: :verbatim}, true
 
+    # Create tags
+    run(:create_tags, cnx_page.tag_defs)
```
jpslav: Instead of this call, should we just be tagging the page with all tags (instead of just LOs like we are doing below)?  If so, then that suggests that the logic in the `Content::Routines::CreateTags` routine should move into `Content::Routines::TagResource` (or possibly `CreateTag` (singular) is called by `TagResource`).

@Dantemss thoughts?

db/migrate/20150411202409_add_path_to_tasks_tasked_exercises.rb
```diff
@@ -0,0 +1,5 @@
+class AddPathToTasksTaskedExercises < ActiveRecord::Migration
+  def change
+    add_column :tasks_tasked_exercises, :path, :string
```
jpslav: If you'd like, you are welcome to rewrite existing migrations instead of doing `add_column`, `rename_column`, etc.  This is allowed for the time being since we don't have any permanent deployed instance of Tutor (and we always drop the db in development and rebuild).

jpslav:
Done with my review @karenc

app/subsystems/content/models/exercise.rb
karenc: It is definitely possible to get the exercise tags using `has_many`, is it possible to also get the teks tags that are linked to the exercise tags?

app/subsystems/content/routines/import_page.rb
Dantemss: I would argue that `cnx_page.los` is not needed anymore, just use the `tag_defs` method (need to ensure all the los for all pages appear in the tag defs). Also this new `Content::Routines::CreateTags` routine can call the TagResource routine internally.

app/routines/task_exercise.rb
Dantemss: I would argue that we don't need this new DB field. You can access this information using `task_step.page.path`.

jivewise: I believe we still need to know the chapter_section (note the table with the blue heading here: http://mockups.openstax.org/Tutor_MVP_v2/add_homework_3.html) for when a teacher is reviewing or editng a homework that they've created.

jpslav: Chatted with @jivewise and folks -- we don't need to store the `chapter_section` in the `TaskedExercise`.

app/subsystems/content/routines/import_page.rb
jpslav: @karenc I made a PR off of your branch to record my thoughts about this part of the code and its tentacles: https://github.com/openstax/tutor-server/pull/206

I think there are a few cases where `cnx_page.los` is still needed, see the PR for how it now just uses the `tag_defs` call.  Also, the PR has changes for `TagResource` etc.

app/subsystems/content/models/exercise.rb
Dantemss: The new Exercise entity gives direct access to the tags and los (the string, not the model). Obviously, TEKS weren't implemented when I did it.

app/subsystems/content/models/lo_teks_tag.rb
Dantemss: Thought we assumed the subsystem to be the same as the current model by default, and you had to specify :none to disable it?

app/subsystems/content/models/exercise.rb
jpslav: Cool, then we can wait to use Dante's new stuff later.

app/subsystems/content/models/lo_teks_tag.rb
jpslav: We do I think, but here karen had to specify the names manually since they can't be inferred from the association name.  Still don't know why don't have specify full class name here.  maybe some rails autoloading thing that looks in same namespace first?

karenc:
Rebased and merged JP's suggestion.

karenc:
The last commit changes the "content" in the exercises api to return a hash instead of a json string.

app/external/openstax/cnx/v1/page.rb
```diff
@@ -132,9 +129,58 @@ def root
...
+      # teks tags
+      doc.css('[class^="ost-standards-def"]').each do |node|
+        name = node.css('[class^="ost-standards-name"]').first.try(:content).try(:strip)
+        description = node.css('[class^="ost-standards-description"]').first.try(:content).try(:strip)
+        tag_value = node.attr('class').split.last
+        @tags[tag_value] = {
+          value: tag_value,
+          name: name,
+          description: description,
+          type: :generic
```
Dantemss: Might wanna add a TEKS type, though that can be done later...

app/subsystems/content/models/exercise.rb
```diff
@@ -6,4 +6,11 @@ class Content::Models::Exercise < Tutor::SubSystems::BaseModel
   has_many :exercise_tags, dependent: :destroy
 
   has_many :tasked_exercises, subsystem: :tasks, primary_key: :url, foreign_key: :url
+
+  has_many :tags, through: :exercise_tags
+
+  def all_tags
```
Dantemss: This name is a bit misleading...

app/models/exercise.rb
```diff
@@ -2,14 +2,14 @@ class Exercise < Entity
 
   wraps Content::Models::Exercise
 
-  exposes :url, :title, :content
+  exposes :url, :title, :content, :all_tags
```
Dantemss: Not sure if you want to expose these arrays of Content::Models::Tag to the outside world...

karenc: It's so I can use it in the representer (I think).

app/external/openstax/cnx/v1/page.rb
karenc: Yeah I thought about it.  I'll add a TEKS type.

app/subsystems/content/models/exercise.rb
karenc: It was just called "tags" before, but there's now already one in app/models/exercise.rb.  Any suggestions?

Dantemss: I'm bad with names, but `tags_with_teks`? Idk, ask JP.

app/models/exercise.rb
Dantemss: I think the representer currently assumes a Content::Models::Exercise so it's probably fine to leave the `all_tags` part out. But if removing it doesn't work, we can take care of it later.

Dantemss:
A few comments above, but otherwise looks good.

app/models/exercise.rb
karenc: Removing it causes this test to fail:

```
   1) Api::V1::CoursesController#exercises should work on the happy path
      Failure/Error: api_get :exercises, user_1_token, parameters: {id: course.id,
      NoMethodError:      
        undefined method `all_tags' for #<Exercise:0x0000000914f560>
```
So I'll just leave this for now.

karenc:
Changes made: (plus some changes in tests not shown)

```diff
diff --git a/app/external/openstax/cnx/v1/page.rb b/app/external/openstax/cnx/v1/page.rb
index 5737333..62648db 100644
--- a/app/external/openstax/cnx/v1/page.rb
+++ b/app/external/openstax/cnx/v1/page.rb
@@ -158,7 +158,7 @@ module OpenStax::Cnx::V1
           value: tag_value,
           name: name,
           description: description,
-          type: :generic
+          type: :teks
         }
       end
 
diff --git a/app/models/exercise.rb b/app/models/exercise.rb
index b356e14..b5c620e 100644
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -2,7 +2,7 @@ class Exercise < Entity
 
   wraps Content::Models::Exercise
 
-  exposes :url, :title, :content, :all_tags
+  exposes :url, :title, :content, :tags_with_teks
 
   def tags
     tag_models.collect{ |t| t.value }
diff --git a/app/representers/api/v1/exercise_representer.rb b/app/representers/api/v1/exercise_representer.rb
index cefa596..2dccc80 100644
--- a/app/representers/api/v1/exercise_representer.rb
+++ b/app/representers/api/v1/exercise_representer.rb
@@ -16,7 +16,7 @@ module Api::V1
                required: true
              }
 
-    collection :all_tags,
+    collection :tags_with_teks,
                as: :tags,
                readable: true,
                writeable: false,
diff --git a/app/representers/api/v1/task_plan_search_representer.rb b/app/representers/api/v1/task_plan_search_representer.rb
index f37f7fb..a79d974 100644
--- a/app/representers/api/v1/task_plan_search_representer.rb
+++ b/app/representers/api/v1/task_plan_search_representer.rb
@@ -2,7 +2,7 @@ module Api::V1
   class TaskPlanSearchRepresenter < OpenStax::Api::V1::AbstractSearchRepresenter
 
     collection :items, inherit: true,
-                       class: Tasks::Models::TaskPlan,
+                       class: ::Tasks::Models::TaskPlan,
                        decorator: TaskPlanRepresenter
 
   end
diff --git a/app/subsystems/content/models/exercise.rb b/app/subsystems/content/models/exercise.rb
index 6b3362c..b43a561 100644
--- a/app/subsystems/content/models/exercise.rb
+++ b/app/subsystems/content/models/exercise.rb
@@ -9,7 +9,7 @@ class Content::Models::Exercise < Tutor::SubSystems::BaseModel
 
   has_many :tags, through: :exercise_tags
 
-  def all_tags
+  def tags_with_teks
     # Include tek tags
     tags.collect { |t| [t, t.teks_tags] }.flatten.uniq
   end
diff --git a/app/subsystems/content/models/tag.rb b/app/subsystems/content/models/tag.rb
index e796899..7f1f26e 100644
--- a/app/subsystems/content/models/tag.rb
+++ b/app/subsystems/content/models/tag.rb
@@ -5,7 +5,7 @@ class Content::Models::Tag < Tutor::SubSystems::BaseModel
   has_many :teks_tags, through: :lo_teks_tags, class_name: 'Tag', source: :teks
 
   # List the different types of tags
-  enum tag_type: [ :generic, :lo ]
+  enum tag_type: [ :generic, :teks, :lo ]
 
   validates :value, presence: true
   validates :tag_type, presence: true
diff --git a/spec/external/openstax/cnx/v1/page_spec.rb b/spec/external/openstax/cnx/v1/page_spec.rb
index 01c86c1..d7ba556 100644
--- a/spec/external/openstax/cnx/v1/page_spec.rb
+++ b/spec/external/openstax/cnx/v1/page_spec.rb
@@ -83,7 +83,7 @@ RSpec.describe OpenStax::Cnx::V1::Page, :type => :external, :vcr => VCR_OPTS do
                   },
                   {
                     value: 'ost-tag-teks-112-39-c-4c',
-                    type: :generic,
+                    type: :teks,
                     name: '4C',
                     description: 'analyze and describe accelerated motion in two dimensions using equations, including projectile and circular examples'
                   }
@@ -116,7 +116,7 @@ RSpec.describe OpenStax::Cnx::V1::Page, :type => :external, :vcr => VCR_OPTS do
                   },
                   {
                     value: 'ost-tag-teks-112-39-c-4d',
-                    type: :generic,
+                    type: :teks,
                     name: '(D)',
                     description: 'calculate the effect of forces on objects, including the law of inertia, the relationship between force and acceleration, and the nature of force pairs between objects'
                   }
```

Dantemss:
Looks good to me

jpslav:
Love it, thanks for all the hard work @karenc!

PR #187, opened 2015-04-10
Homework stats

Dantemss:
# Done:
- Added grade to GET plans API
- Decide who is responsible for hiding the grade for iReadings (have fun, FE)
- Fixed/improved specs


Dantemss:
The code adds the grade for both HW and iReading. We should decide if the BE is responsible for hiding the iReading grade, or the FE.

Dantemss:
Also need to decide on the value returned if there are no gradable (completed) exercises. Could be nil (i.e. key is absent from JSON, 0 or 100).

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -39,8 +39,9 @@ class TaskPlanRepresenter < Roar::Decorator
              writeable: true
 
     property :stats,
-             extend: IReadingStatsRepresenter,
-             getter: ->(args){ args[:stats] },
+             extend: TaskPlanStatsRepresenter,
+             getter: ->(args){ CalculateTaskPlanStats.call(plan: self).outputs.stats },
```
jpslav:
```ruby
getter: -> (*) { CalculateTaskPlanStats[plan: self] }
```

jpslav:
I'm ok with the BE providing extra info that the FE doesn't show.

app/representers/api/v1/task_plan_representer.rb
Dantemss: :)

Dantemss:
@jpslav ready for review

jpslav:
Looks good @Dantemss, gonna merge.  In the back of your mind can you start thinking about how to cache this stuff so we don't have to compute from scratch every time?

PR #188, opened 2015-04-13
Make user an admin

joemsak:
https://www.pivotaltracker.com/story/show/92101070


spec/factories/user_profile.rb
```diff
@@ -12,13 +12,13 @@
     end
 
     after(:build) do |user, evaluator|
-      user.account = FactoryGirl.build(:openstax_accounts_account,
```
Dantemss: Thanks for the `||=`

PR #195, opened 2015-04-15
Entity base class and Exercise entity

Dantemss:
# Done
- Wrapping a single object
- Automatically wrapping relations
- Old specs
- Comments on Entity class
- Decision about the namespaces
- New specs


app/controllers/api/v1/courses_controller.rb
```diff
@@ -55,9 +55,9 @@ def exercises
                                               course)
 
     los = Content::GetLos[params]
-    outputs = SearchLocalExercises.call(tag: los, match_count: 1).outputs
+    results = SearchLocalExercises.call(tag: los, match_count: 1).outputs
```
jpslav: Why rename this var?  Seems a little confusing to use the word we normally use for what comes out of a `call` for that things' `outputs`.

app/external/openstax/exercises/v1/exercise.rb
```diff
@@ -41,22 +41,33 @@ def questions
...
+  def question_answer_ids
+    @question_answer_ids ||= question_answers.collect do |q|
+      q.collect{|a| a['id'].to_s}
+    end
+  end
+
+  def correct_question_answers
+    @correct_question_answers ||= question_answers.collect do |q|
+      q.select do |a|
+        (Float(a['correctness']) rescue 0) >= 1
```
jpslav: you go, man

app/controllers/api/v1/courses_controller.rb
Dantemss: Right... not sure what I was doing here... I'll just undo it.

app/models/exercise.rb
```diff
@@ -0,0 +1,44 @@
+class Exercise < Entity
+  wraps Content::Models::Exercise
+
+  exposes :url, :title, :content
+
+  # Could simplify some of these by building question and answer wrappers
+  # (POROs, not Entity subclasses, since no DB object for them)
+  delegate :uid, :questions, :question_formats, :question_answers, :question_answer_ids,
+           :correct_question_answers, :correct_question_answer_ids,
+           :content_without_correctness, to: :wrapper
```
jpslav: A little confusing to have a thing called `wrapper` that isn't this class.  (This class, `Exercise` wraps `Content::Models::Exercise` but isn't the "`wrapper`").  Rename `wrapper` to `content_wrapper`?

app/models/exercise.rb
```diff
@@ -0,0 +1,44 @@
+class Exercise < Entity
```
jpslav: My initial reaction is that this class be inside the `Content` subsystem because it has knowledge of things in `Content::Models`.  It would be `Content::Exercise` and would be accessible outside of the `Content` SS.

app/models/exercise.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  exposes :url, :title, :content
+
+  # Could simplify some of these by building question and answer wrappers
+  # (POROs, not Entity subclasses, since no DB object for them)
+  delegate :uid, :questions, :question_formats, :question_answers, :question_answer_ids,
+           :correct_question_answers, :correct_question_answer_ids,
+           :content_without_correctness, to: :wrapper
+
+  def self.search(options = {})
```
jpslav: I don't see any reason to have this search method here.  Better to just keep these entities as wrappers of data that is otherwise hidden inside a SS, and then have callers just use top-level routines directly.

app/models/exercise.rb
```diff
@@ -0,0 +1,44 @@
...
+  protected
+
+  # To remove the dependency on the wrapper,
+  # we would have to store the parsed content instead of just storing the JSON
+  def wrapper
+    @wrapper ||= OpenStax::Exercises::V1::Exercise.new(content)
+  end
+
+  def tag_models
+    repository.exercise_tags.includes(:tag).collect{ |et| et.tag }
```
jpslav: Can there be a `has_many :through` here to get straight to the `Tag` objects?

app/models/exercise.rb
jpslav: I can maybe be convinced otherwise tho...

jpslav:
@kevinburleigh75 please review too.

app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -2,42 +2,16 @@ class Tasks::Models::TaskedExercise < Tutor::SubSystems::BaseModel
   acts_as_tasked
 
   belongs_to :exercise, subsystem: :content
```
jpslav: Isn't this the direct link we said we probably didn't want to have?

Dantemss: The whole goal of this PR is to get the Domain code out of the AR classes. The DB can do whatever is necessary behind the scenes. Relations like this one will be hidden (or not) by the wrapper.

app/models/exercise.rb
Dantemss: We can discuss :)

app/models/exercise.rb
Dantemss: I renamed it to `parser`.

app/models/exercise.rb
Dantemss: We can discuss this too :)

app/models/exercise.rb
Dantemss: Possibly.

app/representers/api/v1/tasked_exercise_representer.rb
```diff
@@ -27,20 +27,23 @@ class TaskedExerciseRepresenter < Roar::Decorator
              type: String,
              writeable: false,
              readable: true,
+             getter: -> (*) { Exercise.new(exercise).content_without_correctness },
```
jpslav: Seems a little weird to me to go through `Exercise` in this way.  Also as noted earlier having the `exercise` relation point directly to a `Content` model is an issue and maybe why this is weird.  Might be less weird if `exercise` returned an `::Exercise` and the block here was `{ exercise.content_without_correctness }`.  However, I'm not quite sure about the direction of the dependencies -- i.e. now the new entities use subsystem models, but is it ok for those models to also use the entities?

app/representers/api/v1/tasked_exercise_representer.rb
```diff
@@ -62,18 +65,19 @@ class TaskedExerciseRepresenter < Roar::Decorator
...
 
-    property :can_be_recovered,
-             as: :has_recovery,
-             type: 'boolean',
+    property :correct_answer_id,
+             type: String,
              writeable: false,
              readable: true,
              if: -> (*) { task_step.completed? },
+             getter: -> (*) { Exercise.new(exercise).correct_question_answer_ids[0][0] },
```
jpslav: The `[0][0]` bit seems like a bit of a red flag.

app/routines/recover_tasked_exercise.rb
```diff
@@ -8,19 +8,15 @@ class RecoverTaskedExercise
   protected
 
   def exec(tasked_exercise:)
-    fatal_error(code: :recovery_not_available) \
-      unless tasked_exercise.can_be_recovered?
+    fatal_error(code: :recovery_not_available) unless tasked_exercise.can_be_recovered?
```
jpslav: you're a good man charlie brown.

app/representers/api/v1/tasked_exercise_representer.rb
jpslav: @kevinburleigh75 - thoughts?

app/routines/recover_tasked_exercise.rb
```diff
@@ -50,16 +46,13 @@ def create_task_step_after(task_step:, exercise:)
   def get_recovery_exercise_for(tasked_exercise:,
                                 required_tag_names: ['practice-problem'])
 
-    ex_relation = Content::Models::Exercise.preload(exercise_tags: :tag)
-
     # Randomize LO order
-    los = tasked_exercise.wrapper.los.shuffle
+    los = Exercise.new(tasked_exercise.exercise).los.shuffle
```
jpslav: Weird in the same way the above is -- probably just due to the fact that this top-level routine is being passed a private SS model.  In the end game do you see there being an Entity wrapper for `Tasks::Models::TaskedExercise`?  That might make this routine and this call in particular cleaner.

app/routines/recover_tasked_exercise.rb
```diff
@@ -50,16 +46,13 @@ def create_task_step_after(task_step:, exercise:)
...
 
     # Try to find unassigned exercises first
     taskees = tasked_exercise.task_step.task.taskings.collect{|t| t.role}
     los.each do |lo|
```
jpslav: Still would be good to put these two search calls somewhere on their own (said this before in https://github.com/openstax/tutor-server/pull/148/files#r27745163)

app/routines/recover_tasked_exercise.rb
```diff
@@ -50,16 +46,13 @@ def create_task_step_after(task_step:, exercise:)
...
                                 required_tag_names: ['practice-problem'])
 
-    ex_relation = Content::Models::Exercise.preload(exercise_tags: :tag)
-
     # Randomize LO order
-    los = tasked_exercise.wrapper.los.shuffle
+    los = Exercise.new(tasked_exercise.exercise).los.shuffle
 
     # Try to find unassigned exercises first
     taskees = tasked_exercise.task_step.task.taskings.collect{|t| t.role}
```
jpslav: This line could be a battle for another day, but seems like the `Tasks` SS could provide a routine to get the taskees for a task or a component of a task.  This is too much SS knowledge here.  The solution could be something similar to `Tasks::DoesTaskingExist`.

jpslav: https://github.com/openstax/tutor-server/issues/200

app/routines/recover_tasked_exercise.rb
jpslav: The strictness thing I mentioned would make it so that you wouldn't have to call the search twice like is happening here -- instead, inside the search routine it'd try first with the restriction that the exercises have not been assigned; then if that didn't pan out, it would search without that restriction (that's why the restriction was "loose")

app/subsystems/content/get_exercise.rb
```diff
@@ -6,7 +6,7 @@ class Content::GetExercise
 
   def exec(id:)
     ex = Content::Models::Exercise.find(id)
-    outputs[:exercise] = OpenStax::Exercises::V1::Exercise.new(ex.content)
+    outputs[:exercise] = Exercise.new(ex)
```
jpslav: nice

app/routines/recover_tasked_exercise.rb
Dantemss: Doing that right now.

app/representers/api/v1/tasked_exercise_representer.rb
Dantemss: Moved to TaskedExercise wrapper in the next commit.

app/routines/recover_tasked_exercise.rb
Dantemss: This routine can be moved into the Task SS, I think.

app/routines/recover_tasked_exercise.rb
Dantemss: Ah I see... well could you make an issue for that? :)

app/subsystems/course_membership/models/entity_extensions.rb
```diff
@@ -1,2 +1,4 @@
+require_dependency '../models/entity/role'
+
 Entity::Role.has_many :students, subsystem: :course_membership, foreign_key: 'entity_role_id'
```
jpslav: I know we had this discussion in another PR, and it is probably me who dropped the ball on it, but I still believe that the default foreign key for `has_many` SS associations is wrong.  I'm making a PR.

lib/entity.rb
```diff
@@ -0,0 +1,105 @@
+require 'entity/common'
+require 'entity/relation'
+
+class Entity
```
jpslav: One day promise me you'll comment this bad boy, because yeah I understand what's going on, sure, of course, crystal clear ;-)

Dantemss: I can comment before we merge...

app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -15,14 +15,14 @@ class Api::V1::TaskStepsController < Api::V1::ApiController
   api :GET, '/steps/:step_id', 'Gets the specified TaskStep'
   def show
     task_step = Tasks::Models::TaskStep.find(params[:id])
-    tasked = task_step.tasked
+    tasked = Task::TaskedExercise.temp_hack(task_step.tasked)
```
Dantemss: These lines can be removed once we wrap the entire Task structure...

app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -39,7 +39,7 @@ def self.schema
   def self.add_exercise_step(task:, exercise:)
     step = Tasks::Models::TaskStep.new(task: task)
 
-    TaskExercise[task_step: step, exercise: exercise]
+    Task::TaskedExercise.new(task_step: step, exercise: exercise)
```
Dantemss: This will eventually be a method on either the Task wrapper or the TaskedExercise wrapper that does this...

app/subsystems/course_membership/models/entity_extensions.rb
```diff
@@ -1,2 +1,4 @@
+require_dependency '../models/entity/role'
```
Dantemss: This is necessary because of the name conflict (of Entity). We should decide on what namespaces we want.

app/models/task/tasked_exercise.rb
```diff
@@ -0,0 +1,90 @@
...
+           :content_without_correctness, to: :parser
+
+  # These 2 methods (hacks) will eventually move to Task when we have wrappers for those
+  # so that this class is not directly exposed
+  def initialize(t = {})
+    t.is_a?(Tasks::Models::TaskedExercise) ? super(t) : super(TaskExercise[t])
+  end
+
+  def self.temp_hack(t)
+    t.is_a?(Tasks::Models::TaskedExercise) ? new(t).passthrough : t
```
Dantemss: Passthrough gives you an Entity that exposes all methods from the repo. I did that instead of CRUD entity. We can discuss which is preferable. Regardless of passthrough, the entity still has to delegate `:transaction` for it to work (because `:transaction` is a class method).

Dantemss:
@jpslav ready for review round 2. I updated the comment at the top with the TODO's for this PR.

app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -49,22 +47,18 @@ def completed
...
   def recovery
-    tasked = Tasks::Models::TaskStep.find(params[:id]).tasked
-    OSU::AccessPolicy.require_action_allowed!(:recover,
-                                              current_api_user,
-                                              tasked)
+    tasked = Task::TaskedExercise.temp_hack(Tasks::Models::TaskStep.find(params[:id]).tasked)
+    OSU::AccessPolicy.require_action_allowed!(:recover, current_api_user, tasked)
 
-    result = RecoverTaskedExercise.call(tasked_exercise: tasked)
+    result = tasked.recover
```
jpslav: I prefer to just call the routines instead of calling a method that calls the routine.  I don't see a benefit to doing it this way, but I do see a detriment as we are entangling this wrapper class with a number of routines that don't really have any connection to it.

app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -74,14 +68,17 @@ def recovery
     #{json_schema(Api::V1::RefreshRepresenter, include: :readable)}
   EOS
   def refresh
-    tasked = Tasks::Models::TaskStep.find(params[:id]).tasked
+    tasked = Task::TaskedExercise.temp_hack(Tasks::Models::TaskStep.find(params[:id]).tasked)
     OSU::AccessPolicy.require_action_allowed!(:refresh, current_api_user, tasked)
 
-    result = RefreshTaskedExercise.call(tasked_exercise: tasked)
+    result = tasked.refresh
```
jpslav: Ditto above.

app/models/exercise.rb
```diff
@@ -0,0 +1,25 @@
+class Exercise < Entity
+
+  wraps Content::Models::Exercise
```
jpslav: Can one of these entities wrap more than one SS model?

app/models/exercise.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  exposes :url, :title, :content
+
+  def self.search(options = {})
+    SearchLocalExercises[options]
```
jpslav: So long, farewell, auf Wiedersehen, good night

app/models/task/tasked_exercise.rb
```diff
@@ -0,0 +1,94 @@
...
+          :answer_id, :answer_ids, :free_response, :transaction
+
+  delegate :uid, :questions, :question_formats, :question_answers, :question_answer_ids,
+           :correct_question_answers, :correct_question_answer_ids,
+           :content_without_correctness, to: :parser
+
+  # These 2 methods (hacks) will eventually move to Task when we have wrappers for those
+  # so that this class is not directly exposed
+  def initialize(t = {})
+    t.is_a?(Tasks::Models::TaskedExercise) ? super(t) : super(TaskExercise[t])
```
jpslav: Can you explain what's going on here?  I'm confused.

Also, we probably need to change the name of the `TaskExercise` routine -- it is just too close to `TaskedExercise`

app/controllers/api/v1/task_steps_controller.rb
Dantemss: But they do have a connection: these routines (which should be moved into the Task subsystem) only accept the repository class as an argument, and that is not exposed to the domain layer. The methods perform the conversion.

app/models/exercise.rb
Dantemss: Not currently. I'm trying to think of ways... Maybe passing a block to perform the translation?

app/models/exercise.rb
Dantemss: Fine if you really like your domain routines so much I suppose I can make them return Entities and remove those methods that call them. (instead of moving them into the subsystems)

app/models/task/tasked_exercise.rb
Dantemss: Don't worry, this is a temporary hack to bridge the gap between the `TaskedExercise` (`Task::Exercise`?) and the rest of the Task components, which are not wrapped.

app/models/task/tasked_exercise.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def recover
+    RecoverTaskedExercise.call(tasked_exercise: repository)
+  end
+
+  def refresh
+    RefreshTaskedExercise.call(tasked_exercise: repository)
+  end
+
+  def handle_task_step_completion!
```
jpslav: This should definitely not be here!  My hope for wrappers is that they give safe and convenient access to data; they should not have logic unto themselves.  Logic goes in routines, *especially* logic like this that connects out to a 3rd party service.

app/models/task/tasked_exercise.rb
```diff
@@ -0,0 +1,94 @@
...
+    # "trial" is set to only "0" for now.  When multiple
+    # attempts are supported, it will be incremented to indicate the attempt #
+
+    ## Blatant hack below (the identifier *should* be set to
+    ## the exchange identifier in the current user's profile,
+    ## but the role id is a close temporary proxy):
+    OpenStax::Exchange.record_multiple_choice_answer(identifier, url, trial, answer_id)
+  end
+
+  def inject_debug_content(debug_content:, pre_br: false, post_br: false)
```
jpslav: @Dantemss I know you didn't write the code that puts this method here, but along with @kevinburleigh75 we should think about where this kind of spy / debug code lives (kind of would like it to not be here but instead in some library that reopens the class and adds this method in certain environments)

app/models/task/tasked_exercise.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  def can_be_recovered?
+    repository.can_be_recovered
+  end
+
+  protected
+
+  def parser
+    repository.parser
```
jpslav: `repository` --> `wrapped` or `wrapee` to keep with the theme?

app/models/task/tasked_exercise.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def feedback_for(answer_id)
+    parser.feedback_map[answer_id] || ''
+  end
+
+  def answer_is_correct?(answer_id)
+    correct_question_answer_ids.flatten.include?(answer_id)
+  end
+
+  def identifier
```
jpslav: Like the `handle...` code above that calls this, this shouldn't be here.

app/routines/mark_task_step_completed.rb
```diff
@@ -9,7 +9,7 @@ def exec(task_step:, completion_time: Time.now)
     task_step.save
     transfer_errors_from(task_step, {type: :verbatim}, true)
 
-    task_step.tasked.try(:handle_task_step_completion!)
+    Task::TaskedExercise.temp_hack(task_step.tasked).try(:handle_task_step_completion!)
```
jpslav: I like the `temp_hack` name; very clear :-)

jpslav:
Still looking good Dante.  As noted, I'm still in favor of the wrappers being more data wrappers than wrappers to routine calls or holding logic of their own.  The CRUD entity / passthrough bit could use some discussion.  Gotta wait for that until @kevinburleigh75 has a chance to review.

app/models/task/tasked_exercise.rb
Dantemss: I picked `repository` because it was a standard DDD name.

app/models/task/tasked_exercise.rb
Dantemss: Definitely shouldn't be in the AR class either...

Dantemss:
To allow an entity to wrap multiple classes (or a class to be wrapped by multiple entities), we need to solve 2 problems:

1. How to map from the Entity class to the wrapped class for class methods. The solution could be:
```rb
class_exposes :method_name, from: Class
```

or simply (using exposes for both class and instance methods):

```rb
exposes :method_name, from_class: Class
```

2. How to map from the wrapped class to the Entity when wrapping method results. The solution could be (on the wrapped class):

```rb
wrapped_by do |instance|
  case instance
  when instance.type == 'blah'
    FirstEntityClass
  else
    SecondEntityClass
  end
end
```

Let me know your thoughts.

app/models/exercise.rb
jpslav: name each wrapped model and autogenerate `repository_name1`, `repository_name2` methods?

Not important I guess, it was just one reason why these objects should not be in the SS (i.e. if they wrapped multiple classes across SS's)

Dantemss:
All done.

app/models/exercise.rb
Dantemss: See my comment below for ways this can be done... the main issue with 1 Entity wrapping 2+ classes is determining which class receives class methods, though that can be done on the call to `class_exposes`.

app/external/openstax/exercises/v1/exercise.rb
kevinburleigh75: Why is this `rescue` needed?  We already know we're in an `Exercise` that has `question_answers` - does that not imply the existence of 'correctness'?

app/external/openstax/exercises/v1/exercise.rb
```diff
@@ -41,22 +40,33 @@ def questions
...
   end
 
-  def correct_answer_ids
-    @correct_answer_ids ||= answers.collect do |ans|
-      ans.select do |a|
-        correctness = Float(a['correctness']) rescue 0
-        correctness >= 1
-      end.first['id']
+  def question_answer_ids
+    @question_answer_ids ||= question_answers.collect do |q|
```
kevinburleigh75: This is minor, but it seems that above `q` is being used to represent a 'question' but here (and elsewhere) it's being used to represent a 'question_answer'.

kevinburleigh75: Also I'd personally prefer to avoid single-letter variable names to allow easier searching (`qq` would be fine, IMO).

Dantemss: Fine with me.

app/subsystems/tasks/models/task.rb
```diff
@@ -27,7 +27,7 @@ def init
     self.spaced_practice_algorithm ||= SpacedPracticeAlgorithmDoNothing.new
   end
 
-  def is_shared
+  def is_shared?
```
kevinburleigh75: I've seen this in several places, but should we be naming this `#shared?`?

Dantemss: Possibly. We are not very consistent with our booleans. @jpslav?

jpslav: I kind of like the `is_` prefix.  Agreed that we aren't that consistent and really I'm ok either way.

jpslav:
@Dantemss re wrapping by multiple classes, my earlier question was just wondering about it.  Don't know if we really need it or not, and frankly I don't really get all the implications of `exposes` and `wrapped_by` enough to thoughtfully respond to your proposed changes.  

But when has that ever stopped me from trying anyway?

Why is the `exposes` line separate from the `wraps` line?  If you combine them does that let you more easily see which methods are delegated to which wrapped class? e.g.

```ruby
wraps Content::Models::Exercise,
      exposes: [:content, :title, :url],
      class_exposes: [:blah]
```

Just a thought, do with it what you will :-)
This way you don't have to repeat with `from_class` or similar.

Dantemss:
I'll give it a try...

Dantemss:
Oh ok one issue with them being together: the proposed syntax makes me think I can maybe expose different instance methods for different wrapped classes (which is not currently possible and would lead to an API mess) and expose the same class method to 2 different classes (which is impossible).

So I recommend we keep the current syntax :)

app/access_policies/tasked_access_policy.rb
```diff
@@ -1,9 +1,12 @@
...
   def self.action_allowed?(action, requestor, tasked)
     case action
-    when :read, :create, :update, :destroy, :mark_completed, :recover, :refresh
-      requestor.is_human? &&
-      DoesTaskingExist[task_component: tasked, user: requestor]
+    when :read, :mark_completed, :recover, :refresh
+      requestor.is_human? && DoesTaskingExist[task_component: tasked, user: requestor]
+    when :update
+      requestor.is_human? && DoesTaskingExist[task_component: tasked, user: requestor] && \
+      tasked.can_be_answered? && !tasked.task_step.task.feedback_available?
```
jpslav: These are very exercise-specific methods in a more generic access policy.  However, I think the solution to not having these here is beyond the scope of this PR.  So please just put in a comment that says these checks likely don't belong here.  Kevin would prefer the following comment but it is really up to you:  

```ruby
# congrats!  you finally figured out why that code in the task subsystem isn't getting called like you thought!  woo-hoo!
```

Dantemss: Well I disagree... unlike most of our validations, if this check fails it will immediately raise a SecurityTransgression, which should tell you exactly why your code isn't getting called.

But that's fine I can move this check to some AR validation that will return an error I suppose.

Dantemss:
I suppose https://github.com/openstax/tutor-server/pull/203 handles this issue so I'll just remove this check.

app/access_policies/tasked_access_policy.rb
Dantemss: I suppose https://github.com/openstax/tutor-server/pull/203 handles this issue so I'll just remove this check.

PR #196, opened 2015-04-15
Added Sprint 10 task that creates 3 readings and 3 homeworks and some fake student data

Dantemss:
# Setup
`rake db:drop db:create db:migrate db:seed sprint:010mix`

# Refresh my memory/try another
Logged in as a student, visit the third iReading (the fifth task on the list). The third problem in that reading is a "try another"/"refresh my memory" problem (if you get it wrong).

# Stats
Logged in as teacher:
`GET /api/plans/2`, look under the "stats" key.


Dantemss:
@jpslav all done.

PR #197, opened 2015-04-17
make events API endpoint role-aware

kevinburleigh75:
This PR makes the `/courses/:course_id/events(/role/:role_id)` API endpoint role-aware.  It also fixes some bugs related to accidental usage of the `Entity::Role` `role_type` enum in production code and removes now-dead code.


app/routines/choose_course_role.rb
```diff
@@ -22,44 +22,78 @@ class ChooseCourseRole
...
+
     if role_id
-      role = Entity::Role.find(role_id)
-      fatal_error(code: :invalid_role, message:"Role not found") unless role
-      validate_role_membership(course, user, role)
-    else
-      roles = GetUserCourseRoles[course:course, user:user, types:allowed_role_type]
-      validate_role_listing(roles)
+      user_course_roles = user_course_roles.select{|role| role.id == Integer(role_id)}
+      if user_course_roles.none?
```
jpslav: Remove the line above and change this line to 

```ruby
  if user_course_roles.none?{|role| role.id == Integer(role_id)
```

or alternatively this whole block of code becomes

```ruby
  if role_id && user_course_roles.none?{|role| role.id == Integer(role_id)
    fatal_error(blah)
  end
```

app/routines/choose_course_role.rb
```diff
@@ -22,44 +22,78 @@ class ChooseCourseRole
...
-        fatal_error(code: :multiple_roles, message:"The role must be specified because there is more than one student role available" )
+  private
+
+  def find_unique_role(course:, roles:, type:)
+    role_name = "#{type.to_s}"
+    method_name = "#{role_name}?"
+    matching_roles = roles.select do |role|
+      case type
+      when :teacher
+        CourseMembership::IsCourseTeacher[course: course, roles: role]
```
jpslav: A little bit of a shame to have to run this query again given that we've already down below queried teachers and students separately.

app/routines/get_role_course_events.rb
```diff
@@ -0,0 +1,29 @@
+class GetRoleCourseEvents
+  lev_routine
+
+  uses_routine GetCourseTaskPlans,
```
jpslav: Would you be up for (possibly in another PR) moving `GetCourseTaskPlans` into `Tasks::GetTaskPlans` and changing it to take an `owner` field instead of a course, e.g.

```ruby
class GetCourseTaskPlans
  lev_routine

  protected

  def exec(owner:)
    task_plans = Tasks::Models::TaskPlan.where(owner: owner).to_a
    outputs[:total_count] = task_plans.count
    outputs[:items] = task_plans
  end
end
```

app/subsystems/course_membership/add_student.rb
```diff
@@ -1,13 +1,9 @@
 class CourseMembership::AddStudent
   lev_routine
 
-  uses_routine Role::CreateUserRole
-
   protected
 
   def exec(course:, role:)
-    user = Entity::User.create!
```
jpslav: Yay!

jpslav:
Review done.

app/routines/choose_course_role.rb
kevinburleigh75: `user_course_roles` is mutated here and used later, so I don't think these suggestions would work as you think.

PR #201, opened 2015-04-18
corrected SS has_blah association default foreign key

jpslav:
@Dantemss and @nathanstitt -- we had some earlier discussion in another PR about has_many default foreign keys in SS's.  This fix changes the default to use the `my_subsystem_name` instead of `subsystem_name`, which I believe is correct (and gets rid of the need to specify foreign keys like we were doing).  Let me know what you think.


nathanstitt:
I agree, that seems like what the foreign_key should be referring to.

PR #202, opened 2015-04-21
sprint010 events task

kevinburleigh75:
I think this is ready to go.  There are a few places where seemingly broken code (entity_extension-related mostly) was commented out and replaced with normal code.  Eventually that will need to be remedied.

### DIFF
--- a/app/routines/get_role_course_events.rb
+++ b/app/routines/get_role_course_events.rb
@@ -5,8 +5,7 @@ class GetRoleCourseEvents
                as: :get_plans,
                translations: { outputs: { map: { items: :plans } } }
   uses_routine Tasks::GetTasks,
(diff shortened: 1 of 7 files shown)


kevinburleigh75:
## Rake Task

Running the following command will create the setup described below:
`bundle exec rake db:drop db:create db:migrate sprint:010events`

## The Setup

Three courses are created:

* `Physics - 0 Plans`
* `Physics - 1 Plan`
* `Physics - 4 Plans`

and each course has three associated users:

* a student (`*_plan_student`)
* a teacher (`*_plan_teacher`)
* both a teacher and a student (`*_plan_teacher_student`)

where `*` is one of `{zero,one,four}`.

The indicated number of assignments have been distributed to each student.

PR #204, opened 2015-04-21
Strong ETag and If-Modified-Since

Dantemss:
Support ETags (If-None-Match) and If-Modified-Since based on the updated_at timestamp  for Tasks and TaskSteps (GET only)


Dantemss:
This basically saves us having to render the representers again, not much else...

jpslav:
@philschatz I think that ETag stuff is now good to go.

philschatz:
ooh, exciting! Thanks!

PR #205, opened 2015-04-22
Class summaries for assignments routes /api/plans/:id/stats and /api/plans/:id/review

Dantemss:
# Done
- Moved stats representers into the Tasks::Stats namespace
- Added review route and controller method (to be called for class summaries for assignments)
- Added stats route and controller method (to be called for the calendar popup)
- Modified task plan stats routine to calculate extra details if `details: true` is specified.
- Specs, new and old


Dantemss:
Found some bugs that need fixing...

Dantemss:
I believe this is now done... I'll make another PR to fix those specs, since they are unrelated.

app/representers/api/v1/course_events_representer.rb
```diff
@@ -4,59 +4,13 @@ class CourseEventsRepresenter < ::Roar::Decorator
...
-      property :completed?, as: :complete,
-               type: :boolean,
-               readable: true
-
-    end
-
     collection :plans,
                readable: true,
-               decorator: PlanElements
+               decorator: Api::V1::Tasks::Calendar::TaskPlanRepresenter
```
jpslav: Some of this is probably OBE with my dashboard PR.

app/representers/api/v1/tasks/stats/course_and_period_representer.rb
```diff
@@ -0,0 +1,56 @@
+module Api::V1
+  module Tasks
```
jpslav: I started doing the namespace thing too, except I put mine under the controllers dir (which maybe isn't the best) -- just did that since those representers / routines are only used by that controller.

PR #206, opened 2015-04-22
suggested changes for tagging

jpslav:
@karenc here is at least a sketch of an approach following the discussion in https://github.com/openstax/tutor-server/pull/185/files#r28832810  

I will make my own comments in the PR to explain what I was thinking; was just easier for me to write as code instead of trying to explain in prose.  This code is totally untested and I fear may be out of date with some unpushed changes of yours, @karenc.  But you'll see the idea -- ping me with questions.

/cc @Dantemss


app/external/openstax/cnx/v1/page.rb
```diff
@@ -51,7 +51,7 @@ def initialize(hash: {}, chapter_section: '', is_intro: nil, book_part_title: ni
       @content         = content
       @los             = los
       @fragments       = fragments
-      @tag_defs        = tag_defs
+      @tags            = tags
```
jpslav: We can just call them `tags` from this `OpenStax::Cnx` perspective.

app/external/openstax/cnx/v1/page.rb
```diff
@@ -129,38 +129,52 @@ def root
     end
 
     def los
-      @los ||= root.css(LO_CSS).collect do |node|
-        LO_REGEX.match(node.attributes['class']).try(:[], 1)
-      end.compact.uniq
+      @los ||= tags.select{|value, attributes| attributes[:type] == :lo}
```
jpslav: Instead of having competing ways of getting LOs, just let the `tags` method get them and this convenience `los` method will select from what `tags` returns.

app/external/openstax/cnx/v1/page.rb
```diff
@@ -129,38 +129,52 @@ def root
...
+      @tags = {}
+
+      # Extract tag name and description from .ost-standards-def and .os-learning-objective-def.
+      # Also extract any LOs we find in use in case they weren't appropriately defined with in
+      # their own .os-learning-objective-def block.
+
+      root.css(LO_CSS).each do |node|
+        lo_value = LO_REGEX.match(node.attributes['class']).try(:[], 1)
+        @tags[lo_value] = { value: lo_value, type: :lo } if lo_value.present?
+      end
```
jpslav: Here I've started using the `value` terminology.  I think it is still a good idea to have this LO_CSS search in case there are LOs used in the content that mistakenly don't have a `ost-learning-objective-def` block.

app/external/openstax/cnx/v1/page.rb
```diff
@@ -129,38 +129,52 @@ def root
...
+
+      # Can't use self.root b/c ost-standards-def is inside .os-teacher which was removed in root.
 
-      # ost-standards-def is inside .os-teacher which was removed in root
-      doc = Nokogiri::HTML(content)
```
jpslav: There is already a `doc` method in this class that has this value.

app/external/openstax/cnx/v1/page.rb
```diff
@@ -129,38 +129,52 @@ def root
...
-        @tag_defs[class_name] = {}
-        @tag_defs[class_name][:name] = name
-        @tag_defs[class_name][:description] = description
+        tag_value = node.attr('class').split.last
+        @tags[tag_value] = {
+          value: tag_value,
+          name: name,
+          description: description,
+          type: :generic
+        }
```
jpslav: Switched to using `tag_value` instead of `class_name` -- a little more clear, imo.  Also setting the type here and below so it can be used when creating `Content::Models::Tag`s.

app/external/openstax/cnx/v1/page.rb
```diff
@@ -129,38 +129,52 @@ def root
...
+        teks_value = classes[2]
         name = node.content.strip
         name.gsub!(/\s+/, ' ')
-        @tag_defs[lo_class] = { name: name, teks: teks_class }
+        @tags[lo_value].merge!({
+          value: value,
+          name: name,
+          teks: teks_value,
+          type: :lo
+        })
```
jpslav: Using `merge!` here for the case when an LO was found via the `LO_CSS` approach above.

app/subsystems/content/routines/find_or_create_tag.rb
```diff
@@ -0,0 +1,60 @@
+class Content::Routines::FindOrCreateTag
```
jpslav: This is like `Content::Routines::CreateTags` except that it handles inputs of type String as well as `Content::Models::Tag`.  This cleans up the logic in `TagResource` a bit.

app/subsystems/content/routines/find_or_create_tag.rb
```diff
@@ -0,0 +1,60 @@
...
+      description: hash[:description],
+      tag_type: hash[:tag_type]
+    )
+
+    # If the hash mentions a TEKS tag, link it
+
+    teks_value = hash[:teks]
+
+    if teks_value
+      raise "Can only link TEKS tags to LOs" if tag.tag_type != "lo"
```
jpslav: This should be tested :-)

app/subsystems/content/routines/import_page.rb
```diff
@@ -31,11 +31,8 @@ def exec(cnx_page:, book_part:)
     book_part.pages << outputs[:page] unless book_part.nil?
     transfer_errors_from outputs[:page], {type: :verbatim}, true
 
-    # Create tags
-    run(:create_tags, cnx_page.tag_defs)
-
-    # Tag Page with LO's
-    run(:tag, outputs[:page], cnx_page.los, tag_type: :lo)
+    # Tag the Page
+    run(:tag, outputs[:page], cnx_page.tags)
```
jpslav: Here we are tagging with all tags, not just LOs.  Since these tags have a `type` field, we don't have to pass it in separately.  Note we still pass in the type from `import_exercises.rb` -- those tags are just strings and we need to specify the type.

app/subsystems/content/routines/tag_resource.rb
```diff
@@ -9,21 +9,15 @@ def exec(resource, tags, options = {})
...
 
     outputs[:tags] = []
     outputs[:taggings] = []
-    [tags].flatten.compact.each do |t|
-      tag = t.is_a?(Content::Models::Tag) ? \
-            t : Content::Models::Tag.find_or_initialize_by(value: t.to_s)
-      unless tag.persisted?
-        tag.tag_type = tag_type
-        tag.save!
-      end
```
jpslav: A lot of this logic goes into the FindOrCreateTag routine

app/subsystems/content/routines/tag_resource.rb
```diff
@@ -9,21 +9,15 @@ def exec(resource, tags, options = {})
     tagging_class = options[:tagging_class] || \
                       "#{resource_class_name}Tag".constantize
     resource_field = resource_class_name.underscore.split('/').last.to_sym
-    tag_type = options[:tag_type] || 0
-    outputs[:tag_type] = tag_type
```
jpslav: Goes into FindOrCreate or just get the default from the `Tag` class.

app/subsystems/content/routines/tag_resource.rb
```diff
@@ -9,21 +9,15 @@ def exec(resource, tags, options = {})
...
-      unless tag.persisted?
-        tag.tag_type = tag_type
-        tag.save!
-      end
-
-      outputs[:tags] << tag
-      transfer_errors_from(tag, scope: :tags)
+
+    [tags].flatten.compact.each do |tag|
+      content_tag = run(FindOrCreateTag, input: tag, type: options[:tag_type]).outputs.tag
```
jpslav: Should have a `uses_routine` above probably.

karenc:
Merged into #185

PR #209, opened 2015-04-24
Removed page_id from TaskStep

Dantemss:
# Done
- Removed page_id from TaskStep
- Fixed old specs
- Added search routine to find a Page with a given LO
- Specs for the new routine


jpslav:
@Dantemss please rebase and check with @joemsak that he knows how to use this new approach.

PR #214, opened 2015-04-27
Enabled fake stores in production; added user passwords

Dantemss:
With the changes in this PR, Chris is able to deploy tutor-server with some demo data.


Dantemss:
Fake stores are necessary because the fake BL client is used in the demo deployment and it uses those.

PR #219, opened 2015-04-28
Data for FE and deployment demos

Dantemss:
- Added somewhat meaningful date ranges to assignments that overlap
- Added fake exercise step to third reading for refresh my memory


Dantemss:
The third assignment now has 2 TA/RM exercises (one is not in the book and is artificially inserted).

philschatz:
:+1: I'll test once merged

PR #222, opened 2015-04-28
The return of the return of the deployment data: round 3

Dantemss:


lib/hacks/answer_exercise.rb
```diff
@@ -2,24 +2,27 @@ module Hacks
   class AnswerExercise
     lev_routine
 
+    uses_routine MarkTaskStepCompleted, as: :mark_completed
+
     protected
 
-    def exec(task_step:, is_correct:)
+    def exec(task_step:, is_correct:, free_response: nil, completed: true)
```
jpslav: should the default be '.'  ?

Dantemss: But now it's a sentence that explains all the things!

jpslav: oops missed that — perfect!

PR #224, opened 2015-04-29
added course and user names to setup:001

jpslav:


lib/tasks/setup_001.rb
```diff
@@ -140,4 +138,21 @@ def exec
     end
   end
 
+  def new_profile(username:, name: nil, password: 'password')
+    name ||= Faker::Name.name
```
Dantemss: Can we avoid Faker (maybe after the demo)? No Faker in the production environment so this will cause deployment to fail (since we didn't specify a default name for students)

Alternatively, we can add Faker to production.

Dantemss: Nvm I see you added Faker to production.

PR #225, opened 2015-04-29
change learning guide route from ending in /stats to /guide to match wha...

jpslav:
...t FE is expecting


jpslav:
Already done in a different PR.

PR #230, opened 2015-04-29
fix for empty candidate exercises

kevinburleigh75:
This fixes the problem with:
```
NoMethodError: undefined method `url' for nil:NilClass
```
happening in the sprint:010mix setup.


Dantemss:
Hmm I'm thinking we might want to just do the loop with the minimum of `candidate_exercises.count` and `number`.

kevinburleigh75:
We need to isolate each condition so we can eventually debug this issue.

PR #233, opened 2015-04-30
Added option to use in-memory sqlite database

Dantemss:
- Disabled by default
- Activate by specifying the env variable `OXT_TEST_DB=:memory:` or `OXT_DEV_DB=:memory:` (prepend to `rake` or `export` it or use dotenv)
- Active in the specified environment only
- Saves about 1 minute running specs on my machine: ~7 minutes -> ~6 minutes (most likely will be more if you don't have SSD or have lots of RAM, less if you have low RAM)

### DIFF
--- a/app/subsystems/tasks/models/tasked_exercise.rb
+++ b/app/subsystems/tasks/models/tasked_exercise.rb
@@ -94,11 +94,11 @@ class Tasks::Models::TaskedExercise < Tutor::SubSystems::BaseModel
   def identifier
     # TODO this code (and the blatant hack above) have got to go! Had to add the
     # tries and the secure random to a basic unit test to work.
-    task_step.task.taskings.try(:first).try(:role).try(:id) || SecureRandom.hex
+    task_step.task.taskings.try(:first).try(:role).try(:id).try(:to_s) || SecureRandom.hex
   end
 
   def trial
-    SecureRandom.hex
+    task_step.id.to_s
   end
 
   # Eventually this will be enforced by the exercise substeps
--- a/app/subsystems/tasks/models/tasked_exercise.rb
+++ b/app/subsystems/tasks/models/tasked_exercise.rb
@@ -98,7 +98,7 @@ class Tasks::Models::TaskedExercise < Tutor::SubSystems::BaseModel
   end
 
   def trial
-    task_step.id.to_s
+    SecureRandom.hex
   end
 
   # Eventually this will be enforced by the exercise substeps
--- a/Gemfile
+++ b/Gemfile
@@ -105,6 +105,12 @@ gem 'pg'
 gem 'faker'
 
 group :development, :test do
+  # SQLite adapter
+  gem 'sqlite3'
+
+  # Allows the use of the in-memory SQLite3 database in Rails tests
+  gem 'memory_test_fix'
+
   # Get env variables from .env file
   gem 'dotenv-rails'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -205,6 +205,9 @@ GEM
     mail (2.6.3)
       mime-types (>= 1.16, < 3)
     maruku (0.7.2)
+    memory_test_fix (1.2.2)
+      activerecord (>= 3.2.0, < 4.99.0)
+      railties (>= 3.2.0, < 4.99.0)
     method_source (0.8.2)
     mime-types (2.5)
     mini_magick (4.2.0)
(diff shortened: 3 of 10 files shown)


philschatz:
:+1: if it can be used in development!

Dantemss:
Currently I only enabled it in `test`. Would it help to have it in `dev` too?

Dantemss:
Activated the option to use the in-memory DB in dev (but keep in mind you will lose all your data every time if you do that)

philschatz:
@nathanstitt @pandafulmanda @jivewise @darrenmason @dak: Thoughts on using in-memory DB (rather than postgres) for UI dev?

I usually only kill rails when resetting the dataset or pulling in changes.

jpslav:
Um... sorry I'm late to the party.  I like your initiative Dante and think it is an interesting idea, but I'm wary of moving away from Postgres everywhere; we all just said "hooray" 3 weeks ago when we moved to PG in development because then our dev/test/prod environments were all on the same page.

I'd much rather improve the performance of any slow business logic that is making the tests run slow.  This will speed up our non-test performance as well as our spec speed.  And we'll still developing and testing against the tools we'll be using in production, which I think is a good thing.

Thoughts?

jpslav:
What about configuring PG in the test environment to get speedups (turning off certain safeties you wouldn't want to turn off in production):

http://stackoverflow.com/a/9407940/1664216

First big link from that S/O answer: http://www.postgresql.org/docs/current/static/non-durability.html

jpslav:
Also consider profiling the specs to find trouble spots, e.g.

* https://github.com/sinisterchipmunk/rspec-prof
* https://thejspr.com/blog/profiling-rspec-tests/

nathanstitt:
@philschatz I don't think using in-memory sqlite will work in development for the front-end, since it wouldn't work with the Rake tasks.  Well it'd work, but all the data would be lost as soon as the rake process ended, so not terribly useful :grinning: 

Aside from that I don't think it'd matter if the backend used in-memory sqlite for testing, as long as the Postgresql specs are still ran on Travis.  What I'd be afraid might happen is  a repeat of the mess that we had to straighten out when we migrated to Postgresql, i.e. missing foreign key references and migrations that wouldn't run.

I am also a fan of using Postgresql's advanced features, like binary JSON and array support, which would preclude using sqlite 

I'm also curious what gains might be if we employ the tips @jpslav pointed out.

Dantemss:
Well, the default is still postgres. This is just meant as a way to more quickly run specs for WIP stuff (you should still run the full thing on PG before you pull). Plus, alternating between PG and SQLite exposed some order dependency in some specs that I had to fix.

I'll remove the dev option once again, since it's not that useful without rake tasks.

app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -98,7 +98,7 @@ def identifier
   end
 
   def trial
-    task_step.id.to_s
+    SecureRandom.hex
```
jpslav: Why don't the trial to be something predictable and possibly repeatable?

Dantemss: Done in a separate PR. This change here is just so I can run the tests on SQLite without it barfing all over the place.

Dantemss: SQLite resets the DB for every test, including primary keys, so exchange-ruby was erring out because of the repeated trial.

jpslav: But should we be saving the trial?  (might we not need to know the trial for this TaskedExercise at some point later?)  If it is random on each call of this method, it doesn't seem useful.

Dantemss: Since we don't have homework recovery yet... In https://github.com/openstax/tutor-server/pull/239 I hardcoded it back to "1". (that PR updated exchange-ruby to not error out when you resend the answer with the same identifier/url/trial).

jpslav: The right answer is to use the task_step ID -- we want to be sending good data to Exchange regardless of whether we have certain features implemented in Tutor.  Sounds like we're making a change to business logic because of a test idiosyncrasy.  

Is something like the following sufficient to be repeatable and test-proof?

```ruby
def trial
  "#{task_step.id.to_s}:#{created_at.to_i}"
end
```

Dantemss: The above suggestion will fail because exchange-ruby enforces that you cannot answer the same question twice. task_step id and created_at will both be the same.

Dantemss: But you know what... I'll just revert this here so this can get merged because this fix has nothing to do with the change, other than the fact that the business logic is broken.

PR #234, opened 2015-04-30
WIP: Run specs in parallel, using all cpu cores (really only for discussion)

nathanstitt:
This is the results of an experiment I ran to run the specs in parallel using the "parallel_tests" gem https://github.com/grosser/parallel_tests

It chopped my spec run down from 7:05 to 2:56.  That's on a 8 core macbook, so it divided the specs by 8 and then ran each batch in it's own sub-process.  I'm somewhat disappointed in this, since I'd hoped it'd be close to 8 times faster.

It takes a bit of work to setup, you have to run:
 * `rake parallel:create` (one time after installing the gem)
 * `rake parallel:prepare` (after migrating, to populate all the DBs)
 * and then `rake parallel:spec` to run the specs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -108,6 +108,9 @@ group :development, :test do
   # Get env variables from .env file
   gem 'dotenv-rails'
 
+  # Run specs in parallel
+  gem "parallel_tests"
+
   # Allows 'ap' alternative to 'pp'
   gem 'awesome_print'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -260,6 +260,9 @@ GEM
       lev
       rails (>= 3.1)
     pager (1.0.1)
+    parallel (1.4.1)
+    parallel_tests (1.3.7)
+      parallel
     pg (0.18.1)
     polyamorous (1.1.0)
       activerecord (>= 3.0)
@@ -496,6 +499,7 @@ DEPENDENCIES
   openstax_api (~> 5.3.0)
   openstax_exchange
   openstax_utilities (~> 4.2.0)
+  parallel_tests
   pg
   pry
   pry-nav
--- a/config/database.yml
+++ b/config/database.yml
@@ -18,7 +18,7 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  database: <%= ENV['OXT_TEST_DB'] || 'ox_tutor_test' %>
+  database: <%= ENV['OXT_TEST_DB'] || 'ox_tutor_test' %>_<%= #{ENV['TEST_ENV_NUMBER']} %>
 
 ## Production database is intentionally left unconfigured
 #production:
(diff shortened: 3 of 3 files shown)


nathanstitt:
I also ran `rspec --profile` which shows the 10 slowest specs.  Looks like if we can kill the top 5, it'd shave around 3 minutes off a run.  Particularly the first one which is only testing a rake task and taking a minute and half.

    Top 10 slowest example groups:
      Setup001
        95.39 seconds average (95.39 seconds / 1 example) ./spec/lib/tasks/setup_001_spec.rb:5
      Tasks::Assistants::HomeworkAssistant
        19.91 seconds average (19.91 seconds / 1 example) ./spec/subsystems/tasks/assistants/homework_assistant_spec.rb:4
      Content::ImportBook
        19.8 seconds average (79.2 seconds / 4 examples) ./spec/subsystems/content/import_book_spec.rb:4
      Content::Routines::SearchPages
        17.71 seconds average (17.71 seconds / 1 example) ./spec/subsystems/content/routines/search_pages_spec.rb:4
      Tasks::Assistants::IReadingAssistant
        13.11 seconds average (26.22 seconds / 2 examples) ./spec/subsystems/tasks/assistants/i_reading_assistant_spec.rb:4
      CalculateTaskPlanStats
        9.98 seconds average (39.92 seconds / 4 examples) ./spec/routines/calculate_task_plan_stats_spec.rb:4
      Api::V1::TaskPlanWithDetailedStatsRepresenter
        5.52 seconds average (5.52 seconds / 1 example) ./spec/representers/api/v1/task_plan_with_detailed_stats_representer_spec.rb:4
      Api::V1::TaskPlanWithStatsRepresenter
        5.13 seconds average (5.13 seconds / 1 example) ./spec/representers/api/v1/task_plan_with_stats_representer_spec.rb:4
      Content::Routines::SearchExercises
        4.03 seconds average (4.03 seconds / 1 example) ./spec/subsystems/content/routines/search_exercises_spec.rb:4
      Content::Routines::ImportPage
        3.87 seconds average (38.7 seconds / 10 examples) ./spec/subsystems/content/routines/import_page_spec.rb:4

Dantemss:
The first one ensures the deployment rake task at least runs...
We could mark it as :slow and disable running that by default, but enable on travis.

Dantemss:
Also someone made some law somewhere that says you will never have N times speedup with N cores :D

Dantemss:
But overall I like it :)

nathanstitt:
Yeah, I'm continually surprised by how little multiple cores/processes help.  I "knew" it wouldn't be 8X, but was still hoping for something in the 6-7X range.  But 2.5(ish) speedup isn't bad I guess ...

jpslav:
Any reason to not merge this in?  Seems harmless enough if people don't want to use it.

PR #236, opened 2015-04-30
Performance book API fixes and tests

karenc:
- Write custom query for getting tasks for students in performance book

  Return only homework and reading tasks for students ordered by due date

- Get rid of the dependency on task plan in performance book code

- Include task_steps and tasked when getting tasks in performance book

- Use PerformanceBookRepresenter in performance book api

- Add performance book api tests

### DIFF
--- a/app/subsystems/tasks/get_performance_book.rb
+++ b/app/subsystems/tasks/get_performance_book.rb
@@ -3,10 +3,6 @@ require_relative 'models/entity_extensions'
 class Tasks::GetPerformanceBook
   lev_routine express_output: :performance_book
 
-  uses_routine Tasks::GetTasks,
-               as: :get_tasks,
-               translations: { outputs: { type: :verbatim } }
-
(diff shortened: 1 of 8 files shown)


app/subsystems/tasks/get_performance_book.rb
```diff
@@ -34,30 +30,40 @@ def exec(course:, role:)
...
     }
   end
 
-  def get_data_headings(task_plan)
-    { title: task_plan.title }.merge(class_average(task_plan))
+  def get_tasks_for_student(student)
+    # Return reading and homework tasks for a student ordered by due date
+    Tasks::Models::Task
+      .joins { taskings }
+      .where { taskings.entity_role_id == student }
```
jpslav: `student.id`?

karenc: It seems to work with just `student`, but may be `student.id` makes more sense.

karenc:
@jpslav ready for review

PR #238, opened 2015-04-30
Update the function signature and usage of CreateProfile

joemsak:
https://github.com/openstax/napkin-notes/blob/master/jp/s010_followup.md#fix-createprofile

### DIFF
--- a/app/subsystems/user_profile/create_profile.rb
+++ b/app/subsystems/user_profile/create_profile.rb
@@ -1,31 +1,29 @@
 module UserProfile
   class CreateProfile
-    lev_routine
+    lev_routine express_output: :profile
 
     uses_routine OpenStax::Accounts::CreateTempAccount,
       translations: { outputs: { type: :verbatim } },
       as: :create_account
 
     protected
-    def exec(attrs:)
-      attrs = prepare_attrs_for_profile(attrs)
-      outputs[:profile] = Models::Profile.create!(attrs)
+    def exec(username: nil, password: nil, openstax_uid: nil, access_token: nil,
+             first_name: nil, last_name: nil, full_name: nil, title: nil,
+             entity_user_id: nil, account_id: nil, exchange_identifier: nil)
+      outputs[:profile] = Models::Profile.create!({
+        exchange_identifier: exchange_identifier || OpenStax::Exchange.create_identifier,
+        entity_user_id: entity_user_id || new_entity_user_id,
+        account_id: account_id || new_account_id({ username: username,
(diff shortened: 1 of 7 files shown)


app/subsystems/user_profile/create_profile.rb
```diff
@@ -1,31 +1,29 @@
...
+    def exec(username: nil, password: nil, openstax_uid: nil, access_token: nil,
+             first_name: nil, last_name: nil, full_name: nil, title: nil,
+             entity_user_id: nil, account_id: nil, exchange_identifier: nil)
+      outputs[:profile] = Models::Profile.create!({
+        exchange_identifier: exchange_identifier || OpenStax::Exchange.create_identifier,
+        entity_user_id: entity_user_id || new_entity_user_id,
+        account_id: account_id || new_account_id({ username: username,
+                                                   openstax_uid: openstax_uid,
+                                                   access_token: access_token,
+                                                   first_name: first_name,
```
jpslav: I don't think these parameters are being used in CreateTempAccount.  

I'm looking for someone to have a deeper understanding of the use of `CreateProfile` and what it should be doing internally.  That these parameters have been broken out but appear to be dead ends (at least based on my 5 minute look at the associated code), concerns me.

Is `CreateProfile` only meant to be used in test code?  If so, it should be moved to some dev folder/namespace.  Is `CreateTempAccount` the right thing to be calling?  Is its use of `inputs={}` also problematic (probably since it doesn't then complain about all of these parameters going into it that aren't used.

jpslav:
Also, if specs are missing they should be added.  If present, they should be improved.

app/subsystems/user_profile/create_profile.rb
joemsak: This routine was created as part of the work to have one point in the system to create a profile (and an entity user/account is created as a result, if existing reference IDs are not passed in), and for the Admin Creates User Accounts feature (dev/test thing I suppose)

As for whether the parameters are used in the Account creation or not, I couldn't tell you. I based it solely on the db schema.

Dantemss: `CreateTempAccount` will use either an :email parameter or a :username and (optional) :password, as previously specified.

Dantemss: Not sure if we want to call CreateTempAccount from here, but it does work on the demo script (even if Accounts is not being stubbed).

joemsak:
Okay this has been updated much nicer : )

I also found that if you pass an account_id you don't need a username so I added a use case for that.

PR #242, opened 2015-05-04
Performance book export

joemsak:


app/routines/get_course_profile.rb
```diff
@@ -0,0 +1,3 @@
+class GetCourseProfile
+  lev_routine delegates_to: CourseProfile::GetProfile
```
jpslav: nice

app/subsystems/course_profile/get_profile.rb
```diff
@@ -2,7 +2,6 @@ class CourseProfile::GetProfile
   lev_routine express_output: :profile
 
   protected
-
```
jpslav: don't hate on the whitespace

app/subsystems/tasks/export_performance_book.rb
```diff
@@ -0,0 +1,95 @@
...
+      as: :create_performance_book_export
+
+    protected
+    def exec(role:, course:)
+      run(:get_course_profile, course: course)
+      run(:get_performance_book, course: course, role: role)
+
+      Axlsx::Package.new do |file|
+        file.use_shared_strings = true # OS X Numbers interoperability
+        file.workbook.styles.fonts.first.name = 'Helvetica Neue'
```
jpslav: Do we need to specify a font?  (Is it harmful to do so, like if the computer using the file doesn't have that font?)  What happens if we don't specify a font?

app/subsystems/tasks/export_performance_book.rb
```diff
@@ -0,0 +1,95 @@
...
+    uses_routine CreatePerformanceBookExport,
+      translations: { outputs: { type: :verbatim } },
+      as: :create_performance_book_export
+
+    protected
+    def exec(role:, course:)
+      run(:get_course_profile, course: course)
+      run(:get_performance_book, course: course, role: role)
+
+      Axlsx::Package.new do |file|
```
jpslav: Using `file` here is a bit of a red herring, because it isn't really a `File` object or the like that we're dealing with but an `Axlsx::Package`.  I hunted around a bit trying to find where a real on-disk file was coming into the picture.  Can we rename to `package` or `pp` or `ap` or something that doesn't have a filesystem connotation?

app/subsystems/tasks/export_performance_book.rb
```diff
@@ -0,0 +1,95 @@
...
+        file.use_shared_strings = true # OS X Numbers interoperability
```
jpslav: Cool.

app/subsystems/tasks/export_performance_book.rb
```diff
@@ -0,0 +1,95 @@
...
+      text = Axlsx::RichText.new
+      text.add_run(content, b: true)
+      text
+    end
+
+    def tmp_file_path
+      ['./tmp/', generate_file_name, '.xlsx'].join('')
+    end
+
+    def generate_file_name
```
jpslav: Any race condition issues for someone who accidentally double clicks the "export" button? (i.e. do we need something random in this filename or perhaps a counter added to the end if the filename already exists?)

app/subsystems/course_profile/get_profile.rb
joemsak: It was my code originally if that's fair

app/subsystems/tasks/export_performance_book.rb
joemsak: Helvetica Neue should be universal, it's possible that `fonts` contains fallbacks down to `sans-serif` which is standard web practice... Have to check. It uses `Arial` by default which I think is gaudy

app/subsystems/tasks/export_performance_book.rb
joemsak: :+1: trying to avoid following the doc example of `p` don't particularly agree even if it is a small scope, I'll find an appropriate and descriptive name to use

app/subsystems/tasks/export_performance_book.rb
joemsak: Do we handle double clicks, a UI event, or does front-end have anything to do with that on their button component?

app/subsystems/tasks/export_performance_book.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      Axlsx::Package.new do |file|
+        file.use_shared_strings = true # OS X Numbers interoperability
+        file.workbook.styles.fonts.first.name = 'Helvetica Neue'
+        create_summary_worksheet(file: file)
+        create_data_worksheet(file: file)
+        file.serialize(tmp_file_path)
+      end
+
+      run(:create_performance_book_export, filepath: tmp_file_path,
```
jpslav: Can the few lines from `CreatePerformanceBookExport` just go here?  i.e. replace this call to `run` with:

```ruby
Models::PerformanceBookExport.create!(course: course, role: role, export: File.open(filepath))
```

jpslav: No one besides this routine will then be creating `PerformanceBookExports`, which seems like that's a good encapsulation.

joemsak: :+1:

app/subsystems/tasks/models/performance_book_export.rb
```diff
@@ -0,0 +1,18 @@
+module Tasks::Models
+  class PerformanceBookExport < Tutor::SubSystems::BaseModel
+    mount_uploader :export, ExportUploader
+
+    default_scope { order('created_at DESC') }
```
jpslav: With squeel, I believe this can be:

```ruby
default_scope order{created_at.desc}
```

FWIW

app/uploaders/export_uploader.rb
```diff
@@ -0,0 +1,9 @@
+# encoding: utf-8
+
+class ExportUploader < CarrierWave::Uploader::Base
+  storage :file
+
+  def store_dir
+    "uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
+  end
+end
```
jpslav: love carrierwave :-)

joemsak: :+1:

app/subsystems/tasks/models/performance_book_export.rb
joemsak: :+1:

app/subsystems/queues/export_performance_book.rb
```diff
@@ -0,0 +1,10 @@
+module Queues
```
jpslav: I appreciate the subsystem enthusiasm, but I'm wondering if a subsystem is appropriate here.  If `ActiveJob` didn't exist, this would totally make sense as there'd be a lot of administration involved in making background jobs happen -- but `ActiveJob` already gives us quite a bit of encapsulate of queueing mumbo-jumbo, and I feel that this subsystem will just end up being one-line calls to `perform_later` and job classes that are one-line calls to routines.  

Instead, `Queues::ExportPerformanceBook` could become `Tasks::ExportPerformanceBook` calling `Tasks::Jobs::ExportPerformanceBook`, and I don't think we'd miss any encapsulation from the queues SS.   Later if we extend `lev` to have an option like `lev_routine active_job: :default` then `lev` could even dynamically generate the job class from the routine `exec` signature and we'd be able to get rid of one explicit class definition.

Happy to discuss.

joemsak: :+1: happy to see the pragmatic approach encouraged, actually ... subsystem enthusiasm was more subsystem enthusiasm than consultation

spec/controllers/api/v1/courses_controller_spec.rb
```diff
@@ -5,19 +5,15 @@
 RSpec.describe Api::V1::CoursesController, type: :controller, api: true,
                                            version: :v1, speed: :slow, vcr: VCR_OPTS do
 
-  let!(:application)     { FactoryGirl.create :doorkeeper_application }
   let!(:user_1)          { FactoryGirl.create :user_profile }
   let!(:user_1_token)    { FactoryGirl.create :doorkeeper_access_token,
-                                              application: application,
```
jpslav: nice cleanup!

jpslav:
Looks nice Joe! Review complete.

app/subsystems/course_profile/get_profile.rb
jpslav: NOOOOOOOOO! 

Ok. ;-)

app/subsystems/tasks/export_performance_book.rb
jpslav: Yeah I agree `p` is not good.

app/subsystems/tasks/export_performance_book.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+    end
+
+    def bold_text(content)
+      text = Axlsx::RichText.new
+      text.add_run(content, b: true)
+      text
+    end
+
+    def tmp_file_path
```
jpslav: Are these files truly temporary?  Do we delete after attaching via carrierwave?

app/subsystems/tasks/export_performance_book.rb
jpslav: Yeah, that's a good question.  We can either make sure to remember to get them to prevent it, or our `tmp_file_path` method can become:

```ruby
def tmp_file_path
  ['./tmp/', generate_file_name, '.xlsx'].join('').tap do |filename|
    raise "Filename already exists" if File.exist?(filename)
  end
end
```

app/subsystems/tasks/models/performance_book_export.rb
joemsak: Actually in Rails 4 these kinds of 'squeel' queries are becoming more built-in

`default_scope { order(created_at: :desc) }` is native rails

app/subsystems/queues/export_performance_book.rb
joemsak: `Tasks::ExportPerformanceBook` can't call the job because the job calls said routine, unless you want me to move the routine's logic to the job, but usually a job is just a wrapper around the actual class so that the class can be used outside of the context of a queued job. So, not sure what step to take here.

app/subsystems/tasks/export_performance_book.rb
joemsak: The `tmp` path is typically cleared out on deploys

joemsak:
So I've done the cleanup sans what I commented on, and then there are test errors that I've asked about in the BE dev chat

app/controllers/api/v1/courses_controller.rb
```diff
@@ -151,19 +148,68 @@ def practice_get
...
   end
 
   protected
+  def authorize_response_for_teacher(course)
+    status = if course && current_user_is_teaching?(course)
+               :ok
+             elsif course
+               :forbidden
+             else
+               :not_found
```
jpslav: `course` won't ever be nil here because if `course = Entity::Course.find_by(id: params[:id])` doesn't find the course it raises an exception instead of returning nil.  This exception is already converted to a :not_found status by default Rails mechanisms, and a PR in progress is going to formalize this.

Our (intended) standard approach for doing authorization is to use access policies.  I'd prefer that this method be removed in favor of an access policy approach.

```ruby
def performance_exports
  course = Entity::Course.find_by(id: params[:id])

  OSU::AccessPolicy.require_action_allowed!(:export, current_api_user, course)

  exports = Tasks::GetPerformanceBookExports[course: course,
                                             role: get_course_role]
  respond_with exports, represent_with: Api::V1::PerformanceBookExportsRepresenter
end
```

The `find` call will raise an `ActiveRecord::RecordNotFound` exception which Rails will convert to a response with a 404 status (currently it is an HTML response, the PR in the works makes it a JSON response when appropriate).

The `require_action_allowed!` call will raise a `SecurityTransgression` which will (with the new PR) return a `:forbidden` status.  That Tutor isn't already doing that mapping is our fault -- a bunch of other places in the code assume it is happening where it isn't, so better for you to follow suit with the `require_action_allowed!` approach and we'll get the other fix in.

For `require_action_allowed!` to work, you'll need to modify the `CourseAccessPolicy` like so:

```ruby
class CourseAccessPolicy
    case action
    when :readings
      requestor.is_human?
    when :exercises, :export   # add `:export` here
      UserIsCourseTeacher[user: requestor.entity_user, course: course]
    else
      false
    end
  end
end
```

joemsak: :+1:

app/controllers/api/v1/courses_controller.rb
```diff
@@ -180,4 +207,8 @@ def get_practice_role
     get_course_role(types: :student)
   end
 
+  def current_user_is_teaching?(course)
```
jpslav: Unused method.

joemsak: Done and done, good looking out thank you

app/subsystems/tasks/export_performance_book.rb
Dantemss: We can make a cron later to periodically clear tmp if it becomes an issue.

Dantemss: On another note... did you try to use the Tempfile object? (I know Tempfile objects don't work in all situations though)

Dantemss: Ah nvm... guess since you're using a background job the Tempfile would be deleted too soon.

philschatz:
I'm not sure if this is the right PR to comment on but I **LOVE** the precision of grades in the XLS file :wink: :

image

PR #245, opened 2015-05-05
Return the currently logged in user's name

nathanstitt:
The front-end needs a method to determine the name of the currently logged in user for display on the top navbar.

This is a quick implementation of it for discussion.  If there's no security objections with displaying it, I'll expand it to use a proper representer so the endpoint can be documented by apipie.

The corresponding FE PR is: https://github.com/openstax/tutor-js/pull/183


philschatz:
:+1:

pandafulmanda:
:+1:

PR #246, opened 2015-05-05
Handle multiple versions of the same exercise

Dantemss:
- `Content::Models::Exercise`'s now have number and version integer fields
- Routines that create exercises populate those fields
- Local exercise search routines only return the latest locally cached version of exercises
- No way to fetch updates from Exercises yet (other than re-importing the book)
- Added some specs to ensure the internal search routines return only the latest exercises


jpslav:
@Dantemss looks good.  Rebase then merge.

PR #252, opened 2015-05-06
Fixed BigLearn V1 calls

Dantemss:


Dantemss:
Simple fix and specs are green, so merging.

PR #265, opened 2015-05-12
Send grading events to Exchange

Dantemss:
# Done
- Code to send exercise grade to Exchange when an exercise step is completed
- Released exchange-ruby and updated the Gemfile
- Specs


app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -24,6 +24,10 @@ def handle_task_step_completion!
     # "trial" is set to only "1" for now. When multiple
     # attempts are supported, it will be incremented to indicate the attempt #
     OpenStax::Exchange.record_multiple_choice_answer(identifiers.first, url, trial, answer_id)
+
+    grade = is_correct? ? '1' : '0'
```
jpslav: why is grade a string and not a float?

Dantemss: Changing that one now...

PR #270, opened 2015-05-13
added explicit exception response handling

jpslav:
This PR adds exception handling beyond Rails' defaults.  Certain exception types are logged and emailed to developers.  `SecurityTransgressions` now finally get converted to 403 errors.  An HTML error page is added because who likes Rails' boilerplate 500 page -- not me.

Note to @openstax/tutor-fe-dev  -- check out the recommendation in the README changes below for getting production-like error responses while working with tutor-server in development mode.

Also FE devs note that the production error responses now may include an `error_id` field, which is a string identifier that you may choose to display to end users -- the identifer is also in exception emails and in the log file, so we can work to track down issues.


app/views/errors/any.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    case @status
+    when :forbidden
+      "You are not allowed to access this."
+    when :not_found
+      "We couldn't find what you asked for."
+    when :unprocessable_entity
+      "We didn't understand what you asked for."
+    when :internal_server_error
+      "Sorry, Tutor had some unexpected trouble with your request."
+    else
```
Dantemss: This will never be reached (and is also unnecessary, since message should default to nil)

jpslav: Taking out the else

PR #271, opened 2015-05-14
attempt to fix CI error

jpslav:


spec/subsystems/tasks/get_performance_book_exports_spec.rb
```diff
@@ -22,6 +22,8 @@
 
     exports = described_class[course: course, role: role]
 
+    exports = exports.collect{|export| HashWithIndifferentAccess.new(export.to_hash)}
```
joemsak: I thought this is how the lev routine output sorta worked anyway?

jpslav: lev outputs are hashie mashes, didn't know if the `==` comparison between it and a Hash was messing it up, so I added line 25 below.

joemsak: I was able to fix this problem with @Dantemss' suggestion of using include(hash_including(...)) instead of eq()

PR #272, opened 2015-05-14
personalized exercises in ireadings

kevinburleigh75:
This PR includes some refactoring of PEs in Homeworks and the addition of PEs in IReadings.

There is still some work to be done to robustify(TM) the code and ensure that BigLearn is properly called, but all of the plumbing is now in place.

### DIFF
--- a/app/subsystems/tasks/assistants/homework_assistant.rb
+++ b/app/subsystems/tasks/assistants/homework_assistant.rb
@@ -248,6 +248,9 @@ class Tasks::Assistants::HomeworkAssistant
   end
 
   def self.add_personalized_exercise_steps!(task_plan:, task:, taskee:)
+    task.personalized_placeholder_strategy = Tasks::PlaceholderStrategies::HomeworkPersonalized.new \
+      if num_personalized_exercises > 0
+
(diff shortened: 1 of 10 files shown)


app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -248,6 +248,9 @@ def self.k_ago_map(num_spaced_practice_exercises)
   end
 
   def self.add_personalized_exercise_steps!(task_plan:, task:, taskee:)
+    task.personalized_placeholder_strategy = Tasks::PlaceholderStrategies::HomeworkPersonalized.new \
+      if num_personalized_exercises > 0
```
jpslav: It is an efficiency, but if you have the first line be `return if num_personalized_exercises == 0` then you don't need this check and you skip the `task.save!` below.  Also I don't think the returned `task` is used at the end.

kevinburleigh75: This prevents the `Task` from bothering to do anything later if it doesn't have placeholders.  This gets called once whereas the other code gets called many times.

PR #286, opened 2015-05-19
CLUE works

jpslav:


app/external/openstax/big_learn/v1/real_client.rb
```diff
@@ -50,26 +50,43 @@ def stringify_tag_search(tag_search)
...
     response = request(:get, clue_uri(query))
 
     result = handle_response(response)
 
-    # get the value out of the result
-    raise NotYetImplemented
+    # extract the clue using the knowledge that we have a simplified input (only one
+    # tag query, so we can just pull out the appropriate value).  It could be that there's
+    # no CLUE to give, in which case we return nil.
+    clue = result.try(:[], "aggregates").try(:first).try(:[], "aggregate")
```
kevinburleigh75: Assuming `result` should never be `nil`, this could be simplified to something like:
```ruby
clue = result.fetch("aggregates"){[]}.fetch(0){{}}.fetch("aggregate"){nil}
```
which is a little more expressive (IMO) and doesn't hide other potential issues.

PR #288, opened 2015-05-19
Fixes learning guide order and iReading calendar popup student count

jpslav:


philschatz:
:+1: Works for me!

app/routines/get_course_stats.rb
```diff
@@ -21,7 +21,7 @@ def exec(role:, course:)
     run(:get_course_books, course: course)
     run(:visit_book, book: outputs.books.first, visitor_names: [:toc, :page_data])
 
-    chapters = compile_chapters
+    chapters = compile_chapters.sort_by{|ch| ch[:chapter_section]}
```
kevinburleigh75: It's not obvious to me that page[:chapter_section] is sortable in the way we want - will 10.3 come before 2.4?

app/routines/get_course_stats.rb
```diff
@@ -41,7 +41,7 @@ def compile_chapters
         task_type: :mixed_practice
       )
 
-      pages = compile_pages(task_steps: task_steps)
+      pages = compile_pages(task_steps: task_steps).sort_by{|page| page[:chapter_section]}
```
kevinburleigh75: Same question as L24.

app/routines/get_course_stats.rb
jpslav: I admit that I'm not entirely sure it is 100% right, but it works enough for now.  Longer strings are sorted after shorter strings, and shorter strings are compared by characters, so I **think** it is ok.

PR #294, opened 2015-05-19
Improve learning guide performance

joemsak:
* Adds gem `bullet` for N+1 and unused eager loading detection
  * `tail -f log/bullet.log`
* Adds gem `active_record_query_trace` (disabled by default) to show the app trace of every query
  * set `ENV['TRACE_ACTIVERECORD_QUERIES']` to `true`
* Adds eager loading to various learning guide related routines, as pointed out by using bullet

### DIFF
--- a/config/initializers/trace_activerecord_queries.rb
+++ b/config/initializers/trace_activerecord_queries.rb
@@ -1,0 +1,1 @@
+ActiveRecordQueryTrace.enabled = ((ENV['TRACE_ACTIVERECORD_QUERIES'] || '').downcase == 'true')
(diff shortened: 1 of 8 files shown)


config/initializers/trace_activerecord_queries.rb
```diff
@@ -0,0 +1 @@
+ActiveRecordQueryTrace.enabled = ((ENV['TRACE_ACTIVERECORD_QUERIES'] || '').downcase == 'true')
```
jpslav: Suggest using EnvUtilities in lib (https://github.com/openstax/tutor-server/blob/master/lib/env_utilities.rb) so we can standardize this stuff a bit.

```ruby
ActiveRecordQueryTrace.enabled = EnvUtilities.load_boolean(name: 'TRACE_ACTIVERECORD_QUERIES', default: false)
```

Probably should update `EnvUtilities` with the `downcase` you use here.

jpslav:
Looks good.  One comment to address, then needs a rebase and we can merge.

config/initializers/trace_activerecord_queries.rb
Dantemss: Ugh guess secrets.yml might need this treatment too for the booleans in there... (though we usually don't use those env variables anyway)

PR #295, opened 2015-05-19
Page numbering changes

Dantemss:
- Intro pages are now numbered as 0 (FE to hide their number if needed)
- Chapter-section is now an array everywhere
- Added Page Entity

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -5,17 +5,11 @@ class Page < Entity
   exposes :url, :title, :content, :chapter_section, :book_part, :is_intro?, :fragments
 
   def tags
-    tag_models.collect{ |t| t.value }
+    repository.tags.collect{ |t| t.value }
   end
 
   def los
-    tag_models.select{ |t| t.lo? }.collect{ |t| t.value }
-  end
-
-  protected
-
-  def tag_models
-    repository.page_tags.includes(:tag).collect{ |et| et.tag }
+    repository.tags.select{ |t| t.lo? }.collect{ |t| t.value }
   end
 
 end
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -1,0 +1,21 @@
+class Page < Entity
+
+  wraps Content::Models::Page
+
+  exposes :url, :title, :content, :chapter_section, :book_part, :is_intro?, :fragments
+
+  def tags
+    tag_models.collect{ |t| t.value }
+  end
+
+  def los
+    tag_models.select{ |t| t.lo? }.collect{ |t| t.value }
+  end
+
+  protected
+
+  def tag_models
+    repository.page_tags.includes(:tag).collect{ |et| et.tag }
+  end
+
+end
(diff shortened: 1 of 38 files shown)


Dantemss:
Still can't switch to the final way of determining if a Page is an intro or not because it seems to not be working in archive-staging-tutor.

Dantemss:
cc @jivewise

Intro pages now get an array that ends in 0, like [4, 0]. I can also add a boolean if needed to tell the FE if a Page is an intro. UX wants to hide their numbering.

app/models/page.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def los
+    tag_models.select{ |t| t.lo? }.collect{ |t| t.value }
+  end
+
+  protected
+
+  def tag_models
+    repository.page_tags.includes(:tag).collect{ |et| et.tag }
```
jpslav: I think we've talked for a while about adding a `tags` association on `Content::Models::Page` and `Content::Models::Exercise` so we don't have to go through `page_tags` and `exercise_tags`.  I just tried the following and it appeared to work, can you add it in in both `Content::Models::Page` and `Content::Models::Exercise`?

```ruby
class Content::Models::Page < Tutor::SubSystems::BaseModel
  ...
  has_many :tags, through: :page_tags
  ...
end
```

And then does Rails query caching make it so we don't need to `@tag_models ||= repository.blah` here?

app/subsystems/tasks/assistants/i_reading_assistant.rb
```diff
@@ -100,7 +100,7 @@ def self.create_task!(task_plan:)
   def self.add_core_steps!(task:, cnx_pages:)
     cnx_pages.each do |page|
       # Chapter intro pages get their titles from the chapter instead
-      page_title = page.is_intro? ? page.book_part_title : page.title
+      page_title = page.is_intro? ? page.book_part.title : page.title
```
jpslav: Hmmm.... I don't understand how `page.book_part.title` isn't blowing up... from what I see `OpenStax::Cnx::V1::Page` doesn't have a `book_part` method... are these by chance not `OpenStax::Cnx::V1::Pages` as implied by the `cnx_pages` argument on line 100?

config/initializers/decorator_extensions.rb
```diff
@@ -1,7 +0,0 @@
-class Roar::Decorator
-
-  def self.chapter_section_formatter
```
jpslav: Goodbye!

jpslav:
@Dantemss I imagine this is needed in alpha; if you can get it early tomorrow we can merge it before we make the alpha branch.

app/subsystems/tasks/assistants/i_reading_assistant.rb
Dantemss: Oh yeah, they are Page entities now.

I didn't like the fact that people were adding domain things into Cnx::V1::Page so decided to split them.

Dantemss:
Should be ready for merging @jpslav

PR #303, opened 2015-05-21
Added missing cassettes and fixed some spec order dependency issues

Dantemss:


app/external/openstax/cnx/v1.rb
```diff
@@ -21,9 +21,13 @@ def self.with_archive_url(url:)
     old_url = archive_url_base
     old_ssl_url = archive_url_base(ssl: true)
 
-    set_archive_url_base(url: url)
-    result = yield
-    set_archive_url_base(url: old_url, ssl: old_ssl_url)
+    begin
+      set_archive_url_base(url: url)
+      result = yield
+    ensure
```
jpslav: nice

PR #304, opened 2015-05-21
Timecop

Dantemss:
- Switched Rails cache store to Redis
- Added settings store
- Added timecop logic to freeze time for all processes

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -107,6 +107,9 @@ gem 'pg'
 # Lorem Ipsum
 gem 'faker'
 
+# Key-value store for caching and job queuing
+gem 'redis-rails'
+
 # Background job queueing
 gem 'resque'
 
@@ -131,6 +134,9 @@ group :development, :test do
 
   # Allows the use of the in-memory SQLite3 database in Rails tests
   gem 'memory_test_fix'
+
+  # Fake in-memory Redis for development and testing
+  gem 'fakeredis'
 
   # Resque but for testing
   gem 'resque_spec'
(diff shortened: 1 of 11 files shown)


jpslav:
Looking nice. Pls add redis host port etc as secrets.  Also in anticipation of shared redis instance later can we scope the key names under a secret name (later tutor-production) or (tutor-stage). Change timecop to use travel instead of freeze. This will require setting the travel time and setting the time that travel time was set. Then when you travel you use both those values and time.now to travel to the right time.  Sorry reviewing from phone so this is short.

lakshmivyas:
Shared instance may be sooner than later ;) :)

Dantemss:
Do you expect that different environments (staging, production, demo?) will share the same redis instance?

lakshmivyas:
I think the shared instance will be per envionment for scaled out configurations (think prod stage).  For dev/demo environments, we can think of having a shared instance.   If it doesn't cost us too much I think the best option is to have one instance per environment.  \cc @cnuber

Dantemss:
Done except for admin views

jpslav:
I think tutor-demo and the like will be on one box with redis on same box.  But stage and prod will be shared.

lakshmivyas:
@jpslav - Given the issues we are having with BL, and the fact that we use redis on BL, bl admin (and for background jobs on our other apps), we might gain more benefit from a shared instance.   Same argument for local PG vs shared one.   We should discuss this sometime after alpha. \cc @cnuber .

The shared instances can just be a separate ec2 instance or managed ones (RDS, elasticache).

jpslav:
I'm not overly excited by the idea of different instances for everything below staging. All on one seems ok at least for us for now.

PR #305, opened 2015-05-22
Change big_learn to biglearn

lakshmivyas:
Exterminate the abomination that is "big_learn"


app/external/openstax/biglearn/v1/fake_client.rb
```diff
@@ -14,6 +14,8 @@ def add_exercises(exercises)
     # Iterate through the exercises, storing each in the store, overwriting
     # any with the same ID
 
+    puts "Adding fake exercises"
```
Dantemss: Not sure if I like this... it will print a bunch of messages during our specs... maybe use the logger?

config/initializers/openstax_biglearn.rb
```diff
@@ -6,6 +6,7 @@
   config.secret     = secrets['secret']
 end
 
+puts secrets
```
Dantemss: That's kind of dangerous :)

Dantemss:
I think you also need to rename the BigLearn module/class to Biglearn.

app/external/openstax/biglearn/v1/fake_client.rb
lakshmivyas: I'd like to leave these in and address them in a separate pull request.

lakshmivyas:
@Dantemss - Renamed BigLearn to Biglearn as well.  Thanks.

app/external/openstax/biglearn/v1/fake_client.rb
jpslav: I agree... no "puts" in master.

jpslav:
Um... I think `BigLearn` is more right than `Biglearn`.  Show of hands?

lakshmivyas:
+1 for Biglearn.   Because of repeated usage it has become a single word for me,   I suspect it will be for most people on openstax.  So, typing BigLearn would be unnatural.

app/external/openstax/biglearn/v1/fake_client.rb
lakshmivyas: hmm - I guess I just have to rebase.   I didn't add these puts'.

lakshmivyas: oh, well -may be I did and forgot about it - removing.

PR #306, opened 2015-05-22
Add pagination to user search results

karenc:


app/subsystems/user_profile/search_profiles.rb
```diff
@@ -4,19 +4,10 @@ class SearchProfiles
...
-      outputs[:profiles] = profiles.collect do |profile|
-        {
-          id: profile.id,
-          account_id: profile.account_id,
-          entity_user_id: profile.entity_user_id,
-          full_name: profile.full_name,
-          username: profile.username
-        }
-      end
+      outputs[:profiles] = Models::Profile.joins { account }
```
jpslav: I know this mirrors something that has already been merged, but ideally we shouldn't be passing back an active relation from this routine, but instead something like the collection of hashes that is removed above.

This can be achieved by passing the `page` and `per_page` options into the routine and calling `paginate` inside the routine.  I don't know how well this works with the view tho - might need some changes there.

Can you give it a shot post timecop stuff?

app/subsystems/user_profile/search_profiles.rb
```diff
@@ -4,19 +4,10 @@ class SearchProfiles
...
+                                          .where { (lower(account.username).like search_term) |
+                                                   (lower(account.full_name).like search_term) }
+                                          .includes { account }
```
jpslav: Also, do we need an explicit `order` clause (e.g. order by last name)?  Seems like we would otherwise we could get inconsistently ordered results.

app/subsystems/user_profile/search_profiles.rb
karenc: That's how I originally did it.  The pagination view helper expects `.total_pages` on the collection.  I don't really know how to (or if I should even) add a method to array to get this to work.

jpslav: I think the below completely untested code will work.  We (in other situations) have "Search" routines return the array of results with things like the total count, etc, so makes sense to do it here.  The Hashie::Mash will give method-like accessors.

```ruby
def exec(search_term:, page: nil, per_page: nil)
  profiles = Models::Profile.joins { account }
                            .where { (lower(account.username).like search_term) |
                                     (lower(account.full_name).like search_term) }
                            .includes { account }

  # Don't paginate for callers who don't need pagination
  do_paginate = page.present?

  if do_paginate
    per_page ||= 30
    profiles = profiles.paginate(page: page, per_page: per_page)
  end

  outputs[:profiles] = Hashie::Mash.new(
    total_items: do_paginate ? profiles.total_entries : profiles.count,
    total_pages: do_paginate ? profiles.total_pages : 1,
    items: profiles.collect do |profile|
      {
        id: profile.id,
        account_id: profile.account_id,
        entity_user_id: profile.entity_user_id,
        full_name: profile.full_name,
        username: profile.username
      }
  )
end
```

karenc: @jpslav The view helper `will_paginate` is still not happy with it :(  It complains that :renderer is not specified.  Could you have a look at [admin-user-pages-2](https://github.com/openstax/tutor-server/compare/admin-user-pages-2) admin users index page please?

karenc:
Pagination with JP's fix for will_paginate ready for review.

jpslav:
Two specs fail b/c we took out `total_pages` - can ditch those specs.

jpslav:
Otherwise looks good.

karenc:
Sorry, fixed now

PR #308, opened 2015-05-22
Optimize performance book view

joemsak:
https://www.pivotaltracker.com/story/show/95108188

* Reduces html page view load time by 47%
* Reduces ActiveRecord query time spent in the db by 85%
* Adds `GetStudentProfiles` to help get relationships to make simpler queries
* Eliminates N+1 queries
* Moves queries inside `.each` style loops out to entire `course` scope queries for all data needed
  * Example:
```ruby
students.each do |student|
  student_data = get_data_per_student_one_query_each_time(student.stuff)
  do_stuff_with_the_students_data(student_data)
  #...
end

# VS

the_data = get_all_the_data_for_all_students

students.each do |student|
  student_data = the_data.select { student_filter_criteria }
  do_stuff_with_the_students_data(student_data)
end
```
* Gets nitpicky on formatting encountered while spelunking

Note: Probably easiest just to view `GetPerformanceBook` as it is at HEAD

https://github.com/openstax/tutor-server/blob/optimize-performance-book-view/app/subsystems/tasks/get_performance_book.rb


app/subsystems/tasks/get_performance_book.rb
```diff
@@ -48,12 +48,19 @@ def get_performance_book_for_teacher(course)
...
+              .joins { taskings }
+              .where { taskings.entity_role_id == student.id }
+              .where { task_type.in Tasks::Models::Task.task_types.values_at(:reading, :homework) }
+              .order { due_at }
+              .includes(task_steps: :tasked)
+
+    task_steps_by_tasked_type = tasks.collect(&:task_steps).flatten.group_by(&:tasked_type)
+    exercises = task_steps_by_tasked_type.fetch("Tasks::Models::TaskedExercise",[])
+    preloader = ActiveRecord::Associations::Preloader.new
+    preloader.preload(exercises, :tasked_exercise)
```
joemsak: Here's where I attempted a blogged strategy that did not take

app/subsystems/tasks/get_performance_book.rb
```diff
@@ -86,8 +93,8 @@ def get_student_data(tasks, full_name, role)
 
   def exercise_count(task, index)
     return {} unless task.task_type == 'homework'
-    correct_count = task.task_steps.select { |ts| ts.tasked.is_correct? }.length
-    attempted_count = task.task_steps.select { |ts| ts.completed? }.length
+    correct_count = task.task_steps.select { |ts| ts.tasked_exercise.is_correct? }.length
```
joemsak: Here's where the log is showing what looks like an n+1 query that could be cleaned up, though bullet doesn't detect it and no usual strategy has affected it

joemsak:
Probably easiest just to view `GetPerformanceBook` as it is at HEAD

https://github.com/openstax/tutor-server/blob/optimize-performance-book-view/app/subsystems/tasks/get_performance_book.rb

app/subsystems/user_profile/get_user_full_names.rb
```diff
@@ -5,8 +5,16 @@ class UserProfile::GetUserFullNames
...
-    outputs[:full_names] = OpenStax::Accounts::Account.where { id.in profiles.collect(&:account_id) } \
-                                                      .collect(&:full_name)
+
+    profiles = UserProfile::Models::Profile.includes(:account).where {
+      entity_user_id.in entity_users.map(&:id)
+    }
+
+    outputs[:full_names] = profiles.collect { |profile|
+      {
+        entity_user_id: profile.entity_user_id,
```
jpslav: When accepting a list of input arguments to a routine like is happening here, I totally agree with putting the related ID in the appropriate output like you're doing here -- think we need to do that in more places, and thanks for doing it here. 

A little weird to me tho to have `full_name` attribute inside of a hash of full names -- leads to a slightly unexpected call like above

```ruby
outputs[:teacher_names].collect(&:full_name).sort
```

where `GetFullNames` is being called and then we're mapping the `full_names` output to `teacher_names` and the collecting `full_name` on it.

If we're going to do a `collect(&:full_name)` on the output, can `GetUserFullNames` potentially just be replaced with a more generic `GetProfiles`?  Seems like we're already paying the query expense, might as well report a few more fields.  And if so, a `GetProfiles` routine is going to look pretty similar to the existing `SearchProfiles` routine, and you might consider just merging the two into `SearchProfiles`, e.g.

```ruby
module UserProfile
  class SearchProfiles
    lev_routine express_output: :profiles

    protected
    def exec(search:, order: 'last_name', page: nil, per_page: 30)
      profiles = Models::Profile.joins{ account }.order(order).includes { account }

      case search
      when String
        profiles = profiles.where { (lower(account.username).like search) |
                                    (lower(account.full_name).like search) |
                                    (lower(account.first_name).like search) |
                                    (lower(account.last_name).like search) }
      when Object
        search = [search].flatten
        raise "Can only search by entity users" unless search.first.is_a?(Entity::User)
        profiles = profiles.where{ entity_user_id.in search.map(&:id) }
      end

      do_pagination = page.present?
      profiles = profiles.paginate(page: page, per_page: per_page) if do_pagination

      outputs[:profiles] = Hashie::Mash.new(
        total_items: do_pagination ? profiles.total_entries : profiles.count,
        items: profiles.collect do |profile|
          {
            id: profile.id,
            account_id: profile.account_id,
            entity_user_id: profile.entity_user_id,
            full_name: profile.full_name,
            username: profile.username
          }
        end
      )
    end
  end
end
```

obviously with more private methods probably :-)  Thoughts?

app/routines/get_teacher_names.rb
```diff
@@ -15,6 +15,6 @@ def exec(course_id)
     course = Entity::Course.find(course_id)
     run(:get_teacher_users, course)
     run(:get_full_names, outputs[:teachers])
-    outputs[:teacher_names] = outputs[:teacher_names].sort
+    outputs[:teacher_names] = outputs[:teacher_names].collect(&:full_name).sort
```
jpslav: Like mentioned below, this is a bit odd to me and maybe can be made un-odd by collapsing some similar code together.

app/subsystems/course_membership/get_course_roles.rb
```diff
@@ -1,5 +1,3 @@
-require_relative 'models/entity_extensions'
```
jpslav: gracias!

app/routines/get_teacher_names.rb
joemsak: I completely agree and I think this is a step in that direction. Next PR after merge? Can get on it immediately

app/subsystems/course_membership/get_students.rb
```diff
@@ -4,7 +4,9 @@ class CourseMembership::GetStudents
   protected
 
   def exec(course)
-    ss_maps = CourseMembership::Models::Student.where{entity_course_id == course.id}
-    outputs[:students] = ss_maps.collect{|ss_map| ss_map.role}
+    ss_maps = CourseMembership::Models::Student
+                .includes(:role)
+                .where{entity_course_id == course.id}
+    outputs[:students] = ss_maps.collect(&:role)
```
jpslav: Thanks for cleaning up -- does the `ss_maps` thing do anything for you?  Has always kind of confused me.  

Can it be:

```ruby
outputs[:students] = CourseMembership::Models::Student
                       .includes(:role)
                       .where{entity_course_id == course.id}
                       .collect(&:role)
```

Also, since a `course` relation exists, I think it can be:

```ruby
outputs[:students] = CourseMembership::Models::Student
                       .includes(:role)
                       .where{course == my{course}}
                       .collect(&:role)
```

The `my{...}` being there to disambiguate the `course` relation from the `course` object.

app/subsystems/user_profile/get_student_profiles.rb
```diff
@@ -0,0 +1,35 @@
...
+    uses_routine UserProfile::GetUserFullNames,
+                 as: :get_user_full_names,
+                 translations: { outputs: { type: :verbatim } }
+
+    protected
+    def exec(course: course)
+      student_roles = run(:get_students, course).outputs.students
+      users = run(:get_users_for_roles, student_roles).outputs.users
+      names = run(:get_user_full_names, users).outputs.full_names
+      role_users = Role::Models::User.where(entity_user_id: users.collect(&:id))
```
jpslav: I'm not sure why we need to get this?

app/subsystems/user_profile/get_student_profiles.rb
```diff
@@ -0,0 +1,35 @@
+module UserProfile
+  class GetStudentProfiles
```
jpslav: Fine to have such a routine, but it shouldn't be in the `UserProfile` SS since it combines subsystems.  It should be in `/app/routines`

app/subsystems/tasks/get_performance_book.rb
```diff
@@ -1,99 +1,114 @@
...
 
+      attempted_count = task_steps.select(&:completed?).length
+      correct_count = exercises.select(&:is_correct?).length
+
+      if attempted_count > 0
+        @class_average[index] << (Float(correct_count) / attempted_count)
+      end
+
+      {
+        exercise_count: task_steps.length,
```
jpslav: supposed to be `exercises.length`?

app/subsystems/tasks/get_performance_book.rb
```diff
@@ -1,99 +1,114 @@
...
+          data
+        }
+      }
+    end
+
+    def exercise_count(task_steps, exercises, index)
+      tasked_ids = task_steps.collect(&:tasked_id)
+      exercises = exercises.select { |e| tasked_ids.include?(e.id) }
 
+      attempted_count = task_steps.select(&:completed?).length
```
jpslav: Next steps of putting counts in `Task` should eliminate the need to get the task steps at all I think.  Soon.

app/subsystems/user_profile/get_student_profiles.rb
joemsak: Lines 25 and 26 to get from a student role to a user's account's full name through entity user which is tied by Role::Models::User... this brought up a discussion with kevin that roles are too separated from the entity user which often has other properties we need like in these instances

app/subsystems/tasks/get_performance_book.rb
joemsak: Good eye

joemsak:
@jpslav All updated according to your feedback

PR #313, opened 2015-05-26
WIP:  a delay (true by default) to DistributeTasks to queue in a Job

joemsak:
https://www.pivotaltracker.com/story/show/95180724


jpslav:
See alternative approach in https://github.com/lml/lev/pull/19

PR #314, opened 2015-05-26
Fixes assignment republishing after editing

Dantemss:


app/routines/distribute_tasks.rb
```diff
@@ -15,22 +15,22 @@ def validate_json(schema, object, options = {})
...
     # Validate the given settings against the assistant's schema
     # Intervention settings already included when the task_plan was saved
     err = validate_json(assistant.schema, task_plan.settings)
 
-    fatal_error(code: :invalid_settings, message: 'Invalid settings', data: err) \
-      unless err.empty?
+    fatal_error(code: :invalid_settings, message: 'Invalid settings', data: err) unless err.empty?
+
+    # Delete pre-existing assignments
+    unless task_plan.tasks.empty?
```
jpslav: Um... do we need there to be more checks here, like that the tasks are not yet open?

PR #316, opened 2015-05-27
got rid of N+1 queries in performance book, fixed some bugs...

jpslav:
... and started some other optimizations left as an exercise to the reader.

### DIFF
--- a/app/subsystems/role/get_users_for_roles.rb
+++ b/app/subsystems/role/get_users_for_roles.rb
@@ -6,6 +6,7 @@ class Role::GetUsersForRoles
   def exec(roles)
     role_ids = [roles].flatten.compact.collect{|r| r.id}
     ss_maps = Role::Models::User.includes(:user).where{entity_role_id.in role_ids}
-    outputs[:users] = ss_maps.collect(&:user).uniq
+    outputs[:role_to_user_map] = Hash[ss_maps.collect{|ss_map| [ss_map.entity_role_id, ss_map.user]}]
+    outputs[:users] = outputs[:role_to_user_map].values.uniq
   end
 end
--- a/app/subsystems/tasks/get_performance_book.rb
+++ b/app/subsystems/tasks/get_performance_book.rb
@@ -18,12 +18,10 @@ class Tasks::GetPerformanceBook
   protected
 
   def exec(course:, role:)
-    outputs[:performance_book] = \
-      if CourseMembership::IsCourseTeacher[course: course, roles: [role]]
-        get_performance_book_for_teacher(course)
-      else
-        raise(SecurityTransgression, 'The caller is not a teacher in this course')
-      end
+    raise(SecurityTransgression, 'The caller is not a teacher in this course') \
+      unless CourseMembership::IsCourseTeacher[course: course, roles: [role]]
+
+    outputs[:performance_book] = get_performance_book_for_teacher(course)
   end
 
   private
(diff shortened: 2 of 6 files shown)


app/subsystems/role/get_users_for_roles.rb
```diff
@@ -6,6 +6,7 @@ class Role::GetUsersForRoles
   def exec(roles)
     role_ids = [roles].flatten.compact.collect{|r| r.id}
     ss_maps = Role::Models::User.includes(:user).where{entity_role_id.in role_ids}
-    outputs[:users] = ss_maps.collect(&:user).uniq
+    outputs[:role_to_user_map] = Hash[ss_maps.collect{|ss_map| [ss_map.entity_role_id, ss_map.user]}]
```
jpslav: Nice to return a mapping between inputs and outputs; makes the routine more useful when multiple inputs, I think.  Maybe also do for GetUserFullNames.

Also, please add a spec for this output :-)

app/subsystems/tasks/get_performance_book.rb
```diff
@@ -32,9 +30,15 @@ def get_performance_book_for_teacher(course)
     student_data_list = []
     @class_average = []
     tasks = []
-    run(:get_students, course).outputs.students.each do |student|
+
+    students = run(:get_students, course).outputs.students
+    role_to_user_map = run(:get_users_for_roles, students).outputs.role_to_user_map
+    # TODO get a user_to_full_name_map and build a student_full_name_map to use below
+    # instead of calling that query over and over
```
jpslav: This is the start of running these queries once instead of inside a loop.  Can do what the TODO says to get user names.  Could conceivably get all Tasks at once too, but probably need to make an assumption that there's only one Task per student (no shared Tasks, which is true now, but won't always be).

app/subsystems/tasks/get_performance_book.rb
```diff
@@ -48,19 +52,12 @@ def get_performance_book_for_teacher(course)
...
-              .joins { taskings }
-              .where { taskings.entity_role_id == student.id }
-              .where { task_type.in Tasks::Models::Task.task_types.values_at(:reading, :homework) }
-              .order { due_at }
-              .includes(task_steps: :tasked)
-
-    task_steps_by_tasked_type = tasks.collect(&:task_steps).flatten.group_by(&:tasked_type)
-    exercises = task_steps_by_tasked_type.fetch("Tasks::Models::TaskedExercise",[])
-    preloader = ActiveRecord::Associations::Preloader.new
-    preloader.preload(exercises, :tasked_exercise)
```
jpslav: Took out what you said didn't work.

app/subsystems/tasks/get_performance_book.rb
```diff
@@ -93,13 +90,17 @@ def get_student_data(tasks, full_name, role)
 
   def exercise_count(task, index)
     return {} unless task.task_type == 'homework'
-    correct_count = task.task_steps.select { |ts| ts.tasked_exercise.is_correct? }.length
-    attempted_count = task.task_steps.select(&:completed?).length
+
+    exercise_task_steps = task.task_steps.exercises.includes(:tasked)
```
jpslav: No more N+1 (to my knowledge)

app/subsystems/tasks/models/task.rb
```diff
@@ -67,7 +67,7 @@ def feedback_available?(current_time: Time.now)
   end
 
   def completed?
-    self.task_steps(true).all?{|ts| ts.completed? }
```
jpslav: KB added these `(true)` reloads in, but I don't think they're really needed.  Added a reload in `handle_task_step_completion!` just to be safe.

PR #322, opened 2015-05-28
Split demo:001 task into two: create course and users and everything else

karenc:
So that Ansible can run sparfa in between demo:001 and demo:002.

Add "demo" rake task that runs all demo:* tasks.


lib/tasks/demo.rake
```diff
@@ -0,0 +1,17 @@
...
+    base_name = File.basename(filename, '.rb')
+    puts "Running #{base_name}"
+    require "tasks/#{base_name}"
+    class_name = base_name.classify
+    result = class_name.constantize.call(args.to_h.merge(print_logs: true))
+
+    if result.errors.none?
+      puts "#{base_name} Success!"
+    else
+      result.errors.each{ |error| puts "#{base_name} Error: " + Lev::ErrorTranslator.translate(error) }
```
kevinburleigh75: Should we break out of this loop if there is an error?  Is there any point in continuing?

karenc: I'll create a new PR for breaking out of the loop if there's an error.

karenc: https://github.com/openstax/tutor-server/pull/325

PR #323, opened 2015-05-28
add require timecop gem

cnuber:
add require timecop gem

### DIFF
--- a/config/initializers/timecop.rb
+++ b/config/initializers/timecop.rb
@@ -1,3 +1,4 @@
+require "timecop"
 class Timecop
   def self.enabled?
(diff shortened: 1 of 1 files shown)


jpslav:
merge after tests pass @kevinburleigh75

config/initializers/timecop.rb
```diff
@@ -1,3 +1,4 @@
+require "timecop"
```
jpslav: I think this might explode on real production deploys because there is no gem available.

lakshmivyas: Perhaps this file should be dropped by ansible on demo instances or removed on real prod?

PR #326, opened 2015-05-28
Move gems out of test, into dev, use tracer in dev only

joemsak:
I was having a MAJOR computer issue due to the tracer gem being enabled during tests and rake scripts

I now moved the gems out of the test environment, and made sure the tracer can only be enabled in development

This may not necessarily fix the demo script issue, however, because those run in development and I could still have it enabled. However, in the interim, I can be vigilant about setting that off when running the scripts

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -123,9 +123,6 @@ gem 'axlsx', '~> 2.1.0.pre'
 gem 'will_paginate', '~> 3.0.6'
 
 group :development do
-  # Nail down n+1 queries and unused eager loading
-  gem 'bullet'
-
   # Trace AR queries
   gem 'active_record_query_trace'
 end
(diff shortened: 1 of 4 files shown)


jpslav:
May also want to put in big warning messages in the log or stdout if that tracer is enabled.

joemsak:
It wont be a problem in tests anymore, how can we add a sitewide check in rake?

joemsak:
Or is it enough to check on demo001

joemsak:
Or how about I just turn it off for people automatically

joemsak:
```
 rake demo:001
I, [2015-05-28T13:13:44.222373 #55354]  INFO -- : ActiveRecord Query Trace is enabled! This will be way too slow!
I, [2015-05-28T13:13:44.222451 #55354]  INFO -- : DISABLING ActiveRecord Query Trace
```

jpslav:
@joemsak when I fixed a spec failure issue I also took out the auto shut off (but left in a STDOUT warning).

PR #329, opened 2015-05-29
removed ch4 tasks to improve demo

jpslav:
Pending blessing from @jason-holmes


philschatz:
Should ch3 be split into 2 assignments?

jason-holmes:
The issue is chapter 3 only has 2 sections (3 if you can't the intro, which we don't), so I think it would work better as a single assignment.

jpslav:
ok - @philschatz this can be merged at your discretion

PR #330, opened 2015-05-29
added default human readable names for DOK and Blooms tags

jpslav:
@Dantemss please review for ticket # 95089964


jpslav:
@philschatz I think this PR will work for now, but I think the better long term bet is to bundle give the FE the raw data that can be extracted from each tag and to have you put those data together appropriately based on that tag type -- that way the BE gets out of the business of deciding exactly how things look in the FE.

Note that I didn't see any examples of the "length" tag in the content or the tagging legend, so didn't know what to do there.

jpslav:
Strike that - I did find the "time-short" tags and added default names for them.

PR #337, opened 2015-06-02
practice widget calls biglearn

kevinburleigh75:
Note that I needed to `xit` two specs that were hard-coded to `exercise_source: :local`.


app/controllers/api/v1/courses_controller.rb
```diff
@@ -135,7 +135,7 @@ def practice_post
     consume!(practice, represent_with: Api::V1::PracticeRepresenter)
 
     entity_task = ResetPracticeWidget[
-      role: get_practice_role, condition: :local,
+      role: get_practice_role, exercise_source: :biglearn,
```
jpslav: Thanks for changing away from `condition` which was a holdover from when we actually passed the boolean search.

PR #339, opened 2015-06-02
Admin add teacher to course

karenc:
Change admin add teacher to course page to search for user instead of listing users.


app/controllers/admin/users_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Admin::UsersController < Admin::BaseController
   before_action :get_user, only: [:edit, :update]
+  skip_before_action :verify_authenticity_token, if: :json_request?
```
jpslav: @Dantemss is this not good given that we just added this recently in other APIs?

jpslav:
Looks good pending authenticity token question.

app/controllers/admin/users_controller.rb
Dantemss: Yeah any request that uses the session (current_user etc) and not OAuth needs the authenticity token to avoid XSS. Is jquery-ujs not working with these JS libraries?

Dantemss: And by XSS I really mean CSRF... blargh

Dantemss: This SO thread has plenty of solutions for ajax: http://stackoverflow.com/questions/7203304/warning-cant-verify-csrf-token-authenticity-rails

karenc:
Added csrf token to ajax request.

PR #347, opened 2015-06-03
Allow teacher to use any of the student roles in their course

karenc:


app/routines/get_user_course_roles.rb
```diff
@@ -25,5 +25,16 @@ def exec(course:, user:, types: :any)
     # Intersect the results from above
     outputs[:roles] = outputs["[:get_user_roles, :roles]"] &
                       outputs["[:get_course_roles, :roles]"]
+
+    # Add student roles if user is a teacher
+    if [:any, :student].include?(types)
```
jpslav: I like the conciseness of this change, but I think it can't go in this routine.  This routine is intended to "Return Roles belonging to a User in a Course" per a hard-to-see-comment :-)  I have found at least one use where we need to honor this "ownership" of roles.  `GetCourseUserTasks` calls this routine to figure out a user's course roles so we can then get their tasks in the course.  With this change, a teacher would all of the tasks of all of their students as their own tasks.  So I think this code needs to go into a different routine or directly into the ChooseCourseRole routine or elsewhere.  @kevinburleigh75 can weigh in if he wants.

PR #350, opened 2015-06-03
Added missing /api to the exercises url

Dantemss:


app/external/openstax/exercises/v1/exercise.rb
```diff
@@ -23,7 +23,7 @@ def uid
   end
 
   def url
-    @url ||= "#{@server_url}/exercises/#{uid}"
+    @url ||= "#{@server_url}/api/exercises/#{uid}"
```
jpslav: Isn't without the "api" the URL too?  (where someone goes to look at it) -- why do we need the "api" here?

Dantemss: I guess you're right... even though it doesn't work atm, that should be the canonical URL. I'll close the PR.

PR #352, opened 2015-06-04
Tag visible flag

karenc:
- Add visible flag for content tags
- Only return visible tags in exercise representer
- Fix tags controller update action and order tags by ID
- Make tag visible flag editable in the admin views
- Create more tag types and assign them according to the value
- Auto populate tag visible flag based on tag type


karenc:
Ready for review.

PR #353, opened 2015-06-05
Distribute tasks background jobs

joemsak:
https://www.pivotaltracker.com/story/show/95180724

Need to know how front-end response changes since `TaskPlans#publish` now runs `DistributeTasks` asynchronously

*Front End MUST note:*

```javascript
// TaskPlans#publish response on error is no longer:

response.body.errors[0]

// it is now

response.body
```

After calling for a publish to the task plan, query the task plan for the published_at attribute

```ruby
post '/api/plans/123/publish', course_id: 456

# response:
json: { task_plan: '/api/plans/123' }

get '/api/plans/123' #=> Until you get non-null 'published_at' attribute
```

### DIFF
--- a/app/controllers/api/v1/task_plans_controller.rb
+++ b/app/controllers/api/v1/task_plans_controller.rb
@@ -77,7 +77,7 @@ class Api::V1::TaskPlansController < Api::V1::ApiController
 
     if (settings = CheckValidSettings[validatable: task_plan])[:valid]
       job = DistributeTasks.perform_later(task_plan, validate: false)
-      render json: { id: task_plan.id, job: "/jobs/#{job.job_id}" }, status: :accepted
+      render json: { task_plan: api_task_plan_path(task_plan) }, status: :accepted
     else
       render json: settings[:errors], status: :unprocessable_entity
     end
--- a/app/controllers/api/v1/task_plans_controller.rb
+++ b/app/controllers/api/v1/task_plans_controller.rb
@@ -76,7 +76,7 @@ class Api::V1::TaskPlansController < Api::V1::ApiController
     OSU::AccessPolicy.require_action_allowed!(:publish, current_api_user, task_plan)
 
     if (settings = CheckValidSettings[validatable: task_plan])[:valid]
-      job = DistributeTasks.perform_later(task_plan, validate: false)
+      job = DistributeTasks.perform_later(task_plan)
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/task_plans_controller.rb
```diff
@@ -75,14 +75,11 @@ def publish
     task_plan = Tasks::Models::TaskPlan.find(params[:id])
     OSU::AccessPolicy.require_action_allowed!(:publish, current_api_user, task_plan)
 
-    result = DistributeTasks.call(task_plan)
-    if result.errors.empty?
-      respond_with task_plan, represent_with: Api::V1::TaskPlanRepresenter, location: nil
+    if (settings = CheckValidSettings[validatable: task_plan])[:valid]
```
jpslav: I know I preach routines for all the things, but I'm actually totally fine with a model validating its own internal state which is exactly what this is.  So here can we use quasi-normal style Rails AR validations?

So then we could say

```ruby
if task_plan.valid?
  blah
else
  render json: task_plan.errors, status: :unprocessable_entity # or something to this effect
end
```

app/routines/distribute_tasks.rb
```diff
@@ -1,29 +1,18 @@
...
 
   protected
 
-  def validate_json(schema, object, options = {})
-    options[:insert_defaults] = true if options[:insert_defaults].nil?
-
-    JSON::Validator.fully_validate(schema, object, options)
-  end
+  def exec(task_plan, validate: true)
+    potentially_halt_with_errors!(task_plan) if validate
```
jpslav: Based on above...

```ruby
fatal_error(blah) if !task_plan.valid?
```

app/controllers/api/v1/task_plans_controller.rb
joemsak: Ah okay, since the call is updated to 'perform_later' then the check has to be done before queueing the job, hence it got split out

Want me to move the entire check into the model?

jpslav: Hold on this; Dante may do in larger `TaskPlan` validation work.

Dantemss:
Closing this PR (commits used in https://github.com/openstax/tutor-server/pull/385, thanks Joe)

Dantemss:
And by 385 I actually meant https://github.com/openstax/tutor-server/pull/392

PR #357, opened 2015-06-08
Course Periods WIP

Dantemss:
# Done
- API
- Models
- Period routines
- Specs


app/representers/api/v1/courses/dashboard_representer.rb
```diff
@@ -18,16 +18,6 @@ class Base < Roar::Decorator
                readable: true,
                writeable: false
 
-      property :opens_at,
```
jpslav: I think the `Plan` below still needs some `opens_at` / `due_at` information so it can be displayed on a calendar.  Maybe ask @pandafulmanda what information they want for display on the calendar? (ranges, outer hull of dates, specific dates in a `taskings` list in the `Plan` JSON, etc)

app/representers/api/v1/courses/dashboard_representer.rb
```diff
@@ -43,10 +33,27 @@ class Plan < Base
                type: String,
                readable: true,
                writeable: false
+
+      collection :periods,
```
jpslav: Like above, don't think we really need a list of periods, but either a list of tasking plans or just some `opens_at` / `due_at` information for use on the calendar.

app/representers/api/v1/performance_report_representer.rb
```diff
@@ -58,11 +58,15 @@ class DataHeadings < Roar::Decorator
                type: String,
                readable: true
 
-      property :class_average,
+      property :average,
```
jpslav: Heads up @darrenmason that this field rename is coming down the pike (can't say `class_` any longer because we're not aggregating by class / course)

app/representers/api/v1/students_representer.rb
```diff
@@ -0,0 +1,7 @@
+module Api::V1
+  class RosterRepresenter < Roar::Decorator
```
jpslav: name file to match class name?

app/representers/api/v1/tasking_plan_period_representer.rb
```diff
@@ -0,0 +1,33 @@
+module Api::V1
+  class TaskingPlanPeriodRepresenter < Roar::Decorator
+
+    include Roar::JSON
+    include Representable::Coercion
+
+    property :id,
+             type: String,
+             readable: true,
+             writeable: true,
```
jpslav: Not `true`

app/representers/api/v1/courses/dashboard_representer.rb
Dantemss: Thanks, was using the wrong decorator.

app/representers/api/v1/tasking_plan_period_representer.rb
Dantemss: I think this needs to be true so the FE can set which periods get the assignment.

app/representers/api/v1/students_representer.rb
Dantemss: Oops guess I renamed the file but forgot the class name.

app/representers/api/v1/courses/dashboard_representer.rb
Dantemss: This representer has an array below that includes period id, opens_at and due_at. Let me know if that doesn't work...

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -19,32 +19,23 @@ class TaskPlanRepresenter < Roar::Decorator
...
-                self.due_at = new_time
-             }
-
     property :settings,
              type: Object,
              readable: true,
              writeable: true
 
+    collection :tasking_plans,
+               as: :periods,
```
jpslav: I would very much prefer this to not be hard-wired to `:periods` with a period-ish representer.  `tasking_plans` and `TaskingPlanRepresenter` would be better.  The representer could have a `target` with `id` and `type` and then the data of the target.

app/representers/api/v1/tasks/calendar/task_plan_representer.rb
```diff
@@ -21,16 +21,6 @@ class TaskPlanRepresenter < ::Roar::Decorator
                  readable: true,
                  writeable: false
 
-        property :opens_at,
```
jpslav: Since the courses/events endpoint is no longer being used, want to blow it and its representers away in this PR?

Dantemss: OK

app/routines/split_task_plan_tasking_plans.rb
```diff
@@ -0,0 +1,38 @@
+class SplitTaskPlanTaskingPlans
```
jpslav: Consider renaming `IndividualizeTaskingPlans` or `ConvertToRoleTaskingPlans` and take `def exec(tasking_plans)`?  When I first read "Split" I assumed it was grouping the tasking plans into multiple sets.

app/routines/distribute_tasks.rb
```diff
@@ -30,10 +30,11 @@ def exec(task_plan)
       task_plan.reload
     end
 
-    taskees = run(:get_taskees, task_plan).outputs[:taskees]
+    tasking_plans = run(:get_tasking_plans, task_plan).outputs[:tasking_plans]
 
     # Call the assistant code to create and distribute Tasks
-    outputs[:tasks] = assistant.distribute_tasks(task_plan: task_plan, taskees: taskees)
+    outputs[:tasks] = assistant.distribute_tasks(task_plan: task_plan,
```
jpslav: I know @kevinburleigh75 has some issues with the approach in this routine.  I think the assistant's `distribute_tasks` method should not be taking the full details of the `tasking_plan`, but rather just create the task and let this outer `DistributeTasks` routine create the `Tasking`s and attach them to the `Task`; in which case the assistant's method becomes `create_task` instead of `distribute_tasks`.  Thoughts?

Wonder if the assistant's `distribute_tasks` should not be taking the

app/routines/split_task_plan_tasking_plans.rb
Dantemss: OK

app/routines/distribute_tasks.rb
Dantemss: Sounds good.

Dantemss: I think passing a block to the assistant that assigns each task is the best way to do it.

app/representers/api/v1/tasking_plan_period_representer.rb
jpslav: Ah right, as it is now it would be.  I read this as the immutable ID of the tasking plan.  Like mentioned above, if this could just be TaskingPlanRepresenter with a target_id and target_type, that'd be cool.  B/c sooner than not we'll get the "let's assign this one task to just one student" or "hey we actually want to assign this task to the whole class" or "hey a researcher wants to assign a task to 5 courses".  Since the model is polymorphic, having target here and no hardcoded "period" stuff would be good.  FE could hardcode it for now if they want (always assume periods).

app/routines/distribute_tasks.rb
jpslav: I think I may know what you're referring to, but at the very least I'm looking forward to seeing what you come up with :-)

app/routines/move_user_to_course_period.rb
```diff
@@ -0,0 +1,13 @@
+class MoveUserToCoursePeriod
+  lev_routine
+
+  uses_routine GetUserCourseRoles
+  uses_routine CourseMembership::AddStudent
+
+  protected
+
+  def exec(user:, course:, period_name:)
```
jpslav: I wonder if we should add this yet.  I have a suspicion that if we call this some crazy stuff is going to happen, e.g. in the performance report (where we currently assume all users in a course have all the same tasks, which we know may not be the case with periods in the mix).

app/routines/move_user_to_course_period.rb
```diff
@@ -0,0 +1,13 @@
...
+    roles = run(GetUserCourseRoles, user: user, course: course, types: :student).outputs.roles
```
jpslav: A teacher can have multiple student roles, in which case the `roles.first` thing on the next line isn't what we want.  Might mean we have to take a role in this routine instead of a user.

app/subsystems/course_membership/add_student.rb
```diff
@@ -1,10 +1,16 @@
...
   protected
 
-  def exec(course:, role:)
-    ss_map = CourseMembership::Models::Student.create(entity_course_id: course.id, entity_role_id: role.id)
-    transfer_errors_from(ss_map, {type: :verbatim}, true)
+  def exec(period:, role:)
+    course_periods = period.course.periods.to_a
+    student = CourseMembership::Models::Student.find_by(period: course_periods, role: role)
+    student ||= CourseMembership::Models::Student.new(role: role)
+    student.period = period.to_model
```
jpslav: Hmm.  I worry a bit that `AddStudent` is doing the "move" operation.  Seems like a non-obvious side effect of something named "AddBlah".  Also seems like there might be notifications, etc about a move that would need to be in the "move" routine, but somehow also in this "add" routine because it is actually doing the move.  

Until we hammer out the ramifications of moving a student from one period to another, my vote is to remove the "move" routine and make the "add" routine freak out if the role is currently a student in the course.  

But I've been wrong before. @kevinburleigh75?

app/subsystems/course_membership/create_period.rb
```diff
@@ -0,0 +1,11 @@
+class CourseMembership::CreatePeriod
+  lev_routine
```
jpslav: might as well make it express

app/subsystems/course_membership/get_course_periods.rb
```diff
@@ -0,0 +1,12 @@
+require_relative 'models/entity_extensions'
```
jpslav: Can you remove this and move the entity_extensions into the appropriate entity?  @joemsak already did a number of these in his PR which I merged today, maybe this one is already done too.

app/subsystems/course_membership/get_course_periods.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class CourseMembership::GetCoursePeriods
+  lev_routine
```
jpslav: express me :-)

app/subsystems/course_membership/get_course_periods.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  protected
+
+  def exec(course:)
+    outputs[:periods] = Entity::Relation.new(course.periods)
```
jpslav: wow I don't understand this line!

app/routines/split_task_plan_tasking_plans.rb
```diff
@@ -0,0 +1,38 @@
...
+      when UserProfile::Models::Profile
+        Role::GetDefaultUserRole[target.entity_user]
+      when Entity::User
+        Role::GetDefaultUserRole[target]
+      when Entity::Course
+        CourseMembership::GetCourseRoles.call(
+          course: target, types: :student
+        ).outputs.roles
+      when CourseMemberShip::Models::Period
+        CourseMembership::GetCourseRoles.call(
```
jpslav: `course_membership/get_students.rb` is a more efficient call?

app/subsystems/course_membership/is_course_student.rb
```diff
@@ -4,9 +4,7 @@ class CourseMembership::IsCourseStudent
   protected
 
   def exec(course:, roles:)
-    role_ids = [roles].flatten.collect{|r| r.id}
-    outputs[:is_course_student] =
-      CourseMembership::Models::Student.where{entity_course_id == course.id}
-                                       .where{entity_role_id >> role_ids}.any?
+    outputs[:is_course_student] = course.periods.joins(students: :role)
```
jpslav: I don't think you need to joins all the way to `role` since you are just checking the `entity_role_id` attribute on `students`

app/subsystems/course_membership/remove_student.rb
```diff
@@ -0,0 +1,13 @@
+# Removes the given role from its period/course
+class CourseMembership::RemoveStudent
```
jpslav: Do we really want to destroy students?  Can we just mark them as removed?  Seems like we normally want to keep the record of their existence.

app/subsystems/tasks/assistants/i_reading_assistant.rb
```diff
@@ -82,10 +82,10 @@ def self.assign_task!(task:, taskee:)
     task
   end
 
-  def self.create_task!(task_plan:)
+  def self.create_task!(task_plan:, tasking_plan:)
```
jpslav: we've gotta remove the duplication here w.r.t. the other assistants.  maybe the earlier discussion is related.

jpslav:
it's a Herculean effort @Dantemss and it is looking good.

app/subsystems/course_membership/get_course_periods.rb
Dantemss: Entity::Relation just wraps... any relation :)

app/subsystems/course_membership/remove_student.rb
Dantemss: I think we keep the Role around. The Student just joins the Role to the period.

app/subsystems/course_membership/get_course_periods.rb
Dantemss: It works like an ActiveRecord relation but returns entities instead of AR models (assuming there is an entity for that model).

app/routines/split_task_plan_tasking_plans.rb
Dantemss: Is get_students even used anywhere?

jpslav: Dunno.  Could be used here (one query instead of iterating thru all the periods and collecting students).

app/subsystems/course_membership/get_course_periods.rb
jpslav: Fancy.

app/routines/split_task_plan_tasking_plans.rb
Dantemss: I opted to fix up the existing routines instead. We can get rid of one of them (later).

app/subsystems/tasks/models/assistant.rb
```diff
@@ -9,7 +9,7 @@ class Tasks::Models::Assistant < Tutor::SubSystems::BaseModel
   validate :code_class_existence, if: :code_class_name
 
   # Delegate all real work to the actual implementation (the "worker")
-  delegate :schema, :distribute_tasks, to: :code_class
+  delegate :schema, :create_tasks, to: :code_class
```
jpslav: Now that there's more required structure to the `create_tasks` call (in that it takes a block), do you think this AR `Assistant` should not just simply delegate the `create_tasks` call but instead implement it and call a different method on the `code_class`, e.g.

```ruby
class Tasks::Models::Assistant < Tutor::SubSystems::BaseModel
  def create_tasks(task_plan:, taskees:)
    tasks = code_class.build_tasks(task_plan: task_plan, taskees: taskees)
    tasks.collect do |task|
      yield(task, taskee) if block_given?
      task.save!
      task
  end
end
```

Just a thought, may be some reasons not to.

app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -109,12 +101,11 @@ def self.add_core_steps!(task:, exercises:)
       related_content = get_related_content_for(exercise)
 
       step = add_exercise_step(task: task, exercise: exercise)
-      step.core_group!
```
jpslav: Why write out the assignment instead of using this `!` method provided by AR enums?

app/routines/reset_practice_widget.rb
```diff
@@ -45,10 +45,10 @@ def exec(role:, exercise_source:, page_ids: [], book_part_ids: [])
                   :mixed_practice
                 end
 
-    task = Tasks::CreateTask[task_type: task_type,
-                             title: 'Practice',
-                             opens_at: Time.now,
-                             feedback_at: Time.now]
+    task = Tasks::BuildTask[task_type: task_type,
```
jpslav: Out of curiosity, what prompted the switch from `CreateTask` to `BuildTask`?

app/subsystems/course_membership/remove_student.rb
jpslav: I'm still not quite sure we want to implement move and remove yet, without a larger understanding of what UX expects to happen in addition to what kind of traceability we need.  So I'd vote we either remove these for now or put a very prominent comment in that says that these should be used with care until we really determine what they should be doing.

jpslav:
6/9/2015 review done.

app/subsystems/tasks/assistants/homework_assistant.rb
Dantemss: Because the `!` method calls update! and saves the record, which will fail because the dates are not set.

app/routines/reset_practice_widget.rb
Dantemss: Can't save the record before the dates are set.

app/subsystems/tasks/assistants/homework_assistant.rb
jpslav: That is an excellent reason ;-)

app/subsystems/tasks/models/assistant.rb
Dantemss: One reason might be that Kevin wanted to get rid of the Assistant class at some point :)

Dantemss: Ah how about this: I'll change create_tasks to build_tasks so you don't have to yield.

jpslav: ok :-)

Dantemss:
OK ready for another review/merge.

Dantemss:
Just kidding, found some stashed changes that I had forgotten about... give me a moment...

Dantemss:
OK really done now :)

app/subsystems/tasks/assistants/README.md
```diff
@@ -2,13 +2,11 @@
...
+  1. Implement the `build_tasks(task_plan:, taskees:)`
      singleton method which:
-       - Receives as inputs the TaskPlan being assigned and
-         the taskees (targets)
-       - Creates Task objects and assigns them to the taskees
-         based on the task_plan
-       - Returns the assigned Tasks
+       - Receives as inputs the TaskPlan being assigned and the taskees (targets)
+       - Builds Task objects
+       - Returns an array containing the Tasks to be assigned to the taskees (in order)
```
jpslav: I like the simplification of build_tasks not taking a block, but I just want to take a sec to point out to the team the "in order" requirement so everyone is aware that assistants needs to return tasks in order of the taskees provided to them.  This might be something in the future where we choose for `build_tasks` to return a hash of taskee to task for additional safety.

/cc @openstax/tutor-be-dev

lib/tasks/demo_001.rb
```diff
@@ -41,6 +42,8 @@ def exec(print_logs: true, book_version: :latest, random_seed: nil)
     course = create_course(name: 'Physics I')
     run(:add_book, book: outputs.book, course: course)
 
+    period = create_period(course: course)
```
jpslav: Sometime next week can you add a 2nd period where some students go so there are multiple periods for FE to test out?

PR #364, opened 2015-06-09
Reference Book

karenc:
- Add uuid and version fields to content pages
- Store uuid and version when importing cnx pages
- Add pages API to return content html for a cnx uuid
- Generate random uuid and version for content page factory
- Add cnx_id to pages returned by readings API


app/subsystems/content/routines/create_page.rb
```diff
@@ -4,12 +4,15 @@ class Content::Routines::CreatePage
 
   protected
 
-  def exec(url:, title:, content:, book_part:, chapter_section:)
+  def exec(url:, title:, content:, book_part:, chapter_section:, id:)
```
jpslav: I'm not a big fan of passing in an `id` field to this routine as it seems like it lets the caller set the primary key.  Can we just pass in the `uuid` and `version` and do the split in `ImportPage`?

karenc:
Ready for review.  Changes to links in the content html will be in another PR.

philschatz:
@dak thoughts?

jpslav:
Looks good to me.  Can merge when @dak or @philschatz weigh in (or they can just give the go ahead and we can fix issues in followup PR)

philschatz:
@dak's out and I'm fine with a followup PR

PR #365, opened 2015-06-10
added more variation to demo, not using timecop

jpslav:
The assignments are now set to be due on the day the script is run, and then on every other business day before that (the below images were run on June 9)

pastedgraphic-8

Here are the popups:

pastedgraphic-9

pastedgraphic-10

pastedgraphic-11

pastedgraphic-12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ script:
 branches:
   only:
     - master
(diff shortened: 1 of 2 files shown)


philschatz:
:+1:

PR #366, opened 2015-06-10
CDN/Remote storage of performance report export

joemsak:
https://www.pivotaltracker.com/story/show/96151740


app/uploaders/export_uploader.rb
```diff
@@ -1,9 +1,33 @@
...
+  def filename
+    return if original_filename.blank?
+
+    # Reuse hashed filename for other versions of the same file
+    return model.read_attribute(mounted_as) unless version_name.blank?
+
+    # Don't try to hash uncached files
+    return super unless cached?
+
+    original_filename || "#{content_hash}.#{file.extension}"
```
jpslav: I believe that if `original_filename` is set then `filename`, this code will often return `original_filename`.  Which means that if two different things in the system use the same `original_filename`, they'll collide because they are both stored in the `exports` folder.  Is this expected?  What does all this code do?

jpslav: @joemsak any response?

Dantemss:
PR https://github.com/openstax/tutor-server/pull/462 ate this one. Om nom nom nom.

jpslav:
(rofl) @Dantemss

PR #372, opened 2015-06-11
ExportPerformanceReport with Lev::Routine.perform_later

### LINKED ISSUE #362: Remove ExportPerformanceBookJob in favor of lev automatic jobs
https://github.com/openstax/tutor-server/blob/master/app/subsystems/tasks/jobs/export_performance_book_job.rb

joemsak:
Fix #362


config/initializers/active_job_status.rb
```diff
@@ -0,0 +1 @@
+ActiveJobStatus.store = ActiveSupport::Cache::MemoryStore.new
```
jpslav: Does this mean this gem stores its info in memory?  I suspect that'll work given that we have multiple web server instances...

joemsak: Yeah, apparently this is the only store it supports currently? But this gem is small and easy to follow we could fork it and build on it for our own purposes. I'm sure the author would appreciate it

jpslav: If `ActiveSupport::Cache::MemoryStore` works, I imagine any class that inherits from its parent (`ActiveSupport::Cache::Store`, [docs here](http://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html)) will also work.

https://github.com/redis-store/redis-activesupport seems like a candidate.  

Then we'd just need our initializer to connect to a new one of these stores using the configuration like in cache_store.rb in tutor-server (pulling config from secrets).

Gemfile
```diff
@@ -59,7 +59,7 @@ gem 'apipie-rails'
 gem 'maruku'
 
 # Lev framework
-gem 'lev', '~> 4.2'
+gem 'lev', github: 'lml/lev'
```
jpslav: Can you use the `git: 'https://github.com/rails/rails.git', ref: '4aded'` style so we lock in a version?

app/controllers/api/v1/jobs_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Api
+  module V1
+    class JobsController < ApiController
+      def show
+        render json: Lev::Status.find(params[:id])
```
jpslav: yay :-)

jpslav:
Still need to send 200 when job complete right?  And need to document where the FE can find the URL of the file (do we need a representer for the job so we can document me it via apipie?)

PR #373, opened 2015-06-12
Update links pointing to the book to relative links in page content

karenc:


nathanstitt:
Hey @karenc, this looks great.

Something else we're going to have to do is remove the html `<head>` section and `<body>` tags.  Right now the content that the `/page/<uuid>` endpoint is spitting out is a complete html page, which doesn't render very well when injected into an existing page :grin:

jpslav:
@nathanstitt do you know what cnx webview does for the head and body tags (I assume they deal with them too because they pull from archive)?  @philschatz would you prefer for for BE or FE to do these extra content cleanups?

jpslav:
@karenc looks great - thanks.  will merge.

We might in the future want to reorganize this a bit.  There is related code in https://github.com/openstax/tutor-server/blob/master/app/external/openstax/cnx/v1/page.rb#L102-L136 which may be a better home for this PR's block of CSS processing.  And I believe we could add a `all_uuids` method to `OpenStax::Cnx::V1::Book` which could pull all of the UUIDs out of the top collection's JSON (without needing to gather all the pages), and then we could get that list to the aforementioned conversion method to do everything in one pass during import.  I bring this up as an idea for later, as I know @kevinburleigh75 is already mucking around in some of this code anyway.

PR #380, opened 2015-06-16
Added courses show API

Dantemss:


app/access_policies/course_access_policy.rb
```diff
@@ -1,6 +1,8 @@
 class CourseAccessPolicy
   def self.action_allowed?(action, requestor, course)
     case action
+    when :index, :read
+      !requestor.is_anonymous?
```
jpslav: Needs to be a member of course to read the course -- probably even just the teacher.

jpslav: actually any member ok for now.

jpslav: students will eventually need this endpoint so we know which book cover / banner to display.

PR #382, opened 2015-06-16
Student Roster API

karenc:


app/subsystems/course_membership/get_student_roster.rb
```diff
@@ -0,0 +1,37 @@
+class CourseMembership::GetStudentRoster
```
jpslav: Can you move this to `/app/routines` since it knows about multiple subsystems?

jpslav:
Looks good @karenc -- one comment

app/subsystems/course_membership/get_student_roster.rb
karenc: Sure

PR #384, opened 2015-06-17
Don't change the master server process date on bootup

Dantemss:
Should fix unicorn-worker-killer timeouts in production at the beginning of requests

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -17,4 +17,8 @@ if ENV['RAILS_ENV'] == 'production'
 end
 
 require ::File.expand_path('../config/environment',  __FILE__)
+
+# Ensure that the main server process has the actual time
+Timecop.return
+
 run Rails.application


Dantemss:
I believe that the following was happening: we last booted up unicorn with timecop set to 5 days in the future. So any new workers had their date initially set to that. For some reason, this caused unicorn-worker-killer to trigger and kill them. Resetting the date before the server is done booting seems to fix this, but this requires more testing in tutor-dev.

config.ru
```diff
@@ -17,4 +17,8 @@ if ENV['RAILS_ENV'] == 'production'
 end
 
 require ::File.expand_path('../config/environment',  __FILE__)
+
+# Ensure that the main server process has the actual time
+Timecop.return
```
jpslav: Timecop not always loaded right?  If gem not present (as in production mode), I think this will blow up?

Dantemss: Good point... but config/initializers/timecop.rb does a `require 'timecop'` so I don't think Tutor would work at all in Production without timecop?

Dantemss: Easiest solution would be to put timecop in all envs and rely on the code that checks `Timecop.enabled?`

jpslav: I agree it wouldn't work at all in production.  Your solution works for me.

PR #385, opened 2015-06-18
Changed task_plan publish API to the new format (no background jobs yet)

Dantemss:


app/representers/api/v1/task_plan_representer.rb
```diff
@@ -36,5 +40,10 @@ class TaskPlanRepresenter < Roar::Decorator
                readable: true,
                writeable: true
 
+    property :is_publish_requested,
+             readable: false,
```
jpslav: Since FE is setting this value, Phil wanted it to be readable too -- `getter: -> (*) { !self.published_at.nil? }`

app/routines/distribute_tasks.rb
```diff
@@ -1,34 +1,18 @@
-require 'json-schema'
-
 class DistributeTasks
 
-  lev_routine
+  lev_routine transaction: :no_transaction # The controller already wraps this in a transaction
```
jpslav: We definitely still want this to be in a transaction when called on its own or via a background job, so I'd prefer to not have `:no_transaction` here, even temporarily.  Can we just call this routine after the update in the controller?

app/routines/distribute_tasks.rb
```diff
@@ -1,34 +1,18 @@
...
   def exec(task_plan)
-    owner = task_plan.owner
-    assistant = task_plan.assistant
-
-    # Validate the given settings against the assistant's schema
-    # Intervention settings already included when the task_plan was saved
-    err = validate_json(assistant.schema, task_plan.settings)
+    task_plan.publish_last_requested_at = Time.now
 
-    fatal_error(code: :invalid_settings, message: 'Invalid settings', data: err) unless err.empty?
```
jpslav: Too much to leave as 

```ruby
fatal_error(code: :invalid_settings, message: 'Invalid settings', data: err) unless task_plan.valid?
```

app/representers/api/v1/task_plan_representer.rb
jpslav: I admit this reads a bit funny once the publish is complete... you might check with @philschatz about his preferences here.

app/subsystems/tasks/models/task_plan.rb
```diff
@@ -15,4 +19,28 @@ class Tasks::Models::TaskPlan < Tutor::SubSystems::BaseModel
   validates :owner, presence: true
   validates :type, presence: true
 
+  before_save :validate_settings
```
jpslav: Why not just `validate :validate_settings`?

app/subsystems/tasks/models/task_plan.rb
```diff
@@ -15,4 +19,28 @@ class Tasks::Models::TaskPlan < Tutor::SubSystems::BaseModel
...
+
+  def publish_last_requested_at
+    published_at
+  end
+
+  def publish_last_requested_at=(val)
```
jpslav: seems a bit weird to me to double book the `published_at` timestamp to hold the `requested_at` field too?

jpslav:
Review complete

app/routines/distribute_tasks.rb
Dantemss: Nah, because the standard_x methods print the output, so it needs to happen before they return (or we need to not use them)

app/routines/distribute_tasks.rb
Dantemss: Maybe just need a transfer_errors_from at the end?

app/subsystems/tasks/models/task_plan.rb
Dantemss: Oh yeah guess I could do that :)

app/subsystems/tasks/models/task_plan.rb
Dantemss: Only temporarily...

app/representers/api/v1/task_plan_representer.rb
Dantemss: Is it ok for this to return true for the current request, then be hidden afterwards?

app/routines/distribute_tasks.rb
jpslav: Then I think we need a `SaveTaskPlan` routine that replaces the `standard_blahs` here and `uses_routine DistributeTasks` (and `DistributeTasks` reverts to not having this `:no_transaction`).

app/representers/api/v1/task_plan_representer.rb
philschatz: Ooh, the outcome of this value might answer "What should happen when teacher clicks Publish and publish hasn't finished yet". 

Should teacher be able to:

1. Save as Draft
2. Publish again
3. Watch paint dry until Publish completes or fails

@jivewise thoughts?

app/routines/distribute_tasks.rb
jpslav: Hmmm... if the settings are invalid I think we want to bail out before doing the expensive work of actually distributing the tasks?  

I see that my suggested code is dumb because I'm saying to give a specific fatal error if anything invalid in the task plan.  I guess if the settings are invalid the assistant will puke early on anyway?  (Actually how do assistants give back errors?)

app/subsystems/tasks/models/task_plan.rb
jpslav: ok

app/subsystems/tasks/models/task_plan.rb
jpslav: :-)

app/routines/distribute_tasks.rb
Dantemss: Sigh... I think I'll just readd the transaction... hopefully it will work fine even with a nested transaction... as long as lev doesn't try to set the isolation level.

philschatz:
@jivewise I think this should relate to your work on re-enabling drafts, no?

jivewise:
Yep, keeping an eye on these changes.

Dantemss:
@jpslav ready for second review/merge (when FE is ready)

jivewise:
The front end for this PR: https://github.com/openstax/tutor-js/pull/406

philschatz:
Is this PR only waiting on FE changes?

Dantemss:
I believe so, yes... although I'm still working on making publish go to a background job.

Dantemss:
The API change itself doesn't depend on the background job.

jivewise:
Okay, I'll try to finish this up today on the FE.

jpslav:
All looks good to me -- can be merged when @philschatz and @jivewise are ready for it.

Fredasaurus:
Is my input needed on this?

jpslav:
@jivewise and @philschatz can we merge this in?

jivewise:
https://github.com/openstax/tutor-server/pull/385

philschatz:
Corresponding  openstax/tutor-js#424 is merged

PR #387, opened 2015-06-18
Performance report by periods

karenc:
- Change performance report to be reverse chronological (latest first)
- Change performance report APIs to be broken down by period
 - Performance report API returns a list, one set of data per period
 - Performance report exports have one worksheet per period
- Change period hash to period_id in performance report API
- Update SetupPerformanceReportData to do as the comments say
- Use exact match in performance API test to test calculation of average
- Add more students and periods to performance API test
-  Fix code for getting exercises in Tasks::GetPerformanceReport

### DIFF
--- a/app/subsystems/course_membership/get_students.rb
+++ b/app/subsystems/course_membership/get_students.rb
@@ -3,10 +3,10 @@ class CourseMembership::GetStudents
 
   protected
 
-  def exec(course)
+  def exec(period:)
     outputs[:students] = CourseMembership::Models::Student
                            .includes(:role)
-                           .where(period: course.periods)
+                           .where(period: period)
                            .collect(&:role)
   end
 end
--- a/spec/controllers/api/v1/courses_controller_spec.rb
+++ b/spec/controllers/api/v1/courses_controller_spec.rb
@@ -694,7 +694,7 @@ RSpec.describe Api::V1::CoursesController, type: :controller, api: true,
               {
                 type: 'homework',
                 id: kind_of(Integer),
-                status: 'in_progress',
+                status: 'completed',
                 exercise_count: 4,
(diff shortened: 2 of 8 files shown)


philschatz:
Adding @darrenmason since he is working on the corresponding `tutor-js` PR

karenc:
Ready for review.

karenc:
WIP Changing the period hash to just period_id.

karenc:
Ready for review.

app/subsystems/course_membership/get_students.rb
```diff
@@ -3,10 +3,10 @@ class CourseMembership::GetStudents
 
   protected
 
-  def exec(course)
+  def exec(period)
```
jpslav: Can you make this a named argument so that folks don't accidentally pass in a course?

spec/controllers/api/v1/courses_controller_spec.rb
```diff
@@ -681,10 +681,11 @@
...
+          period_id: course.periods.first.id.to_s,
           data_headings: [
-            { title: 'Homework task plan', average: kind_of(Float) },
+            { title: 'Homework 2 task plan', average: kind_of(Float) },
             { title: 'Reading task plan' },
-            { title: 'Homework 2 task plan', average: kind_of(Float) }
+            { title: 'Homework task plan', average: kind_of(Float) }
           ],
           students: [{
             name: kind_of(String),
```
jpslav: Any chance you could put in what the names should be and add some more student name checks?
Also check for the presence of another period?

Will be doing some optimizations of the perf report (again) and would be good for spec to be working with these other checks in place.

jpslav:
Looks good karen, just two comments.

PR #388, opened 2015-06-18
rename :trouble to :is_trouble in dashboard represented

kevinburleigh75:
Yup.  That's what it does.


jpslav:
This PR is trouble.

PR #391, opened 2015-06-22
GuidesController#student and #teacher with periods

joemsak:
https://github.com/openstax/napkin-notes/blob/master/phil/s014-periods-talk.md


app/controllers/api/v1/courses_controller.rb
```diff
@@ -109,15 +109,15 @@ def dashboard
...
-    #{json_schema(Api::V1::CourseStatsRepresenter, include: :readable)}
+    #{json_schema(Api::V1::CourseGuideRepresenter, include: :readable)}
   EOS
-  def stats
+  def student_guide
     course = Entity::Course.find(params[:id])
     role = get_course_role(types: :student)
-    course_stats = GetCourseStats[role: role, course: course]
-    respond_with course_stats, represent_with: Api::V1::CourseStatsRepresenter
+    course_guide = GetCourseGuide[role: role, course: course]
```
jpslav: Thanks for renaming to `GetCourseGuide` -- much clearer.

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,163 @@
...
+            children: chapters
+          }
+        }
+      end
+    else
+      course_guide = {
+        periods: outputs[:periods].collect do |period|
+          chapters = compile_chapters # cant memoize
+
+          {
```
jpslav: See https://github.com/openstax/napkin-notes/pull/17

Just a period_id with the other stats inline, not nested under `stats:`

app/controllers/api/v1/courses_controller.rb
```diff
@@ -109,15 +109,28 @@ def dashboard
...
 
-  api :GET, '/courses/:id/stats(/role/:role_id)', 'Returns course stats for Learning Guide'
+  api :GET, '/courses/:id/student_guide',
+            'Returns course guide for Learning Guide'
   description <<-EOS
-    #{json_schema(Api::V1::CourseStatsRepresenter, include: :readable)}
+    #{json_schema(Api::V1::CourseGuideRepresenter, include: :readable)}
   EOS
-  def stats
+  def student_guide
```
jpslav: Can you go ahead and move student_guide and teacher_guide into their own `GuidesController` or something to that effect? (to start reducing what is all lopped into `CoursesController`)

joemsak: Great idea, a good step addressing our concerns about this controller and its spec. Thanks

app/routines/get_course_guide.rb
joemsak: K thanks!

app/controllers/api/v1/courses_controller.rb
joemsak: Thank you

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,147 @@
...
+    as: :is_teacher
+
+  protected
+  def exec(role:, course:)
+    @role, @course = role, course
+
+    get_task_steps
+
+    run(:get_course_books, course: course)
+    run(:get_periods, course: course, roles: role)
```
joemsak: It now gets the correct periods for the role @jpslav

app/subsystems/course_membership/get_course_periods.rb
```diff
@@ -1,10 +1,23 @@
...
 
-  def exec(course:)
-    outputs[:periods] = Entity::Relation.new(course.periods)
-  end
+    protected
+    def exec(course:, roles: [])
+      roles = [roles].flatten
+      periods = Entity::Relation.new(course.periods)
 
+      outputs[:periods] = if roles.any?
```
joemsak: @jpslav Here is where I made the update to get course periods for a provided set of roles

app/controllers/api/v1/guides_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      description <<-EOS
+        #{json_schema(Api::V1::CourseGuideRepresenter, include: :readable)}
+      EOS
+      def student
+        course = Entity::Course.find(params[:id])
+        guide = GetCourseGuide[role: role(course, :student), course: course]
+        respond_with guide, represent_with: Api::V1::CourseGuideRepresenter
+      end
+
+      api :GET, '/courses/:id/teacher_guide(/role/:role_id)',
```
jpslav: Think the `(/role/:role_id)` documentation needs to be on the other endpoint?

app/controllers/api/v1/guides_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      private
+      def role(course, types = :any)
+        result = ChooseCourseRole.call(user: current_human_user.entity_user,
+                                       course: course,
+                                       allowed_role_type: types,
+                                       role_id: params[:role_id])
+        if result.errors.any?
+          raise(IllegalState, result.errors.map(&:message).to_sentence)
```
jpslav: This will result in a 500 error via https://github.com/openstax/tutor-server/blob/b1d912d1702b12f9d2015dc8281ce47a0b758072/config/initializers/controllers.rb#L24-L66

So instead we probably want to raise `SecurityTransgression` for errors with `:code` of `:invalid_user` or `:invalid_role`.  For other things we can raise `StandardError` or something and trigger an appropriate 500.

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,147 @@
...
+  private
+  attr_reader :role, :course
+
+  def compile_course_guide
+    if run(:is_teacher, course: course, roles: role).outputs.is_course_teacher
+      outputs[:periods].collect do |period|
+        { period_id: period.id }.merge(course_stats)
+      end
+    else # only 1 period for student role
+      [{ period: { id: outputs[:periods].last.id },
```
jpslav: @joemsak - I think we'll want to be consistent with how we refer to Periods from the student or teacher learning guides, and per our recent pow-wow about this we'll probably want `period_id` instead of a `period` hash.

@philschatz @darrenmason, two questions:

1. Do you want the period information included in the student learning guide JSON?
2. If so, how do you feel about the following JSON:

Student Guide
```ruby
{
  period_id: "42",
  # guide stats here
}
```

Teacher Guide
```ruby
[
  {
    period_id: "23",
    # guide stats here
  },
  # continue with more periods here
]
```

If you don't want the period information in the student guide JSON, we'd just leave it out and it'd otherwise look like the above.

I think this is inline with our discussions, but just want to double check since the student guide isn't mentioned in the periods napkin note.

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,147 @@
...
+         stats: course_stats }]
```
jpslav: Also per the periods pow-wow I don't think the course stats are supposed to be nested inside of a `stats` key any longer.  @philschatz and @darrenmason can confirm.

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,147 @@
...
+    end
+  end
+
+  def course_stats
```
jpslav: Hmmm... looking around.  Seems the stats aren't filtered by period?  They need to be.

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,147 @@
...
+    { title: outputs.toc.title, # toc is root book
+      page_ids: compile_chapters.collect { |cc| cc[:page_ids] }.flatten.uniq,
+      children: compile_chapters }
+  end
+
+  def get_task_steps
+    if run(:is_teacher, course: course, roles: role).outputs.is_course_teacher
+      run(:get_course_roles, course: course, types: :student)
```
jpslav: Maybe the period filtering happens around here?

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,147 @@
...
+      run(:get_role_task_steps, roles: outputs.roles)
+    else
+      run(:get_role_task_steps, roles: role)
+    end
+  end
+
+  def compile_chapters
+    @chapters ||= task_steps_grouped_by_book_part.collect { |book_part_id, task_steps|
```
jpslav: With multiple calls to `course_stats` I don't think we can memoize here.

app/routines/get_course_guide.rb
jpslav: Actually if you don't need the `period_id` in the student guide JSON that'd be good to know.

app/routines/get_course_guide.rb
jpslav: Reading `GetCoursePeriods` I don't understand this call when `role` is a teacher role.  If it is a teacher requesting the guide, they'll want all periods.  `GetCoursePeriods` in this case seems to want to return the periods that the teacher is in, but they won't be in any periods (only students are in periods).

If a student is requesting the guide, we don't even really care about periods, we just want that student's tasks to drive the generation of the stats.

app/subsystems/course_membership/get_course_periods.rb
```diff
@@ -1,10 +1,28 @@
...
+
+      outputs[:periods] = if roles.any?
+                            periods_in_roles(course.periods, roles)
+                          else
+                            Entity::Relation.new(course.periods)
+                          end
+    end
 
+    private
+    def periods_in_roles(periods, roles)
```
jpslav: I don't quite understand the expression "periods_in_roles" (as roles are members of periods but not the other way around).

app/subsystems/course_membership/get_period_roles.rb
```diff
@@ -1,28 +1,28 @@
-class CourseMembership::GetPeriodRoles
```
jpslav: Please revert the indentation change :-D

jpslav:
Review complete.

app/subsystems/course_membership/get_period_roles.rb
joemsak: For this kind of change, I thought we were trying to encourage this pattern:

```ruby
module SubsystemName
  class RoutineName
  end
end
```
 is that okay?

app/routines/get_course_guide.rb
joemsak: Yeah I wasn't sure either... I was going back and forth on it. Looping code is confusing that way ;-P

app/routines/get_course_guide.rb
joemsak: Oh, okay, I was still referencing this note: https://github.com/openstax/napkin-notes/blob/master/phil/s014-periods-talk.md#formats is it out of date?

app/routines/get_course_guide.rb
joemsak: Ah, okay, I believe I was fixing this part to work with the current representer correctly, it expects it as `period: { id: }`

app/routines/get_course_guide.rb
joemsak: Ah okay perhaps I assumed the teacher role would get all periods from that sub routine?

app/controllers/api/v1/guides_controller.rb
joemsak: This looks like the code was already existing, as I don't personally remember writing anything to do with `IllegalState` however I can totally make the changes

joemsak: Yeah, I just copy pasted existing code from CoursesController

app/routines/get_course_guide.rb
joemsak: But I think in the response it turns it into period_id?

app/routines/get_course_guide.rb
joemsak: Ah right I think this is where I left off, saying that it's still usable by FED to get it wired up but the data need to be connected to periods/book/chapter

app/routines/get_course_guide.rb
jpslav: That napkin note was the discussions (that was converted to a now-obsolete [google doc](https://docs.google.com/document/d/1d671s6B97V4d5qwvqMmB8p1ujKOXWtErWvQESxfjTa0/edit)) and eventually made into a PR [commit](https://github.com/openstax/napkin-notes/commit/5d742b3aaac6b3ff7c67883bc7e086f66a12fe79) into a [sprint 13 periods napkin note](https://github.com/openstax/napkin-notes/blob/master/phil/s013-course-periods.md)

app/routines/get_course_guide.rb
jpslav: Yep, the representer will likely need to change to fit what phil and darren are looking for (which may include no `period_id` in the student JSON).

spec/support/setup_course_guide.rb
```diff
@@ -24,13 +24,12 @@ class SetupCourseStats
...
+
     puts "=== Creating course ==="
     outputs.course = course || run(:create_course, name: 'Physics').outputs.course
 
-    puts "=== Creating a course period ==="
-    outputs.period = period || run(:create_period, course: course).outputs.period
-
     puts "=== Fetch & import book ==="
     run(:fetch_and_import_book, id: '93e2b09d-261c-4007-a987-0b3062fe154b')
     run(:visit_book, book: outputs.book, visitor_names: :page_data)
```
jpslav: Minor but can do both visits at once to save a little time:

```ruby
run(:visit_book, book: outputs.book, visitor_names: [:page_data, :toc])
```

app/representers/api/v1/course_guide_representer.rb
```diff
@@ -14,6 +14,7 @@ class CourseStatsRepresenter < Roar::Decorator
 
     collection :children,
                readable: true,
-               writeable: false
+               writeable: false,
+               extend: CourseGuideChildrenRepresenter
```
jpslav: I think this can just be `CourseGuideChildRepresenter` and you can get rid of the `CourseGuideChildrenRepresenter`... ?

app/representers/api/v1/course_guide_child_representer.rb
```diff
@@ -0,0 +1,51 @@
+module Api::V1
+  class CourseGuideChildRepresenter < Roar::Decorator
+    include Roar::JSON
+    include Representable::Coercion
+
+    property :id,
```
jpslav: Remove this from here and if populating it elsewhere get rid of it there too (unless used in some way I'm not seeing).  The "id" field is not in https://github.com/openstax/napkin-notes/blob/master/phil/s009-learning-guide.md and while it is still in the tutor-js dummy JSON, Darren says they are not using it.

spec/controllers/api/v1/guides_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      let!(:student_3_role) {
+        AddUserAsPeriodStudent.call(period: period,
+                                    user: user_3.entity_user).outputs[:role]
+      }
+
+      it 'returns the student guide for the logged in user' do
+        expect(get_course_guide).to receive(:[])
+                                    .with(role: student_role, course: course)
+                                    .and_return(course_guide)
```
jpslav: helping me out with spec-land -- why will this return `course_guide`?

joemsak: The test is expecting that the guide method `[]` was called with the correct arguments and in order to not blow up what the controller needs to do after the call, I had it return something. It's possible I could use `.and_call_original`

joemsak: Looks like there's some leftover stubbing up above that I carried over from courses controller spec and perhaps didn't take enough ownership in analyizing

joemsak:
Replaced by #428 with new branch name

PR #392, opened 2015-06-22
Publish Tasks in background jobs

Dantemss:
Ready for review.


app/controllers/api/v1/jobs_controller.rb
```diff
@@ -1,21 +1,30 @@
...
+    short_description 'Represents queued jobs in the system'
+    description <<-EOS
+      ActiveJob jobs description to be written...
+    EOS
+  end
+
+  api :GET, '/jobs/:id', 'Returns job statuses'
+  description <<-EOS
+    Returns queued job statuses in the system
+    { status: \\[:queued, :working, :complete\\] }
```
jpslav: There is more in the job json.  @joemsak can you help tell Dante what there is?  (e.g. there's a url field I think which is used by the perf book export).  Would be nice to have a representer so we can have descriptions of the fields for the FE folks.

jpslav: also I think the status output is probably a string rather than a symbol.

Dantemss: Uh the only thing I changed on this line was adding the `\\` so Maruku wouldn't complain... the rest was in Joe's commits.

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -29,6 +34,12 @@ class TaskPlanRepresenter < Roar::Decorator
              readable: true,
              writeable: false
 
+    property :progress_url,
```
jpslav: To be explicit about the type, can this be `publish_job_url`?

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -29,6 +34,12 @@ class TaskPlanRepresenter < Roar::Decorator
...
+             type: String,
+             readable: true,
+             writeable: false,
+             getter: ->(*) { "/api/jobs/#{publish_job_uuid}" }
```
jpslav: If `publish_job_uuid` is not set, this will return `/api/jobs/` which isn't a valid route.

Can it be

```ruby
getter: ->(*) { publish_job_uuid? ? "/api/jobs/#{publish_job_uuid}" : nil }
```

app/controllers/api/v1/jobs_controller.rb
jpslav: FYI asked @joemsak to spruce the job stuff up in https://github.com/openstax/tutor-server/issues/406

app/subsystems/tasks/models/task_plan.rb
```diff
@@ -35,4 +37,12 @@ def valid_settings
     false
   end
 
+  def not_open
+    return if tasks.none? { |tt| tt.opens_at.nil? || tt.opens_at <= Time.now } || \
```
jpslav: Move logic to `Task` and make it `tasks.none?(&:open?)`

app/subsystems/tasks/models/task_plan.rb
```diff
@@ -35,4 +37,12 @@ def valid_settings
     false
   end
 
+  def not_open
```
jpslav: Since this is more than just `not_open` can we rename to `changes_are_allowed` or something?

jpslav:
Review complete.

jpslav:
Still need to have a way to start / monitor / restart resque workers.  Any deep thoughts?

app/representers/api/v1/task_plan_representer.rb
Dantemss: I think Phil had requested `progress_url`? But I'll change it.

app/representers/api/v1/task_plan_representer.rb
Dantemss: Fixed in https://github.com/openstax/tutor-server/pull/397

app/subsystems/tasks/models/task_plan.rb
Dantemss: OK

app/subsystems/tasks/models/task_plan.rb
Dantemss: OK

Dantemss:
If resque doesn't come with its own, use god/monit (ask LN/Chris)

Dantemss:
If the background stuff is not ready, we could use the inline_adapter for resque in the meantime.

jpslav:
Yep @Dantemss I did what you recommended in https://github.com/openstax/tutor-server/commit/590634edf4a1ea4da6c3ad187888907ec6b6eed1

PR #394, opened 2015-06-23
Always include timecop (since it's required in an initializer)

Dantemss:


jpslav:
Can you add a `before_filter { Timecop.enabled? }` to the admin views that deal with mucking with time?  (and maybe block the link from the admin console too) -- just so we don't go accidentally changing time on the production server? :-)

Dantemss:
@jpslav done

PR #396, opened 2015-06-23
Specify page with version as separate route

nathanstitt:
Rails routes seemingly cannot use the '@' symbol as a separator.  The original really seems like it should have worked but it would never set the `:version` parameter.

Previously a request for `/api/pages/1bb611e9-0ded-48d6-a107-fbb9bd900851@2` would set `:uuid` to `1bb611e9-0ded-48d6-a107-fbb9bd900851@2` and leave `:version` nil.  Since there isn't a page with the `@2` as part of it's `:uuid`, it would always return 404.

I also played with using regex based constraints.  They worked, but I thought this was a bit clearer.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -70,7 +70,10 @@ Rails.application.routes.draw do
       end
     end
 
-    get 'pages/:uuid(@:version)', controller: :pages, action: :get_page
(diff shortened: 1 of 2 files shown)


karenc:
Looks good to me.

jpslav:
Love it - thanks @nathanstitt!

PR #399, opened 2015-06-25
Admin CRUD Districts, Schools

joemsak:
https://www.pivotaltracker.com/story/show/97198934

https://www.pivotaltracker.com/story/show/97198938

https://www.pivotaltracker.com/story/show/97198936


app/subsystems/course_detail/build_district.rb
```diff
@@ -0,0 +1,10 @@
+module CourseDetail
+  class BuildDistrict
```
jpslav: appears to be unused

app/subsystems/course_detail/get_district.rb
```diff
@@ -0,0 +1,16 @@
+module CourseDetail
+  class GetDistrict
+    lev_routine express_output: :district
+
+    protected
+    def exec(id:, action: :read, caller:)
+      district = Models::District.find(id)
+
+      raise SecurityTransgression unless DistrictAccessPolicy.action_allowed?(
```
jpslav: Authorization checks should be restricted to controllers and handlers; don't want them in routines.

app/subsystems/course_detail/models/district.rb
```diff
@@ -0,0 +1,6 @@
+module CourseDetail
+  module Models
+    class District < Tutor::SubSystems::BaseModel
```
jpslav: Needs `validates :name, presence: true` probably also `unique: true` -- with DB constraints to match.

app/routines/create_district.rb
```diff
@@ -0,0 +1,3 @@
+class CreateDistrict
+  lev_routine delegates_to: CourseDetail::CreateDistrict
```
jpslav: I can be ok with these `app/routines` wrappers, but I seem to recall a discussion where we decided that top-level SS routines (i.e. those in the SS folder, not those in the SS `routines` folder) should be equally as accessible by outside code as code in `app/routines`, so these wrappers might not be strictly necessary.

I obviously need to write an `app/subsystems/README.md`.

app/subsystems/course_detail/get_district.rb
jpslav: This check and others like it can be moved into handlers' `authorized?` method or directly into the controller if there is no handler.

app/access_policies/district_access_policy.rb
```diff
@@ -0,0 +1,14 @@
+class DistrictAccessPolicy
+  def self.action_allowed?(action, requestor, district)
+    case action
+    when :index
+      true
```
jpslav: This is a weird case since the admin controllers already check for admin?, but I think `:index` and `:read` could also be `requestor.is_admin?` with checking through handler `authorized?` methods.

jpslav:
well-written PR, @joemsak

app/subsystems/course_detail/models/district.rb
joemsak: Done, thanks

joemsak:
Thanks @jpslav I updated it to your comments

jpslav:
Creating a school with the blank district selected fails with an exception:

```
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"[FILTERED]", "school"=>{"name"=>"Milby High School", "course_detail_district_id"=>""}, "commit"=>"Save"}
  ...
An exception occurred: PG::ForeignKeyViolation [465649] ERROR:  insert or update on table "course_detail_schools" violates foreign key constraint "fk_rails_c8ebd13d90"
DETAIL:  Key (course_detail_district_id)=(0) is not present in table "course_detail_districts".
```

because the district ID is blank string?

jpslav:
http://localhost:3001/admin/schools needs a page heading -- check other examples, setting `<% @page_header = "Heading Name Here" %>` will do it.

app/controllers/admin/courses_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Admin::CoursesController < Admin::BaseController
-  before_action :get_users, only: [:new, :edit]
+  before_action :get_users, except: [:index, :destroy]
+  before_action :get_schools, except: [:index, :destroy]
```
jpslav: Change this to `only: [:new, :edit]` -- not needed in index, destroy, create, students, update

app/controllers/admin/courses_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Admin::CoursesController < Admin::BaseController
-  before_action :get_users, only: [:new, :edit]
+  before_action :get_users, except: [:index, :destroy]
```
jpslav: Take this out, as well as the `get_users` method -- it is unused.

app/handlers/admin/courses_create.rb
```diff
@@ -15,6 +16,7 @@ def authorized?
   end
 
   def handle
-    run(CreateCourse, name: course_params.name)
+    run(CreateCourse, name: course_params.name,
```
jpslav: Need to set the course time zone (from form input)

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,8 @@
 module ApplicationHelper
+  def bootstrap_flash
```
jpslav: nice

app/routines/create_course.rb
```diff
@@ -5,8 +5,10 @@ class CreateCourse
     translations: { outputs: { type: :verbatim } },
     as: :create_course_profile
 
-  def exec(name: 'Unnamed')
+  def exec(name: 'Unnamed', school_id: nil)
```
jpslav: Remove the name default unless there's a compelling reason to keep it.  On the second call without a name this will error on the name non-uniqueness, and I don't see us really ever using the default.

app/subsystems/course_detail/create_district.rb
```diff
@@ -0,0 +1,15 @@
+module CourseDetail
+  class CreateDistrict
+    lev_routine express_output: :district
+
+    protected
+    def exec(name: 'Unnamed')
```
jpslav: Ditto on unnamed here - not seeing the practical advantage.

app/subsystems/course_detail/delete_school.rb
```diff
@@ -0,0 +1,15 @@
...
+  class DeleteSchool
+    lev_routine
+
+    uses_routine GetSchool,
+      translations: { outputs: { type: :verbatim } },
+      as: :get_school
+
+    protected
+    def exec(id:)
+      run(:get_school, id: id)
```
jpslav: Since you're in the SS here, you don't need to use the routine to get the school, just do like you did in `DeleteDistrict`

app/subsystems/course_detail/models/school.rb
```diff
@@ -0,0 +1,16 @@
+module CourseDetail
+  module Models
+    class School < Tutor::SubSystems::BaseModel
+      belongs_to :district, class_name: 'CourseDetail::Models::District',
```
jpslav: `belongs_to :district` should work (should do the SS stuff automatically because you're in a `Tutor::SubSystems::BaseModel`).

app/subsystems/course_detail/update_school.rb
```diff
@@ -0,0 +1,19 @@
...
+  class UpdateSchool
+    lev_routine express_output: :school
+
+    uses_routine GetSchool,
+      translations: { outputs: { type: :verbatim } },
+      as: :get_school
+
+    protected
+    def exec(id:, attributes: {})
+      run(:get_school, id: id)
```
jpslav: In SS, don't need to call routine to get model

app/subsystems/course_profile/models/profile.rb
```diff
@@ -1,7 +1,12 @@
 class CourseProfile::Models::Profile < Tutor::SubSystems::BaseModel
   belongs_to :course, subsystem: :entity
+  belongs_to :school, class_name: 'CourseDetail::Models::School', foreign_key: :course_detail_school_id
```
jpslav: `belongs_to :school, subsystem: :course_detail`

jpslav: oh yeah and one SS having a direct relation to another SS is against the rules ;-) 

Looks like the only reason this relation exists is to provide `school_name` which is used by a view.  I'd rather `:school` was just another option passed to `collect_extended_course_info`, following the similar pattern when we want to aggregate more course information, especially when the use is just for display purposes.

app/subsystems/course_profile/models/profile.rb
```diff
@@ -1,7 +1,12 @@
...
 
   validates :name, presence: true
   validates :timezone, presence: true,
-                       inclusion: { in: ActiveSupport::TimeZone.all.collect{ |tz| tz.name } }
+                       inclusion: { in: ActiveSupport::TimeZone.all.collect(&:name) }
```
jpslav: JP looks the other way

app/views/admin/districts/edit.html.erb
```diff
@@ -0,0 +1 @@
+<%= render 'form', url: admin_district_path(@district.id), method: :put %>
```
jpslav: like above, this template (and others) need to set the `@page_header`

app/subsystems/course_detail/models/district.rb
```diff
@@ -0,0 +1,7 @@
...
+      validates :name, presence: true, uniqueness: true
+    end
```
jpslav: as a safety mechanism, don't allow districts with schools to be destroyed.

ditto don't allow schools with courses to be destroyed.

The contents of these things should be emptied before they are destroyed

app/controllers/admin/courses_controller.rb
joemsak: Create/Update need it if there are errors and the form is re-rendered... Rendering an action won't pick up the filters or instance variables set in that action

app/handlers/admin/courses_create.rb
joemsak: Sorry, not sure I see where the time zone param is coming from in `/app/views/admin/courses/_form`

app/helpers/application_helper.rb
joemsak: :)

app/routines/create_course.rb
joemsak: Yah fair

app/subsystems/course_profile/models/profile.rb
joemsak: lol sorry it got noticed while I was in there and i just gotta clean it up i mean look at it it's so nice to look at.... :)

app/views/admin/districts/edit.html.erb
joemsak: k but i'm just gonna add them in the controller actions instead of embedded ruby code

Dantemss:
I think the issue with the FK is that the select box is returning a 0 for a blank district, when you really want a nil.

karenc:
In order for `has_many :profiles, subsystem: :course_profile` to work, I had to do this:

```diff
diff --git a/config/initializers/01-subsystems.rb b/config/initializers/01-subsystems.rb
index 06c709c..3d45c19 100644
--- a/config/initializers/01-subsystems.rb
+++ b/config/initializers/01-subsystems.rb
@@ -7,4 +7,4 @@ require 'entity/role'
 # Only these namespaces have been configured and should be extended
 # Once they are all configured, we can remove this line entirely,
 # but will still need to require the subsystems on rails boot, before any models are loaded
-Tutor::SubSystems.valid_namespaces = %w(content course_content course_profile entity tasks role course_membership)
+Tutor::SubSystems.valid_namespaces = %w(content course_content course_profile entity tasks role course_membership course_detail)
```

jpslav:
Good one @karenc

joemsak:
@Dantemss I think this is something that normally rails handles gracefully for the developer. Is our way of using handlers/routines interfering with it, you think?

PR #400, opened 2015-06-25
Use strings in student data IDs

joemsak:
Request from FED


jpslav:
Instead of changing the IDs to strings in the routines, you can just change the `type: Integer` to `type: String` in the representer (https://github.com/apotonick/representable#coercion).  Pretty sure that works (at least appears to be how we're doing it elsewhere).

PR #407, opened 2015-06-27
Added API routes for adding, moving and deleting students

Dantemss:
Merge https://github.com/openstax/tutor-server/pull/401 first.


app/controllers/api/v1/students_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+      args = @payload.to_hash.symbolize_keys.merge(period: period)
+      @result = CreateStudent.call(args)
+      @student = @result.outputs[:student]
+      OSU::AccessPolicy.require_action_allowed!(:create, current_api_user, @student)
+      raise SecurityTransgression \
+        unless @student.period.entity_course_id.to_s == params[:course_id].to_s
+    end
+
+    if @result.errors.any?
+      render json: { errors: @result.errors }, status: :conflict
```
jpslav: Any reason to not use `render_api_errors(@result.errors, :conflict)`?

Dantemss: Changed, though I left the conflict out because idk where it came from.

Gemfile
```diff
@@ -46,7 +46,8 @@ gem 'openstax_utilities', '~> 4.2.0'
 gem 'whenever'
 
 # OpenStax Accounts integration
-gem 'openstax_accounts', '~> 5.1.2'
+gem 'openstax_accounts', github: 'Dantemss/accounts-rails',
```
jpslav: Want to update to recently released version?

Dantemss: Sure

nathanstitt:
I've implemented support for student deletion and updating periods and that seems to work well. I'm unsure how creating a new student and resetting a password is supposed to work.  It appears to me that that's not implemented yet, but perhaps I've just missed those functions?

jpslav:
Resetting a password and changing names is going to be an admin function in Accounts

PR #409, opened 2015-06-29
Ensure consistent Courses API

joemsak:
From @jpslav 

@joemsak can you look into this?

https://github.com/openstax/tutor-server/blob/master/app/controllers/api/v1/courses_controller.rb#L20 seems to imply that /api/user/courses has both roles and periods, but that's not what @philschatz is seeing.

Can you make sure that both routes are showing periods and roles?

Can you also make sure that the get of /api/courses/1 is only accessible to people who are admins or who have a role in the course?

### DIFF
--- a/app/access_policies/course_access_policy.rb
+++ b/app/access_policies/course_access_policy.rb
@@ -6,9 +6,10 @@ class CourseAccessPolicy
     when :read, :readings, :task_plans
       # readings should be readable by course teachers and students because FE
       # uses it for the reference view
-      requestor.is_human? && \
-      (UserIsCourseStudent[user: requestor.entity_user, course: course] || \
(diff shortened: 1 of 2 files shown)


joemsak:
@jpslav @philschatz Hey guys, does this

https://github.com/openstax/tutor-server/blob/master/spec/controllers/api/v1/courses_controller_spec.rb#L133-L139

Clarify that the endpoint returns roles and periods when the requestor is a teacher of the course?

jpslav:
@joemsak 

I think this: https://github.com/openstax/tutor-server/blob/master/spec/controllers/api/v1/courses_controller_spec.rb#L278-L325

shows that the roles are not where we want them (at least it doesn't test for them and they aren't present when I hit `/api/courses/1` manually)

PR #410, opened 2015-06-29
Biology content

nathanstitt:
* [x] Import content
 * [x] setup courses and users for it
 * [x] Add readings
 * [x] And exercises


app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -274,7 +274,7 @@ def self.add_personalized_exercise_steps!(task_plan:, task:, taskee:)
   end
 
   def self.num_personalized_exercises
-    1
+    0
```
jpslav: I imagine this is 0 just for debugging?

nathanstitt: It's because BigLearn doesn't have any suggestions for Biology, which causes the demo script to fail when it marks exercises as complete and BigLearn then returns null.

Dantemss: Yep, there are no exercises in exercises-hsp for bio at all with the required tag (I think it's os-practice-concepts?)

lakshmivyas: @nathanstitt, @Dantemss - you can import questions into biglearn-dev (or whatever environment you use) using the admin panel and then run full refresh.

Dantemss: We know... the issue is that bio requires different tags for personalized (there are no questions with the same tags as physics personalized)

app/subsystems/tasks/placeholder_strategies/i_reading_personalized.rb
```diff
@@ -35,17 +34,16 @@ def populate_placeholders(task:)
...
+    # DONT YOU JUDGE ME!
+    if task.task_plan.owner.name =~ /biology/i
+       { _and: [
+           'apbio', 'ost-chapter-review', 'review', 'time-short',
+           { _or: ['k12phys', 'os-practice-concepts'] + task.los }
+       ] }
+    else
+      { _and: [ 'os-practice-concepts', { _or: task.los } ] }
+    end
   end
```
nathanstitt: What should be done here?  

I think the content that generates a course (the book) should set some BigLearn query conditions on the course.  Then when this routine is called later, it will read them off the course and submit them to BL.

Any other ideas?

app/subsystems/tasks/assistants/homework_assistant.rb
lakshmivyas: ah, I see.   Thanks for explaining @Dantemss.

nathanstitt:
This is basically working now, but leaving WIP until we can remove the nasty course name regex workaround.

app/subsystems/tasks/placeholder_strategies/i_reading_personalized.rb
```diff
@@ -35,17 +32,21 @@ def populate_placeholders(task:)
     task
   end
 
-  def biglearn_condition(los)
-    condition = {
+  def biglearn_condition(task)
```
jpslav: @kevinburleigh75 can you bless this code?

lib/tasks/demo_002.rb
```diff
@@ -38,10 +182,10 @@ def exec(print_logs: true, book_version: :latest, random_seed: nil)
                   55,
                   77,
                   88,
-                  %w( 1 1 1 1 1 1 0 1 1 1 ),
+                  21,
```
Dantemss: Seems kind of low? Old one was like 90.

lib/tasks/demo_002.rb
```diff
@@ -74,32 +218,32 @@ def exec(print_logs: true, book_version: :latest, random_seed: nil)
       step_types: %w( r i e r r e r r r e r e e e ),
       entries: [
                   94,
-                  %w( 1 1 0 1 1 0 1 1 1 1 1 1 0 1 ),
+                  %w( 1 1 0 1 1 0 1 1 1 1 1 1 1 0 ),
```
Dantemss: Seems odd to me that the physics assignments changed somehow?

lib/tasks/demo_001.rb
```diff
@@ -23,41 +23,64 @@ def exec(print_logs: true, book_version: :latest, random_seed: nil)
...
-      ''
-    when :stable
-      "@#{STABLE_VERSION}"
-    else
-      book_version.blank? ? '' : "@#{book_version.to_s}"
-    end
+                     when :latest
+                       ''
+                     when :stable
+                       "@#{STABLE_VERSION}"
```
Dantemss: I think each book will need a different one.

lib/tasks/demo_001.rb
```diff
@@ -23,41 +23,64 @@ def exec(print_logs: true, book_version: :latest, random_seed: nil)
...
-      run(:import_book, id: cnx_book_id)
-      log("Imported book #{cnx_book_id} from #{archive_url}.")
-    end
 
-    course = create_course(name: 'Physics I')
-    run(:add_book, book: outputs.book, course: course)
-
-    period_1 = create_period(course: course)
-    period_2 = create_period(course: course)
+    cnx_books = {
```
Dantemss: Ask JP first, but wouldn't it be better to make a separate demo script for bio only?
And if you want both courses, just run the bio demo and the physics demo?

lib/tasks/demo_001.rb
```diff
@@ -23,41 +23,64 @@ def exec(print_logs: true, book_version: :latest, random_seed: nil)
...
+      username = "student#{(ii + 1).to_s.rjust(2,'0')}"
+      user = new_user_profile(username: username).entity_user
+
+      student_courses.map do | course |
+        run(AddUserAsPeriodStudent, period: course.periods[ii%2], user: user)
+      end
 
+    end
+
+    log("Added #{teacher_profile.account.full_name} as a teacher and added #{students.flatten.count} students.")
```
Dantemss: This only lists one of the teachers (could move that part into the loop above).

Dantemss: Oh nvm, I guess the teacher is the same :)

lib/tasks/demo_001.rb
nathanstitt: I think it'd make sense to add a `book` parameter to `demo:002` rake script.  That way it could be executed as `rake demo:002[biology]`, `rake demo:002[physics]` or `rake demo:002[all]` with `all` being the default.

@jpslav thoughts?

jpslav: That seems like a good idea to me.  Please make it so @nathanstitt

lib/tasks/demo_001.rb
jpslav: I agree.

lib/tasks/demo_001.rb
jpslav: While I think NS's idea is good, I'm starting to wonder if the better long term plan would be to leave the code alone and put all of the book-specific details into config files (eg. `biology.yml`) and read those files from the generic demo code.

nathanstitt:
I've added the book parameter.  A pleasant side effect is that it can import much faster if only a partial book import is performed.

@jpslav ready for review

PR #415, opened 2015-07-06
Authorization audit

karenc:
- Restrict courses readings API to course teachers and students
  - Courses readings API was originally going to be restricted to course
    teachers in commit cfb649ae.  Course students need access to it too
    because the readings API is also used for the reference view in the FE.
- Add security transgression tests for courses exercises API
- Restrict courses plans API to course teachers and students
- Add illegal state tests for courses dashboard API
- Add illegal state tests for courses stats API
- Add illegal state tests for courses practice (GET) API
- Add security transgression tests for tasks API
- Add security transgression for task steps show API

### DIFF
--- a/app/access_policies/course_access_policy.rb
+++ b/app/access_policies/course_access_policy.rb
@@ -3,12 +3,12 @@ class CourseAccessPolicy
     case action
     when :index
       !requestor.is_anonymous?
-    when :read
+    when :read, :readings
+      # readings should be readable by course teachers and students because FE
+      # uses it for the reference view
       requestor.is_human? && \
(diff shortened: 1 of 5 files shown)


jpslav:
I believe the course readings API is also used by the reference view (to get the TOC), so that will need to be open to anyone on the course -- you might double check me on that.

karenc:
Does the jobs API need to be restricted?

jpslav:
Looks good to me @karenc -- still WIP?

karenc:
Ready for review.

Removed extra tests for courses students API because the test was moved in master and those tests have already been added.

Added security transgression tests for tasks and task steps API.

PR #416, opened 2015-07-07
Extract AP LO's from the Bio content and use them

Dantemss:
- Requires small FE change in the assignment builder (it has to support tag type: "aplo", probably displaying it as "AP LO: blah blah")


Dantemss:
@nathanstitt I believe this is yours to review, since I'm making a PR into your branch :)

app/representers/api/v1/tag_representer.rb
```diff
@@ -18,12 +18,8 @@ class TagRepresenter < Roar::Decorator
...
              type: String,
              readable: true,
-             writeable: false
-
-    property :description,
-             type: String,
-             readable: true,
-             writeable: false
+             writeable: false,
+             getter: ->(*) { "#{name} #{description}".strip }
```
jpslav: Why not let the FE decide what to do with the name and description (seems like we should leave them separate since we have that distinction from the content, and FE can choose how to combine or not combine).

jpslav:
Yep @nathanstitt this one is yours (I just made one comment)

PR #421, opened 2015-07-07
Single grasp check

Dantemss:
If more than 1 exercise in a feature, enable try another and only create 1 exercise in the assignment


Dantemss:
This does not affect any of the stuff in the tests, so I need more info to test it (which module is affected by this?)

jpslav:
cc @kjdav who may be able to answer Dante's question

kjdav:
@Dantemss can you test in Chapter 3, Acceleration? (I believe that is chapter 3 module 2). Check out Watch Physics, should have 2 grasp checks in it.

also you can @ me @kjdav. I only saw this bc JP alerted me. :-)

Dantemss:
@kjdav I think that was removed from the demo content?
3.2 has no video
3.1 has a video, but it has only 1 exercise

jpslav:
looks good to me @Dantemss but I can't tell if you need me to wait so that you can test it.

kjdav:
OK. I think we should test it on real content if possible... Copying @alinams -- @Dantemss says we need an example of a feature with multiple grasp checks in the content. @Dantemss when do you need this by?

Dantemss:
Whenever, but maybe we have something from the old Physics content? I just can't find it anymore...

openstaxalina:
Based on Edward's analysis, this only happens in 2 times in the first 10 chapters.  And they're both in chapter 3.  Here's a collection with that content: https://legacy-staging-tutor.cnx.org/content/col10027/latest/

Module titled "Acceleration" (first non-intro module): there are 2 grasp checks on the Watch Physics: Acceleration feature.

Module titled "Representing Acceleration with Equations and Graphs" (second non-intro module): there are 2 grasp checks on the Virtual Physics: The Moving Man (Part 2) feature. 

@Dantemss I'm not sure this is all you need...do you need this imported into Tutor?  If so I probably need to restructure the collection...

Dantemss:
For future reference, the archive content is at https://archive-staging-tutor.cnx.org/contents/bc9f0535-6101-4ca2-8d06-9cebbb0f9a68@6.1

Dantemss:
OK tested locally with the content Alina provided, fixed it up and seems to be working now.

PR #422, opened 2015-07-08
Task counts caching

kevinburleigh75:


app/subsystems/tasks/models/task.rb
```diff
@@ -114,34 +114,68 @@ def core_task_steps_completed_at
...
   end
 
   def actual_and_placeholder_exercise_count
-    task_steps(true).includes(:tasked).select{ |task_step|
-      task_step.exercise? || (task_step.placeholder? && task_step.tasked.exercise_type?)
-    }.length
+    exercise_steps_count + placeholder_exercise_steps_count
   end
 
   def completed_exercise_count
```
jpslav: This comment could be addressed in a follow-on PR...
--------
Get rid of this and replace it's one use in the `dashboard_representer` with `completed_exercise_steps_count (it already has an `as:` and everything so the JSON output won't change).  

Ditto for `correct_exercise_count`, which may need a bit more work to make sure output API doesn't change (an `as:`).

Ok there are a number of these aliased methods for counts -- I would prefer to eliminate them -- you?

app/subsystems/tasks/models/task.rb
```diff
@@ -114,34 +114,68 @@ def core_task_steps_completed_at
...
-    exercise_steps.select{|step| step.tasked.is_correct?}.count
+    correct_exercise_steps_count
   end
 
   def exercise_steps
-    task_steps(true).includes(:tasked).select{|task_step| task_step.exercise?}
+    task_steps.preload(:tasked).select{|task_step| task_step.exercise?}
   end
 
   def completed_exercise_steps
```
jpslav: FWIW, this is no longer used anywhere (follow-on cleanup PR)

app/subsystems/tasks/models/task.rb
```diff
@@ -156,4 +190,46 @@ def opens_at_or_due_at
...
+    self.completed_steps_count = task_steps.count(&:completed?)
+  end
+
+  def update_core_steps_count(task_steps:)
+    self.core_steps_count = task_steps.count(&:core_group?)
+  end
+
+  def update_completed_core_steps_count(task_steps:)
+    self.completed_core_steps_count = task_steps.select(&:core_group?)
+                                                .count(&:completed?)
```
jpslav: Minor, but instead of iterating through the list of steps twice (once in `select`, once in `count`) just do them together:

```ruby
self.completed_core_steps_count = task_steps.count{|ts| ts.core_group? && ts.completed?}
```

app/subsystems/tasks/models/task.rb
```diff
@@ -156,4 +190,46 @@ def opens_at_or_due_at
...
+  end
+
+  def update_exercise_steps_count(task_steps:)
+    self.exercise_steps_count = task_steps.count(&:exercise?)
+  end
+
+  def update_completed_exercise_steps_count(task_steps:)
+    self.completed_exercise_steps_count = task_steps.select(&:exercise?)
+                                                    .count(&:completed?)
```
jpslav: Ditto collapse into one iteration

app/subsystems/tasks/models/task.rb
```diff
@@ -156,4 +190,46 @@ def opens_at_or_due_at
...
+  end
+
+  def update_correct_exercise_steps_count(task_steps:)
+    self.correct_exercise_steps_count = task_steps.select(&:exercise?)
+                                                  .select{|step| step.tasked.is_correct?}
+                                                  .count
```
jpslav: collapse from 3 to 1 iterations

app/subsystems/tasks/models/task.rb
```diff
@@ -156,4 +190,46 @@ def opens_at_or_due_at
...
+  end
+
+  def update_placeholder_steps_count(task_steps:)
+    self.placeholder_steps_count = task_steps.count(&:placeholder?)
+  end
+
+  def update_placeholder_exercise_steps_count(task_steps:)
+    self.placeholder_exercise_steps_count = task_steps.select(&:placeholder?)
+                                                      .select{|step| step.tasked.exercise_type?}
+                                                      .count
```
jpslav: 3 to 1 iteration

app/subsystems/tasks/models/task.rb
```diff
@@ -114,34 +114,68 @@ def core_task_steps_completed_at
...
+
+    update_steps_count(task_steps: steps)
+    update_completed_steps_count(task_steps: steps)
+    update_core_steps_count(task_steps: steps)
+    update_completed_core_steps_count(task_steps: steps)
+    update_exercise_steps_count(task_steps: steps)
+    update_completed_exercise_steps_count(task_steps: steps)
+    update_correct_exercise_steps_count(task_steps: steps)
+    update_placeholder_steps_count(task_steps: steps)
+    update_placeholder_exercise_steps_count(task_steps: steps)
```
jpslav: I guess it'd might be a smallish performance improvement to just iterate once? (maybe not worth it?)

```ruby
[:steps_count, :completed_steps_count, ...].each{|count_attribute| self.send("#{count_attribute}=",0)}

task_steps.each do |step|
  self.steps_count += 1
  self.completed_steps_count +=1 if step.completed?
  ... etc ...
end
```

app/subsystems/tasks/models/task.rb
```diff
@@ -114,34 +114,68 @@ def core_task_steps_completed_at
...
+      update_step_counts!
     end
   end
 
+  def update_step_counts!
+    # puts "#{self.id} updating step counts"
+
+    steps = self.task_steps.to_a
+
+    update_steps_count(task_steps: steps)
```
jpslav: "update" sounds like "increment" to me.  does it to you?  "reset"?

app/subsystems/tasks/models/task.rb
```diff
@@ -114,34 +114,68 @@ def core_task_steps_completed_at
...
+
+    if core_task_steps_completed? && placeholder_steps_count > 0
+      # puts "#{self.id} running placeholder strategy"
+
+      strategy = personalized_placeholder_strategy
+      unless strategy.nil?
+        strategy.populate_placeholders(task: self)
+      end
+
+      update_step_counts!
```
jpslav: [There is at least one other place where we're added task steps](https://github.com/openstax/tutor-server/blob/4b2cd9c1de7fe3c4f2fd9ab6ebccb17785d9b3aa/app/subsystems/tasks/recover_task_step.rb#L30-L39) and I'm concerned it'll be easy to forget to update the step counts everywhere (since we don't have a `HeyIJustAddedATaskStep` routine to parallel `MarkTaskStepCompleted`).

While I'm normally not a fan, I could be a fan of a `before_save :update_step_counts!` (we'd want to be careful that `update_step_counts!` returns something callback friendly, i.e. not falsy).

This doesn't entirely save us because it isn't required for the `Task` to be saved when a new step is added (a new step can be added by just creating a new step with the task foreign key).  Maybe there's some Rails-way to do this (@Dantemss?) or maybe we can just introduce a routine or other method that must be called by convention whenever steps are updated.

app/subsystems/tasks/models/task_step.rb
```diff
@@ -18,21 +19,7 @@ class Tasks::Models::TaskStep < Tutor::SubSystems::BaseModel
...
-  end
-
-  def init_labels
-    self.settings['labels'] ||= []
-  end
-
-  def init_related_content
-    self.settings['related_content'] ||= []
-  end
+  after_save { task.update_step_counts! }
```
jpslav: Ah :-)  If you immediately know the candlelight is fire the meal was cooked a long time ago

app/subsystems/tasks/models/task_step.rb
```diff
@@ -4,7 +4,8 @@ class Tasks::Models::TaskStep < Tutor::SubSystems::BaseModel
 
   enum group_type: [:default_group, :core_group, :spaced_practice_group, :personalized_group]
 
-  serialize :settings, JSON
+  serialize :related_content, Array
+  serialize :labels, Array
```
jpslav: nice split

app/subsystems/tasks/models/task_step.rb
jpslav: Probably also need an `after_destroy`? (e.g. placeholder steps that are destroyed)

jpslav: This is maybe where a more fine-tuned update approach could happen:

```ruby
after_create {
  task.steps_count += 1
  task.core_steps_count += 1 if core_group?
  ... etc ...
}

after_save {
  task.completed_steps_count += 1 if completed?
  ... etc ...
}

after_destroy {
  task.steps_count -= 1
  ... etc ...
}
```

`reset_counts!` type calls in `Task` would still be good, of course.  Not saying things have to be this way, just brainstorming.

jpslav:
Really like the PR

app/subsystems/tasks/models/task.rb
kevinburleigh75: If we knew that all decisions will always be made on a single step in isolation, collapsing could work.  But we should also consider that testing would become more complex and that the option to update individual counts would be lost.

app/subsystems/tasks/models/task.rb
kevinburleigh75: `calculate_` makes me think no side effects; `update_` means update, but is maybe too unclear.  `reset_` might be a good...

app/subsystems/tasks/models/task_step.rb
kevinburleigh75: I didn't do an `after_destroy` because we always need to replace the destroyed `TaskStep` with a new one (which will trigger the `after_save`).  But if that changes we will need to protect ourselves..

This is also an argument in favor of controlled access to components through an aggregate root.

app/subsystems/tasks/models/task.rb
kevinburleigh75: It makes sense to me that they should eventually be replaces with the more canonically-named `_count` variants.  But since (I assume) they are used by other code, I didn't eliminate them at this time.

app/subsystems/tasks/models/task.rb
jpslav: yep

app/subsystems/tasks/models/task_step.rb
jpslav: aggregate root needs a less scary name

PR #425, opened 2015-07-08
Import demo content from yaml files

nathanstitt:
Main changes:
 * `entries` on the assignments has been renamed to `scores`.  I'm loading the data into a Hashie::Mash, and `entries` is an existing method that lists all the key/value pairs.
 * In order to make the YAML a bit more brief and allow one-liner arrays of `scores`, the named values are renamed.  `incomplete` => `i` and `not_started` => `ns`


jpslav:
Can the number of periods, and list of students in each period be in the YAML?

nathanstitt:
I'm having trouble with the task plan for physics "Read 3.2 Acceleration Pt2". 

The tasks that are generated from it contain a different number of steps.  Sometimes the tasks have 13 steps, and sometimes 14.  This causes an exception in demo_base's `work_tasks`  I've added debugging info to verify what's going on, but not sure how that could occur.  

Would it be safe to change the exception to fire only if the available responses are less than the step count?

nathanstitt:
This changes how students are created.  There's a students list containing the student initials and names and the course yaml files are set to read the students using their initials.  This allows students to be shared between the courses so the same student can be assigned to multiple courses and periods.

However this has the side effect of forcing a change to usernames.  Previously 10 students where created and the numbers were appending to make the username be studentXX.  Now it's not as deterministic so usernames are setup using the student's initials and they are named "student-hp" for "Harry Potter".

If the integer based usernames are important, we can switch it to use an internal counter to re-create that scheme.  However there'd be no guarantee that student01 would be the same if the course yaml was edited

nathanstitt:
This now works, but something really weird is occurring with the task plans for physics.  If the order of assignments is changed in `physics.yml` so they're in the "natural" order of:

 * Read 3.1 Acceleration Pt1
 * Read 3.2 Acceleration Pt2
 * Read 4.1-4.2 Force & Motion Pt1

The step counts change for the exercises.  "Pt1" changes from 12 steps to 10.  "Pt2" suddenly need 14 steps, even though it has only 13 now.   Then it suddenly changes back to 13 on it's 10th task.   This causes an exception in demo_base's `work_tasks` method.

Dantemss:
Maybe spaced practice etc? The only weird thing is that the total should be the same...

jpslav:
I'm fine with non-numeric student usernames (as long as they themselves are unique, no Harry Potters and Henry Penchantforbuggery)

kevinburleigh75:
Certainly re-ordering the assignments could affect the number of exercises due to spaced practice.  And currently I'd expect it to change by only +/-1 exercise.  The step printout for the assignment should show there the last *exercise* appearing/disappearing when it moves (the last step is a personalized placeholder, and should not be affected).

nathanstitt:
Yes, there were a few students who's initials did clash, so I had to change the the initials to be slightly inaccurate, i.e. `lb` is Leopold Bloom, but Lily Bart is `lh`.  

The initials are just intended to be something that's a bit easier to remember than an ID number but are still uniquely identify a student across assignments.

lib/tasks/demo/biology.yml
```diff
@@ -0,0 +1,104 @@
...
+    students: [an,cg,cd,dp,fa,jr,js,lh,ms,rb,wh]
+
+assignments:
+
+  - type: reading
+    step_types: [ r, r, e, r, r, e, r, r, e, r, r, e, r, e, r, p, e, e, r, r, e ]
+    chapter_sections: [[1, 1, 0], [1, 1, 1], [1,1,2]]
+    title: Read Unit 1. The Chemistry of Life
+    periods:
+      - index: 0 # index of the title above
```
kevinburleigh75: In a future PR, we might want to add 'aliases' for periods (like we have for students) and use them here instead of the index.

What's the "title above" anyways?

lib/tasks/demo/biology.yml
```diff
@@ -0,0 +1,104 @@
...
+  - name: 3rd Period
+    students: [ag,ap,bs,eh,hc,hg,iu,iw,nz,rl,sk]
+  - name: 4th Period
+    students: [an,cg,cd,dp,fa,jr,js,lh,ms,rb,wh]
+
+assignments:
+
+  - type: reading
+    step_types: [ r, r, e, r, r, e, r, r, e, r, r, e, r, e, r, p, e, e, r, r, e ]
+    chapter_sections: [[1, 1, 0], [1, 1, 1], [1,1,2]]
```
Dantemss: The numbers will have to change once the "no units" PR is merged.
This assignment is also really "Read Chapter 1. The Study of Life", not unit 1.
Check out https://github.com/openstax/tutor-server/pull/435/files

lib/tasks/demo/biology.yml
```diff
@@ -0,0 +1,104 @@
...
+        students: {ag: 100,ap: 82,bs: 71,eh: 90,hc: 78,hg: 87,iu: i,iw: i,nz: 90,rl: 85,sk: ns}
+
+      - index: 3 # index of the title above
+        due_at: <%= due_today %>
+        students: {an: 100,cg: 82,cd: 71,dp: 90,fa: 78,jr: 87,js: i,lh: i,ms: 90,rb: 85,wh: ns}
+
+  - type: reading
+    step_types: [ r, r, r, r, r, r, r, e, r, r, r, p ]
+    chapter_sections: [[2, 3, 0], [2, 3, 1h], [2, 3, 2]]
+    title: Read Unit 2. The Cell
```
Dantemss: This assignment is really "Read Chapter 6. Metabolism" (based on the current numbers)

jpslav:
Lovin this PR @nathanstitt

lib/tasks/demo/biology.yml
```diff
@@ -0,0 +1,104 @@
...
+          js: [ 1, 1, 0, 1, 1, 1, 1, 0, 1, 1, 1, 0 ]
+          lh: i
+          ms: i
+          rb: [ 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1 ]
+          wh: [ 1, 0, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1 ]
+
+  - type: homework
+    step_types: [ e, e, e, e, e, e, e, e, e, e, p ]
+    chapter_sections: [[1, 1, 0], [1, 1, 1], [1,1,2]]
+    title: HW The Chemistry of Life
```
Dantemss: Same thing here, this assignment is all chapter 1, not the other chapters in unit 1 (study of life instead of chemistry of life)

lib/tasks/demo/biology.yml
nathanstitt: The `index` refers to the index of the period above, so it kind of is an alias.  I failed to update the comment, it was referring to when the periods above was just an array of titles.

nathanstitt:
Hm, I think we may have an issue with Travis.  It's failing because a spec run now takes longer than 10 minutes with two courses and biology having 4 periods.  The Travis error message says "No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself."  Maybe we can extend the time if we turn on logging for the demo script that way it'll periodically output something?

PR #427, opened 2015-07-08
iReading and HW personalized strategies now have different logic for physics and bio

Dantemss:
Ready for review.


jpslav:
Looks good to me.  Might merge after demo unless particularly needed before it.

PR #428, opened 2015-07-08
Course period learning guide

joemsak:


app/controllers/api/v1/guides_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+        short_description 'Represents course guides in the system'
+        description <<-EOS
+          Course guide description to be written...
+        EOS
+      end
+
+      api :GET, '/courses/:id/guide(/role/:role_id)',
+                'Returns a student course guide for Learning Guide'
+      description <<-EOS
+        #{json_schema(Api::V1::CourseGuideRepresenter, include: :readable)}
```
jpslav: Change this to match real representer name

app/controllers/api/v1/guides_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      def student
+        course = Entity::Course.find(params[:id])
+        guide = GetCourseGuide[role: role(course, :student), course: course]
+        respond_with guide, represent_with: Api::V1::CourseGuidePeriodRepresenter
+      end
+
+      api :GET, '/courses/:id/teacher_guide',
+                'Returns course guide for Learning Guide for teachers'
+      description <<-EOS
+        #{json_schema(Api::V1::CourseGuideRepresenter, include: :readable)}
```
jpslav: Change this to match real representer name

app/controllers/api/v1/guides_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      EOS
+      def teacher
+        course = Entity::Course.find(params[:id])
+        guide = GetCourseGuide[role: role(course, :teacher), course: course]
+        respond_with guide, represent_with: Api::V1::CourseGuideRepresenter
```
jpslav: Can this be `TeacherCourseGuideRepresenter`?

app/representers/api/v1/course_guide_child_representer.rb
```diff
@@ -0,0 +1,51 @@
...
+             type: Integer,
+             readable: true,
+             writeable: false,
+             schema_info: { required: true }
+
+    collection :page_ids,
+               readable: true,
+               writeable: false,
+               getter: -> (*) { page_ids && page_ids.map(&:to_s) },
+               schema_info: { items: { type: 'string' } }
```
jpslav: I don't think this schema_info value is used in any way (did you see this style somewhere else?).

app/representers/api/v1/teacher_course_guide_representer.rb
```diff
@@ -0,0 +1,26 @@
+module Api::V1
+  class CourseGuidePeriodRepresenter < Roar::Decorator
```
jpslav: Can you make this its own file?

app/representers/api/v1/course_guide_child_representer.rb
jpslav: if you want to say they are strings, you can use the `description` field which is used in the API docs, e.g. https://goo.gl/8QSTsO

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,161 @@
...
+        page_ids: pages.collect { |pp| pp[:id] },
+        children: pages
+      }
+    }.sort_by { |ch| ch[:chapter_section] }
+  end
+
+  def filter_task_steps_by_period_roles(task_steps)
+    role_ids = outputs.roles.flatten.collect(&:id)
+
+    task_steps.select do |task_step|
```
jpslav: Simplify to

```ruby
task_steps.select do |task_step|
  task_step.task.taskings.any?(entity_role_id: role_ids)
end
```

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,161 @@
...
+
+  def completed_practices(task_steps, task_type)
+    task_ids = task_steps.collect(&:tasks_task_id).uniq
+    tasks = Tasks::Models::Task.where(id: task_ids, task_type: task_type)
+    tasks.select(&:completed?)
+  end
+
+  def get_current_level(task_steps)
+    lo_tags = get_lo_tags(task_steps)
+    OpenStax::Biglearn::V1.get_clue(roles: role, tags: lo_tags)
```
jpslav: `role` here is the role of the requestor.  When getting the CLUE, we want to pass in the roles of the students in question, which for the student guide is the one student role and for the teacher guide is all of the roles in a period.  

So I recommend renaming the main `role` variable to `requestor_role` (or something more specific like that) and then make sure to pass in the right set of roles here.

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,161 @@
...
+    translations: { outputs: { type: :verbatim } },
+    as: :is_teacher
+
+  protected
+  def exec(role:, course:)
+    @role, @course = role, course
+
+    run(:get_course_books, course: course)
+    run(:get_periods, course: course, roles: role)
+    run(:visit_book, book: outputs.books.first, visitor_names: [:toc, :page_data])
```
jpslav: Man we need to start caching the toc and page_data.  @Dantemss we might soon want to dust off your ancient PR on this.

app/routines/get_course_guide.rb
jpslav: I have a lingering concern that this is too many queries to do to filter (especially when we're talking a 200 students having worked 30 assignments with 20 exercises each) by period when we could just be calling get_role_task_steps once for each period passing in each period's roles.

BUT.... I would also like to flip this whole routine on its head pretty soon and start using the cached step counts that @joemsak started and @kevinburleigh75 finished.  Which might mean we can get away from being TaskStep-centric here and save ourselves a ton of queries.

jpslav: maybe we can't flip to Task-centric after all because each Task has exercises from different pages....  hmmm....

jpslav:
Please add in some practice tasks so we can test for non-zero counts

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,164 @@
...
+        children: pages
+      }
+    }.sort_by { |ch| ch[:chapter_section] }
+  end
+
+  def filter_task_steps_by_period_roles(task_steps)
+    role_ids = outputs.roles.flatten.collect(&:id)
+
+    task_steps.select do |task_step|
+      task_step.task.taskings.any? { |t| role_ids.include?(t.entity_role_id) }
```
jpslav: Sorry should have been:

```ruby
task_step.task.taskings.where(entity_role_id: role_ids).any?
```

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,164 @@
...
+  def book_parts_by_id
+    @book_parts_by_id ||= Hash[book_parts.map { |part| [part.id, part] }]
+  end
+
+  def book_parts
+    [outputs.toc, outputs.toc.children].flatten
+  end
+
+  def is_teacher?
+    run(:is_teacher, course: course, roles: role).outputs.is_course_teacher
```
jpslav: memoize

app/routines/get_course_guide.rb
```diff
@@ -0,0 +1,164 @@
...
+    role_ids = outputs.roles.flatten.collect(&:id)
+
+    task_steps.select do |task_step|
+      task_step.task.taskings.any? { |t| role_ids.include?(t.entity_role_id) }
+    end
+  end
+
+  def exercises_grouped_by_book_part
+    outputs.task_steps.flatten.select { |t| t.exercise? && t.completed? }.group_by do |t|
+      Content::Routines::SearchPages[tag: get_lo_tags(t)].first.content_book_part_id
```
jpslav: @kevinburleigh75 I'm pretty sure I heard you pontificating in the area of linking things directly to the pages from which they sprang instead of navigating to them through their LOs?  Do you agree that it'd be a great thing (performance- and simplicity-wise) to record the page (and chapter) that each step is sourced from?

Would help a lot in quick look analytics too.  https://goo.gl/njkmle

jpslav:
@philschatz & @darrenmason -- are you ok with us merging this learning guide PR (the period-aware one)?  A few spec changes are still needed but otherwise it is good to go and I want the team to start working on some optimizations on it.

app/representers/api/v1/course_guide_period_representer.rb
```diff
@@ -0,0 +1,24 @@
+module Api::V1
+  class CourseGuidePeriodRepresenter < Roar::Decorator
```
jpslav: isn't there supposed to be a period_id here (or somewhere)?

joemsak: thought for student guide they don't need it, so it's in the teacher guide

PR #435, opened 2015-07-10
Completely removed units from the bio book during import

Dantemss:
Including the reference view etc. You can debate whether or not this is the right thing to do.

Also fixes chapter numbering in the presence of units.

### DIFF
--- a/app/external/openstax/cnx/v1/book_part.rb
+++ b/app/external/openstax/cnx/v1/book_part.rb
@@ -1,15 +1,18 @@
 module OpenStax::Cnx::V1
   class BookPart
 
-    def initialize(hash: {}, chapter_section: [], title: nil, contents: nil, parts: nil)
-      @hash            = hash
-      @chapter_section = chapter_section
-      @title           = title
-      @contents        = contents
-      @parts           = parts
+    def initialize(hash: {}, chapter_section: [], title: nil,
+                   contents: nil, parts: nil, initial_child_book_part_index: 1)
+      @hash                       = hash
+      @chapter_section            = chapter_section
+      @title                      = title
+      @contents                   = contents
+      @parts                      = parts
+      @next_child_book_part_index = initial_child_book_part_index
     end
 
-    attr_reader :hash, :chapter_section
+    attr_reader :hash
+    attr_accessor :chapter_section
 
     def title
       @title ||= hash.fetch('title') { |key|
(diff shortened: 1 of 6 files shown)


jpslav:
Before I review, @philschatz @kathi-fletcher is this what we want?

philschatz:
:+1: That's my understanding

app/external/openstax/cnx/v1/book_part.rb
```diff
@@ -1,15 +1,18 @@
...
-      @contents        = contents
-      @parts           = parts
+    def initialize(hash: {}, chapter_section: [], title: nil,
+                   contents: nil, parts: nil, initial_child_book_part_index: 0)
+      @hash                       = hash
+      @chapter_section            = chapter_section
+      @title                      = title
+      @contents                   = contents
+      @parts                      = parts
+      @last_child_book_part_index = initial_child_book_part_index
```
jpslav: The names `last_...`, `initial_...`, and `next_child_book_part_index` are confusing me.  The "initial" seems like the first one that should be used, but on line 47 below we add one to it.  Also on line 61 "next" is the "last".  Can you clarify / simplify the naming?

app/subsystems/content/routines/import_book_part.rb
```diff
@@ -11,21 +11,25 @@ class Content::Routines::ImportBookPart
...
+      )
+
+      parent_book_part.child_book_parts << book_part unless parent_book_part.nil?
+    end
 
     cnx_book_part.parts.each do |part|
       if part.is_a?(OpenStax::Cnx::V1::BookPart)
-        child_book_part = run(:import_book_part, cnx_book_part: part,
-                              parent_book_part: book_part).outputs.book_part
-        book_part.child_book_parts << child_book_part
```
jpslav: any idea why were doing this (and old line 28) before?

Dantemss: It was working fine before, but with the changes above, this made each child be included twice (and error out etc) since the import_book_part already adds the child to the parent.

app/external/openstax/cnx/v1/book_part.rb
Dantemss: No way, I barely understand it myself...
This is what makes chapters be numbered independently of units.
I could rename `@last_child_book_part_index` to `@current_child_book_part_index` but idk if that's more clear or not.

Dantemss:
Closing due to https://github.com/openstax/tutor-server/pull/439

PR #436, opened 2015-07-10
External assignments student deidentifier and templatized URL

karenc:
- Add auto generated unique deidentifier field to course students

- Allow templatized url in external assignment urls

  At the moment, only `deidentifier` will work in templatized url.

  The syntax looks like this:

  `https://www.example.org/survey?id={{deidentifier}}`

  For a student with deidentifier "1433f416", they will get this url:

  `https://www.example.org/survey?id=1433f416`

- Add admin view to show students for a course

  Shows username, name and deidentifier

- Add deidentifier to student representer

### DIFF
--- a/app/controllers/admin/students_controller.rb
+++ b/app/controllers/admin/students_controller.rb
@@ -1,0 +1,15 @@
+class Admin::StudentsController < Admin::BaseController
+  before_action :get_course, only: [:index]
+
+  def index
+    @students = GetStudentRoster[course: @entity_course]
+    @students.sort! { |a, b| a.username <=> b.username }
+  end
+
+  protected
+
+  def get_course
+    @entity_course = Entity::Course.find(params[:course_id])
+    @course = GetCourseProfile[course: @entity_course]
+  end
+end
--- a/app/subsystems/tasks/assistants/external_assignment_assistant.rb
+++ b/app/subsystems/tasks/assistants/external_assignment_assistant.rb
@@ -16,17 +16,24 @@ class Tasks::Assistants::ExternalAssignmentAssistant
   end
 
   def self.build_tasks(task_plan:, taskees:)
-    url = task_plan.settings['external_url']
-    taskees.collect do |taskee|
-      build_external_task(task_plan: task_plan, taskee: taskee, url: url)
+    students = taskees.collect { |taskee| taskee.students.first }.compact
+
+    raise StandardError, 'External assignment taskees must all be students'\
+      if students.length != taskees.length
+
+    taskees.collect.with_index do |taskee, i|
+      build_external_task(task_plan: task_plan,
+                          taskee: taskee,
+                          student: students[i])
     end
   end
 
   protected
-  def self.build_external_task(task_plan:, taskee:, url:)
+  def self.build_external_task(task_plan:, taskee:, student:)
     task = build_task(task_plan: task_plan)
     step = Tasks::Models::TaskStep.new(task: task)
-    Tasks::Models::TaskedExternalUrl.new(task_step: step, url: url)
+    tasked_external_url(task_step: step, taskee: taskee, student: student,
+                        url: task_plan.settings['external_url'])
     task.task_steps << step
     task
   end
@@ -41,5 +48,14 @@ class Tasks::Assistants::ExternalAssignmentAssistant
       title: title,
       description: description
     ]
+  end
+
+  def self.tasked_external_url(task_step:, taskee:, student:, url:)
+    {
+      deidentifier: student.try(:deidentifier)
+    }.each do |key, value|
+      url = url.gsub("{{#{key}}}", value)
+    end
+    Tasks::Models::TaskedExternalUrl.new(task_step: task_step, url: url)
   end
 end
--- a/app/models/entity/role.rb
+++ b/app/models/entity/role.rb
@@ -7,5 +7,5 @@ class Entity::Role < Tutor::SubSystems::BaseModel
   has_one :role_user, dependent: :destroy, subsystem: :role
   has_one :user, through: :role_user
 
-  delegate :username, :first_name, :last_name, :full_name, to: :user
+  delegate :username, :first_name, :last_name, :full_name, :name, to: :user
 end
--- a/app/models/entity/user.rb
+++ b/app/models/entity/user.rb
@@ -4,5 +4,5 @@ class Entity::User < Tutor::SubSystems::BaseModel
   has_many :role_user, subsystem: :role
   has_many :roles, through: :role_user
 
-  delegate :username, :first_name, :last_name, :full_name, to: :profile
+  delegate :username, :first_name, :last_name, :full_name, :name, to: :profile
 end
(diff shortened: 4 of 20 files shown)


app/subsystems/user_profile/models/profile.rb
```diff
@@ -56,4 +58,11 @@ def undelete
   alias_method :user_id, :id
   def user; self; end
 
+  protected
+
+  def generate_research_id
+    self.research_id = SecureRandom.hex(4) if self.research_id.blank?
```
jpslav: Because I'm neurotic, can you add a check to make sure the research_id is not yet used?

Like in https://github.com/lml/ost/blob/master/app/models/user.rb#L185-L190 but keeping 8 hex chars like you have now.  

Also I believe this line can be 

```ruby
self.research_id ||= SecureRandom.hex(4)
```

which only applies if this code doesn't completely change based on my first comment

db/migrate/20140724184630_create_user_profile_profiles.rb
```diff
@@ -13,6 +14,7 @@ def change
       t.index :deleted_at
       t.index :exchange_read_identifier, unique: true
       t.index :exchange_write_identifier, unique: true
+      t.index :research_id, unique: true
```
karenc: Isn't this enough for checking uniqueness?

jpslav: yes but it'll blow up if there's a conflict. slightly better to avoid blowing up in the model code and then just let this DB constraint catch a race condition.

this is all impacted by whether or not we decide to use the last 8 chars of the read only exchange ID

app/subsystems/user_profile/models/profile.rb
```diff
@@ -56,4 +58,13 @@ def undelete
...
   def user; self; end
 
+  protected
+
+  def generate_research_id
+    begin
+      research_id = SecureRandom.hex(4)
+    end while UserProfile::Models::Profile.exists?(research_id: research_id)
+    self.research_id ||= research_id
+  end
```
karenc: Updated `generate_research_id` to check existing records :)

karenc:
Ready for review.

app/controllers/admin/courses_controller.rb
```diff
@@ -30,6 +30,23 @@ def update
                 })
   end
 
+  def show_students
+    course = Entity::Course.find(params[:id])
+    @course = GetCourseProfile[course: course]
+    students = CourseMembership::Models::Student.joins(:course).where {
```
jpslav: Instead of all this code, just call `GetStudentRoster` and add the other fields you need there (username, deidentifier)

app/subsystems/tasks/assistants/external_assignment_assistant.rb
```diff
@@ -16,17 +16,40 @@ def self.schema
   end
 
   def self.build_tasks(task_plan:, taskees:)
+    raise NotImplementedError unless task_plan.owner.is_a?(Entity::Course)
```
jpslav: hmm... why have this?

app/subsystems/tasks/assistants/external_assignment_assistant.rb
```diff
@@ -42,4 +65,13 @@ def self.build_task(task_plan:)
       description: description
     ]
   end
+
+  def self.tasked_external_url(task_step:, taskee:, student:, url:)
+    {
+      deidentifier: student.try(:deidentifier)
+    }.each do |key, value|
```
jpslav: pretty

app/controllers/admin/courses_controller.rb
```diff
@@ -30,6 +30,23 @@ def update
                 })
   end
 
+  def show_students
```
jpslav: Can this be an index method in `admin/students_controller.rb`?  Just to separating some things out of this controller before it gets too big.

app/views/admin/courses/show_students.html.erb
```diff
@@ -0,0 +1,20 @@
+<h2>Students for course "<%= @course.name %>"</h2>
+
+<table class="table table-striped">
+  <thead>
+    <tr>
+      <th>Username</th>
+      <th>Name</th>
+      <th>deidentifier</th>
```
jpslav: Capitalize

app/subsystems/tasks/assistants/external_assignment_assistant.rb
```diff
@@ -16,17 +16,40 @@ def self.schema
...
 
   def self.build_tasks(task_plan:, taskees:)
+    raise NotImplementedError unless task_plan.owner.is_a?(Entity::Course)
+
     url = task_plan.settings['external_url']
+
+    # taskees may be default roles and not be linked to
+    # CourseMembership::Models::Student, need to get the user, find all the
+    # roles for the user and get all the CourseMembership::Models::Student for
+    # each taskee
```
jpslav: Uhoh.  This may be a case of me saying "do it this way" but not realizing the acrobatics that are required to get there -- and acrobatics are normally a bad sign ... let me think about this...

jpslav: Ok, I'm in a better place now.  :-)

While it is in general true that taskees derived from a task plan may be any kind of role, not just students, it is ok for us to declare that the external assignments are just for students.  Meaning if the incoming roles are not associated to students we can justifiably freak out.

I think this method could look like the following:

```ruby
def self.build_tasks(task_plan:, taskees:)
  # Incidentally I think role should has_one :student instead of has_many, but we don't 
  # have to solve that right now.
  students = taskees.collect{ |taskee| taskee.students.try(:first) }

  raise StandardError, "External assignment taskees must all be students" if students.size != taskees.size

  url = task_plan.settings['external_url']

  taskees.collect.with_index do |taskee, ii|
    build_external_task(task_plan: task_plan,
                        taskee: taskee,
                        student: student[ii],
                        url: url) # don't really need to pass this in since task_plan is passed
  end
end
```

Thanks tho for hearing what I said and executing it; sorry I was speaking too generally and you had to go down a circuitous path.

jpslav:
Review complete.

app/controllers/admin/courses_controller.rb
jpslav: And it'd be **super** great if, after you do this, you also add a deidentifier property into the `Api::V1::StudentsRepresenter`.  That representer is fed by the data from `GetStudentRoster`.

app/subsystems/tasks/assistants/external_assignment_assistant.rb
karenc: Is `task_plan.owner` always a course?

jpslav: See comment below, should make this line not needed.

app/controllers/admin/students_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Admin::StudentsController < Admin::BaseController
+  before_action :get_course, only: [:index]
+
+  def index
+    @students = GetStudentRoster[course: @entity_course]
```
jpslav: beautiful :-)

PR #444, opened 2015-07-12
README updates, NewRelic example, Bullet configurable

jpslav:
@nathanstitt can you review to make sure I'm not lying in the README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,20 +2,6 @@
 [![Build Status](https://travis-ci.org/openstax/tutor-server.png?branch=master)](https://travis-ci.org/openstax/tutor-server)
 [![Coverage Status](https://img.shields.io/coveralls/openstax/tutor-server.svg)](https://coveralls.io/r/openstax/tutor-server)
 
-# OpenStax testing logins
-
-The demo tasks populate the server with a few students:
-
-  * student-ak
-  * student-hp
-  * student-hg
-  * student-cd
-  * student-nz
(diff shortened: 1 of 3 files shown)


nathanstitt:
+1 I think that reads well and has a bunch of additional details that'll be useful.  I like the mention of the `USE_REAL_BACKGROUND_JOBS` in particular, I'll have to play with that.

A tiny quibble might be that it's not really necessary to run the entire `:drop, :create, :migrate, :seed` sequence.  

`rake db:reset demo` is equivalent and a bit shorter.
http://edgeguides.rubyonrails.org/active_record_migrations.html#resetting-the-database

lakshmivyas:
For deployment, we use db:schema:load.  Perhaps we should use the same thing here.
https://github.com/openstax/tutor-deployment/blob/master/tasks/add_demo_data.yml#L12

BTW, are there any changes required to deployment because of these recent changes?

jpslav:
@lakshmivyas yeah I think the tutor demo task should be dropping the database, which db:schema:load is not doing (wonder how things have been working up through this point?).

jpslav:
thanks @nathanstitt for pointing that out -- updated the PR

lakshmivyas:
@jpslav - I can't find the link (and doing too many things to search properly) - but `db:schema:load` turned out to be ideal for our setup as the rails user does not have (and should not have) privileges to drop a database.

PR #240, opened 2015-05-01
WIP Use visitors in import; Cache ToC

Dantemss:
# Done
- Some visitor cleanup

# TODO
- Cache book ToC
- Routine to get all pages in a content_book
- Routine to get all exercises in a content_book
- Use visitors in book import

PR #445, opened 2015-07-12
Cache book visitors

jpslav:
Adds caching to book visitors, so we only have to get the `:toc` or `:page_data` once (which is good since books don't change (yet)).

Incidentally simplified `Content::GetLos` to use direct relations instead of book visitation.

### DIFF
--- a/app/subsystems/content/get_los.rb
+++ b/app/subsystems/content/get_los.rb
@@ -7,12 +7,9 @@ class Content::GetLos
   def exec(options = {})
     page_ids = [options[:page_ids]].flatten.compact
     book_part_ids = [options[:book_part_ids]].flatten.compact
-    book_part_ids.each do |book_part_id|
-      book_part = Content::Models::BookPart.find(book_part_id)
-      page_ids += Content::VisitBookPart[book_part: book_part,
-                                         visitor_names: 'page_data']
-                    .collect{|info| info[:id]}
-    end
+
+    page_ids += Content::Models::Page.where(content_book_part_id: book_part_ids)
+                                     .pluck(:id)
 
     outputs[:los] = Content::Models::Tag.lo
                                         .joins(:page_tags)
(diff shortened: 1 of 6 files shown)


jpslav:
@Dantemss can you review?

jpslav:
Someone should also run the `toc` and `page_data` visitors during book import (since we're already doing a visit there for exercises)

Dantemss:
Maybe we can have a list of visitors that the importer runs or something...

Dantemss:
Do you want the stuff in https://github.com/openstax/tutor-server/pull/240 or should I close? Mainly it makes all BookParts have an entity_book_id pointing to the root so that visiting (can) be less expensive.

jpslav:
@Dantemss is it that #240 adds `null: false` to the DB schema?  That's the difference I see w.r.t. `entity_book_id`.

Dantemss:
The one other difference would be that the book is always passed to all book_part children in https://github.com/Dantemss/tutor-server/blob/7a4bb5dcebe45d1f6ccada79d2c86ff442122c28/app/subsystems/content/routines/import_book_part.rb

Dantemss:
That would allow visitors to assume the book tree is flat if they don't care about the tree structure (you can get all parts in the book with BookPart.where(book: book)). To get the root BookPart, you'd also have to check that parent_book_part is nil.

jpslav:
Ok, so we already do this (set the `book` on all BookParts), but this change just makes it mandatory? (can’t be nil)

Dantemss:
Currently only the root gets the Book: https://github.com/openstax/tutor-server/blob/master/app/subsystems/content/routines/import_book_part.rb

jpslav:
Does this cover the functionality we want to preserve (for now) from #240?

https://github.com/openstax/tutor-server/pull/456 <https://github.com/openstax/tutor-server/pull/456>

Dantemss:
Looks good to me.

PR #451, opened 2015-07-12
bootstrapped the admin views

jpslav:
was getting tired of no styling :-)


jpslav:
merging as there is no risk of breaking main code since just changed admin templates

jpslav:
image

PR #456, opened 2015-07-13
requires book to be set in book part

jpslav:


jpslav:
@Dantemss please review -- I think these are some of the changes you want to preserve from #240

Dantemss:
Seems good to me.

PR #457, opened 2015-07-14
Admin import books

karenc:
- Store archive url in `Content::Models::BookPart.url`

  Store archive url in `OpenStax::Cnx::V1::Book` and pass it to
  `Content::ImportBook` in order to store it in
  `Content::Models::BookPart.url`.
- Add uuid and version fields to content_book_parts table
- Store cnx uuid and version in root `Content::Models::BookPart`

  Store cnx uuid and version in `OpenStax::Cnx::V1::Book`, then pass it to
  `Content::Models::BookPart` in `Content::ImportBook`.
- Add routine Content::ListBooks that returns all book info
- Add admin books index page which lists all the books in the database
- Add test for books controller #index
- Make content pages unique on "content_book_part_id, url"

  There was a uniqueness constraint on "url" and it became a problem when
    a page is created from another book.  It has the same cnx url but has
    different chapter sections, content book part etc.
    
    There was talk of creating a join table like book part pages but this is
    how we're going to implement this for now.
- Add admin page to import a cnx book
- Add test for admin books controller post #import
- Add remove other books option to AddBookToCourse
- Add form to select a course book on admin edit course page
- Add tests for admin courses controller post #set_book
- Add test for admin courses controller #edit


karenc:
Ready for review.

app/views/admin/books/import.html.erb
```diff
@@ -0,0 +1,13 @@
+<% @page_header = 'Import a Book' %>
+
+<%= form_tag '', method: :post do %>
+  <div class="form-group">
+    <%= label_tag :archive_url %>
+    <%= text_field_tag :archive_url, params[:archive_url], class: 'form-control', placeholder: @default_archive_url %>
```
jpslav: Make the default archive URL the value of the text field instead of a placeholder so people don't have to type it in all the time.

jpslav:
On the books index page, put the UUID and version in the table

app/views/admin/books/import.html.erb
karenc: People don't.  That's what we use if people don't type anything in.

I made it a placeholder so that people can always revert to using the archive URL.

jpslav:
Disable the import button after it has been pressed.  Add a message near it that says import will likely take a while.

jpslav:
When I tried to import a book that was already imported, I got a quick success message but there was a ROLLBACK in the logs.

jpslav:
Should be allowed to import different versions of the same book, but not a version that already exists

app/controllers/admin/books_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def import
+    @default_archive_url = OpenStax::Cnx::V1.archive_url_base
+    import_book if request.post?
+  end
+
+  protected
+  def import_book
+    archive_url = params[:archive_url].present? ? params[:archive_url] : @default_archive_url
```
jpslav: Along with the change below relating to the placeholder for the archive URL, I'd be fine just requiring it in the form (no default needed here).

app/views/admin/books/import.html.erb
karenc: Should I put something like "<default-cnx-url> is used if this is left empty" in the placeholder?

jpslav:
For convenience, on the book index page put the link to the book on CNX (could be the archive URL).  This likely necessitates storing the archive URL somewhere (like Content::Models::BookPart, and only being on the root book part ok)

app/views/admin/courses/_select_book.html.erb
```diff
@@ -0,0 +1,10 @@
+<% method ||= :post %>
+
+<%= form_tag(url, method: method, class: 'form-inline') do |f| %>
+  <div class="form-group">
+    <%= label_tag :book_id %>
+    <%= select_tag :book_id, options_from_collection_for_select(@books, :id, :title, params[:book_id] || @selected_book), include_blank: true, class: 'form-control' %>
```
jpslav: Would be nice to list out at least a portion of the UUID@version string so we don't get ourselves confused e.g.:

`AP Biology (ccbc51fa...@5.63)`

app/controllers/admin/courses_controller.rb
```diff
@@ -53,6 +55,21 @@ def students
     redirect_to edit_admin_course_path(params[:id])
   end
 
+  def select_book
+    course = Entity::Course.find(params[:id])
+    book = Entity::Book.find(params[:book_id])
+    if CourseContent::Models::CourseBook.where { entity_course_id == course.id }
```
jpslav: `if course.course_book.try(:book) == book`

jpslav: or you could add `has_one :book, through: :course_book` to `Entity::Course` and say `if course.book == book`

app/controllers/admin/courses_controller.rb
```diff
@@ -53,6 +55,21 @@ def students
...
+                                        .where { entity_book_id == book.id }.exists?
+      flash[:notice] = "Course book \"#{book.book_part.title}\" is already selected for \"#{course.profile.name}\""
+    else
+      # remove the existing course book
+      CourseContent::Models::CourseBook.where { entity_course_id == course.id }.delete_all
```
jpslav: I think we want this line down deeper in the business logic.  

Add...

```ruby
class CourseContent::SetBookOnCourse
  lev_routine

  protected

  def exec(course:, book:)
    # remove the existing course book then add the new one
    course.course_book.destroy_all
    course_book = CourseContent::Models::CourseBook.create(course: course, book: book)
    transfer_errors_from(course_book, {type: :verbatim}, true)
  end
end
```

And then just call that routine from here with `SetBookOnCourse[course: course, book: book]`

app/controllers/admin/courses_controller.rb
```diff
@@ -78,4 +95,14 @@ def create_user(user)
                                first_name: first_name, last_name: last_name,
                                full_name: full_name]
   end
+
+  def get_books
```
jpslav: This code is also up above in the books index action.  Create a `Content::ListBooks` routine that does this.  Per one of my other comments, the hashes should also probably contain the UUID / version.  Make sure to have `lev_routine express_output: :books` so that you can just say `@books = Content::ListBooks[]` instead of `@books = Content::ListBooks[].outputs.books`.

app/controllers/admin/courses_controller.rb
```diff
@@ -78,4 +95,14 @@ def create_user(user)
...
+    books = Content::Models::BookPart.where { parent_book_part_id == nil }.order(:title)
```
jpslav: This line will get moved into the routine I described above, but so that we don't spread the `parent_book_part_id == nil` logic around too much, add a scope to `BookPart`:

```ruby
scope :roots, -> { where(parent_book_part_id == nil) }
```

so can say `Content::Models::BookPart.roots`

app/models/entity/book.rb
```diff
@@ -1,3 +1,4 @@
 class Entity::Book < Tutor::SubSystems::BaseModel
+  has_one :book_part, subsystem: :content
```
jpslav: `root_book_part` for clarify

app/models/entity/course.rb
```diff
@@ -1,5 +1,6 @@
 class Entity::Course < Tutor::SubSystems::BaseModel
   has_one :profile, subsystem: :course_profile
+  has_one :course_book, subsystem: :course_content
```
jpslav: The code currently assumes only one course book, but really since it is a join table it could be many... so probably would be more correct to say `has_many :course_books`

app/views/admin/books/import.html.erb
jpslav: Ah -- I had to type it in every time because the default isn't right :-)  (see https://github.com/openstax/tutor-server/issues/480 which I just wrote).

jpslav: If you leave it a placeholder you don't have to have a <default-cnx-url> message

app/views/layouts/_flash_messages.html.erb
```diff
@@ -1,9 +1,10 @@
 <% flash.each do |name, msg| %>
+  <% background = { 'notice' => 'success', 'error' => 'danger' } %>
   <% if msg.kind_of?(Array) %>
     <% msg.each do |m| %>
-      <%= content_tag :div, m, class: "flash_#{name}" %>
+      <%= content_tag :div, m, class: "flash-message flash_#{name} bg-#{background[name]}" %>
     <% end %>
   <% else %>
-    <%= content_tag :div, msg, class: "flash_#{name}" %>
+    <%= content_tag :div, msg, class: "flash-message flash_#{name} bg-#{background[name]}" %>
```
jpslav: FWIW, @joemsak's #399 has [a helper](https://github.com/openstax/tutor-server/blob/admin-creates-districts/app/helpers/application_helper.rb#L2-L6) for dealing with bootstrap flash messages.

karenc: Ok, I'll just remove my changes.

karenc:
Ready for review again.

app/subsystems/content/routines/import_page.rb
```diff
@@ -21,13 +21,18 @@ class Content::Routines::ImportPage
...
+    # so only update the page if the page doesn't exist
+    outputs[:page] = Content::Models::Page.where(url: cnx_page.url).first
     run(:create_page, url: cnx_page.url,
                       title: cnx_page.title,
                       content: cnx_page.converted_content,
                       book_part: book_part,
                       chapter_section: chapter_section,
                       uuid: uuid,
-                      version: version)
+                      version: version) if outputs[:page].nil?
```
jpslav: @Dantemss and @kevinburleigh75 -- what do you think of this optimization @karenc added to not import a page that has already been imported?  My concern is that one page can be in two collections and have different `chapter_section`s, so we shouldn't do this... is that right?

jpslav:
Looks good @karenc -- just one question to resolve with Kevin and Dante

app/subsystems/content/routines/import_page.rb
karenc: Oh yeah... that's possibly a problem.  I added it because pages have to have unique `url`s.  (from acts_as_resource)  

Makes me wonder if we should remove `book_part` and `chapter_section` from `Content::Models::Page` and do something else.

jpslav: No I think those things belong in `Content::Models::Page`, as that is our model for a page in one of our books (as Kevin points out, we are very book-centric, so our models should reflect that mindset).  If anything, what you're trying to do here could be made the responsibility of the `OpenStax::Cnx` part of the code -- it could cache CNX content locally so as to be able to skip a network request for something that was already downloaded once.  

So I'd say remove the `if outputs[:page].nil?` optimization here.

karenc: I wasn't adding an optimization on purpose... I was trying to get around
the unique constraint on url.

karenc: Actually this is a problem in the reference book pages API as well...  We're looking up the page by uuid and version, and changing the links in the page if the links point to another page in the book... 

There could be several results for the same uuid and version and we don't know which book from the API.

jpslav: Excellent point @karenc.  Steps to take:

1. Remove the uniqueness constraint on the URL (as we are saying we can have multiple copies of the same module in different books).
2. Remove the `if outputs[:page].nil?` since this uniqueness constraint is gone.
3. Look into changing the reference view to pull by `Content::Models::Page` id instead of UUID and version (because we want one reference book's local links to stay in that particular book and not jump out).

@kevinburleigh75 @Dantemss y'all ok with this?

Dantemss: To me the url uniquely identifies a page, so I'd rather change how the page links to the book parts instead. (i.e. make a join table like BookPartPages).

karenc: We're going to change the unique constraint on `Content::Models::Page` from just `url` to `url, entity_book_id`.

karenc:
Removed 98bc55c1 and added 77134026

PR #462, opened 2015-07-14
Store performance book export round 3

Dantemss:
- Included Joe's PR
- Made ExportUploader properly hash files
- Changed fog-aws to fog (fog-aws require carrierwave 1.0)
- Added S3 template to example secrets file


Dantemss:
I *think* this is done but honestly it needs some testing in the dev deployment. And some UI.

PR #465, opened 2015-07-14
Remove periods

karenc:
Add delete button to periods on admin edit course page

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -91,7 +91,7 @@ Rails.application.routes.draw do
       member do
         post 'students'
       end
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -91,7 +91,7 @@
       member do
         post 'students'
       end
-      resources :periods, except: :destroy
+      resources :periods
```
jpslav: Can you make this `resources :periods, shallow: true`?

That way the edit, delete, show routes won't need the course ID (which isn't really used).

karenc:
Ready for review again.

PR #466, opened 2015-07-14
Required opens_at for TaskingPlans and Tasks

Dantemss:


app/subsystems/tasks/models/tasking_plan.rb
```diff
@@ -1,23 +1,29 @@
...
   belongs_to :task_plan, inverse_of: :tasking_plans
   belongs_to :target, polymorphic: true
 
   validates :target, presence: true
   validates :task_plan, presence: true, uniqueness: { scope: [:target_type, :target_id] }
 
-  validates :due_at, presence: true
-  validates :due_at, timeliness: { on_or_after: :opens_at }, if: :opens_at
+  validates :opens_at, presence: true, timeliness: { type: :date }
+  validates :due_at, presence: true, timeliness: { type: :date }
```
jpslav: This can have `timeliness: { on_or_after: :opens_at }` (since both required) and we can get rid of `due_at_on_or_after_opens_at`, I believe.

jpslav:
Looks good, just one change needed.

app/subsystems/tasks/models/tasking_plan.rb
Dantemss: I don't like that method because it keeps barfing when opens_at is nil, which happens in tests (so it pollutes the presence tests etc)

PR #475, opened 2015-07-15
Rename role-user relationship to role_user

karenc:
It used to be user and we end up with something like

```
entity_course.teachers.includes(role: { user: { user: { profile: :account } } })
```

the double user is kind of confusing, after renaming:

```
entity_course.teachers.includes(role: { role_user: { user: { profile: :account } } })
```


karenc:
Closing this, going to do #484 instead.

PR #476, opened 2015-07-15
Faster assignments

Dantemss:
Sped up assignment creation by ~3x (from ~4.5s to ~1.3s per student).

Sorry JP, couldn't resist.

### DIFF
--- a/app/routines/distribute_tasks.rb
+++ b/app/routines/distribute_tasks.rb
@@ -6,6 +6,20 @@ class DistributeTasks
 
   protected
 
+  def efficiently_save(task)
+    steps = task.task_steps.to_a
+    task.task_steps = []
+    task.save!
+    steps.each do |step|
+      tasked = step.tasked
+      tasked.task_step = nil
+      tasked.save!
+      step.tasked = tasked
+      task.task_steps << step
+      step.save!
+    end
+  end
+
   def exec(task_plan)
     # Delete pre-existing assignments
     task_plan.tasks.destroy_all unless task_plan.tasks.empty?
@@ -30,7 +44,7 @@ class DistributeTasks
       task.opens_at = opens_ats[ii]
       task.due_at = due_ats[ii] || (task.opens_at + 1.week)
       task.feedback_at ||= task.due_at
-      task.save!
+      efficiently_save(task)
     end
 
     outputs[:tasks] = tasks
--- a/lib/tasks/demo_base.rb
+++ b/lib/tasks/demo_base.rb
@@ -267,16 +267,16 @@ class DemoBase
     raise "Invalid number of responses for #{task.title}" +
           "(responses,task_steps) = (#{responses.count}, #{task.task_steps.count})\n" +
           "(task = #{print_task(task: task)}) " \
-      if responses.count != task.task_steps.count
+      if responses.count != task.steps_count
 
-    core_task_steps = task.core_task_steps
+    core_task_steps = task.core_task_steps.to_a
     core_task_steps_count = core_task_steps.count
 
     core_task_steps.each_with_index do |step, index|
       work_step(step, responses[index])
     end
 
-    spaced_practice_task_steps = task.spaced_practice_task_steps
+    spaced_practice_task_steps = task.spaced_practice_task_steps.to_a
     spaced_practice_task_steps_count = spaced_practice_task_steps.count
 
     spaced_practice_task_steps.each_with_index do |step, index|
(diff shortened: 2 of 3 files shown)


app/routines/distribute_tasks.rb
```diff
@@ -6,6 +6,20 @@ class DistributeTasks
 
   protected
 
+  def efficiently_save(task)
```
Dantemss: My guess is that Rails is very inefficient with its callbacks...

app/routines/distribute_tasks.rb
```diff
@@ -6,6 +6,20 @@ class DistributeTasks
...
+    steps = task.task_steps.to_a
+    task.task_steps = []
+    task.save!
+    steps.each do |step|
+      tasked = step.tasked
+      tasked.task_step = nil
+      tasked.save!
```
jpslav: With the `tasked`'s `task_step` set to nil, `TaskedExercise#no_feedback` won't trigger a bunch of queries like it normally would in the original save call.  What happens speed-wise if you comment out that validation and call it the normal way?  

`TaskedExercise#[valid_answer, valid_state]` could also potentially be skipped in this case for extra speed (so we don't end up parsing the exercise JSON over and over)

app/routines/distribute_tasks.rb
```diff
@@ -6,6 +6,20 @@ class DistributeTasks
...
+      step.tasked = tasked
+      task.task_steps << step
+      step.save!
```
jpslav: since `task` is already saved, the `<< step` call [will save the step](http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many), making the extra `save!` on line 19 unnecessary.

jpslav: `TaskStep` has an `after_save { task.update_step_counts }`, which means that not-so-cheap method is belng called once each on lines 18 and 19 each time we iterate through this look (and each call gets slower and slower because more steps are added to recount).  Can we just count once at the end (safely)?

lib/tasks/demo_base.rb
```diff
@@ -267,16 +267,16 @@ def work_task(task:, responses:[])
...
-    core_task_steps = task.core_task_steps
+    core_task_steps = task.core_task_steps.to_a
     core_task_steps_count = core_task_steps.count
 
     core_task_steps.each_with_index do |step, index|
       work_step(step, responses[index])
     end
 
-    spaced_practice_task_steps = task.spaced_practice_task_steps
+    spaced_practice_task_steps = task.spaced_practice_task_steps.to_a
```
jpslav: I imagine you're doing the `to_a` call so that the steps get fully loaded and cached.  

You could instead not do this and take out line 280, then replace `spaced_practice_task_steps_count` on line 292 with `spaced_practice_task_steps.size`.  Since the full records would have already been used on line 282, the `size` call will call `length` on the loaded records instead of another DB count.  If for some reason the records hadn't been loaded, `size` will call count and do the query efficiently.  

Just a thought for these `to_a` uses here and elsewhere... you can leave like this if you want.

lib/tasks/demo_base.rb
```diff
@@ -285,7 +285,7 @@ def work_task(task:, responses:[])
 
     return unless task.core_task_steps_completed?
 
-    personalized_task_steps = task.personalized_task_steps
+    personalized_task_steps = task.personalized_task_steps.to_a
     personalized_task_steps_count = personalized_task_steps.count
```
jpslav: This line is unused. (also means the `to_a`) is unnecessary

app/routines/distribute_tasks.rb
Dantemss: Commenting out the validations made the saving about 30% faster. Nowhere near the gains with this method. I think Rails must have been doing non-validation callbacks multiple times before.

Dantemss:
@jpslav ready

app/routines/distribute_tasks.rb
Dantemss: The save is apparently necessary or else the step counts will be incorrect.

app/routines/distribute_tasks.rb
Dantemss: The save is apparently necessary or else the step counts will be incorrect.

app/routines/distribute_tasks.rb
Dantemss: Sorry I meant the other save :)

PR #485, opened 2015-07-17
Rename role user

karenc:
- Add direct link between entity roles and users through role_users

  Now we can do get all the roles that a user has by doing
  `entity_user.roles` and get the user that has a role by doing
  `entity_role.user`.

- Rename `Role::Models::User` to `Role::Models::RoleUser`


app/controllers/admin/courses_controller.rb
```diff
@@ -17,8 +17,8 @@ def edit
     entity_course = Entity::Course.find(params[:id])
     @course = GetCourseProfile[course: entity_course]
     @periods = entity_course.periods
-    teachers = entity_course.teachers.includes(role: { user: { user: { profile: :account } } })
-    @teachers = teachers.collect { |t| t.role.user.user.profile }
+    teachers = entity_course.teachers.includes(role: { user: { profile: :account } })
+    @teachers = teachers.collect { |t| t.role.role_user.user.profile }
```
jpslav: `t.role.user`  ?

karenc: yes, good catch : )

PR #489, opened 2015-07-17
need type in data heading for performance report to set different headings for external

pandafulmanda:
Is this allowed?

Wanted for openstax/tutor-js#477 so that there's a way to differentiate external heading from reading and homework heading and avoid linking to stats review


jpslav:
Looks good to me :-)

pandafulmanda:
dang it, fixing the spec for real

jpslav:
just an order problem

pandafulmanda:
YAY :cake:

Issue #463, opened 2015-07-14
Potential improvements to demo script YAML 


kevinburleigh75:
* `periods` should not be referenced by indices because it is super error-prone
 * see [this example](https://github.com/openstax/tutor-server/blob/master/lib/tasks/demo/biology.yml#L24)
 * instead, use a handle (much like how `students` are...uh...handled)
    * handles could be defined [here](https://github.com/openstax/tutor-server/blob/master/lib/tasks/demo/biology.yml#L7)
    * recycling `name` for this purpose would probably not be a great solution

* separate completion from assignment
 * not sure if this necessarily implies separation in the YAML structure, but it might

* we should consider whether or not the same `student` role could be in multiple periods, and whether we want the demo script to support that now or not

* if we add a `latest content` flag to the command line, we'll almost certainly need to add a `no task auto-completion` flag to prevent the auto-completion safeguard from interfering with assignment distribution

jpslav:
cc @nathanstitt @Dantemss as a heads up, not that this needs to be done now, but the next time someone is updating the demo code this is a good writeup to reference

nathanstitt:
+1, good points.   I also think we should add a method to specify the configuration on the command line, either as a Rake argument or an ENV var.

Dantemss:
- Agree
- Not totally sure what this means
- I think we decided not to worry about this for now
- Agree

kevinburleigh75:
clarification on the second point:

Assignment auto-completion can cause the demo script to abort before all content has been imported.  Usually this failure is due to a mismatch in the number of task steps, and not due to content.

By separating the import and auto-completion phases, script users can still create their own assignments from the imported content.

(This issue should be lessened now that the demo script uses fixed content versions, and - presumably - whoever updates those versions will actually fix any issues before committing it.)

Dantemss:
I agree with separating assignment creation/completion into their own rake tasks (that are still all run together by `rake demo`).

kevinburleigh75:
The @Dantemss has a legit point.  Definitely worth considering.

Dantemss:
Done by Nathan in https://github.com/openstax/tutor-server/pull/496

PR #496, opened 2015-07-21
Demo import enhancements

nathanstitt:
As discussed in #463, this:

 * Updates Physics to version with all 10 chapters
 * Sets an id for periods, and refers to them by that vs the error prone index method.
 * separates completion from assignment
 * Adds a :version argument to the Rake tasks
 * Replaces the exception when response counts do not match the exercise counts with an warning.   If an response is not found for an exercise index, it will not be worked.


lib/tasks/demo/content_configuration.rb
```diff
@@ -124,7 +126,7 @@ def course
   end
 
   def self.config_directory
-    Thread.current[:config_directory] || CONFIG_DIR
+    Thread.current[:config_directory] || ENV['CONFIG'] || CONFIG_DIR
```
kevinburleigh75: I feel like this is error-prone.  Is it not possible to just take a constructor argument or something and defer this logic to the caller?  Or something more concrete?

nathanstitt: Good point.  After discussion I don't think there's anything gained by exposing this a public getter method. Instead we should just perform the above logic once and then pass the result into the config instances when they're created.

PR #498, opened 2015-07-21
Lateness in performance report exports

joemsak:
https://www.pivotaltracker.com/story/show/99330300

* Adds `Task#late?` where `#last_worked_at` is present and later than `due_at`
* Moves `ExportPerformanceReport` xlsx formatting into `PerformanceReport::ExportXlsx`, for future proofing
  * Probably better to view final file https://github.com/openstax/tutor-server/blob/lateness-in-performance-report-export/app/subsystems/tasks/export_performance_report.rb
* Grades are now formatted in student rows as `91%` instead of `0.9191919191919191`
* Late assignments are marked yellow with a `Late` comment by the author `OpenStax`


app/subsystems/tasks/models/task.rb
```diff
@@ -88,14 +88,14 @@ def set_last_worked_at(time:)
   end
 
   def status
-    # task is "completed" if all steps are completed,
-    #         "in_progress" if some steps are completed and
-    #         "not_started" if no steps are completed
-    if self.completed?
-      'completed'
+    if completed? && last_worked_at > due_at
```
jpslav: Don't want to mix lateness with completeness.

A task can be incomplete and late, complete and late, incomplete and not late, or complete and not late.  "late" just means worked on after due.

task can have `def late?; last_worked_at > due_at; end`

could be that `status` is really `completion_status`

joemsak: oh yeah doy :-P

app/subsystems/tasks/models/task.rb
```diff
@@ -88,17 +88,25 @@ def set_last_worked_at(time:)
   end
 
   def status
-    # task is "completed" if all steps are completed,
-    #         "in_progress" if some steps are completed and
-    #         "not_started" if no steps are completed
-    if self.completed?
+    if late?
```
jpslav: Do not mix lateness into status (status method should be unchanged by this PR).

app/subsystems/tasks/performance_report/write_to_file/xlsx.rb
```diff
@@ -0,0 +1,101 @@
+module Tasks
+  module PerformanceReport
+    module WriteToFile
```
jpslav: Like the extraction and preparation for different formats.  Why this extra `WriteToFile` module/directory?  Maybe we could get rid of one layer?

joemsak: Just an attempt at obvious naming... will work it out... yeah I looked at this and realized this is a strategy pattern waiting to happen

joemsak: I'd like to say `PerformanceReport::XlsxFormatter` but it doesn't really fit as a routine name. It's good for a service class you'd instantiate with `.new()` and call some instance method on, but for routines, it's better to name it as what it does... hence `WriteToFile` namespace...

`PerformanceReport::ExportXlsx` perhaps?

jpslav:
Thanks @joemsak!

Issue #495, opened 2015-07-20
Don't send tasks to student if opens_at > current time


Dantemss:

nathanstitt:
Continuing the discussion from #383,  I can't think of any reason the student dashboard would **currently** need to know about unopened tasks.  With that said, if the BE want's to continue sending them, it's trivial to filter them out so they don't display.

jpslav:
Would be fine to stop sending them.  If we don't want kids to know about the tasks, might as well not send them so enterprising students don't go hitting our API and surprising our teachers with unexpected knowledge.

Dantemss:
Done by Nathan

PR #499, opened 2015-07-21
Don't send unopened tasks to students

nathanstitt:
Per discussion in #495

### DIFF
--- a/app/controllers/api/v1/courses/dashboard.rb
+++ b/app/controllers/api/v1/courses/dashboard.rb
@@ -18,8 +18,17 @@ class Api::V1::Courses::Dashboard
 
     raise SecurityTransgression if role_type.nil?
 
-    load_tasks(role)
-    load_plans(course) if :teacher == role_type
+    tasks = load_tasks(role)
+    if :teacher == role_type
+      load_plans(course)
+    end
+    if :student == role_type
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/courses/dashboard.rb
```diff
@@ -18,8 +18,17 @@ def exec(course:, role:)
...
 
-    load_tasks(role)
-    load_plans(course) if :teacher == role_type
+    tasks = load_tasks(role)
+    if :teacher == role_type
+      load_plans(course)
+    end
+    if :student == role_type
+      # students are not allowed to view un-opened tasks
+      Time.use_zone(course.profile.timezone) do
```
jpslav: Do you really need to use the time zone?  Hasn't the `opens_at` already been adjusted for the course time zone, making it directly comparable against `Time.now`?  (I hope so anyway)

jpslav: As much as possible I'd prefer for the logic that is here to go into the `load_tasks` private method, a la Joe's style, so the main exec can just be a simple list of private method calls.

nathanstitt: I guess I don't :grin:   I wasn't aware that it was already adjusted.  I'll remove it.

Issue #378, opened 2015-06-15
add first_name and last_name fields to perf report


darrenmason:
Currently student name endpoint returns combined field for student name.  firstname and lastname fields would allow sorting by lastname in report table.

jpslav:
@karenc this should be a pretty small change and should have a good impact (teachers will expect this to be able to sort by name, etc).  Can you make this change in the API (for now just add first_name and last_name, don't remove the full name so we don't break FE).  Also, while you're at it, do this split in the performance report export too.

philschatz:
Adding @darrenmason, @nathanstitt, and @pandafulmanda  so the following are updated as well:

- Perf Report
- Roster
- HW Review page (both class-aggregate Free responses and the "Viewing As..." in the bottom-right corner)

jpslav:
Thanks @philschatz

jpslav:
@philschatz @pandafulmanda @nathanstitt this should be addressed now I think with the exception of the "HW Review page" bullet above -- if that is the detailed task plan stats that has a collection of strings called `names`, please let us know how you want that API to change.

PR #501, opened 2015-07-21
Add first and last names to performance report

karenc:
- Remove GetStudentProfiles and CourseMembership::GetStudents

  No longer in use.

- Add `CourseMembership::Models::Period.active_enrollments` association

- Add first and last names to performance report

- Add first and last names to performance report exports

- Remove extra delegate in CourseMembership::Models::Student


app/routines/get_student_profiles.rb
```diff
@@ -1,32 +1,19 @@
 class GetStudentProfiles
```
jpslav: This routine is now very similar to `GetStudentRoster`.  Could it maybe be used in place of this routine?  (would have to add a group_by period call). If so, we can ditch `GetStudentProfiles`.

app/subsystems/course_membership/models/period.rb
```diff
@@ -16,7 +19,7 @@ class CourseMembership::Models::Period < Tutor::SubSystems::BaseModel
   default_scope { order(:name) }
 
   def student_roles
-    enrollments.latest.active.includes(student: :role).collect{ |en| en.student.role }
+    active_enrollments.includes(student: :role).collect{ |en| en.student.role }
```
jpslav: for my own benefit, can you explain how this call works without a `period` being passed as indicated on line 11 above?

jpslav:
Looks good.  One substantive comment to address and then a rebase.

app/routines/get_student_profiles.rb
karenc: Student roster has period information, maybe we can just filter it?

app/subsystems/course_membership/models/period.rb
karenc: I think `has_many` with lambda has an optional argument which is the object itself (a `CourseMembership::Models::Period` instance) if I understand it correctly.

For example:
`CourseMembership::Models::Period.first.student_roles`, the `active_enrollments` `period` argument is `CourseMembership::Models::Period.first`.

Don't know if this answers your question.

app/routines/get_student_profiles.rb
karenc: Oh turns out... after a rebase onto master, there's no code using `GetStudentProfiles` anymore. ;D  I'll remove it.

jpslav:
This partially addresses #378

PR #502, opened 2015-07-22
Add needed fields to performance report

nathanstitt:
This adds `due_at` and `last_worked_at` to the tasks, and `type` to data_headings.

### DIFF
--- a/app/subsystems/tasks/get_performance_report.rb
+++ b/app/subsystems/tasks/get_performance_report.rb
@@ -58,6 +58,7 @@ module Tasks
       tasks.collect.with_index { |t, i|
         { title: t.title,
           plan_id: t.tasks_task_plan_id,
+          type: t.task_type,
           due_at: t.due_at
         }.merge(average(t, i))
       }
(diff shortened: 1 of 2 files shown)


jpslav:
Pretty obvious, but just need to add the new fields to the performance report spec in the courses controller: https://github.com/openstax/tutor-server/blob/master/spec/controllers/api/v1/courses_controller_spec.rb#L781-L917

(no get_performance_report spec but do have this controller spec -- and either Karen or Joe will be moving this code into a PerformanceController, fyi)

nathanstitt:
I now see that I worded that horribly last night :grin: 

I intended to ask if I should create a `get_performance_report_spec` even if it slows down the specs quite a bit?  Or if I should just test in the controller as you pointed out?

jpslav:
I would say don't for now.  Hopefully we can get to a place where we can build up fake data (faster), at which point we can write that spec.  For the moment, the coverage in the courses controller is enough.

So just need a rebase.

PR #503, opened 2015-07-22
Use time cop to work tasks before due date unless otherwise specified

nathanstitt:
By default the importer was working tasks after the due_at.  This adds a late setting so some students can be marked late, and other's will be worked 1 minute before due_at.


joemsak:
Merged into my branch, thanks

PR #505, opened 2015-07-22
Speed up learning guide

Dantemss:
Ready for review.


Dantemss:
Oh I realize that there's some potential to combine a lot of BL queries, so those can be even more optimized (I only did local optimizations, didn't have time to optimize BL calls).

nathanstitt:
Very cool.  Loads in 2.1 seconds for me now, used to take over 40.

app/routines/reset_practice_widget.rb
```diff
@@ -54,6 +54,7 @@ def exec(role:, exercise_source:, page_ids: [], book_part_ids: [], randomize: tr
       step = Tasks::Models::TaskStep.new(task: task)
 
       step.tasked = TaskExercise[exercise: exercise, task_step: step]
+      step.add_related_content(get_related_content_for(exercise)) unless exercise_source == :fake
```
kevinburleigh75: Why don't we need related content for exercises from the `:fake` source?

app/routines/reset_practice_widget.rb
```diff
@@ -69,6 +70,22 @@ def exec(role:, exercise_source:, page_ids: [], book_part_ids: [], randomize: tr
...
+    { title: page.title, chapter_section: page.chapter_section }
+  end
+
+  def content_exercise_page(content_exercise)
+    los = content_exercise.los + content_exercise.aplos
+
+    pages = Content::Models::Page.joins{page_tags.tag}
+                                 .where{page_tags.tag.value.in los}
+
+    raise "#{pages.count} pages found for exercise #{content_exercise.url}" unless pages.one?
```
kevinburleigh75: Are we planning to change this before we allow multiple book versions to be imported?  Will it explode if we don't?

Dantemss: Yeah it'll go kaboom.

app/routines/reset_practice_widget.rb
Dantemss: `:fake` exercises are not even persisted in the DB so they have no page... I suppose we could create a fake page, but that's weird too...

app/subsystems/tasks/get_tasks.rb
```diff
@@ -8,16 +8,8 @@ class Tasks::GetTasks
...
-                     where{entity_role_id.in role_ids}.
-                     collect{|tasking| tasking.task}
-
-    # This was temporarily removed due to a problem (with entity_extensions?):
-    # outputs[:tasks] = Entity::Task.joins{taskings}
-    #                               .where{taskings.entity_role_id.in role_ids}
-
-    outputs[:tasks] = entity_tasks
+    outputs[:tasks] = Entity::Task.joins{taskings}
+                                  .where{taskings.entity_role_id.in role_ids}
```
kevinburleigh75: Yay!

Dantemss: Funny enough I think in the end I removed this routine from the learning guide routine.

PR #515, opened 2015-07-24
Hide uncertain clues from the learning guide

Dantemss:
This does not handle "never hide a CLUE once it is displayed", though the rules from BL ensure that a CLUE is never hidden after 5 questions are answered.

### DIFF
--- a/app/external/openstax/biglearn/v1.rb
+++ b/app/external/openstax/biglearn/v1.rb
@@ -8,7 +8,23 @@ module OpenStax::Biglearn::V1
     tags = [tags].flatten.compact
     roles = [roles].flatten.compact
 
-    client.get_clue(roles: roles, tags: tags)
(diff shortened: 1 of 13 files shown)


Dantemss:
This does not handle "never hide a CLUE once it is displayed", though the rules from BL ensure that a CLUE is never hidden after 5 questions are answered.

Dantemss:
@lakshmivyas You should take a look at this take a look before Chris deploys

jpslav:
Ok, I'm merging this.  @lakshmivyas you should still take a look so we can head off any problems early in the day.  @cnuber heads up that this is in (in case it has BL deployment ramifications, though I'm guessing LN's stuff is in place)

lakshmivyas:
Looks great.  @Dantemss - Thanks so much. You rock :+1: .

PR #518, opened 2015-07-24
Update Biology to version with latest chapters

nathanstitt:
Also rearranges the physics content so the assignments
are made in logical order, i.e. section 3.1 and then 4

And a bit better logging output.  With a bonus screenshot solely because of the content:

screen shot 2015-07-24 at 3 07 52 pm


Dantemss:
How did that hippo... nevermind.

jpslav:
hahahahaha.... hippo

PR #519, opened 2015-07-24
Change the Lev status store to use Redis

Dantemss:
The namespace setting in secrets.yml is not really required, since Lev will also add its own lev_status namespace.

### DIFF
--- a/config/initializers/lev.rb
+++ b/config/initializers/lev.rb
@@ -1,3 +1,9 @@
+redis_secrets = Rails.application.secrets['redis']
+
 Lev.configure do |config|
   config.raise_fatal_errors = false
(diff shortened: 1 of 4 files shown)


config/initializers/lev.rb
```diff
@@ -1,3 +1,9 @@
+redis_secrets = Rails.application.secrets['redis']
+
 Lev.configure do |config|
   config.raise_fatal_errors = false
+  config.status_store = Lev::RedisStore.new(
```
jpslav: Per the comment in https://github.com/lml/lev/pull/35, hopefully this can be 

```ruby
config.status_store = # some existing redis cache store or a new one
```

PR #520, opened 2015-07-25
Speed up demo/specs

Dantemss:
- Sped up the book import (about 2x) by reusing more tag objects during the import
- Removed setup_001 and spec


app/external/openstax/exercises/v1/exercise.rb
```diff
@@ -37,12 +37,30 @@ def tags
     @tags ||= content_hash['tags']
   end
 
+  def tag_hashes
```
jpslav: Can you make this `protected` to prevent it from leaking out?

jpslav: oops it is already out :-D

app/subsystems/content/routines/find_or_create_tags.rb
```diff
@@ -0,0 +1,79 @@
...
+  #       teks:        a raw value of a (possibly non-existent) TEKS tag to link
+  #                    to (type must be "lo")
+  #
+  def exec(input:)
+    tags = input.select{ |ii| ii.is_a?(Content::Models::Tag) }
+    hash_array = input.select{ |ii| ii.is_a?(Hash) }
+    outputs[:tags] = tags + find_or_create_tags_from_hash_array(hash_array)
+  end
+
+  def find_or_create_tags_from_hash_array(hash_array)
```
jpslav: I'm finding it difficult to understand what's going on here.  Any chance some of the code below could be DRYed up with a helper class like:

```ruby
class HashArray
  def initialize(hash_array)
    @array
  end

  def remove_existing_tags!(tag_value_hash_field:)
    values = @array.collect{ |hash| hash[tag_value_hash_field] }.compact.uniq
    existing_tags = Content::Models::Tag.where(value: values)
    existing_tag_values = existing_tags.collect{ |tag| tag.value }
    @array.reject!{|hash| !existing_tag_values.include?(hash[:value])}
    existing_tags
  end
end
```

and then maybe the code in the big method could be a bit simpler? a la...

```ruby
def find_or_create_tags_from_hash_array(hash_array)

  remaining_hashes = HashArray.new(hash_array)

  # Find existing tags
  existing_tags = remaining_hashes.remove_existing_tags!(tag_value_hash_field: :value)

  # Create new TEKS tags first
  existing_teks_tags = remaining_hashes.remove_existing_tags!(tag_value_hash_field: :teks)

  ...
```

If it can't, so be it, just seems like there's some repeated stuff going on here that could be pulled out.

jpslav: (all that said, the comments did help so thanks for that)

Dantemss: Uh ok

app/external/openstax/exercises/v1/exercise.rb
```diff
@@ -37,12 +37,30 @@ def tags
...
+             (!APLO_REGEX.match(tag).nil? ? :aplo : :generic)
+      }
+    end
+  end
+
+  def lo_hashes
+    @lo_hashes ||= tag_hashes.select{ |hash| hash[:type] == :lo }
+  end
+
+  def aplo_hashes
```
jpslav: let's keep this area of the code in mind when we think about where encapsulation can be improved -- while I can see arguments both ways, part of me thinks that LOs and APLOs should not be something that OX Exercises code should be aware of (anything in `/app/external/openstax/exercises` should only be aware of things we'd be ok with `github.com/openstax/exercises` being aware of)

app/subsystems/tasks/placeholder_strategies/i_reading_personalized.rb
```diff
@@ -17,7 +17,7 @@ def populate_placeholders(task:)
       difficulty:        0.5,
       allow_repetitions: true
     )
-
+debugger if exercise_urls.empty?
```
jpslav: howdy Mr. debugger statement

jpslav:
review done

app/subsystems/tasks/placeholder_strategies/i_reading_personalized.rb
Dantemss: Oops :)

Dantemss:
@jpslav done

PR #527, opened 2015-07-28
Handle external task for perf report export as well

pandafulmanda:
Maybe this will work?

I noticed locally, an error was yelling about how there isn't a way to handle the external tasks

### DIFF
--- a/app/subsystems/tasks/performance_report/export_xlsx.rb
+++ b/app/subsystems/tasks/performance_report/export_xlsx.rb
@@ -96,7 +96,7 @@ module Tasks
         when 'reading'
           data.status.humanize
         when 'external'
-          data.status.humanize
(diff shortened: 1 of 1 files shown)


Dantemss:
Yep exports in tutor-dev for bio are failing because of the external assignments now. We need this.

app/subsystems/tasks/performance_report/export_xlsx.rb
```diff
@@ -95,10 +95,12 @@ def score(data)
           '%.2f' % (data.correct_exercise_count/data.exercise_count.to_f)
         when 'reading'
           data.status.humanize
+        when 'external'
+          data.status.humanize
```
jpslav:
```ruby
data.status == "not_started" ? "Not clicked" : "Clicked"
```

PR #530, opened 2015-07-28
Report on whether a student is active or inactive in roster

jpslav:


jpslav:
Pairs with https://github.com/openstax/tutor-js/pull/551.  Can be merged before that.

Issue #480, opened 2015-07-16
Make CNX archive a URL in secrets


jpslav:
Currently the default CNX archive URL is hardcoded in `OpenStax::Cnx::V1#set_archive_url`.  Instead, get it from secrets like we do for all of the other 3rd party sites.

Make an `openstax_cnx.rb` initializer to pull out the URL from secrets and set it in the method above.  There's still an override in `environments/test.rb` which is fine.

Coordinate with @cnuber @lakshmivyas to get the secret into deployment setup as needed.

Dantemss:
Do we still want this?

Dantemss:
Done in #531

Issue #494, opened 2015-07-17
Move configurable classes from app to lib


Dantemss:
Or else they will break when the code is reloaded in dev.

For example, the Biglearn client.

PR #531, opened 2015-07-28
Added CNX archive URL to the secrets file

Dantemss:
Need to coordinate with deployment team.

### DIFF
--- a/app/external/openstax/cnx/v1.rb
+++ b/app/external/openstax/cnx/v1.rb
@@ -1,41 +1,38 @@
+require 'addressable/uri'
 require 'open-uri'
 
 module OpenStax::Cnx::V1
 
   # Sets the archive URL base.  'url' is nominally the non-SSL URL,
-  # tho may be SSL.  An explicit SSL URL can be passed in `ssl`, or
+  # tho may be SSL. An explicit SSL URL can be passed in `ssl`, or
   # by default the SSL URL will be guessed from the URL.
-  def self.set_archive_url_base(url:, ssl: :automatic)
-    @archive_url_base = url
-    @ssl_archive_url_base = (ssl == :automatic) ?
-                            url.gsub('http://','https://') :
-                            ssl
+  def self.set_archive_url_base(url:, ssl: nil)
+    uri = Addressable::URI.parse(url)
+    uri.scheme = 'http'
+    @archive_url_base = uri.to_s
+    uri = Addressable::URI.parse(ssl) unless ssl.nil?
+    uri.scheme = 'https'
+    @ssl_archive_url_base = uri.to_s
   end
 
   def self.archive_url_base(ssl: false)
-    set_archive_url_base(url: 'http://archive.cnx.org/contents/') if @archive_url_base.nil?
     ssl ? @ssl_archive_url_base : @archive_url_base
   end
 
-  def self.with_archive_url(url:)
+  def self.with_archive_url(url:, ssl: nil)
     old_url = archive_url_base
     old_ssl_url = archive_url_base(ssl: true)
 
     begin
-      set_archive_url_base(url: url)
-      result = yield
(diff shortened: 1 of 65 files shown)


Dantemss:
Closes https://github.com/openstax/tutor-server/issues/480

Dantemss:
Requires https://github.com/openstax/tutor-deployment/pull/114

app/external/openstax/cnx/v1.rb
```diff
@@ -1,41 +1,38 @@
...
-  def self.with_archive_url(url:)
+  def self.with_archive_url(url:, ssl: nil)
     old_url = archive_url_base
     old_ssl_url = archive_url_base(ssl: true)
 
     begin
-      set_archive_url_base(url: url)
-      result = yield
+      set_archive_url_base(url: url, ssl: ssl)
+      yield
```
jpslav: why no longer return the result?

app/external/openstax/cnx/v1.rb
```diff
@@ -1,41 +1,38 @@
...
+    uri = Addressable::URI.parse(url)
+    uri.scheme = 'http'
+    @archive_url_base = uri.to_s
+    uri = Addressable::URI.parse(ssl) unless ssl.nil?
+    uri.scheme = 'https'
+    @ssl_archive_url_base = uri.to_s
   end
 
   def self.archive_url_base(ssl: false)
-    set_archive_url_base(url: 'http://archive.cnx.org/contents/') if @archive_url_base.nil?
```
jpslav: With this method not lazily setting an `@archive_url_base`, I worry that calls like https://github.com/Dantemss/tutor-server/blob/ab5568280df9a704195ded9deb17ebbd9f051b3f/app/controllers/admin/books_controller.rb#L7 are going to get `nil`?

jpslav: Or is that what the initializer below is for?

config/secrets.yml.example
```diff
@@ -61,6 +64,9 @@ test:
       secret: <%= ENV["OPENSTAX_ACCOUNTS_SECRET"] %>
       url: <%= ENV["OPENSTAX_ACCOUNTS_URL"] || 'http://localhost:2999' %>
       stub: <%= ENV["OPENSTAX_ACCOUNTS_STUB"] %>
+    cnx:
+      archive_url: <%= ENV["OPENSTAX_CNX_ARCHIVE_URL"] || \
+                       'http://archive-staging-tutor.cnx.org/contents/' %>
```
jpslav: Can the default URLs all be https so that none of the cassettes change?  (and in general we want https, right?)

app/external/openstax/cnx/v1.rb
Dantemss: the return value of the ensure block is ignored :)

app/external/openstax/cnx/v1.rb
Dantemss: The initializer should take care of this (but keep https://github.com/openstax/tutor-server/issues/494 in mind)

config/secrets.yml.example
Dantemss: Could change them to https here, but due to https://github.com/openstax/tutor-server/pull/531/files#diff-91d11ba6b2322b1ea56c4399cebad8c1R11 it wouldn't make any difference.

We could change SSL to be the default in OpenStax::Cnx::V1::RealClient instead...

Dantemss: Not sure if it will matter too much, though, since CNX is publicly accessible.

jpslav: I think SSL should be the default so we get SSL links inside CNX content (I know we change them, but I think we want SSL all the time in general)

Dantemss:
Cnx stuff now defaults to `https`.

PR #529, opened 2015-07-28
Absolutize exercise urls in reference view pages API

karenc:
- Move update links in book code in UpdatePageContent into a function
- Update exercise links to absolute urls in reference view pages API

PR #533, opened 2015-07-29
Absolutize exercise urls as api

pandafulmanda:
For FE to have API url of exercise

Modified from #529


jpslav:
@philschatz I'm good with this update of @karenc's #529.  You can merge with https://github.com/openstax/tutor-js/pull/554 whenever you want.

I did ask Karen to clarify something in the spec (more comments), but she can do that in some other PRl

PR #534, opened 2015-07-29
Configure generators not to add stuff we don't normally mess with

joemsak:
Is the team cool with this

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -42,5 +42,12 @@ module Tutor
 
     # Use delayed_job for background jobs
     config.active_job.queue_adapter = :delayed_job
+
+    config.generators do |g|
+      g.skip_routes true
+      g.helper false
+      g.assets false
+      g.view_specs false
+    end
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -43,8 +43,8 @@ module Tutor
     # Use delayed_job for background jobs
     config.active_job.queue_adapter = :delayed_job
 
+    # Skip helper, asset and view spec generation when generating scaffolds
     config.generators do |g|
-      g.skip_routes true
       g.helper false
       g.assets false
       g.view_specs false


Dantemss:
Not too sure about skipping routes and view specs, but definitely skip helpers and assets.

joemsak:
Do we write view specs?

Skipping routes because the generator always writes them in a way that I never write them, and usually I've already added the route through TDD, as it's the undefined method of that named route in the view that comes up before the uninitialized constant FooController

joemsak:
When I did `git clean -df spec/views` it deleted that entire folder so I'd wager that no we don't write them ;-P

Dantemss:
Just because we don't doesn't mean we shouldn't :)
At least for the view stuff that is in the BE like the admin views...

joemsak:
I use feature specs to cover the experience... personally have not ever messed with view specs, haven't met even advanced rails devs who do.. I'm not saying we should or shoudln't, but we don't and so why have the generator produce pending test files that the team isn't using?

Dantemss:
In my view (HAH!) view specs are supposed to be simple, maybe just a check that the view doesn't blow up. And we do have one: https://github.com/openstax/tutor-server/blob/master/spec/views/webview/index.html.erb_spec.rb :)

jpslav:
I'm ambivalent here.  I think because of our atypical file structure we don't really use generators all that much, so when we do I'm cool with just blowing away generated files that aren't needed.  If we don't use all these pieces and if generators might not work for our filesystem, do we really need them?

jpslav:
Anyone else in @openstax/tutor-be-dev use generators and want to weigh in before we decide on this?

Dantemss:
I do use generators btw, though admittedly most of the time I use the `rails g model name column:type` syntax, since I usually don't want views etc.

And then I have to move the files into the subsystem folders anyway...

jpslav:
yeah that's how I use them too.

joemsak:
Should I just close..

Dantemss:
I found this lost branch and I think this is useful so I'm reopening...

PR #540, opened 2015-07-30
Provide publishing and jobs info on teacher dashboard 

pandafulmanda:
add more info to the teacher dashboard on plan publishing and jobs so that the FE can check on jobs

- [x] update specs?

### DIFF
--- a/app/representers/api/v1/courses/dashboard_representer.rb
+++ b/app/representers/api/v1/courses/dashboard_representer.rb
@@ -35,11 +35,29 @@ module Api::V1::Courses
                readable: true,
                writeable: false
 
+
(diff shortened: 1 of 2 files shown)


jpslav:
A test on the [dashboard representer spec](https://github.com/openstax/tutor-server/blob/master/spec/representers/api/v1/courses/dashboard_representer_spec.rb) would probably be enough.  Adding data to the fake hashie at the top and testing for it in the represented output to make sure there is no funkiness in the representer.

jpslav:
I added some spec, @pandafulmanda

pandafulmanda:
oh dude, thanks!

PR #542, opened 2015-07-30
Made DistributeTasks lock the TaskPlan to prevent concurrent updates

Dantemss:
I wrote a test but couldn't get it to fail, so it's probably not useful... Anyway here it is:

```rb
require 'rails_helper'

RSpec.describe 'TaskPlanLocking', type: :request, api: true, version: :v1 do

  PUBLISH_THREADS = 3
  UPDATE_THREADS = 3

  before(:all) do
    DatabaseCleaner.strategy = :truncation
    DatabaseCleaner.start

    application = FactoryGirl.create :doorkeeper_application
    course_profile = FactoryGirl.create :course_profile_profile
    course = course_profile.course
    teacher_user = FactoryGirl.create :user_profile
    AddUserAsCourseTeacher.call(user: teacher_user.entity_user, course: course)
    @teacher_token = FactoryGirl.create :doorkeeper_access_token,
                                        application: application,
                                        resource_owner_id: teacher_user.id

    period = CreatePeriod[course: course]
    student_user = FactoryGirl.create :user_profile
    AddUserAsPeriodStudent.call(user: student_user.entity_user, period: period)

    @task_plan = FactoryGirl.create(:tasks_task_plan, owner: course)
    tasking_plan = FactoryGirl.create(:tasks_tasking_plan, target: period.to_model,
                                                           task_plan: @task_plan)
  end

  after(:all) do
    DatabaseCleaner.clean
    DatabaseCleaner.strategy = :transaction
  end

  context 'concurrent requests' do
    it 'does not error out or deadlock' do
      @threads = []
      @results = []

      PUBLISH_THREADS.times do
        @threads << Thread.new do
          @results << DistributeTasks.call(@task_plan)
        end
      end

      UPDATE_THREADS.times do
        @threads << Thread.new do
          api_patch "/api/plans/#{@task_plan.id}",
                    @teacher_token,
                    raw_post_data: { opens_at: Time.now - (1..10).to_a.sample.seconds }.to_json
          @results << Hashie::Mash.new(response.body_as_hash)
        end
      end

      @threads.each{ |thread| thread.join }
      expect(@results.size).to eq PUBLISH_THREADS + UPDATE_THREADS
      @results.each do |result|
        expect(result.errors).to be_blank
      end
    end
  end

end
```

### DIFF
--- a/app/controllers/api/v1/task_plans_controller.rb
+++ b/app/controllers/api/v1/task_plans_controller.rb
@@ -97,6 +97,7 @@ class Api::V1::TaskPlansController < Api::V1::ApiController
     OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, task_plan)
     course = task_plan.owner
     Time.use_zone(course.profile.timezone) do
+      # Lock the TaskPlan to prevent concurrent update/publish
       task_plan.with_lock do
         consume!(task_plan, represent_with: Api::V1::TaskPlanRepresenter)
         OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, task_plan)
--- a/app/routines/distribute_tasks.rb
+++ b/app/routines/distribute_tasks.rb
@@ -52,9 +52,9 @@ class DistributeTasks
       efficiently_save(task)
     end
 
-    outputs[:tasks] = tasks
-
     task_plan.update_column(:published_at, Time.now)
+
+    outputs[:tasks] = tasks
   end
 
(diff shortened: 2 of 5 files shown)


Dantemss:
If you think it could be useful, I can add it...

Dantemss:
This PR will ensure that 2 publications/updates don't happen simultaneously, but they will all still happen (order is not guaranteed after the first one).

PR #546, opened 2015-07-31
Add errors when any task steps are empty while distributing tasks

### LINKED ISSUE #538: When distributing tasks, error out if any tasks have no steps
Occasionally, for a variety of reasons (including lack of content, BE bugs, etc), a Task in Tutor can end up with no steps.  This is understandably confusing to students.  

Add a check [right after this line in DistributeTasks](https://github.com/openstax/tutor-server/blob/6caf77e9a10d5837c064f953206c45ea5300385f/app/routines/distribute_tasks.rb#L36) to check if any tasks has 0 steps.  If so add a fatal error:

```ruby
fatal_error(code: :empty_tasks, message: 'Tasks could not be published because some tasks were empty')
```

This will send the error information to Lev's status, which the FE can check under the TaskPlan's publish job -- the status will be "failed" and the `errors` field will be populated.  FE can use the error code to make their own error message if they choose.

cc @philschatz @jason-holmes

joemsak:
Closes #538

### DIFF
--- a/app/routines/distribute_tasks.rb
+++ b/app/routines/distribute_tasks.rb
@@ -49,6 +49,12 @@ class DistributeTasks
 
     # Call the assistant code to create Tasks, then distribute them
     entity_tasks = assistant.build_tasks(task_plan: task_plan, taskees: untasked_taskees)
+
+    if entity_tasks.flat_map(&:task).any? { |t| t.task_steps.empty? }
+      fatal_error(code: :empty_tasks,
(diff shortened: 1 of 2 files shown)


jpslav:
Errors here @joemsak

Dantemss:
Rebasing might solve at least some of the errors.

Dantemss:
I ended up making my own PR since this was never finished. Closing.

PR #549, opened 2015-07-31
Ecosystems

Dantemss:
# Done
- Renamed Entity::Book to Content::Models::Book
- Add main Ecosystems table
- Move Exercise/Page wrappers into Ecosystems subsystem
- Finish other Ecosystem wrappers
- Create ExercisePool tables/models
- Get rid of direct references to Content models (mostly...)
- Update specs
- Renamed Ecosystem:: to Content::


Dantemss:
Feel free to suggest more TODO's.

app/subsystems/course_content/get_course_books.rb
```diff
@@ -5,7 +5,7 @@ class CourseContent::GetCourseBooks
 
   def exec(course:)
     course_books = CourseContent::Models::CourseBook.where(course: course.id)
-    book_ids = course_books.collect(&:entity_book_id)
-    outputs[:books] = Entity::Book.find(book_ids)
+    book_ids = course_books.collect(&:content_book_id)
+    outputs[:books] = Content::Models::Book.find(book_ids)
```
jpslav: The idea, I believe, is that nothing in `Content::Models` will leave the Content SS.  For models that we want to leave the SS, those will be POROs that live at the top level, e.g. `Content::Book`, `Content::Page`, `Content::Exercise`.  They (at least for this SS) will be read-only and not be activerecords.  @kevinburleigh75 to verify

jpslav: I also realize you may not be there yet in this WIP, just want to bring it up before it is too late.  And also I'm now realizing that it may be `Ecosystem::Book` instead of `Content::Book`, eventually.

Dantemss: It's just WIP.

app/routines/add_ecosystem_to_course.rb
```diff
@@ -0,0 +1,12 @@
+class AddEcosystemToCourse
+  lev_routine
+
+  uses_routine CourseEcosystem::AddEcosystemToCourse,
```
jpslav: Can reduce to:

```ruby
class AddEcosystemToCourse
  lev_routine delegates_to: CourseEcosystem::AddEcosystemToCourse
end
```

app/subsystems/content/import_book.rb
```diff
@@ -9,12 +9,14 @@ class Content::ImportBook
...
+
   protected
 
-  # Imports and saves a Cnx::Book as an Entity::Book
+  # Imports and saves a Cnx::Book as an Content::Models::Book
   # Returns the Book object, Resource object and collection JSON as a hash
-  def exec(cnx_book:)
-    outputs[:book] = Entity::Book.create!
+  def exec(cnx_book:, ecosystem:)
+    outputs[:book] = Content::Models::Book.create!(ecosystem_ecosystem_id: ecosystem.id)
```
jpslav: I know this is still WIP, so feel free to ignore any comment that is about code that is going to change before be un-WIP'd.

I believe the output of this routine should not be a `::Models::` object.  Can it not be?

app/subsystems/ecosystem/book.rb
```diff
@@ -0,0 +1,41 @@
...
+      raise_collection_class_error(
+        collection: title,
+        klass:      String,
+        error:      ::Ecosystem::StrategyError
+      )
+
+      title
+    end
+
+    def chapters
```
jpslav: I'm seeing this in our future :-)

```ruby
delegate_to_strategy :chapters, expect_collection_of: ::Ecosystem::Chapter
```
```

File: app/subsystems/ecosystem/wrapper.rb [hunk 5]
```diff
@@ -0,0 +1,15 @@
...
+
+    def initialize(strategy:)
+      @strategy = strategy
+    end
+
+    protected
+
+    def raise_collection_class_error(collection:, klass:, error:)
+      raise error if [collection].flatten.compact.detect{|obj| !obj.is_a? klass}
+    end
```
jpslav: The `collection` part seems non-meaningful since we pass in non-collections.  Maybe something like...

```ruby
def verify_type(of:, type:, error: ::Ecosystem::StrategyError)
  if type.class.is_a? Array
    raise error if !of.is_a? Array
    raise error if of.flatten.compact.any?{|obj| !obj.is_a? klass}
  else
    raise error if !of.is_a? klass
  end
end
```

with 

```ruby
verify_type of: chapters, type: [::Ecosystem::Chapter]
```

or 

```ruby
verify_type of: title, type: String
```

app/subsystems/ecosystem/ecosystem.rb
```diff
@@ -0,0 +1,129 @@
...
+    end
+
+    def reading_dynamic_exercises(pages:)
+      pages_arr = Array(pages).flatten.compact
+
+      raise_collection_class_error(
+        collection: pages_arr,
+        klass:      ::Ecosystem::Page,
+        error:      ArgumentError
+      )
```
jpslav: maybe there's a DRYer way here too similar to...

```ruby
pages = arrayify_and_verify_type of: pages, type: ::Ecosystem::Page
```

app/subsystems/content/import_book.rb
Dantemss: I can remove the book output, since it shouldn't be used anymore.

app/subsystems/ecosystem/wrapper.rb
Dantemss: Where did klass come from? :)

app/subsystems/content/import_book.rb
kevinburleigh75: Certainly `Content::` might want to use `Books` internally, so returning it isn't terrible per se (though I wonder why we'd want to use a `Routine` in that scenario).  But the rest of `Tutor` should be using `Ecosystem::Book`, etc., so there won't be a need to expose this output going forward.

app/subsystems/ecosystem/book.rb
kevinburleigh75: Probably.  I'm just waiting for the dust to settle.

app/subsystems/ecosystem/ecosystem.rb
kevinburleigh75: probably - internal detail not worth the effort at this exact moment

(also, it's hard to test-drive that sort of thing - better left as a refactoring)

app/subsystems/ecosystem/exercise.rb
```diff
@@ -0,0 +1,125 @@
+module Ecosystem
+  class Exercise < Wrapper
```
kevinburleigh75: Why are these being make subclasses of other stuff?

app/subsystems/ecosystem/models/ecosystem.rb
```diff
@@ -0,0 +1,11 @@
+module Ecosystem
+  module Models
```
kevinburleigh75: I am confused as to why this is here.  I thought the plan was to store all the models, etc., in `Content::`, and just overlay the strategies on that.  Why do we need any models in `Ecosystem::` then?

app/subsystems/ecosystem/wrapper.rb
```diff
@@ -0,0 +1,15 @@
...
+  class Wrapper
+
+    def initialize(strategy:)
+      @strategy = strategy
+    end
+
+    protected
+
+    def raise_collection_class_error(collection:, klass:, error:)
+      raise error if [collection].flatten.compact.detect{|obj| !obj.is_a? klass}
```
kevinburleigh75: I believe the use of `compact` here is actually a bug, since it allows `nils` to pass through undetected.

app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -43,10 +43,8 @@ def self.build_tasks(task_plan:, taskees:)
   end
 
   def self.collect_exercises(task_plan:)
-    exercises = task_plan.settings['exercise_ids'].collect do |exercise_id|
-      Content::GetExercise.call(id: exercise_id).outputs.exercise
-    end
-    exercises
+    exercise_ids = task_plan.settings['exercise_ids']
+    Ecosystem::Exercise.find(exercise_ids)
```
kevinburleigh75: There is an open debate about whether or not querying by multiple ids should preserve order or not.  In this case, it seems that we should preserve order (or sort the results), but I'm not sure this code will do that.

kevinburleigh75: I also think that it'd be better to refactor the assistant to instantiate an `Ecosystem::Ecosystem` early in the process and have this read more like:
```
exercises = ecosystem.exercises_by_id(exercise_ids)
```
That makes it way easier to test, IMO.

app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -72,32 +70,17 @@ def self.build_task(task_plan:)
...
-
-    pages = Content::Routines::SearchPages[tag: exercise_los, match_count: 1]
-    outs = Content::GetLos.call(page_ids: pages.map(&:id)).outputs
-    los = outs.los
-    aplos = outs.aplos
-
-    task.los = los
-    task.aplos = aplos
+    task.los = exercises.map(&:los).flatten
+    task.aplos = exercises.map(&:aplos).flatten
```
kevinburleigh75: This might be an existing bug as well, but it seems like there should be some `uniq`s in there somewhere.

kevinburleigh75: It is also possible that, once we get everything updated, the need to store `los` in a `Task` will be OBE.

app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -72,32 +70,17 @@ def self.build_task(task_plan:)
...
 
   def self.add_core_steps!(task:, exercises:)
     exercises.each do |exercise|
-      related_content = get_related_content_for(exercise)
-
       step = add_exercise_step(task: task, exercise: exercise)
       step.group_type = :core_group
-
-      step.add_related_content(related_content)
+      step.add_related_content(exercise.related_content)
```
kevinburleigh75: It is not obvious to me that `Ecosystem::Exercise` should have a `#related_content` method, per se.  Worth thinking about eventually.

app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -181,14 +148,14 @@ def self.get_taskee_homework_history(task:, taskee:)
...
 
   def self.get_exercise_history(tasks:)
-    exercise_history = tasks.collect do |task|
+    # TODO: Do this wrapping somewhere else? Or punt until we have some Task subsystem wrappers?
+    tasks.collect do |task|
       exercise_steps = task.task_steps.select{|task_step| task_step.exercise?}
-      content_exercises = exercise_steps.collect do |ex_step|
-        content_exercise = Content::Models::Exercise.where{url == ex_step.tasked.url}
+      exercise_steps.collect do |step|
+        strategy = Ecosystem::Strategies::Direct::Exercise.new(step.tasked.exercise)
```
kevinburleigh75: I don't think we should be instantiating `Strategies` this deep in the code.  Instead, it should be done much higher (like `#build_tasks` or even perhaps even in `DistributeTasks` - some refactoring will make it more obvious I think).

I also think this whole method could be rewritten as:
```
tasks.collect do |task|
  task.exercises
end
```
or something similar.

app/subsystems/tasks/assistants/i_reading_assistant.rb
```diff
@@ -242,31 +244,10 @@ def self.num_personalized_exercises
...
-    related_content = self.get_related_content_for(exercise)
-    step.add_related_content(related_content)
-    step
-  end
-
   def self.related_content_for_page(page:, title: page.title)
     { title: title, chapter_section: page.chapter_section }
   end
 
-  def self.get_related_content_for(content_exercise)
```
kevinburleigh75: While I agree that this code will need to be moved/updated, it's not clear to me that it is the responsibility of `Ecosystem::` to determine what constitutes 'related content' - to could vary by `Task` type or UX design, etc.  Certainly `Ecosystem::*` classes could have helper methods to get content-based info to ease the burden, though.

app/subsystems/ecosystem/exercise.rb
Dantemss: Could make them include a module instead... no idea why you guys hate inheritance anyway... The module would have to monkey patch the class though, because of the initializer stuff.

app/subsystems/ecosystem/models/ecosystem.rb
Dantemss: Could put this in Entity if you really want. I don't think it belongs in Content.

app/subsystems/ecosystem/wrapper.rb
Dantemss: I'll add a comment saying nils are allowed to pass through.

Dantemss: Or even better, add an allow_nil: option.

app/subsystems/tasks/assistants/homework_assistant.rb
Dantemss: Yeah I know... I added the TODO because I don't know the answer yet.

app/subsystems/tasks/assistants/i_reading_assistant.rb
Dantemss: Seems like a domain thing and I thought ecosystem was supposed to hold domain code?
Any suggestions as to where it should go instead?

app/subsystems/tasks/assistants/homework_assistant.rb
Dantemss: I'll add the uniqs.

app/subsystems/tasks/assistants/homework_assistant.rb
Dantemss: Sure I can do that.

app/subsystems/ecosystem/book.rb
Dantemss: I don't think this is a good idea. No good way to specify the method arguments and Kevin would prob kill me if I just passed `*args` around.

app/subsystems/ecosystem/wrapper.rb
Dantemss: OK I made something like this (will commit soon).

app/subsystems/ecosystem/ecosystem.rb
Dantemss: I reduced these methods to one line now... should be good enough hopefully.

app/subsystems/ecosystem/models/ecosystem.rb
Dantemss: Well fine, I'll move to Content.

app/subsystems/ecosystem/book.rb
```diff
@@ -0,0 +1,19 @@
+module Ecosystem
+  class Book
+
+    include Wrapper
+
+    def title
+      verify_and_return @strategy.title, klass: String
```
jpslav: nice!

app/routines/fetch_and_import_book.rb
```diff
@@ -10,7 +10,8 @@ class FetchAndImportBook
   # Returns a Cnx::Book based on a CNX ID
   def exec(id:)
     outputs[:cnx_book] = OpenStax::Cnx::V1.book(id: id)
-    run(:import_book, cnx_book: outputs[:cnx_book])
+    outputs[:ecosystem] = Ecosystem::Ecosystem.create!
```
jpslav: What's the difference between `Ecosystem::Ecosystem` and `Content::Models::Ecosystem`, or rather why aren't both in the same SS?

app/subsystems/ecosystem/chapter.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def title
+      verify_and_return @strategy.title, klass: String
+    end
+
+    def pages
+      verify_and_return @strategy.pages, klass: ::Ecosystem::Page
+    end
+
+    def chapter_section
```
jpslav: going with `book_location`?

app/subsystems/ecosystem/ecosystem.rb
```diff
@@ -0,0 +1,51 @@
+module Ecosystem
+  class Ecosystem
+
+    include Wrapper
+
+    def self.create(*args, strategy_class: ::Ecosystem::Strategies::Direct::Ecosystem)
+      new(strategy: strategy_class.create(*args))
+    end
+
+    def self.create!(*args, strategy_class: ::Ecosystem::Strategies::Direct::Ecosystem)
```
jpslav: What's the plan with the `args` array?  Not used in the place where you call this method, and don't know what its use will be.

app/subsystems/ecosystem/wrapper.rb
```diff
@@ -0,0 +1,32 @@
...
+                          allow_blank: false, allow_nil: false)
+      return object if klass == Array && object.is_a?(Array)
+
+      if allow_blank
+        return object if object.blank?
+      elsif allow_nil
+        return object if object.nil?
+      end
+
+      raise error if [object].flatten.any?{ |obj| !obj.is_a? klass }
```
jpslav: Since the default `error` is somewhat generic, might be nice to add a message...

```ruby
raise(error, "Tested argument was not of class '#{klass}' as expected") if ...
```

app/subsystems/ecosystem/book.rb
Dantemss: Woohoo

app/subsystems/ecosystem/chapter.rb
Dantemss: Eventually...

app/routines/fetch_and_import_book.rb
Dantemss: For the first question, `Ecosystem::Ecosystem` is the wrapper and `Content::Models::Ecosystem` is the implementation.

For the second question, ask KB.

app/subsystems/ecosystem/ecosystem.rb
Dantemss: Could remove, since Ecosystem has no attributes I suppose...

app/subsystems/ecosystem/wrapper.rb
Dantemss: Sure...

app/subsystems/ecosystem/exercise.rb
Dantemss: Nvm I guess it works without the monkeypatch.

app/routines/fetch_and_import_book.rb
jpslav: @kevinburleigh75 are you expecting both an `Ecosystem` SS and a `Content` SS?

kevinburleigh75: Since we are changing `Content::` instead of adding `Ecosystem::`, it's not clear to my why we'd have any tables in `Ecosystem::`...

app/external/openstax/biglearn/v1.rb
```diff
@@ -31,6 +31,16 @@ def self.add_exercises(exercises)
     client.add_exercises(exercises)
   end
 
+  def self.add_pools(pools)
+    uuids = client.add_pools(pools).collect{ |rs| rs['pool_id'] }
+    pools.each_with_index{ |pool, ii| pool.uuid = uuids[ii] }
```
jpslav: Do these lines represent that Biglearn is setting the UUIDs?

Dantemss: Yes.

Dantemss:
@jpslav @kevinburleigh75 ready to merge!

jpslav:
@Dantemss we had to merge a few other things before this bad boy for the upcoming demo.  Can you rebase then merge?  I would try but I can't push to your branch.... :-)

Dantemss:
@jpslav @kevinburleigh75 rebase done

PR #550, opened 2015-07-31
Better free response checking

jpslav:
switched blank free response validation to when first updated -- too late to do it when m/c response chosen


Dantemss:
Could also instead just add a method that tells the FE what is the next expected step.
Something like `enum next_step: [:mc, :fr, :done]`

jpslav:
Still need the validation to prevent blank free responses when they are required I think.

jpslav:
@karenc check a pivotal card that describes this PR and your takeover of it :-)

PR #551, opened 2015-08-02
Changed import students to use username/password and allow existing users

jpslav:


Gemfile
```diff
@@ -46,7 +46,7 @@ gem 'openstax_utilities', '~> 4.2.0'
 gem 'whenever'
 
 # OpenStax Accounts integration
-gem 'openstax_accounts', '~> 5.2.0'
+gem 'openstax_accounts', git: 'https://github.com/openstax/accounts-rails.git', ref: 'c903f1ed5a1297424f17040bf9cfa4efddf3cd0e'
```
karenc: Is the new version of openstax_accounts released now?

karenc:
Looks good but wait until the next version of openstax_accounts gem is released?

Gemfile
jpslav: Not yet, PR is in and @Dantemss needs to review and release the new version

jpslav:
yes - will wait on accounts-rails release

PR #552, opened 2015-08-03
Add link "Terms" which goes to /fine_print/contracts

karenc:
"Terms" in admin console: (as expected)

admin-terms-1

The "terms" page: (unstyled page)

admin-terms-2

Should I change the page to look more like the other admin pages?

### DIFF
--- a/app/views/layouts/admin.html.erb
+++ b/app/views/layouts/admin.html.erb
@@ -47,6 +47,7 @@
             <% if Timecop.enabled? %>
               <li><%= link_to 'Time Travel', admin_timecop_path %></li>
             <% end %>
+            <li><%= link_to 'Terms', fine_print_path %></li>
           </ul>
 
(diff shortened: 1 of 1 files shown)


jpslav:
If you can, then yes.  Sometimes it is hard to get the layout into an engine like this.

karenc:
@jpslav I was going to ask you how to do it :p

jpslav:
In legacy Tutor we did it like this:  

https://github.com/lml/ost/blob/63bdec720e21840ca9a600c70a03b833886308ee/config/initializers/fine_print.rb#L39-L43

karenc:
I just tried with `layout 'layouts/admin'`, it can't find the paths (```NameError (undefined local variable or method `admin_districts_path' for #<#<Class:0x0000000d11eb78>:0x0000000d11e178>)```)...

PR #562, opened 2015-08-07
Provide bootstrap data to app via script tag

nathanstitt:
Goes along with FE PR [#577](https://github.com/openstax/tutor-js/pull/577)  More details are listed there.

In order for the FE to determine what views should be visible to each user it needs to know the course roles very early in it's lifecycle, even before it fires of any XHR requests.

Therefore the only way for it to obtain that information is for the BE to provide it to the app as it is initialized.

This is only an exploration of a possible approach.  No specs because I'm not sure this is the way to go with it yet.   Comments are welcomed on both PR's.

### DIFF
--- a/app/helpers/webview_helper.rb
+++ b/app/helpers/webview_helper.rb
@@ -1,0 +1,13 @@
+module WebviewHelper
+
+  # Generates data for the FE to read as it boots up
+  def bootstrap_data
+    {
+      user: Api::V1::UserProfileRepresenter.new(current_user),
+      courses: Api::V1::CoursesRepresenter.new(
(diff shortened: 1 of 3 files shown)


jpslav:
I'm more or less ok with this; kind of wish there were a way to just delegate off to the two endpoints this helper is mimicking, but there doesn't appear to be an easy way to do this and it isn't that much code duplication.  If the bootstrapped data grows, we may want to consider alternatives to keep the code dry.

Per [this blog post](http://jfire.io/blog/2012/04/30/how-to-securely-bootstrap-json-in-a-rails-view/) we might want to use `json_escape` instead of `raw`.

nathanstitt:
@jpslav the corresponding FE PR has been merged, we can move forward with this one now

PR #526, opened 2015-07-28
Use the new cached task counts in CalculateTaskPlanStats

karenc:
Close #470

PR #565, opened 2015-08-10
Stats improvements

karenc:
- Make .preload(:tasked) in Task#core_task_steps optional

  Add argument `preload_tasked` to .preload(:tasked) only if it's necessary

  Towards CalculateTaskPlanStats optimization

- Make .preload(:tasked) in Task#spaced_practice_task_steps optional

  Add argument `preload_tasked` to .preload(:tasked) only if it's necessary

  Towards CalculateTaskPlanStats optimization

- Speed up CalculateTaskPlanStats by eager loading

  Part of #526

  Using the demo rake task to populate the database, running
  `CalculateTaskPlanStats.call(plan: Tasks::Models::TaskPlan.find(1))`
  in `rails console` gives the following improvements:

  master (5938dd5fa) makes 1016 queries and takes about 654ms.

  This branch makes 13 queries and takes about 106ms.

### DIFF
--- a/app/routines/calculate_task_plan_stats.rb
+++ b/app/routines/calculate_task_plan_stats.rb
@@ -84,20 +84,15 @@ class CalculateTaskPlanStats
 
   def get_tasked_exercises_from_task_steps(task_steps)
     tasked_exercise_ids = task_steps.flatten.select{ |t| t.exercise? }.collect{ |ts| ts.tasked_id }
-    Tasks::Models::TaskedExercise.where(id: tasked_exercise_ids)
+    Tasks::Models::TaskedExercise.joins { task_step }
+                                 .where { id.in tasked_exercise_ids }
+                                 .order { task_step.number }
                                  .eager_load([{exercise: :page},
                                               {task_step: {task: {taskings: :role}}}]).to_a
   end
 
-  def get_page_for_tasked_exercise(tasked_exercise)
-    content_exercise = tasked_exercise.exercise
-    strategy = ::Content::Strategies::Direct::Exercise.new(content_exercise)
-    ecosystem_exercise = ::Content::Exercise.new(strategy: strategy)
-    ecosystem_exercise.page
-  end
-
   def group_tasked_exercises_by_pages(tasked_exercises)
-    tasked_exercises.group_by{ |te| get_page_for_tasked_exercise(te) }
+    tasked_exercises.group_by { |te| te.exercise.page }
   end
 
   def generate_page_stats_for_task_steps(task_steps)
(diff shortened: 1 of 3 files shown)


app/routines/calculate_task_plan_stats.rb
```diff
@@ -85,15 +83,23 @@ def mean_grade_percent(tasks)
...
-    run(:search_pages, tag: tasked_exercise.los + tasked_exercise.aplos, match_count: 1).outputs.items.first
+    tasked_exercise_ids = task_steps.flatten.select { |ts| ts.exercise? }
+                                            .collect { |ts| ts.tasked_id }
+    Tasks::Models::TaskedExercise
+      .joins { task_step }
+      .where { id.in tasked_exercise_ids }
+      .order { task_step.number }
+      .eager_load(
+        [{ task_step: { task: :taskings } },
+         { exercise: { tags: { page_tags: :page } } }])
```
jpslav: you're going to have some definite rebasing unfortunately, after Dante's PR.

app/subsystems/tasks/models/task.rb
```diff
@@ -113,16 +113,18 @@ def practice?
     page_practice? || chapter_practice? || mixed_practice?
   end
 
-  def core_task_steps
-    self.task_steps.preload(:tasked).to_a.select(&:core_group?)
+  def core_task_steps(preload_tasked: false)
```
jpslav: why make it false by default?

karenc: I'm not sure, can make it true by default?

Dantemss: I'm fine with false by default :)

app/routines/calculate_task_plan_stats.rb
```diff
@@ -97,7 +99,11 @@ def get_page_for_tasked_exercise(tasked_exercise)
   end
 
   def group_tasked_exercises_by_pages(tasked_exercises)
-    tasked_exercises.group_by{ |te| get_page_for_tasked_exercise(te) }
+    tasked_exercises.group_by do |te|
+      te.exercise.tags.collect do |tag|
```
Dantemss: Each Content::Models::Exercise now has 1 page so you can call `te.exercise.page` (no need to preload the tags anymore)

app/subsystems/tasks/models/task.rb
kevinburleigh75: Having flags like this forces all users of this class to understand what's happening under the hood.  If we really need this, we should rethink where/how certain processing is being done.

Issue #360, opened 2015-06-08
Combine user profile-getting routines into one glorious construction of code so great the gods themselves will kneel before your terminal


joemsak:

joemsak:
https://github.com/openstax/tutor-server/pull/308#discussion_r31826766

joemsak:
Closed by PR#617

PR #566, opened 2015-08-10
User profile routine mashups

joemsak:
Addressing #360 

* Removes `UserProfile::GetUserFullNames`
* Merges notion of getting profiles with entity users into `SearchProfiles`
  * `SearchProfiles` now takes a `search:` argument instead of `search_term:`
  * `search:` can be a `String`, an `EntityUser`, or a collection of `EntityUser`s
* Cautions against not having a `params[:page]` removed as `paginate` handles this elegantly already


app/subsystems/user_profile/search_profiles.rb
```diff
@@ -29,5 +21,31 @@ def exec(search_term:, order: 'last_name', page: nil, per_page: 30)
         end
       )
     end
+
+    private
+    def find_profiles(search)
+      case search
+      when String
+        profiles_by_search_term(search)
+      else
```
jpslav: Can you add some checks somewhere to verify that the search really is a single or array of Entity::User?  We could pass in a course and line 44's `collect(&:id)` would chug along without raising a fuss.

app/routines/get_teacher_names.rb
```diff
@@ -1,20 +1,20 @@
 class GetTeacherNames
-  lev_routine
+  lev_routine express_output: :teacher_names
```
jpslav: Can you add a spec for this routine?  I think it is only covered in a dashboard spec with only one name, at least want to know it works with multiple teachers as well.

jpslav:
Looking good other than two comments

joemsak:
@jpslav Updated

app/subsystems/user_profile/search_profiles.rb
```diff
@@ -29,5 +21,35 @@ def exec(search_term:, order: 'last_name', page: nil, per_page: 30)
...
+                           (lower(account.last_name).like term)}
+    end
+
+    def profiles_by_entity_users(users)
+      if (users = [users].flatten).all? { |u| u.respond_to?(:profile) &&
+                                                u.respond_to?(:username) }
+        profiles_query.where{entity_user_id.in users.collect(&:id)}
+      else
+        raise IllegalArgument, "Search must be a String or collection of EntityUsers"
+      end
```
jpslav: Don't need to talk about strings here.  Also, just check that users are `Entity::User` instead of checking that it has two methods? (which some other object could have).

```ruby
users = [users].flatten
raise IllegalArgument, "Expected input to be 'Entity::User's" \   # if this is triggered it is a coding error
  if users.any?{|u| !u.is_a?(Entity::User)}
...
```

joemsak: I don't condone of checking the class of an object, I prefer checking it responds as the correct type of thing, in case the class of the thing should ever change

Also I mention string because it's still true: send me a string or a collection

joemsak: there are not other types of objects that both have a profile and username method, except entity users

i'd love a better more stable method to check but as such entity models aren't setup to be much more than connector records with little to no data/behavior

jpslav: Sometimes all you need is for something to quack like a duck, but here we actually want `users` to be a duck.  Only if `users` is a duck does line 45 do what we intend. 

See the [verify_and_return](https://github.com/openstax/tutor-server/blob/master/app/subsystems/content/wrapper.rb#L35-L63) code that has been added in with Dante and Kevin's content work.  These methods support more explicit type checking on input and output arguments.  This is exactly what we want to do here too, so please extract this code into an includable module, and use it here like:

```ruby
def profiles_by_entity_users(users)
  users = verify_and_return([users].flatten, klass: Entity::User)
  profiles_query.where{entity_user_id.in users.collect(&:id)}
end
```

In the extracted library, the default `error` should be [`TypeError`](http://ruby-doc.org/core-2.2.0/TypeError.html) instead of the `::Content::StrategyError` used now.

Dantemss: Oh maybe we should just use TypeError everywhere...

joemsak: I just abstracted it to `VerifiedCollections` with `VerifiedCollections::TypeError`

lib/verified_collections.rb
```diff
@@ -0,0 +1,36 @@
...
+        raise(error,
+              "Tested argument was of class '#{obj.class}' instead of the expected '#{klass}'.",
+              caller) unless obj.is_a? klass
+      end
+
+      object
+    end
+  end
+
+  class TypeError < StandardError
```
jpslav: Already exists in Ruby http://ruby-doc.org/core-2.1.3/TypeError.html

lib/verified_collections.rb
```diff
@@ -0,0 +1,36 @@
+module VerifiedCollections
```
jpslav: the input to `verify_and_return` doesn't need to be a collection, so a different name would probably be better, e.g. `TypeVerification` or `TypeVerifiers`

joemsak:
@jpslav Updated according to latest comments

Dantemss:
Absorbed into #617 and merged into master.

PR #567, opened 2015-08-11
return all BL clue info to client

jpslav:


jpslav:
@nathanstitt please review

cc @kevinburleigh75

PR #577, opened 2015-08-12
Version merging

Dantemss:
# Done
- Made tags ecosystem-specific
- Admin edit tag screen now sees a list of all tags
- Make Content::Map wrapper that can map an array of exercises from any ecosystem to `{ pages in a target ecosystem => exercises in that page }`
- Routine to get the map for a course
- Pages now hold pool id's instead of the other way around
- Learning guide now uses the ecosystems map


Dantemss:
Merge https://github.com/openstax/tutor-server/pull/575 first.

jpslav:
Looking good so far.

app/routines/fetch_and_import_book_and_create_ecosystem.rb
```diff
@@ -11,7 +11,7 @@ class FetchAndImportBookAndCreateEcosystem
   def exec(id:)
     cnx_book = OpenStax::Cnx::V1.book(id: id)
     outputs[:ecosystem] = Content::Ecosystem.create!(
-      title: "#{cnx_book.title} v#{cnx_book.version}"
+      title: "#{cnx_book.title} (#{cnx_book.uuid}@#{cnx_book.version})"
```
jpslav: Do we need the "at <timestamp>" or some other differentiator when the same version is reimported?

app/routines/get_course_ecosystems_map.rb
```diff
@@ -0,0 +1,25 @@
+class GetCourseEcosystemsMap
+  lev_routine express_output: :ecosystems_map
+
+  protected
+
+  def exec(course:, ecosystem_strategy_class: ::Content::Strategies::Direct::Ecosystem,
+                    map_strategy_class: ::Content::Strategies::Generated::Map)
+    # The first ecosystem is the latest
+    to_content_ecosystem = course.ecosystems.first
+    to_ecosystem_strategy = ecosystem_strategy_class.new(to_content_ecosystem)
```
jpslav: toss up for @Dantemss or @kevinburleigh75 -- why do strategies take `Content::Ecosystem`s in addition to `Content::Ecosystem`s taking strategies?  I would expect only one of these things to happen

app/subsystems/content/list_ecosystems.rb
```diff
@@ -4,17 +4,17 @@ class Content::ListEcosystems
   protected
 
   def exec
-    ecosystems = Content::Ecosystem.all.sort_by{ |es| es.books.first.title.downcase }
+    ecosystems = Content::Ecosystem.all
     outputs[:ecosystems] = ecosystems.collect do |ecosystem|
```
jpslav: We don't have to change this now, but I imagine these Hashies could be swapped out for an array of `Content::Ecosystem` right? (no longer need the hashie since we have POROs that have the same fields.

app/subsystems/content/map.rb
```diff
@@ -0,0 +1,62 @@
...
+  class Map
+
+    include Wrapper
+
+    class << self
+      # Create a new map or return nil if it cannot be created
+      def create(from_ecosystems:, to_ecosystem:,
+                 strategy_class: ::Content::Strategies::Generated::Map)
+        from_arr = verify_and_return [from_ecosystems].flatten.compact,
+                                     klass: ::Content::Ecosystem, error: ArgumentError
```
jpslav: Dante you'll be happy to know (I think) that I have asked Joe to pull out `verify_and_return` into its own module and make the default error be Ruby's standard `TypeError`, which looks like what you're getting at here with `ArgumentError`.

app/subsystems/content/strategies/generated/map.rb
```diff
@@ -0,0 +1,97 @@
...
+            ecosystem_exercise = exercise_map[content_exercise.id]
+            ecosystem_pages = content_pages.collect{ |cp| page_map[cp.id] }.compact
+            ecosystem_pages.each do |ecosystem_page|
+              hash[ecosystem_page] ||= []
+              hash[ecosystem_page] << ecosystem_exercise
+            end
+          end
+        end
+
+        def valid?
```
jpslav: neat

app/subsystems/course_content/add_ecosystem_to_course.rb
```diff
@@ -14,5 +17,23 @@ def exec(course:, ecosystem:, remove_other_ecosystems: false)
...
+
+    # Create a mapping from the old course ecosystems to the new one and validate it
+    from_ecosystems = course.course_ecosystems.collect do |course_ecosystem|
+      strategy = ecosystem_strategy_class.new(course_ecosystem.ecosystem)
+      ::Content::Ecosystem.new(strategy: strategy)
+    end
+    to_ecosystem = case ecosystem
+    when ::Content::Ecosystem
+      ecosystem
+    else
```
jpslav: Can we test the type of `ecosystem` here instead of just `else` to make sure it is what we expect (I assume a strategy)?

app/views/admin/contents/index.html.erb
```diff
@@ -22,5 +22,5 @@
   </tbody>
 </table>
 
-<%= link_to 'Import a Book into a new Ecosystem', import_admin_ecosystems_path,
+<%= link_to 'Import a Book into a new Ecosystem', import_admin_contents_path,
```
jpslav: this will rebase out, Karen fixed it separately

app/subsystems/content/wrapper.rb
```diff
@@ -17,24 +17,24 @@ def _strategy
 
     # Wrappers are equal if the strategies are equal
     def ==(other)
-      self.class == other.class && _strategy == other._strategy
+      self.class == other.class && @strategy == other._strategy
```
jpslav: Later @kevinburleigh75 and I might like to hear why two wrapped objects need their strategies to be equal for them to be equal (and why these methods aren't just in the POROs themselves instead of being inherited)

app/routines/fetch_and_import_book_and_create_ecosystem.rb
Dantemss: I think just add the timestamp to the admin view if needed...
But really why would you want to reimport the same version? :D

app/routines/get_course_ecosystems_map.rb
Dantemss: Not sure what you mean... strategies can only take `Content::Models::Ecosystem`

app/subsystems/content/list_ecosystems.rb
Dantemss: That is true.

app/subsystems/content/map.rb
Dantemss: I'll leave the error types up to KB

app/subsystems/course_content/add_ecosystem_to_course.rb
Dantemss: Works for both `Content::Models::Ecosystem` and the strategy and honestly I'm ok with it blowing up if you pass something weird.

app/views/admin/contents/index.html.erb
Dantemss: ok

app/subsystems/content/wrapper.rb
Dantemss: Uhm I think the strategy should determine when they are equal? idk...

Dantemss:
So I looked at the performance report and it seems that it wouldn't care about ecosystems. So this PR is ready for testing. @jpslav @kevinburleigh75

app/routines/fetch_and_import_book_and_create_ecosystem.rb
jpslav: Importing at different time could give different exercises.

app/routines/get_course_ecosystems_map.rb
jpslav: Oh I guess it was unclear to me that `from_content_ecosystem` was a `Content::Models::Ecosystem` (I read it as a "from `Content::Ecosystem`).  Ok.

app/routines/fetch_and_import_book_and_create_ecosystem.rb
kevinburleigh75: I agree that we should be prepared to import the same book version multiple times to fix issues with exercises.  Certainly the admin will need to be able to clearly tell these imports apart, and ultimately we should probably allow a name to be assigned (instead of just a timestamp) because the exercise manifests might be the differentiator and they don't have meaningful names.

But for now I think we're mostly ok.

app/subsystems/content/import_book.rb
```diff
@@ -29,21 +29,30 @@ def exec(cnx_book:, ecosystem:)
...
 
     if objective_page_tags.empty?
       outputs[:exercises] = []
     else
       outputs[:exercises] = nil
-      run(:import_exercises, page: page_block, query_hash: query_hash)
+      run(:import_exercises, ecosystem: ecosystem, page: page_block, query_hash: query_hash)
     end
 
     # Need a double reload here for it to work for some reason
```
kevinburleigh75: Uh...scary?

app/subsystems/content/wrapper.rb
kevinburleigh75: It's just not clear from this code that we're doing anything except comparing whether or not they have the same strategy is all...

app/subsystems/course_content/add_ecosystem_to_course.rb
```diff
@@ -14,5 +17,23 @@ def exec(course:, ecosystem:, remove_other_ecosystems: false)
...
+      strategy = ecosystem_strategy_class.new(ecosystem)
+      ::Content::Ecosystem.new(strategy: strategy)
+    end
+
```
kevinburleigh75: Is there a specific case where we pass something that's not a `::Content::Ecosystem` into this routine?  If not, seems like we should either assume it's an `Ecosystem` or flag the error early on...

app/subsystems/content/wrapper.rb
Dantemss: Oh, the strategies I make override the `==`, `eql?` and `hash` method to provide more meaningful comparisons (they basically delegate to AR)

app/subsystems/course_content/add_ecosystem_to_course.rb
Dantemss: Some spec somewhere passes a `Content::Models::Ecosystem`... I suppose I could change it.

PR #580, opened 2015-08-13
Adds book_id to Course JSON

jpslav:


jpslav:
@nathanstitt heads up that I merged this

nathanstitt:
Thanks @jpslav.  I guess we could have merged it earlier, there's no reason that it needed to wait on the FE PR (which is taking me forever....)

PR #584, opened 2015-08-14
Update biology to 3.1

nathanstitt:
Don't mind the branch name, I can't count.


nathanstitt:
wip until I hear if physics should also be updated

PR #590, opened 2015-08-17
Extra assignments

karenc:
- Find snap lab notes from a cnx page
- Add Content::Models::Page#snap_labs
- Add Snap Lab Representer
- Add snap_labs to readings API
- Add extra assignment assistant

  The extra assignment assistant builds tasks from snap_lab_ids which are
  in the readings API.  (The format of the snap_lab_ids in the readings
  API is "#{page_id}:#{snap_lab_note_id}".)

  The snap lab notes include a reading and an exercise (at least from the
   two I've seen in Physics).
- Add specs for extra assignment assistant


karenc:
Rebased.

app/representers/api/v1/page_toc_representer.rb
```diff
@@ -45,5 +45,16 @@ class PageTocRepresenter < Roar::Decorator
                description: 'The chapter and page in the book, e.g. [5, 2]'
              }
 
+    collection :snap_labs,
+               getter: -> (*) { snap_labs.collect { |snap_lab| Hashie::Mash.new(snap_lab) } },
+               readable: true,
+               writeable: false,
+               decorator: Api::V1::SnapLabRepresenter,
+               if: lambda { |args| snap_labs.present? },
```
jpslav: I think this is the default? (that the collection won't be output snap_labs is nil)

app/representers/api/v1/page_toc_representer.rb
```diff
@@ -45,5 +45,16 @@ class PageTocRepresenter < Roar::Decorator
                description: 'The chapter and page in the book, e.g. [5, 2]'
              }
 
+    collection :snap_labs,
```
jpslav: Can you remind me why the snap labs are output in the TOC?   And when they are output here, do they include all of the snap lab content too?  Kind of thought snap labs might be accessible by some search like exercises are.

jpslav: And I now see that the `SnapLabRepresenter` doesn't output the content.

app/subsystems/tasks/assistants/extra_assignment_assistant.rb
```diff
@@ -0,0 +1,218 @@
...
+
+  def tasked_reading(reading_fragment:, page:, step:, title: nil)
+    Tasks::Models::TaskedReading.new(task_step: step,
+                                     url: page.url,
+                                     book_location: page.book_location,
+                                     title: title,
+                                     content: reading_fragment.to_html)
+  end
+
+  def tasked_exercise_choice(exercise_choice_fragment:, step:, title: nil)
```
jpslav: to dry up these methods copied from the ireading assistant, will likely want to pull this code out and have it be used from both places.  @Dantemss might be able to provide some suggestions here with an eye towards the future refactoring of the Task SS.

app/representers/api/v1/page_toc_representer.rb
karenc: But it will output if snap_labs is an empty array, I think.

app/representers/api/v1/page_toc_representer.rb
karenc: It is because when teachers choose a snap lab for the extra assignment, the interface is like choosing chapters for iReading, so I thought I'll just add snap labs to the page TOC.

app/subsystems/tasks/assistants/extra_assignment_assistant.rb
karenc: That will be very helpful.  I can move these methods to a module (`FragmentTaskedHelper`?) and then do `include <module-name>` in both classes?

Dantemss: Hm sorry I missed this. Yeah I believe that would do it.

PR #597, opened 2015-08-18
Sort page stats by page.book_location; Sort exercises by average step number

nathanstitt:
WIP because I'm not sure this is the correct order

### DIFF
--- a/app/representers/api/v1/tasks/stats/exercise_representer.rb
+++ b/app/representers/api/v1/tasks/stats/exercise_representer.rb
@@ -20,6 +20,11 @@ module Api::V1
                    readable: true,
                    decorator: Api::V1::Tasks::Stats::AnswerRepresenter
 
+        property :average_step_number,
+                 type: Float,
+                 writeable: false,
+                 readable: true
+
       end
     end
   end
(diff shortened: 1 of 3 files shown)


nathanstitt:
@pandafulmanda I don't think this worked correctly for you? Should we try another sort or just close it for now until we can pick it back up?

Dantemss:
I think this could work (I know you can sort ruby hashes, but I feel bad doing that):

```rb
pages = page_hash.keys.sort_by{ |page| page.book_location }
pages.collect do |page|
  tasked_exercise = page_hash[page]
  generate_page_stats(page, tasked_exercises)
end
```

Dantemss:
You prob just need to sort by `book_location` instead of by `number` (number resets for every chapter)

Dantemss:
Ah I suppose it's fine, sort_by will give you an array of arrays... just use `book_location` instead of `number`.

nathanstitt:
@pandafulmanda Do you remember what the order should be here?

Dantemss:
@nathanstitt @pandafulmanda ready

pandafulmanda:
:+1:  yayayayayaya

PR #598, opened 2015-08-18
Send exercise URL minus protocol as question_id to Biglearn

Dantemss:
This and https://github.com/openstax/exchange/pull/39 are needed to fix the CLUE's.


app/routines/get_ecosystem_exercises_from_biglearn.rb
```diff
@@ -7,13 +7,14 @@ class GetEcosystemExercisesFromBiglearn
...
 
-    numbers = OpenStax::Biglearn::V1.get_projection_exercises(
+    urls = OpenStax::Biglearn::V1.get_projection_exercises(
       role:              role,
       pools:             biglearn_pools,
       count:             count,
       difficulty:        difficulty,
       allow_repetitions: allow_repetitions
     )
+    numbers = urls.collect{ |url| url.chomp('/').split('/').last.split('@').first }
```
nathanstitt: I'd personally reach for a regex over the chomp, split, split, something like: `parts = url=~/(\d)@(\d)[\/]$/`
(because that's so much clearer)

PR #601, opened 2015-08-18
Fixes bug where duplicate exercises would only be counted once in the LG

Dantemss:
Fixes intermittent get_teacher_guide spec failures

### DIFF
--- a/app/subsystems/content/strategies/generated/map.rb
+++ b/app/subsystems/content/strategies/generated/map.rb
@@ -30,16 +30,13 @@ module Content
           @to_ecosystem = to_ecosystem
         end
 
-        def group_exercises_by_pages(exercises:)
+        def map_exercises_to_pages(exercises:)
           all_pages = @to_ecosystem.pages
           page_map = all_pages.each_with_object({}) do |page, hash|
             hash[page.id] = page
           end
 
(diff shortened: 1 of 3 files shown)


app/subsystems/content/strategies/generated/map.rb
```diff
@@ -53,42 +50,31 @@ def group_exercises_by_pages(exercises:)
...
-
-            ecosystem_exercise = exercise_map[content_exercise.id]
             ecosystem_pages = content_pages.collect{ |cp| page_map[cp.id] }.compact
-            ecosystem_pages.each do |ecosystem_page|
-              hash[ecosystem_page] ||= []
-              hash[ecosystem_page] << ecosystem_exercise
-            end
+
+            # We only allow each exercise to map to 1 page
+            hash[content_exercise.id] = ecosystem_pages.size == 1 ? ecosystem_pages.first : nil
```
jpslav: Why don't we blow up if more than 1 page?

Dantemss: We blow up, just not here... (AddEcosystemToCourse blows up)

Dantemss: `valid?` will return false, which will cause `create!` to blow up

PR #606, opened 2015-08-19
styled terms, fixed error pages

jpslav:
Since we may need to have many of our students sign their own terms of use in the fall, this PR adds styling to what was the totally unstyled terms agreement pages (they were pretty bad looking).

I stole the styling from the home page.  Since the home page doesn't have a state where a user is signed in, I couldn't easily steal the username dropdown menu from the FE.  So when a user is logged in and on terms or error pages, their header bar looks like the one on the home page but "Login" instead says "Log Out" (the screenshots below have "Sign out!" which isn't there any more).

image

The "I Agree" button is disabled until the user checks the box

image

If there are multiple terms, they are presented serially.  Likely that for the fall for certain districts we'll combine them into one site terms to sign.

image

I went ahead and cleaned up the placeholder "/terms" page -- but note that we don't yet have a link to get there.  @Fredasaurus et al can decide where to put such a link (I imagine in the footer) and how to style it.

image

Error pages also now have the basic styling from above.  Also, if the user reaches an error page that doesn't have an error reference number, we no longer make any mention of it.

image

### DIFF
--- a/app/assets/javascripts/ui.js.coffee
+++ b/app/assets/javascripts/ui.js.coffee
@@ -22,4 +22,25 @@ Tutor.Ui = {
         return '<span class="' + cls + '">' + match + '</span>'
     )
 
+  disableButton: (selector) ->
+    $(selector).attr('disabled', 'disabled')
+    $(selector).addClass('ui-state-disabled ui-button-disabled')
+    $(selector).attr('aria-disabled', true)
+
+  enableButton: (selector) ->
+    $(selector).removeAttr('disabled')
+    $(selector).removeAttr('aria-disabled')
+    $(selector).removeClass('ui-state-disabled ui-button-disabled')
+    $(selector).button()
+
+  enableOnChecked: (targetSelector, sourceSelector) ->
+    $(document).ready =>
+      @disableButton(targetSelector)
+
+    $(sourceSelector).on 'click', =>
+      if $(sourceSelector).is(':checked')
+        @enableButton(targetSelector)
+      else
+        @disableButton(targetSelector)
+
 }
--- a/app/assets/stylesheets/global.scss
+++ b/app/assets/stylesheets/global.scss
@@ -1,0 +1,332 @@
+
+
+
+@mixin tutor-button(){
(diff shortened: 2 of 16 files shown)


app/assets/stylesheets/global.scss
```diff
@@ -0,0 +1,332 @@
...
+  padding: 18px 40px;
+  font-size: 28px;
+  font-weight: 600;
+  border: none;
+  background: #f47642;
+  -webkit-transition: all .15s ease-in-out;
+  -moz-transition: all .15s ease-in-out;
+  -o-transition: all .15s ease-in-out;
+  transition: all .15s ease-in-out;
+}
```
nathanstitt: There's no reason to use the vendor prefixes for `transition`.  It's been fully supported for several years. http://caniuse.com/#feat=css-transitions

app/assets/javascripts/ui.js.coffee
```diff
@@ -22,4 +22,25 @@ Tutor.Ui = {
...
+    $(selector).attr('disabled', 'disabled')
+    $(selector).addClass('ui-state-disabled ui-button-disabled')
+    $(selector).attr('aria-disabled', true)
+
+  enableButton: (selector) ->
+    $(selector).removeAttr('disabled')
+    $(selector).removeAttr('aria-disabled')
+    $(selector).removeClass('ui-state-disabled ui-button-disabled')
+    $(selector).button()
+
```
nathanstitt: Since these are all using the same selector, you could chain them.

```javascript
 $(selector)
  .removeAttr('disabled')
  .removeAttr('aria-disabled')
  .removeClass('ui-state-disabled ui-button-disabled')
  .button()
```

nathanstitt:
Looks good.  These are pretty small and I'm guessing the styles might have came from elsewhere.

jpslav:
Good comments @nathanstitt -- I'm going to go ahead and merge and we can clean up later.

Fredasaurus:
Looks good. We can finesse later if need be.

PR #608, opened 2015-08-19
Refactor Performance Report code to fix doomsday bugs

Dantemss:
- Order of tasks should now match the table headings
- Tasks ordered by due date
- Students ordered by last name
- Average now divides by all problems in the homework, rather than just completed problems
- Missing tasks (period change etc) are now sent as nils
- `PerformanceReportExport` now returns the uploaded file path instead of the tempfile path
- Cleaned up Performance Report specs that left temp files behind


jpslav:
Looks good to me.

app/representers/api/v1/performance_report_representer.rb
```diff
@@ -3,6 +3,20 @@ class PerformanceReportRepresenter < Roar::Decorator
...
+        nil
+      end
+
+      def to_json(*args)
+        'null'
+      end
+
+    end
+
     class StudentData < Roar::Decorator
```
nathanstitt: @Dantemss I think maybe you mis-understood the "null" I suggested you send the FE.  I mean that the property for the student would be missing, not a string "null". This works though, so if you're happy with it we can roll this way.

Dantemss: It comes out as a javascript/json null, I believe (no quotes).

nathanstitt: Oh I see.  Cool, I thought you were returning the string "null".  Thanks for explaining :)

PR #609, opened 2015-08-20
Pull in changes from demo import

nathanstitt:
The demo import has a better setup for the assignments in physics and a new `demo:show` task that prints all the students assignments.

### DIFF
--- a/lib/tasks/demo/demo_base.rb
+++ b/lib/tasks/demo/demo_base.rb
@@ -357,23 +357,26 @@ class DemoBase
     puts "#{message}\n" if @print_logs
   end
 
-  def print_task(task:)
-    types = task.task_steps.collect do |step|
-      step_code = case step.tasked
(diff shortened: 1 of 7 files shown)


lib/tasks/demo/demo_base.rb
```diff
@@ -386,10 +389,10 @@ def print_task(task:)
         'o'
       end
 
-      "#{group_code}#{step.id}#{step_code}"
+      "#{group_code}#{step.id}#{step_code(step)}"
     end
-
-    "Task #{task.id} / #{task.task_type} / #{types.join(' ')}"
+    codes = task.task_steps.collect{ |step| step_code(step) }
+    "Task #{task.id} / #{task.task_type}\n#{codes.join(', ')}\n#{types.join(' ')}"
```
nathanstitt: By printing the codes this way, it's easy to copy & paste them into the demo yaml steps if they need to be updated.

PR #613, opened 2015-08-21
moved course asst creation from demo to CreateCourse

jpslav:
The demo script was doing business logic not present in the normal flow, so courses created outside of the demo script didn't have assistants linked to them.

### DIFF
--- a/app/controllers/api/v1/task_plans_controller.rb
+++ b/app/controllers/api/v1/task_plans_controller.rb
@@ -69,8 +69,9 @@ class Api::V1::TaskPlansController < Api::V1::ApiController
         consume!(task_plan, represent_with: Api::V1::TaskPlanRepresenter)
         task_plan.assistant = Tasks::GetAssistant[course: course, task_plan: task_plan]
 
-        raise(IllegalState, "No assistant for task plan of type #{task_plan.type}") \
-          if task_plan.assistant.nil?
+        raise(IllegalState,
+          "No assistant for task plan of type #{task_plan.type}"
+        ) if task_plan.assistant.nil?
 
         OSU::AccessPolicy.require_action_allowed!(:create, current_api_user, task_plan)
         uuid = distribute_or_update_tasks(task_plan)
--- a/app/controllers/api/v1/task_plans_controller.rb
+++ b/app/controllers/api/v1/task_plans_controller.rb
@@ -68,7 +68,10 @@ class Api::V1::TaskPlansController < Api::V1::ApiController
       Tasks::Models::TaskPlan.transaction do
         consume!(task_plan, represent_with: Api::V1::TaskPlanRepresenter)
         task_plan.assistant = Tasks::GetAssistant[course: course, task_plan: task_plan]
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/task_plans_controller.rb
```diff
@@ -68,7 +68,10 @@ def create
       Tasks::Models::TaskPlan.transaction do
         consume!(task_plan, represent_with: Api::V1::TaskPlanRepresenter)
         task_plan.assistant = Tasks::GetAssistant[course: course, task_plan: task_plan]
-        return head :unprocessable_entity if task_plan.assistant.nil?
+
+        raise(IllegalState, "No assistant for task plan of type #{task_plan.type}") \
```
jpslav: Changed the 422 error to a more descriptive error condition which will generate notifications and leave a trail

app/controllers/api/v1/task_plans_controller.rb
```diff
@@ -68,7 +68,10 @@ def create
...
+          if task_plan.assistant.nil?
+
         OSU::AccessPolicy.require_action_allowed!(:create, current_api_user, task_plan)
```
nathanstitt: When the if statement is at the end of a long statement, I like to write it as:
```ruby
     raise(IllegalState, 
       "No assistant for task plan of type #{task_plan.type}"
     ) if task_plan.assistant.nil?
```

Of course I'm only saying this since I couldn't find anything else to point out. :grin:

jpslav: I like it.

spec/support/setup_performance_report_data.rb
```diff
@@ -4,10 +4,8 @@ class SetupPerformanceReportData
   protected
   def exec(course:, teacher:, students: [], ecosystem:)
     students = [students].flatten
-    reading_assistant = FactoryGirl.create(:tasks_assistant,
-      code_class_name: 'Tasks::Assistants::IReadingAssistant')
-    homework_assistant = FactoryGirl.create :tasks_assistant,
-      code_class_name: 'Tasks::Assistants::HomeworkAssistant'
+    reading_assistant = get_assistant(course: course, task_plan_type: 'reading')
+    homework_assistant = get_assistant(course: course, task_plan_type: 'homework')
```
nathanstitt: I see this same pattern repeated 3 times in the PR.  That be moved to the course as a `get_assistant` method, or even an extension to the `course_assistants` association.  

You could create a module say `CourseAssistantGetter` with a single method named `for` as described in http://guides.rubyonrails.org/association_basics.html#association-extensions

That way you could just call `course.course_assistants.for(:reading)` here.

As always, feel free to ignore. It's only one line that is repeated so is probably not worth abstracting out any further.

jpslav: I think when we refactor the Task SS this stuff is likely to go away.   Really it could be the `Tasks::GetAssistant` routine (which exists), but the signature is slightly different.

Dantemss:
I agree. Remove this when we refactor.

PR #619, opened 2015-08-24
Expire sessions after an hour of inactivity

nathanstitt:
According to the tests, 1 hour should be used but that might be too tight of a window?

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,3 +1,5 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_tutor_session'
+Rails.application.config.session_store :cookie_store,
+                                       key: '_tutor_session',
+                                       expire_after: 1.hour


jpslav:
Note that expiring a tutor cookie doesn't really log you out (b/c you are still logged in on accounts, and clicking "sign in" on tutor will log you in without requiring a password).  on the flip side, we might not want an automatic logout on accounts because that would have to apply to all OX products (might not make sense on CNX)

Dantemss:
I believe we should make all OX products have around the same logout time.

Dantemss:
Maybe ask CNX what they use...

Dantemss:
Also even if accounts logs them out after 30 min, they will still be logged into CNX if CNX wants to keep them logged on for longer.

jpslav:
Good point Dante.

nathanstitt:
Actually if we have accounts set to 1 hour to match Tutor, accounts will always log out first because the accounts cookie will be set only once (when they login).  Tutor's cookie will progressively increment the 1 hour limit every time a request is made.  It'll only expire 1 hour after the last request.

I think that'd be the best behavior though.  What we wouldn't want is the situation that JP outlined in the first comment, where Tutor logs a user out and they go to login and just get redirect right back due to accounts not have logged them out.

nathanstitt:
Closing this since we've decided to forgo auto-logout until v2

PR #622, opened 2015-08-25
Sort performance report students by case-insensitive last name

joemsak:
* Needs Front-end to remove their sorting code, as their sort is overriding the BE sort. Sorting should only have to happen in one place. It is now correctly sorting by last name downcased to avoid lowercase names being pushed past uppercase 'Z' names


Dantemss:
Seems good to me.

PR #624, opened 2015-08-26
notify devs and log for background job exceptions

jpslav:
Uses http://edgeguides.rubyonrails.org/active_job_basics.html#exceptions to intercept exceptions so we can send dev notifications and log them.  

Seems like we want something like this -- y'all agree?  Thoughts?

1. Docs talks about doing this in the job concrete class, but should work globally in `ActiveJob::Base`, don't you think?
2. Does doing it globally prevent us from calling it with more specific logic from concrete job classes?
3. For different kinds of exceptions, we could potentially retry the job.

### DIFF
--- a/config/initializers/active_job.rb
+++ b/config/initializers/active_job.rb
@@ -1,0 +1,22 @@
+module ActiveJob
+  class Base
+
+    rescue_from(Exception) do |exception|
+      @error_id = "%06d" % SecureRandom.random_number(10**6)
+
+      ExceptionNotifier.notify_exception(
+        exception,
+        data: {
+          error_id: @error_id,
+          message: "A background job exception occurred"
+        }
+      )
+
+      Rails.logger.error {
+        "A background job exception occurred: #{exception.class.name} [#{@error_id}] " +
+        "#{exception.message}\n\n#{exception.backtrace.join("\n")}"
+      }
+    end
+
+  end
+end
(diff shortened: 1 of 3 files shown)


config/initializers/active_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      )
+
+      Rails.logger.error {
+        "A background job exception occurred: #{exception.class.name} [#{@error_id}] " +
+        "#{exception.message}\n\n#{exception.backtrace.join("\n")}"
+      }
+    end
+
+  end
+end
```
nathanstitt: I wonder if the same rational applies to rescuing `Exception` (don't do it) here?  Maybe it should be `rescue_from(StandardError)` https://robots.thoughtbot.com/rescue-standarderror-not-exception

joemsak: :+1: @nathanstitt

jpslav: I think if anything exceptional happens in a background job, we want to hear about it (death to silent failures)

Dantemss: I agree with JP, but you definitely want to reraise the exceptions.

nathanstitt: Yeah, but if you rescue Exception, you're also rescuing from: `Interrupt`, prevents the using CTRL-C to stop process, `SignalException` - prevents the program from responding to sig-quit.  It will be unkillable except by kill -9.  If we really want to rescue Exception then we need to re-raise those at a minimum.  There's probably others that I'm not familiar with as well.

Dantemss:
I would say add the lev update and the reraise and this PR is good.

PR #630, opened 2015-08-28
Improve exception emails

joemsak:
https://www.pivotaltracker.com/story/show/102044672

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -88,10 +88,9 @@ Rails.application.configure do
   config.active_record.dump_schema_after_migration = false
 
   # Send email to developers when users encounter exceptions
-  deployed_to = Rails.application.secrets.server_name || '%SERVER_NAME%'
   config.middleware.use ExceptionNotification::Rack,
     :email => {
-      :email_prefix => "[Tutor] (#{deployed_to}) ",
+      :email_prefix => "[Tutor] (#{DeployUtils.server_nickname}) ",
       :sender_address => %{"OpenStax Tutor" <noreply@openstax.org>},
       :exception_recipients => %w{tutor-notifications@openstax.org}
     }
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -88,10 +88,10 @@ Rails.application.configure do
   config.active_record.dump_schema_after_migration = false
 
   # Send email to developers when users encounter exceptions
-  env = Rails.application.secrets.mail_site_url.match(/tutor-(\w+)\./)[1]
+  deployed_to = Rails.application.secrets.server_name
   config.middleware.use ExceptionNotification::Rack,
     :email => {
-      :email_prefix => "[Tutor] (#{env.upcase}) ",
+      :email_prefix => "[Tutor] (#{deployed_to}) ",
       :sender_address => %{"OpenStax Tutor" <noreply@openstax.org>},
       :exception_recipients => %w{tutor-notifications@openstax.org}
     }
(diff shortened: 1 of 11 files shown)


config/routes.rb
```diff
@@ -150,4 +150,5 @@
 
   match '/*other', via: [:get, :post, :put, :patch, :delete], to: 'webview#index'
 
+  get 'bad_action' => 'test#bad_action'
```
jpslav: I'd prefer to not have test routes in the real routes file.  Can you try something like this instead? http://makandracards.com/makandra/18761-rails-3-4-how-to-add-routes-for-specs-only

config/initializers/controllers.rb
```diff
@@ -22,23 +22,24 @@ def set_app_date_header
...
+                           AbstractController::ActionNotFound
+                        [:not_found, false]
+                      when ActionController::InvalidAuthenticityToken,
+                           Apipie::ParamMissing
+                        [:unprocessable_entity, false]
+                       when ActionView::MissingTemplate
+                        [:bad_request, false]
+                      else
+                        [:internal_server_error, true]
+                      end
```
jpslav: Please no indentation changes for indentation's sake alone

config/environments/production.rb
```diff
@@ -88,9 +88,10 @@
   config.active_record.dump_schema_after_migration = false
 
   # Send email to developers when users encounter exceptions
+  env = Rails.application.secrets.mail_site_url.match(/tutor-(\w+)\./)[1]
```
jpslav: This will not work on production, b/c the site on production is not `tutor-production.openstax.org` but instead just `tutor.openstax.org`.  I think this will result in a nil error when `upcase` called on no match below, or at the very least `[Tutor] ()`.  You could do `|| 'production'` at the end here but that might not be as safe in case there's some random environment that doesn't match this pattern.

Here are some mappings I see, with the latter ones for some random dev or test deployments

`tutor.openstax.org` --> `[Tutor] (production)`
`tutor-kev.openstax.org` --> `[Tutor] (kev)`
`blah-blah.openstax.org` --> `[Tutor] (blah-blah)`
`something-else.com` --> `[Tutor] (something-else.com)`

Three other points:
1. Can we use a variable other than `env`?  When I see that I think `Rails.env` and I think Rails environments.  `deployment` or `deployed_to` are candidates.  
2. Can we put this matcher in some method elsewhere and just call from here, e.g. `Rails.deployment` which returns the strings in parens above.
3. Don't think we need to upcase it -- downcase it if you want something standard, that will at least make it shorter in email clients so we can see more of the subject.

jpslav: Also if we stick this code in a helper like `Rails.deployment`, we can spec it independently for the different cases, make sure we don't get nils, etc.

jpslav:
So that we can test exception emails on various environments, can you add to the admin console the ability to generate exceptions on demand, e.g.:

https://github.com/lml/ost/blob/master/app/controllers/admin_controller.rb#L19-L49
https://github.com/lml/ost/blob/master/config/routes.rb#L73-L80
https://github.com/lml/ost/blob/master/app/views/admin/index.html.erb#L15-L25

config/routes.rb
joemsak: Cool, ya, we agree about this, I didn't know I could do that

config/initializers/controllers.rb
joemsak: A reluctant :+1: ; - P

jpslav:
Looks great.  Can merge on spec pass.

PR #633, opened 2015-08-31
Clear lev jobs after 1 week

joemsak:
https://www.pivotaltracker.com/story/show/102045172


jpslav:
Looks good -- will be backed up by my work of persisting final job status in DB until associated objects deleted (so won't lose anything important by cleaning up redis on this schedule).  Gonna hold off merging for a bit though until that other work in and until we have our gameplan for merges going forward.

PR #634, opened 2015-08-31
loading using the async tag prevents some dom events from being fired

pandafulmanda:
needed to work with openstax/tutor-js#708, but not ready to merge yet, since FE may change the way we solve the problem


pandafulmanda:
closing in favor of solving this on the FE

Dantemss:
Would this still be useful or delete?

PR #636, opened 2015-08-31
Fix missing TEKS and HW builder API performance

Dantemss:
- TEKS are now shown in the homework builder
- Improved homework builder API performance

### DIFF
--- a/app/representers/api/v1/exercise_representer.rb
+++ b/app/representers/api/v1/exercise_representer.rb
@@ -31,7 +31,9 @@ module Api::V1
     collection :tags,
                readable: true,
                writeable: false,
-               getter: -> (*) { tags.select { |tag| tag.visible? } },
+               getter: -> (*) {
+                 (tags + tags.flat_map(&:teks_tags)).select{ |tag| tag.visible? }.uniq
+               },
                decorator: TagRepresenter,
                schema_info: { required: true,
                               description: 'Tags for this exercise' }
--- a/lib/entity/relation.rb
+++ b/lib/entity/relation.rb
@@ -18,11 +18,9 @@ class Entity
             :quoted_primary_key, :connection, :columns_hash
 
     # The constructor for Entity::Relation only accepts an ActiveRecord::Relation object
-    # The relation is modified and set as readonly
     def initialize(repository)
       # Don't dup the repository relation, since that's a huge performance hit
       @repository = repository
-      @repository.readonly!
     end
 
     # Enumerable API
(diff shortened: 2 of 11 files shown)


app/representers/api/v1/exercise_representer.rb
```diff
@@ -31,7 +31,9 @@ class ExerciseRepresenter < Roar::Decorator
     collection :tags,
                readable: true,
                writeable: false,
-               getter: -> (*) { tags.select { |tag| tag.visible? } },
+               getter: -> (*) {
+                 (tags + tags.flat_map(&:teks_tags)).select{ |tag| tag.visible? }.uniq
```
jpslav: Why are we no longer filtering non-teks tags by visibility?

lib/entity/relation.rb
```diff
@@ -18,9 +18,11 @@ class Relation < Entity
             :quoted_primary_key, :connection, :columns_hash
 
     # The constructor for Entity::Relation only accepts an ActiveRecord::Relation object
-    # The relation is set as readonly
+    # The relation is modified and set as readonly
     def initialize(repository)
-      @repository = repository.readonly(true)
+      # Don't dup the repository relation, since that's a huge performance hit
+      @repository = repository
+      @repository.readonly!
```
jpslav: Since `repository` and `@repository` are just references, doesn't adding the `!` on `@repository` also do it for `repository`?  (if so, I don't understand your comment and why lines 24 and 25 are separate)

jpslav:
Reviewed

lib/entity/relation.rb
Dantemss: `repository.readonly` dups the relation, which apparently gets rid of all the preloading stuff and is a huge pain. So I did `readonly!` instead.

app/representers/api/v1/exercise_representer.rb
Dantemss: We are, the select is outside the parens :)

lib/entity/relation.rb
jpslav: ok, but does wherever this is called end up with a readonly'd relation? and is that ok?

Dantemss:
Commented!

lib/entity/relation.rb
Dantemss: Upon further testing, it seems `readonly!` doesn't actually do anything...

Dantemss: I'll change `to_a` to call `readonly!` on all the records, but the relation itself won't be `readonly`... Although the methods that modify relations are hidden anyway.

Dantemss: Well that didn't work... I'm thinking I'll just remove this `readonly` stuff and rely on the fact that `Entity` and `Entity::Relation` already hide methods that modify the records.

app/representers/api/v1/exercise_representer.rb
jpslav: :-)

jpslav:
@Dantemss after you rebase you can merge

PR #646, opened 2015-09-09
WIP: Admin export student activity

joemsak:
https://www.pivotaltracker.com/story/show/102761524

We need an admin view ability to export a spreadsheet that contains a dump of student interactions from tutor.  One day this might form the basis of a research report (where we use research IDs instead of names), but for now it is going to be used for business purposes only, so names are ok.  

The idea is to make a spreadsheet, per course, where each row is a task step, and we dump as much task step information as we can in those rows.


app/controllers/admin/courses_controller.rb
```diff
@@ -86,6 +86,14 @@ def set_ecosystem
     redirect_to edit_admin_course_path(params[:id], anchor: 'content')
   end
 
+  def export_activity
+    @course = Entity::Course.find(params[:id])
+    formatted_name = @course.name.downcase.gsub(' ', '_')
+    formatted_time = Time.current.strftime('%Y-%m-%d-%H-%M-%S-%L')
+      # year - month - day - 24-hour clock hour - minute - second - millisecond
+    @filepath = "/admin/exports/#{formatted_name}_#{formatted_time}.csv"
```
jpslav: You're probably not there yet, but eventually we'll want the export to be a background job just like the performance report -- the admin interface doesn't need to be slick, just functional.

Dantemss:
@jpslav is this OBE (since you made a different research report)?

PR #647, opened 2015-09-09
Updated Rails patch version to 4.2.4 to fix memory leak and updated ruby to 2.2.3 for speedup

Dantemss:
- Replaced `eager_load` with `preload` whenever many records are loaded to reduce memory spikes.
- Updated Rails patch version to 4.2.4 to fix memory leak when creating records in transactions.
- Updated ruby to 2.2.3 for some speedup.
- Fixed warnings given by ruby 2.2.3 about duplicate keys in hashes and circular references in keyword arguments.


lib/tasks/demo/demo_base.rb
```diff
@@ -186,7 +186,7 @@ def get_assistant(course:, task_plan_type:)
     course.course_assistants.where{tasks_task_plan_type == task_plan_type}.first.assistant
   end
 
-  def get_ecosystem(course: course)
```
jpslav: what was with all of these default arguments?  typo?

PR #652, opened 2015-09-11
handle bad mapping exception with error message

jpslav:


jpslav:
Verified in local dev that I get the flash message when switching to unmappable content.

jpslav:
@Dantemss can you review and merge?

app/controllers/admin/courses_controller.rb
```diff
@@ -61,17 +61,22 @@ def students
   def set_ecosystem
     if params[:ecosystem_id].blank?
       flash[:error] = 'Please select a course ecosystem'
-      return redirect_to edit_admin_course_path(params[:id])
```
Dantemss: Don't you need the redirect to display the error message?

Dantemss: Oh I guess it all goes to the same page anyway so it doesn't matter :)

jpslav: yeah

PR #655, opened 2015-09-11
Updated accounts-rails so that it syncs with Accounts every minute

Dantemss:


jpslav:
Merge when travis succeeds

PR #660, opened 2015-09-14
Added exception cause to logs and exception email

Dantemss:
- Exception causes are now printed in logs and emails, if present
- First line of backtrace added to main exception and cause
- Specs

### DIFF
--- a/config/initializers/controllers.rb
+++ b/config/initializers/controllers.rb
@@ -1,5 +1,38 @@
 ActionController::Base.class_exec do
 
+  # See https://github.com/rack/rack/blob/master/lib/rack/utils.rb#L453 for error names/symbols
+  EXCEPTION_STATUS_MAP = Hash.new(:internal_server_error).merge({
+    'SecurityTransgression' => :forbidden,
+    'ActiveRecord::RecordNotFound' => :not_found,
+    'ActionController::RoutingError' => :not_found,
+    'ActionController::UnknownController' => :not_found,
+    'AbstractController::ActionNotFound' => :not_found,
+    'ActionController::InvalidAuthenticityToken' => :unprocessable_entity,
+    'Apipie::ParamMissing' => :unprocessable_entity,
+    'ActionView::MissingTemplate' => :bad_request,
+  })
+
+  NON_NOTIFYING_EXCEPTIONS = Set.new [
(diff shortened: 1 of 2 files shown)


config/initializers/controllers.rb
```diff
@@ -21,6 +21,17 @@ def set_app_date_header
...
 
+  def get_exception_cause(exception)
+    cause = exception.cause
+    return if cause.nil?
+
+    {
+      class: cause.class.name,
+      message: cause.message,
+      cause: get_exception_cause(cause)
+    }
```
jpslav: Can you add at least the first line of the backtrace (where the causing exception was raised?)

config/initializers/controllers.rb
```diff
@@ -67,7 +78,9 @@ def rescue_from_exception(exception)
         env: request.env,
         data: {
           error_id: @error_id,
+          class: exception.class.name,
           message: exception.message,
+          cause: get_exception_cause(exception),
```
jpslav: Can you add the cause to the log message too?

spec/controllers/application_controller_spec.rb
```diff
@@ -122,4 +122,126 @@
       expect(header_time).to be_within(1.second).of(t)
     end
   end
+
+  context 'rescue_from_exception' do
```
jpslav: Thanks!

config/initializers/controllers.rb
Dantemss: Want this on the main one too?

config/initializers/controllers.rb
Dantemss: What part? Or the whole thing?

config/initializers/controllers.rb
jpslav: The main one has the full backtrace via ExceptionNotifier, I believe.

config/initializers/controllers.rb
jpslav: The whole thing, class, message, first line of backtrace, next level down cause.

config/initializers/controllers.rb
Dantemss: Hm I like consistency though so I think I'll add it there too :)

PR #661, opened 2015-09-15
Fixes the lack of APLO's in Bio

Dantemss:
https://www.pivotaltracker.com/n/projects/1156756

- Fixed APLO detection, so they are included in the tags for pages.
- Fixed request URI too long from Exercises (because of the new APLO's).
- Regenerated cassettes for affected specs


Dantemss:
APLO's are now found, but still don't show up in the assignment builder, because there is no intersection with the page APLO's @kjdav

Issue #638, opened 2015-09-01
Tutor should not display courses/error out for inactive students


Dantemss:

jpslav:
Right now `/api/user/courses` returns courses that students have dropped (where the Student has a populated `inactive_at`).  This is causing issues because students are trying to access inactive courses from their dashboard, and since they have no active role they are getting errors.  While the long term solution might be to report all courses for a user and specify which are dropped/inactive, for right now let's just hide dropped courses for students.  

@Dantemss feel free to provide some guidance on where changes should happen.

Dantemss:
I think the issue is mostly in `CollectCourseInfo`. An easy fix would be to just not return courses where the user has no roles.

Also for consistency, maybe `Role::GetUserRoles` shouldn't return inactive student roles (roles that have a `student` and the `student` object is inactive).

PR #667, opened 2015-09-17
(Some) Role cleanup

Dantemss:
- Teachers can now impersonate dropped students.
- Courses where the user is a dropped student should no longer be displayed for selection.

Solves https://www.pivotaltracker.com/story/show/103137354
Solves https://github.com/openstax/tutor-server/issues/638


Dantemss:
Confirmed the above 2 behaviors through manual testing. However, there is no way yet to indicate in the student scores report that some student has been dropped.

PR #668, opened 2015-09-17
Admins can now create users in the admin console

Dantemss:
And they are given a few more attribute fields to set.

Solves https://www.pivotaltracker.com/story/show/100222408


Dantemss:
New user is asked to reset their password on login, but otherwise goes straight to Tutor.

PR #669, opened 2015-09-18
Content Analyst role

Dantemss:
Merge https://github.com/openstax/tutor-server/pull/668 first.

- Added UsersUpdate handler to admin console users update action
- Made the UserProfile subsystem into a real subsystem, like the others
- Created ContentAnalyst model and migration
- Admins can create Admins and ContentAnalysts
- ContentAnalysts can read all ecosystems
- User profile has an `is_content_analyst` flag to let the FE know that a user is a content analyst and should see the proper menu


jpslav:
Epic, @Dantemss.

What about getting rid of `Entity::User` like you did for things like `Entity::Book`?  That is, changing `UserProfile` to `Profile` and introducing a `User::User` PORO?  Would still have all the `models` you do now, but possible the `Profile` model would lose a few methods that would move to the PORO (e.g. `is_content_analyst?`)

Dantemss:
I'll give it a shot.

PR #671, opened 2015-09-22
Set the default cache store to Redis (initializer did not work)

Dantemss:
You can verify this fix works because `Rails.cache` in the console is an instance of the Redis cache class after the fix.


Dantemss:
Chris deployed this branch to QA so I'll go ahead and merge...

PR #673, opened 2015-09-22
Add ecosystem info to endpoints for use in FE spy mode

nathanstitt:
Currently using the ecosystem's title since that seems to be nicely formatted.  Please advise if that's not appropriate.

### DIFF
--- a/app/representers/api/v1/page_representer.rb
+++ b/app/representers/api/v1/page_representer.rb
@@ -9,10 +9,10 @@ module Api::V1
              readable: true,
              writeabel: false
 
-    property :ecosystem_title,
-             type: String,
+    property :spy,
+             type: Object,
              readable: true,
-             getter: ->(*) { ecosystem.title },
+             getter: ->(*) { {title: ecosystem.title} },
              writeable: false
   end
 end
--- a/app/representers/api/v1/page_representer.rb
+++ b/app/representers/api/v1/page_representer.rb
@@ -12,7 +12,7 @@ module Api::V1
     property :spy,
              type: Object,
              readable: true,
-             getter: ->(*) { {title: ecosystem.title} },
+             getter: ->(*) { {ecosystem_title: ecosystem.title} },
              writeable: false
   end
 end
--- a/app/representers/api/v1/page_representer.rb
+++ b/app/representers/api/v1/page_representer.rb
@@ -9,5 +9,10 @@ module Api::V1
              readable: true,
              writeabel: false
 
+    property :ecosystem_title,
+             type: String,
+             readable: true,
+             getter: ->(*) { ecosystem.title },
+             writeable: false
   end
 end
--- a/app/representers/api/v1/task_representer.rb
+++ b/app/representers/api/v1/task_representer.rb
@@ -70,10 +70,9 @@ module Api::V1
                  description: "The steps which this Task is composed of"
                }
 
-    property :ecosystem_title,
-             type: String,
+    property :spy,
+             type: Object,
              readable: true,
-             getter: ->(*) { ecosystem.try(:title) },
              writeable: false
   end
 end
--- a/app/representers/api/v1/task_representer.rb
+++ b/app/representers/api/v1/task_representer.rb
@@ -70,10 +70,12 @@ module Api::V1
                  description: "The steps which this Task is composed of"
                }
 
-    property :ecosystem_info,
+    property :ecosystem_title,
              type: String,
              readable: true,
-             getter: ->(*) { task_plan ? task_plan.ecosystem.title : '' },
+             getter: ->(*) {
+      binding.pry
+      task_plan ? task_plan.ecosystem.title : '' },
              writeable: false
   end
 end
--- a/app/representers/api/v1/task_representer.rb
+++ b/app/representers/api/v1/task_representer.rb
@@ -73,7 +73,7 @@ module Api::V1
     property :ecosystem_title,
              type: String,
              readable: true,
-             getter: ->(*) { ecosystem.title },
+             getter: ->(*) { ecosystem.try(:title) },
              writeable: false
   end
 end
--- a/app/representers/api/v1/task_representer.rb
+++ b/app/representers/api/v1/task_representer.rb
@@ -70,5 +70,10 @@ module Api::V1
                  description: "The steps which this Task is composed of"
                }
 
+    property :ecosystem_info,
+             type: String,
+             readable: true,
+             getter: ->(*) { task_plan ? task_plan.ecosystem.title : '' },
+             writeable: false
   end
 end
--- a/app/representers/api/v1/task_representer.rb
+++ b/app/representers/api/v1/task_representer.rb
@@ -73,9 +73,7 @@ module Api::V1
     property :ecosystem_title,
              type: String,
              readable: true,
-             getter: ->(*) {
-      binding.pry
-      task_plan ? task_plan.ecosystem.title : '' },
+             getter: ->(*) { ecosystem.title },
              writeable: false
   end
 end
--- a/app/routines/reset_practice_widget.rb
+++ b/app/routines/reset_practice_widget.rb
@@ -25,24 +25,26 @@ class ResetPracticeWidget
 
     # Gather 5 exercises
     count = 5
-    exercises = case exercise_source
-                when :fake
-                  pools = []
-                  get_fake_exercises(count)
-                when :local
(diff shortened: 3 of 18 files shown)


Dantemss:
If that's just for QA, showing the title is fine.

Dantemss:
But then again I'd name it `ecosystem_title` or something in the API (more specific than "info").

nathanstitt:
Yep, `ecosystem_title` is better.

I'm a bit unsure how to get the ecosystem info for a practice tasks since they lack a `task_plan`.  I've discovered I can obtain it by looking at the steps, i.e. `task_steps.first.tasked.exercise.ecosystem.title`  That seems really bad both from a performance and maintainability aspect though.

Maybe we should just output the ecosystem info on the steps and not on the task itself?

Dantemss:
Ah I see. For practice tasks, it's the course ecosystem at the time the practice task was created and we don't store it anywhere. The good way would be to add a `content_ecosystem_id` to the Task model. But I don't mind that hack you wrote above (for now).

app/routines/reset_practice_widget.rb
```diff
@@ -25,24 +25,26 @@ def exec(role:, exercise_source:, page_ids: nil, chapter_ids: nil, randomize: tr
...
+      ecosystem, pools = get_ecosystem_and_pools(page_ids, chapter_ids, role)
+      exercises = run(:get_ecosystem_exercises_from_biglearn, ecosystem: ecosystem,
+                      count: count,
+                      role: role,
+                      pools: pools)
+                  .outputs.ecosystem_exercises
+    else
+      raise ArgumentError,
+            "exercise_source: must be one of [:fake, :local, :biglearn]"
+    end
```
nathanstitt: The ecosystem variable was being set on all the cases except `:fake`.  This adds it to that case so it can be used below.  After that change, the code was depending on variables `ecosystem` and `exercises` that are set inside the case.  I removed the top `exercises` assignment to hopefully make it clear that was going on.

app/representers/api/v1/task_representer.rb
```diff
@@ -70,5 +70,10 @@ class TaskRepresenter < Roar::Decorator
                  description: "The steps which this Task is composed of"
                }
 
+    property :ecosystem_title,
+             type: String,
+             readable: true,
+             getter: ->(*) { ecosystem.try(:title) },
+             writeable: false
```
nathanstitt: The `.try` is needed because old tasks didn't have the ecosystem recorded on them and it can be nil.

app/routines/reset_practice_widget.rb
Dantemss: Good enough for a hack I guess.

Dantemss:
Looks good to me.

db/migrate/20150923202225_add_task_content_ecosystem.rb
```diff
@@ -0,0 +1,6 @@
+class AddTaskContentEcosystem < ActiveRecord::Migration
+  def change
+    add_reference :tasks_tasks, :content_ecosystem,
```
Dantemss: Hmm might be a good idea to either add `index: true` or add a comment to the belongs_to saying that the relation is not indexed (so don't define the inverse relation without adding an index).

app/subsystems/tasks/models/task.rb
```diff
@@ -6,6 +6,7 @@ class Tasks::Models::Task < Tutor::SubSystems::BaseModel
                    :page_practice, :mixed_practice, :external]
 
   belongs_to :task_plan, inverse_of: :tasks
+  belongs_to :ecosystem, subsystem: :content
```
jpslav: I'm not really so keen on adding a relation from the Task model to the Content SS, especially for a spy value.  Currently a Task can live if the Content SS disappeared; with a relation, it gets stickier.  And when we refactor the Task SS to get rid of Entity::Task, we'll probably be trying to find ways to get rid of this relationship.  

I'd be happier having a `spy` field on `Tasks::Models::Task` which is a serialized hash of various primitives (similar to how `related_content_array` doesn't have real objects but rather is a hash of strings).  That way we preserve the information we need but don't have an explicit dependency (at least not from the point of view of the Task model).  The flip side is that the representer or some other helper code would need to look up the ecosystem title by the id in the spy hash.

Thoughts?

app/representers/api/v1/page_representer.rb
```diff
@@ -9,5 +9,10 @@ class PageRepresenter < Roar::Decorator
              readable: true,
              writeabel: false
 
+    property :ecosystem_title,
```
jpslav: A la my other comment below, while this could be a fine property, might instead want `property :spy`.  Could even envision a `SpyRepresenter` that could flesh out a hash of spy values (e.g. seeing `ecosystem: { id: 42 }` and converting it to `ecosystem: { id: 42, title: 'Ha ha ho ho' }`).

app/subsystems/tasks/models/task.rb
Dantemss: Ugh I'm usually in favor of ID's instead of serialized hashes, so I would like the ID more. Maybe @kevinburleigh75 can be the tie breaker? :)

nathanstitt:
@jpslav I like your suggestion regarding not adding a fk reference to the ecosystem just for spy mode and instead just storing the info we need onto it.  I've now done so.

I added the column as type `hstore`.  I don't see any use of that column type yet, if it's an issue I guess we could use yaml encoded strings (bleh).

Dantemss:
I suppose hstore is fine now that we all use postgres...

jpslav:
Thanks @nathanstitt I like the changes.  Can we go one extra step and remove the knowledge of ecosystems as far as we can from code within the Task SS that otherwise doesn't need it?

```ruby
# app/routines/add_spy.rb
class Spy
  lev_routine

  def exec(on:, add: ,save: false)
    add.each do |key, value|
      on.spy[key] = expand(value)
    end
    on.save! if save
  end

  def expand(value)
    # This being spy code, I don't care how much it touches other code and
    # subsystems; in fact, I prefer to have all that extra debug knowledge
    # here in one place and not spread around in other SS's.
    case value
    when Content::Ecosystem
      {
        id: value.id,
        title: value.title
      }
    else
      raise IllegalArgument
      # or could have some default behavior like just spying out the ID
    end
  end
end
```

And then take all the passings of `ecosystem` variables out of `Tasks::BuildTask` and `CreatePracticeWidget` by just adding the spy call right after the task is made in various circumstances, e.g.:

```ruby
# app/routines/reset_practice_widget.rb

run(:create_practice_widget_task, exercises: exercises,
                                  task_type: task_type,
                                  ecosystem: ecosystem,
                                  related_content_array: related_content_array)
```

becomes...

```ruby
# app/routines/reset_practice_widget.rb

run(:create_practice_widget_task, exercises: exercises,
                                  task_type: task_type,
                                  related_content_array: related_content_array)

Spy[on: outputs.task, add: {ecosystem: ecosystem}, save: true]
```

And similarly in the assistant we get things like:

```ruby
 def build_homework_task(taskee:, exercises:)
    task = build_task
    Spy[on: task, add: {ecosystem: ecosystem}]
```

By pulling the spy stuff back in this way, we separate it as much as possible from the actual domain logic.  And turning off spying is as easy as commenting out one line or configurably disabling all spying from within the `Spy` routine.

Could be that `Spy` just takes an array of things to "add" to the spy hash, and it deduces the key name from the type of thing being added:

```ruby
Spy[on: outputs.task, add: ecosystem]
```

Can we do this or a close variant of this?

nathanstitt:
@jpslav finished adding spy routine

app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -75,12 +75,13 @@ def build_task
...
     description = @task_plan.description
 
-    Tasks::BuildTask[
+    task = Tasks::BuildTask[
       task_plan:   @task_plan,
       task_type:   :homework,
       title:       title,
       description: description
     ]
+    AddSpyInfo[to: task, from: @ecosystem]
```
jpslav: I see how this does in fact return the `task` from the `build_task` method, but it isn't entirely obvious that that's the case in the way that having `BuildTask[...]` be the last command does.  I'm going to merge the PR because it is great :-), but if folks think the return value is confusing at all, we could change to:

```ruby
...
Tasks::BuildTask[
  ...
].tap do |task|
  AddSpyInfo[to: task, from: @ecosystem]
end
```

I don't know, maybe not clearer...  great PR either way, thanks @nathanstitt

PR #674, opened 2015-09-23
Improved teacher learning guide performance some more

Dantemss:
- Reuse ecosystem mappings already computed across periods
- Map now returns a hash that maps exercise ids to pages for convenience

### DIFF
--- a/app/subsystems/content/strategies/generated/map.rb
+++ b/app/subsystems/content/strategies/generated/map.rb
@@ -77,7 +77,7 @@ module Content
 
           all_map = map_exercises_to_pages(exercises: all_exercises)
 
-          # The hash returned has all exercises given as keys
+          # The hash returned has all given exercise ids as keys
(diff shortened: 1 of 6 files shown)


app/subsystems/content/strategies/generated/map.rb
```diff
@@ -26,53 +26,61 @@ def create!(from_ecosystems:, to_ecosystem:)
...
-        def map_exercises_to_pages(exercises:)
-          all_pages = @to_ecosystem.pages
-          page_map = all_pages.each_with_object({}) do |page, hash|
+          @page_id_to_page_map = to_ecosystem.pages.each_with_object({}) do |page, hash|
             hash[page.id] = page
           end
+          @exercise_id_to_page_map = {}
+        end
 
+        def map_exercises_to_pages(exercises:)
```
Dantemss: This diff looks weird for some reason, but basically the initializer got a couple extra things in it and `map_exercises_to_pages` moved down.

PR #676, opened 2015-09-24
More teacher learning guide performance improvements

Dantemss:
- Combines period-specific queries into one query for the teacher guide
- Optimizes some preloading that was being done incorrectly
- Maps all exercises at once to minimize the number of queries


Dantemss:
I think this is about as fast as it gets without precomputing the learning guide and storing it somewhere...

Dantemss:
Merging so we can test tomorrow...

PR #677, opened 2015-09-24
Share Google Analytics snippet between views

nathanstitt:
This way FE code can access it as well.

### DIFF
--- a/app/views/layouts/webview.html.erb
+++ b/app/views/layouts/webview.html.erb
@@ -7,6 +7,7 @@
   <script src="//cdn.mathjax.org/mathjax/2.5-latest/MathJax.js?config=TeX-MML-AM_HTMLorMML-full&amp;delayStartupUntil=configured" async></script>
   <script type='text/javascript' src='<%= Rails.application.secrets[:js_url] %>' async></script>
   <%= csrf_meta_tags %>
+  <%= render 'webview/google_analytics' %>
 </head>
 <body>
   <script id="tutor-boostrap-data" type="application/json"><%=raw json_escape bootstrap_data.to_json %></script>
--- a/app/views/webview/_google_analytics.html.erb
+++ b/app/views/webview/_google_analytics.html.erb
@@ -1,0 +1,11 @@
+<% if is_real_production_site? %>
+  <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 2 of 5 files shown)


app/views/webview/_google_analytics.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if is_real_production_site? %>
+  <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+
+    ga('create', 'UA-66552106-1', 'auto');
+    ga('send', 'pageview');
```
nathanstitt: Should this variable be read from `secrets.yml` ?

jpslav: Sure can be for convenience. Isn't secret.

nathanstitt:
I'm now storing the tracking code in secrets and am including the JS snippet if it's present.  This will allow us to track other properties as well if we need to.

It'll also allow the FE to test tracking events putting a test site's code into local secrets.

app/views/layouts/webview.html.erb
```diff
@@ -7,6 +7,7 @@
   <script src="//cdn.mathjax.org/mathjax/2.5-latest/MathJax.js?config=TeX-MML-AM_HTMLorMML-full&amp;delayStartupUntil=configured" async></script>
   <script type='text/javascript' src='<%= Rails.application.secrets[:js_url] %>' async></script>
   <%= csrf_meta_tags %>
+  <%= render 'webview/google_analytics' %>
```
jpslav: Do you want to stick the tracking code in the bootstrap data instead or is this really sufficient?  (imagine you'll need the code separately later to send other events to GA)

jpslav:
Looks good to me, agree need @cnuber to weigh in from the tutor-deployment side.

app/views/layouts/webview.html.erb
nathanstitt: This is sufficient for now.  The analytics code sets a global "ga" variable that the FE PR checks for the existence of.  If we use other trackers or get fancier we can revisit it.

PR #672, opened 2015-09-22
User subsystem

Dantemss:
Merge https://github.com/openstax/tutor-server/pull/669 first.

- Renamed user_profile subsystem to user
- Added wrappers and strategies to the user subsystem
- Removed `Entity::User`
- Controller `current_user` etc methods now return a `User::User`
- Fixed specs
- Using `to_model` method for assignments to external polymorphics and FinePrint
- Renamed `Content::StrategyError` to `StrategyError` and moved to lib

PR #678, opened 2015-09-25
Customer service role

Dantemss:
Includes https://github.com/openstax/tutor-server/pull/672

- Added customer service role model and migration
- Admin can add customer service role to users
- Customer service role can access show/index actions in the admin console
- Added model spec
- FE now gets is_customer_service in the user json
- Copied and modified relevant admin controllers and views
- Copied/modified controller/view specs and added some more


Dantemss:
Do we want to reuse the admin controllers (like in the PR right now), or make separate ones for customer service?

jpslav:
Can you rebase now that 672 is merged?  

I haven't looked at the admin views as a customer service user, but are the views getting complicated with a lot of "if customer_service_user do blah" or "do blah if admin" then we may want to consider dedicated screens for each role.

Dantemss:
I think the only thing that would need a lot of ifs would be the console itself. And the courses `show` page is already completely new (because usually all this course info is only displayed on the `edit` page).

Maybe a separate console, but reuse the other pages then?

jpslav:
Seems reasonable. That way we can also put specific messaging and options all in one place for customer service folks

Dantemss:
OK I decided to just make them separate... Customer support views were simple enough that it wasn't too bad.

Dantemss:
All that FE needs to provide is a Customer Support button or redirect. You can test the control panel at `/customer_support`.

app/views/customer_service/targeted_contracts/index.html.erb
```diff
@@ -0,0 +1,24 @@
+<% @page_header = "Targeted Contracts" %>
```
jpslav: Don't want customer service folks doing anything with targeted contracts -- you can take these out.

config/routes.rb
```diff
@@ -149,6 +149,27 @@
...
 
+  namespace 'customer_service' do
+    root to: 'console#index'
+
+    resources :courses, only: [:index, :show] do
+      resources :periods, only: [:index], shallow: true
+      resources :students, only: [:index], shallow: true
+    end
+
+    resources :licenses, only: :index
```
jpslav: I think licenses is something copied from OX Exercises -- can you remove?

jpslav: here and admin routes

jpslav:
Ok, Dante, don't kill me for this... I think there's too much duplicated code here.  

Can we go the other way so there is only one set of "management" or "console" views that both admins and customer service folks can access, but with admins being the only ones who can see edit capabilities and use those edit capabilities?

The routes namespace can be "console" or "manage" or "management" and the controllers / views can use the same word so we don't have customer service people operating under something named "admin".

While the current approach has some benefits, I fear we're going to get stuck evolving two different consoles and always needing to fix bugs in two places, etc, and it will become an increasing pain in the rear.

Thoughts?

config/routes.rb
Dantemss: ok

app/views/customer_service/targeted_contracts/index.html.erb
Dantemss: ok

Dantemss:
@jpslav ok I split the shared partials out.

app/assets/javascripts/customer_service.js
```diff
@@ -0,0 +1,58 @@
...
+  var tab = window.location.hash;
+  if (tab) {
+    $('a[href="' + tab + '"]').click();
+  }
+});
+
+// === Customer Service/Jobs === //
+$(function(){
+  $(document).on('click', '.filter_job_status', function(e) {
+    var desiredStatus = this.href.replace(/^\S+#/, ''),
```
jpslav: Can this JS be in a partial?

app/assets/stylesheets/customer_service.scss
```diff
@@ -0,0 +1,28 @@
...
+  color: red;
+}
+
+#jobs {
+  .completed {
+    display: none;
+  }
+}
+
+#search_by_id {
```
jpslav: Ditto for this jobs-specific CSS (make it shared too)

app/controllers/admin/jobs_controller.rb
```diff
@@ -14,6 +15,7 @@ def show
       end
 
       @page_header = "#{@job.status.titleize} job : #{@job.id}"
```
jpslav: Can you move the @page_headers into the partials?

app/controllers/customer_service/courses_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  private
+
+  def get_schools
+    @schools = SchoolDistrict::ListSchools[]
+  end
+
+  def get_course_details
+    entity_course = Entity::Course.find(params[:id])
+    @course = GetCourseProfile[course: entity_course]
```
jpslav: I still don't like that these 15 lines are copied.

app/assets/javascripts/customer_service.js
Dantemss: Probably...

app/controllers/admin/jobs_controller.rb
```diff
@@ -14,6 +15,7 @@ def show
       end
```
jpslav: What about moving these other lines into the views too?

app/controllers/admin/jobs_controller.rb
Dantemss: I don't think this belongs in partials? It's a page-wide thing to be set by the base view...

app/controllers/customer_service/courses_controller.rb
Dantemss: I can include a module or something

app/controllers/admin/jobs_controller.rb
Dantemss: Too much business logic for the views... I'll make another module.

Dantemss:
@jpslav how about now?

jpslav:
Beautiful - thanks Dante

PR #679, opened 2015-09-28
Trouble flag

Dantemss:
- Shows up on assignment stats page, for each period in the entire assignment
- Also visible for each page in the assignment (but currently that would probably not be used by the FE)
- Trouble if >50% incorrect and >25% problems worked (completed)
- Added specs


jpslav:
Looks good.  Can merge when @darrenmason or @philschatz  gives the go ahead

jpslav:
Doesn't provide trouble flag for the dashboard endpoint -- doing that separately

PR #680, opened 2015-09-28
Add unique_learner_count to the Performance Forecast API

Dantemss:
- Pass the `unique_learner_count` from Biglearn to the FE
- Updated specs


jpslav:
Looks good.  Rebase and we can merge.

PR #682, opened 2015-09-29
Threaded demo task creation/distribution

nathanstitt:
Non working code that throws:

`Circular dependency detected while autoloading constant <class>`


Dantemss:
Fixed now :)

Dantemss:
Now with 100% more forks.

Dantemss:
@nathanstitt: should now work without any hiccups

PR #592, opened 2015-08-17
Add EventAssistant

joemsak:
* Adds `EventAssistant`

PR #683, opened 2015-09-29
Event Assistants Part Deux

jpslav:
Takes over for #592


jpslav:
cc @joemsak, I moved your work into a new branch to avoid rebase perdition

PR #686, opened 2015-09-30
event assistant in course

pandafulmanda:
add event assistance to create course assistance, update docs for plan endpoint, make event assistant match other assistants' styles?


app/subsystems/tasks/create_course_assistants.rb
```diff
@@ -24,6 +24,13 @@ def exec(course:)
       code_class_name: "Tasks::Assistants::ExternalAssignmentAssistant",
       task_plan_type: 'external'
     )
+
+    create_course_assistant(
+      course: course,
+      assistant_name: "Event Assignment Assistant",
```
jpslav: Can just be "Event Assistant".  "Assignment" was only for "External Assignment", and events aren't really assignments, per se.

pandafulmanda: oops, yes

PR #687, opened 2015-09-30
Fix agreeing to fine_print terms of service

nathanstitt:
Looks like a `.to_model` was missed because it didn't have test coverage :)

### DIFF
--- a/app/controllers/terms_controller.rb
+++ b/app/controllers/terms_controller.rb
@@ -11,7 +11,7 @@ class TermsController < ApplicationController
   end
 
   def agree
-    signature = FinePrint.sign_contract(current_user, params[:contract_id]) if params[:i_agree]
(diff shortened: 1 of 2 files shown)


app/controllers/terms_controller.rb
```diff
@@ -11,7 +11,7 @@ def pose
   end
 
   def agree
-    signature = FinePrint.sign_contract(current_user, params[:contract_id]) if params[:i_agree]
+    signature = FinePrint.sign_contract(current_user.to_model, params[:contract_id]) if params[:i_agree]
```
jpslav: @Dantemss should fine print be changed to use a proc to do the `to_model` mapping?  I know it already does so for tutor-server in the `current_user_proc`.

Dantemss: It does normally, but since this controller calls its internal methods manually, it bypasses the mapping.

PR #688, opened 2015-09-30
Made course assistants load lazily

jpslav:


jpslav:
@Dantemss please review.  This pairs with @pandafulmanda's https://github.com/openstax/tutor-server/pull/686 to create missing assistants for previously-created courses.

Dantemss:
Seems good but how about just deleting the model entirely? Maybe register the assistants in an initializer or require the assistants and they can self-register in some module? :)

jpslav:
Eh.... let's delete them later.  I know there has been a desire to do this... but I still want the ability to control which courses get which assistants to support future research work.

PR #692, opened 2015-10-02
CLUe caching in cron task

Dantemss:
- Biglearn client now has a limit of 1000 (students times pools)
- Requests exceeding the limit will be chopped up and done in parallel
- Biglearn client now accepts either 1 role or 1 period and caches the result appropriately
- Biglearn client can now be told to ignore answers since the clue was cached
- Biglearn client can now be told to force a cache miss
- Teacher learning guide now queries Biglearn by period (ignores answers since the clue was cached)
- Added rake task to cache period clues for the latest course ecosystem (forces a cache miss)
- Added whenever schedule to run rake task every minute
- Added specs


Dantemss:
And... threads are actually useful here because it's waiting for IO (from Biglearn).

app/external/openstax/biglearn/v1.rb
```diff
@@ -50,17 +50,23 @@ def self.get_projection_exercises(role:, pools:, count: 1,
   # Each CLUe refers to one specific pool, but uses all roles given.
   # May return nil if no CLUe is available
   # (e.g. no exercises in the pools or confidence too low).
-  def self.get_clues(roles:, pools:)
+  def self.get_clues(pools:, period: nil, role: nil,
+                     force_cache_miss: false, ignore_answer_times: period.present?)
+    raise ArgumentError, 'You must provide either a period or a role but not both', caller \
+      if period.present? && role.present?
```
jpslav: This does not check that either the period or role is provided, as suggested by the error message.

app/external/openstax/biglearn/v1.rb
```diff
@@ -50,17 +50,23 @@ def self.get_projection_exercises(role:, pools:, count: 1,
...
     pools = [pools].flatten.compact
-    roles = [roles].flatten.compact
 
     # No pools given: empty map
     return {} if pools.blank?
 
     # No roles given: map all pools to nil
-    return pools.each_with_object({}) { |pool, hash| hash[pool.uuid] = nil } if roles.blank?
+    return pools.each_with_object({}) { |pool, hash| hash[pool.uuid] = nil } \
+      if role.nil? && period.nil?
```
jpslav: And this says that it is not an error to provide neither role or period

app/external/openstax/biglearn/v1.rb
Dantemss: I can change the error message, I suppose.

app/routines/update_period_clues.rb
```diff
@@ -0,0 +1,34 @@
...
+              ecosystem.pages.collect(&:all_exercises_pool)
+
+      course.periods.collect{ |period| [period, pools] }
+    end
+
+    threads = clue_queries.collect do |period, pools|
+      Thread.new do
+        OpenStax::Biglearn::V1.get_clues(pools: pools, period: period,
+                                         force_cache_miss: true, ignore_answer_times: true)
+      end
```
jpslav: To limit the other classes that `OpenStax::Biglearn` is aware of, can we remove its knowledge of periods?

```ruby
OpenStax::Biglearn::V1.get_clues(pools: pools, roles: period.active_enrollments,
                                 cache_for: period,
                                 force_cache_miss: true, ignore_answer_times: true)
```

Inside biglearn you can set the `learner_cache_key` as:

```ruby
learner_cache_key = 
  if cache_for.present? 
    cache_for.global_id.to_s
  else
    if learners.one?
      learners.first
    else
      nil # don't cache; alternatives: raise exception OR combine learners into one key
    end
  end
```

Dantemss: ok

Dantemss: I think I'll use `.cache_key.split('-').first` so I don't have to include another gem for the global_id.

PR #696, opened 2015-10-06
Performance fix - Refresh recent CLUes only

Dantemss:
- Student CLUes can now be updated with period CLUes
- CLUe update routine now accepts a `:type` parameter that can specify only `:recent` CLUes or `:all` CLUes should be updated
- Biglearn RealClient forgot how to handle answer times, since the CLUe update routine now updates student CLUes
- Specs


config/schedule.rb
```diff
@@ -2,5 +2,9 @@
 
 every 1.minute do
   rake 'openstax:accounts:sync:all'
-  rake 'openstax:biglearn:update_period_clues'
+  rake 'openstax:biglearn:clues:update:recent'
+end
+
+every 1.day do
```
Dantemss: We could also make this run at 2 AM, or more often than once daily.

jpslav:
Looks good, merge when travis passes

PR #697, opened 2015-10-06
Speed up generation of scores report

nathanstitt:


app/subsystems/tasks/get_performance_report.rb
```diff
@@ -70,7 +72,7 @@ def sort_tasking_plans(taskings, course, period)
     def get_taskings(period)
       task_types = Models::Task.task_types.values_at(:reading, :homework, :external)
       # Return reading, homework and external tasks for a student
-      period.taskings.preload(task: {task: {task_plan: :tasking_plans}},
+      period.taskings.preload(task: {task: {task_plan: {tasking_plans: :target} }},
                               role: {profile: :account})
```
nathanstitt: This is the real performance gain.  The other stuff is just cleanups so I didn't get confused by variables being shadowed in the block

nathanstitt:
The current issue is that the `task_plan.tasks` on line 101 isn’t preloaded.  So every time it’s called a queries executed.  The query that is preloaded does preload those, but in the opposite order.  The `:inverse_of` on `task`'s `task_plan` isn't taking effect for some reason

app/subsystems/tasks/get_performance_report.rb
Dantemss: Most likely because it's polymorphic, don't think inverse_of likes polymorphics.

Dantemss:
Oops I totally commented on the wrong thing :) guess no polymorphic here.

Dantemss:
You can totally preload in reverse order, like `{task: {task_plan: :tasks}}`, but I swear I didn't see this being necessary before.

app/subsystems/tasks/get_performance_report.rb
Dantemss: Good catch.

PR #698, opened 2015-10-06
added admin ability to trigger test exceptions

jpslav:
For testing that our exception handling capability works without waiting for a real exception :-)

### DIFF
--- a/app/controllers/admin/console_controller.rb
+++ b/app/controllers/admin/console_controller.rb
@@ -1,2 +1,19 @@
 class Admin::ConsoleController < Admin::BaseController
+
+  def test_raise
+    type = params[:type] || 'standard_error'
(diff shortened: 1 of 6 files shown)


app/controllers/admin/console_controller.rb
```diff
@@ -1,2 +1,19 @@
...
+      raise ActiveRecord::RecordNotFound, msg
+    when 'argument_error'
+      raise ArgumentError, msg
+    when 'standard_error'
+      raise StandardError, msg
+    else
+      raise StandardError, msg
+    end
+  end
+
```
nathanstitt: If you wanted to get fancy you replace all but the first type with: 

`raise (type.classify.safe_constantize || StandardError), msg`

Not sure if that's better though

PR #699, opened 2015-10-06
Fixed UpdateClues to not make unnecessary requests to Biglearn

Dantemss:
UpdateClues was making too many requests to Biglearn (number students that worked problems) * (number of pools worked).

This PR fixes it so it explicitly keeps track of who worked what problems and only makes the minimum number of requests needed.


Dantemss:
JP says to merge :)

PR #700, opened 2015-10-06
Speed up teacher's performance forecast

nathanstitt:
The nested flat_map's was taking most of the time to find all the `TaskSteps`.  This queries for them directly and then also loads the other relations

### DIFF
--- a/app/routines/get_teacher_guide.rb
+++ b/app/routines/get_teacher_guide.rb
@@ -13,14 +13,18 @@ class GetTeacherGuide
   private
 
   def completed_exercise_task_steps_for_course_by_period(course)
-    course.periods.preload(
-      active_enrollments: {student: {role: {taskings: {task: {task: :task_steps}}}}}
-    ).to_a.each_with_object({}) do |period, hash|
-      hash[period.id] = get_completed_exercise_task_steps(
-        period.active_enrollments.flat_map do |ae|
-          ae.student.role.taskings.flat_map{ |ts| ts.task.task.task_steps }
-        end
-      )
+    course.periods.each_with_object({}) do | period,hash |
+      hash[period.id] = Tasks::Models::TaskStep.exercises.complete
+        .preload([
+          {tasked: {exercise: {}}},
+          {task: {taskings: {role: {student: {enrollments: :period}}}}}
+        ])
+        .joins(
+          task: {taskings: {role:{ student:{ enrollments: :period}}}}
+        )
+        .where(
+          task: {taskings:{ role: {student:{ enrollments: {period:{id: period.id}}}}}}
+        ).to_a
     end
   end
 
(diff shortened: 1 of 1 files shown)


app/routines/get_teacher_guide.rb
```diff
@@ -31,14 +35,11 @@ def gather_period_stats(period_id, course, tasked_exercises, exercise_id_to_page
...
+
     period_to_completed_exercise_task_steps_map = \
       completed_exercise_task_steps_for_course_by_period(course)
 
-    all_completed_exercise_task_steps = period_to_completed_exercise_task_steps_map.values.flatten
-
-    task_step_to_tasked_exercise_map = get_tasked_exercises_from_completed_exercise_task_steps(
-      all_completed_exercise_task_steps
-    )
+    task_step_to_tasked_exercise_map = task_steps.map(&:tasked).group_by{ |te| te.task_step.id }
```
Dantemss: Err where are the `task_steps` coming from?

app/routines/get_teacher_guide.rb
```diff
@@ -13,14 +13,18 @@ def exec(role:)
...
-      hash[period.id] = get_completed_exercise_task_steps(
-        period.active_enrollments.flat_map do |ae|
-          ae.student.role.taskings.flat_map{ |ts| ts.task.task.task_steps }
-        end
-      )
+    course.periods.each_with_object({}) do | period,hash |
+      hash[period.id] = Tasks::Models::TaskStep.exercises.complete
+        .preload([
+          {tasked: {exercise: {}}},
+          {task: {taskings: {role: {student: {enrollments: :period}}}}}
```
Dantemss: You can try to remove the second line of this preload (since I think you are not using it anymore?)... ofc leave the `joins` in, though.

nathanstitt: Yeah, I tried it and it does need 'em.  Without it they aren't loaded and Rails goes into meltdown.

app/routines/get_teacher_guide.rb
```diff
@@ -13,14 +13,18 @@ def exec(role:)
...
-    ).to_a.each_with_object({}) do |period, hash|
-      hash[period.id] = get_completed_exercise_task_steps(
-        period.active_enrollments.flat_map do |ae|
-          ae.student.role.taskings.flat_map{ |ts| ts.task.task.task_steps }
-        end
-      )
+    course.periods.each_with_object({}) do | period,hash |
+      hash[period.id] = Tasks::Models::TaskStep.exercises.complete
+        .preload([
+          {tasked: {exercise: {}}},
```
Dantemss: Actually not sure this line is needed either... since the course_guide_methods does another separate query to get the `taskeds`.

app/routines/get_teacher_guide.rb
Dantemss: Ah I see... you bypassed the routine in `course_guide_methods` that was doing the preloading (the  `get_tasked_exercises_from_completed_exercise_task_steps`) that's why it's needed here now.

Dantemss: Maybe `get_tasked_exercises_from_completed_exercise_task_steps` is OBE?

nathanstitt: It's still used by the `app/routines/get_student_guide`.  I left that version alone for now, do you think we should poke at it also?

Dantemss: Could do a quick poke just to get rid of that method call...

Dantemss:
Thanks for the improvements!

app/routines/get_teacher_guide.rb
nathanstitt: Yeah, I took a try at it, and the student_guide needs to load the TaskExercises there.  Probably best to leave it as-is.

PR #703, opened 2015-10-07
Trouble flag on the calendar API

Dantemss:
- Added trouble flag to the calendar (logic in the `GetCourseTaskPlans` routine)
- Renamed dashboard routine and moved it into the routines folder
- Specs

### DIFF
--- a/app/routines/get_course_task_plans.rb
+++ b/app/routines/get_course_task_plans.rb
@@ -4,10 +4,14 @@ class GetCourseTaskPlans
 
   protected
 
-  def exec(course:)
-    task_plans = Tasks::Models::TaskPlan.where(owner: course).to_a
-    outputs[:total_count] = task_plans.count
-    outputs[:items] = task_plans
+  def exec(course:, include_trouble_flags: false)
+    outputs[:plans] = Tasks::Models::TaskPlan.where(owner: course)
+    return unless include_trouble_flags
+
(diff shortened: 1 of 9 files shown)


app/routines/get_course_task_plans.rb
```diff
@@ -4,10 +4,14 @@ class GetCourseTaskPlans
...
-    outputs[:total_count] = task_plans.count
-    outputs[:items] = task_plans
+  def exec(course:, include_trouble_flags: false)
+    outputs[:plans] = Tasks::Models::TaskPlan.where(owner: course)
+    return unless include_trouble_flags
+
+    outputs[:trouble_plan_ids] = Set.new Tasks::Models::TaskPlan.joins(:tasks).group(:id).having{
+      (sum(tasks.completed_exercise_steps_count) > sum(tasks.exercise_steps_count)/4) & \
+      (sum(tasks.correct_exercise_steps_count) < sum(tasks.completed_exercise_steps_count)/2)
+    }.pluck(:id)
```
jpslav: While this is awesome, I just realized it may not be correct.  When someone looks in the quick look analytics, they see trouble flags for specific pages within specific periods.  I think the calendar level flag should only show up if there are trouble flags on pages (for certain periods) in the plan.  So I think there could be a flag on one period's page but the calculation above wouldn't see it (it could be lost in the noise of all of the other pages and periods).  Right?

darrenmason:
corresponding FE PR:  https://github.com/openstax/tutor-js/pull/776

PR #704, opened 2015-10-08
Period enrollment code

joemsak:
https://www.pivotaltracker.com/n/projects/1156756/stories/105130938

### DIFF
--- a/db/migrate/20151008154744_add_enrollment_code_to_course_membership_periods.rb
+++ b/db/migrate/20151008154744_add_enrollment_code_to_course_membership_periods.rb
@@ -3,8 +3,8 @@ class AddEnrollmentCodeToCourseMembershipPeriods < ActiveRecord::Migration
     add_column :course_membership_periods, :enrollment_code, :string
     add_index :course_membership_periods, :enrollment_code, unique: true
 
-    reversible do |dir|
-      dir.up do
+    reversible do |direction|
+      direction.up do
         CourseMembership::Models::Period.where(enrollment_code: nil).find_each do |period|
           period.send(:generate_enrollment_code)
           period.save!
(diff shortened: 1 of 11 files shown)


spec/controllers/api/v1/courses_controller_spec.rb
```diff
@@ -213,10 +197,6 @@ def add_book_to_course(course:)
       it 'returns both roles with the course' do
         api_get :show, user_1_token, parameters: { id: course.id }
         expect(response.body_as_hash).to match a_hash_including(
-          id: course.id.to_s,
-          name: course.profile.name,
-          periods: [{ id: zeroth_period.id.to_s, name: zeroth_period.name },
-                    { id: period.id.to_s, name: period.name }],
```
joemsak: This and the similar tests above were updated because instead of adding in the period information to every single one, it seemed better to remove information from each test that was already tested the first time around, and leave only the part that's in question for the test.

db/migrate/20151008154744_add_enrollment_code_to_course_membership_periods.rb
```diff
@@ -0,0 +1,17 @@
+class AddEnrollmentCodeToCourseMembershipPeriods < ActiveRecord::Migration
+  def change
+    add_column :course_membership_periods, :enrollment_code, :string
+    add_index :course_membership_periods, :enrollment_code, unique: true
+
+    reversible do |dir|
+      dir.up do
+        CourseMembership::Models::Period.where(enrollment_code: nil).find_each do |period|
```
jpslav: thanks for find_each

joemsak: chalk it up to habit :)

PR #705, opened 2015-10-08
Fixing sporadic Exchange communication error

Dantemss:
- **Switched the ActiveJob backend to DelayedJob**
- Moved code to send TaskedExercise answers to Exchange to a routine
- Make routine run as a background job
- Specs

DelayedJob pros and cons

Pros:
- DelayedJob has retry built-in (though Resque has a perfectly fine plugin, it requires you to run an extra scheduler process)
- DelayedJob uses the DB, which we actually backup
- DelayedJob would write the job inside the transaction, so the worker wouldn't see it until the transaction is done

Cons:
- More DB access by DelayedJob's polling (one extra query per worker per 5 seconds).

### DIFF
--- a/config/initializers/delayed_job.rb
+++ b/config/initializers/delayed_job.rb
@@ -1,0 +1,23 @@
+# Defaults:
+# Delayed::Worker.destroy_failed_jobs = true
+# Delayed::Worker.sleep_delay = 5
+# Delayed::Worker.max_attempts = 25
+# Delayed::Worker.max_run_time = 4.hours
+# Delayed::Worker.read_ahead = 5
+# Delayed::Worker.default_queue_name = nil
+# Delayed::Worker.delay_jobs = true
+# Delayed::Worker.raise_signal_exceptions = false
+# Delayed::Worker.logger = Rails.logger
+
+# Keep failed jobs for later inspection
+Delayed::Worker.destroy_failed_jobs = false
+
+# Should be longer than the longest background job (that actually uses this gem)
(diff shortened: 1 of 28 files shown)


jpslav:
Looking good.  Will want to do pubsub stuff later so we can remove the explicit dependence between TE and this new routine, but waaay happier to have all that logic at least physically removed from the TE model (and to have the background job!)

Dantemss:
Need deployment PR: https://github.com/openstax/tutor-deployment/pull/164

config/initializers/delayed_job.rb
```diff
@@ -0,0 +1,23 @@
...
+# Delayed::Worker.default_queue_name = nil
+# Delayed::Worker.delay_jobs = true
+# Delayed::Worker.raise_signal_exceptions = false
+# Delayed::Worker.logger = Rails.logger
+
+# Keep failed jobs for later inspection
+Delayed::Worker.destroy_failed_jobs = false
+
+# Should be longer than the longest background job (that actually uses this gem)
+Delayed::Worker.max_run_time = 10.minutes
```
jpslav: Is 10 minutes sufficient?  Can this be a deployment parameter so can be changed without code change?  e.g.

```ruby
Delayed::Worker.max_run_time = Rails.application.secrets.background_worker_timeout_minutes || 10.minutes
```

jpslav:
Looks good.  One comment.

config/initializers/delayed_job.rb
Dantemss: Sure

PR #706, opened 2015-10-12
Fixes the teacher guide when students have been moved between periods

Dantemss:
Teacher guide was getting confused and picking the wrong period, since it was being passed students that had been moved away.


Dantemss:
Tested this fix manually in dev.

jpslav:
Looks good.

PR #708, opened 2015-10-12
Concept Coach "Magic Map"

Dantemss:
- Added CC "Magic Map" (TagGenerator)
- Added new CC tag type
- Importer now uses CC tag to search for exercises
- Separated tagging stuff out of the exercise wrapper
- Specs


app/subsystems/content/import_book.rb
```diff
@@ -15,7 +15,7 @@ class Content::ImportBook
 
   # Imports and saves a Cnx::Book as an Content::Models::Book
   # Returns the Book object, Resource object and collection JSON as a hash
-  def exec(cnx_book:, ecosystem:, exercise_uids: nil)
+  def exec(cnx_book:, ecosystem:, exercise_uids: nil, concept_coach_tag: nil)
```
jpslav: Nothing else in `Content` is going to know about "concept coach" so I'm wondering if we can make this new argument more generic.

```ruby
FetchAndImportBookAndCreateEcosystem[
  book_cnx_id: params[:cnx_id], tag_generator: ConceptCoachTagGenerator.new(tag_root: params[:cc_tag])
]
```

```ruby
def exec(cnx_book:, ecosystem:, exercise_uids: nil, tag_generator: nil)
```

```ruby
# import_page.rb

tags = cnx_page.tags
tags = tags + tag_generator.generate(book_location: book_location) if tag_generator
# instead of checking tag_generator presence, could make default tag_generator always return [] 
```

```ruby
# /app/subsystems/content/models/concept_coach_tag_generator.rb

class Content::Models::ConceptCoachTagGenerator
  def initialize(tag_root)
    @tag_root = tag_root
  end

  def generate(book_location: book_location)
    return [] if book_location.size != 2
    chapter_tag = "#{@tag_root}-ch#{"%02d" % book_location.first}"
    section_tag = "#{chapter_tag}-s#{"%02d" % book_location.last}"
    [concept_coach_tag, chapter_tag, section_tag]
  end
end
```

Also makes the string manipulating lines easier to unit test

jpslav:
Review complete.  Looking good.

app/subsystems/content/import_book.rb
Dantemss: Hm ok

app/subsystems/content/routines/import_page.rb
```diff
@@ -44,8 +47,11 @@ def exec(cnx_page:, chapter:, number: nil, book_location: nil, save: true)
 
     return unless save
 
-    # Get Exercises from OpenStax Exercises that match the LO's and AP LO's
-    objective_tags = outputs[:tags].select{ |tag| tag.lo? || tag.aplo? }.collect{ |tag| tag.value }
+    # Get Exercises from OpenStax Exercises that match the LO's or AP LO's, plus CC tags
+    objective_tags = outputs[:tags].select do |tag|
```
jpslav: Maybe add an `objective?` method to `Content::Models::Tag` and collapse this to

```ruby
objective_tags = outputs.tags.select(&:objective?).collect(&:value)
```

app/subsystems/content/import_book.rb
```diff
@@ -23,11 +23,14 @@ def exec(cnx_book:, ecosystem:, exercise_uids: nil)
...
                                      content_ecosystem_id: ecosystem.id)
 
-    run(:import_book_part, cnx_book_part: cnx_book.root_book_part, book: book, save: false)
+    run(:import_book_part, cnx_book_part: cnx_book.root_book_part, book: book,
+                           save: false, tag_generator: tag_generator)
 
     Content::Models::Book.import! [book], recursive: true
 
-    objective_page_tags = outputs[:page_taggings].select{ |pt| pt.tag.lo? || pt.tag.aplo? }
+    objective_page_tags = outputs[:page_taggings].select do |pt|
```
jpslav: Like other comment...

```ruby
objective_page_tags = outputs.page_taggings.select(&:objective?)
```

app/subsystems/content/import_book.rb
```diff
@@ -35,7 +38,7 @@ def exec(cnx_book:, ecosystem:, exercise_uids: nil)
     else
       outputs[:exercises] = []
       page_block = ->(exercise_wrapper) {
-        tags = Set.new(exercise_wrapper.los + exercise_wrapper.aplos)
+        tags = Set.new(exercise_wrapper.los + exercise_wrapper.aplos + exercise_wrapper.ccs)
```
jpslav:
```ruby
tags = exercise_wrapper.objective_tags
```
?

app/external/openstax/exercises/v1/exercise.rb
```diff
@@ -6,6 +6,9 @@ class OpenStax::Exercises::V1::Exercise
   # This Regex finds the AP LO's within the exercise tags
   APLO_REGEX = /[\w-]+-aplo-[\w-]+/
 
+  # This Regex finds the CC section tag within the exercise tags
+  CC_REGEX = /[\w-]+-ch\d{2}-s\d{2}/
```
jpslav: Would be kind of cool if this lived in that new `ConceptCoach` namespace right next to where those tags are generated.

jpslav: If each objective type had a "Tagger" object with "generate" and "match" methods, that might clean up this class a bit and make the regex's reusable from other places like the book import (i.e. to check that an APLO is in fact properly formatted for a book location).  There could also be a top level "get_type" method that returns `:cc`, `:aplo`, `:lo`, or `:generic`.

app/subsystems/content/import_book.rb
Dantemss: Is objective really the proper name anymore, considering these can be LO's or APLO's or CC tags?

app/external/openstax/exercises/v1/exercise.rb
Dantemss: Hmm need some more explanation... keep in mind these tags (in the exercise wrapper) are coming from the spreadsheet, tutor didn't make them.

Dantemss:
Going to rebase this off of https://github.com/openstax/tutor-server/pull/711 because I know they will conflict :)

PR #709, opened 2015-10-13
WIP Load JS API from accounts for FE use

nathanstitt:
Goes along with https://github.com/openstax/accounts/pull/202 and https://github.com/openstax/tutor-js/pull/810

In order to test logins you need a real accounts instance configured with the above PR and it's oath linked to tutor-server.  Accounts stubbing must also be turned off in `secrets.yml`

screen shot 2015-10-15 at 9 35 17 am


nathanstitt:
Replaced by #774

Issue #684, opened 2015-09-30
Dry up User::User factory creation in specs


jpslav:
See https://github.com/openstax/tutor-server/pull/672#discussion_r40735119 

Search through the code (or just look in the PR above) to find the places where we need to collapse 3 lines into one.

s019

PR #712, opened 2015-10-14
Teachers manage sections (periods) of courses

joemsak:


app/controllers/api/v1/periods_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  resource_description do
+    api_versions "v1"
+    short_description 'Represents course periods in the system'
+    description <<-EOS
+      Period description to be written...
+    EOS
+  end
+
+  api :POST, '/courses/:course_id/periods(/role/:role_id)',
```
jpslav: Don't need the optional role arguments  here

app/controllers/api/v1/periods_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  private
+  def find_course
+    @course = Entity::Course.find(params[:course_id])
+  end
+
+  def find_period
+    @period = GetPeriod[id: params[:id]].to_model
+  end
+
+  def ensure_requestor_is_course_teacher
```
jpslav: this belongs in an access policy and then each action needs a variant of:

```ruby
OSU::AccessPolicy.require_action_allowed!(:create, current_api_user, period)
```

app/controllers/api/v1/periods_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  api :POST, '/courses/:course_id/periods(/role/:role_id)',
+             'Returns a new course period for given course'
+  description <<-EOS
+    #{json_schema(Api::V1::PeriodRepresenter, include: :readable)}
+  EOS
+  def create
+    period = CreatePeriod[course: @course, name: period_params[:name]]
+    respond_with period, represent_with: Api::V1::PeriodRepresenter, location: nil
```
jpslav: This (amazingly) may be one of first times that we need to hand back an error to the FE (if the period name is invalid, already taken, etc).  Ideas on how to address this? might need to powwow with FE

app/routines/get_period.rb
```diff
@@ -0,0 +1,3 @@
+class GetPeriod
+  lev_routine delegates_to: CourseMembership::GetPeriod
```
jpslav: We no longer need to wrap routines in this way, at least in this circumstance.  Since the `CourseMembership` SS now exposes a `Period` PORO which is returned by the new `CourseMembership::GetPeriod`, that routine can be used directly.

We still want to write routines at the `app/routines` level when they do (or might soon do) more than what a SS knows.

app/subsystems/course_membership/get_period.rb
```diff
@@ -0,0 +1,9 @@
+class CourseMembership::GetPeriod
+  lev_routine express_output: :period
+
+  protected
+  def exec(id:)
+    period = CourseMembership::Models::Period.find(id)
+    outputs.period = CourseMembership::Period.new(period)
```
jpslav: because this routine doesn't expose a model object to the outside world (directly), it can be called directly without an `app/routines/get_period.rb`.

app/controllers/api/v1/periods_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+                                 responder: ResponderWithPutContent
+  end
+
+  private
+  def find_course
+    @course = Entity::Course.find(params[:course_id])
+  end
+
+  def find_period
+    @period = GetPeriod[id: params[:id]].to_model
```
jpslav: Don't think we should have `to_model` here -- the `CourseMembership::Models::Period` should ideally only be used within the `CourseMembership` SS.  Change this method to return the PORO and change `UpdatePeriod` to accept the PORO as an argument.  `UpdatePeriod` can hide the model use from this code.

spec/controllers/api/v1/periods_controller_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'rails_helper'
+
+RSpec.describe Api::V1::PeriodsController, type: :controller, api: true, version: :v1 do
+  let(:course) { CreateCourse[name: 'Biology I'] }
+  let(:other_course) { CreateCourse[name: 'Other course'] }
+
+  let(:teacher_profile) { FactoryGirl.create(:user_profile) }
+
+  let(:teacher_user) {
+    User::User.new(strategy: User::Strategies::Direct::User.new(teacher_profile))
```
jpslav: Will you be working https://github.com/openstax/tutor-server/issues/684 soon so we don't have to have these kinds of user creation lets?

spec/support/without_exception.rb
```diff
@@ -0,0 +1,7 @@
+module WithoutException
+  def rescuing_exceptions(&block)
+    OpenStax::RescueFrom.configure { |c| c.raise_exceptions = false }
+    yield
+    OpenStax::RescueFrom.configure { |c| c.raise_exceptions = true }
+  end
```
jpslav: Cool :-)

`OpenStax::RescueFrom` should by definition be rescuing exceptions and not reraising them, but are there any cases where an exception might still be raised (e.g. say the logger is misconfigured in the gem and it raises an exception when logging is attempted)?

If so, this might need to be something like

```ruby
begin
  OpenStax::RescueFrom.configure { |c| c.raise_exceptions = false }
  yield
ensure
  OpenStax::RescueFrom.configure { |c| c.raise_exceptions = true }
end
```

And come to think of it, we shouldn't be setting `raise_exceptions` to true at the end, we should be setting it back to whatever it was when we entered this method:

```ruby
original_raise_exceptions = OpenStax::RescueFrom.configuration.raise_exceptions
begin
  OpenStax::RescueFrom.configuration.raise_exceptions = false
  yield
ensure
  OpenStax::RescueFrom.configuration.raise_exceptions = original_raise_exceptions
end
```

The above also sets / gets the values directly without using a block.

When this settles, can you add this module to the rescue_from gem itself, so that in all projects we can say:

```ruby
RSpec.configure do |config|
  config.include OpenStax::RescueFrom::WithoutException
```

jpslav:
Looking good, @joemsak.  Will wait to merge until comments addressed and the endpoints handle errors in the client input (name uniqueness, etc).

app/controllers/api/v1/periods_controller.rb
joemsak: Sweet, wasn't sure if I'd have to add the support or drop the docs, was waiting on your word ;)

app/controllers/api/v1/periods_controller.rb
joemsak: Right, right, I forgot:)

spec/controllers/api/v1/periods_controller_spec.rb
joemsak: You know it!

spec/support/without_exception.rb
joemsak: The best part is the name of it :D

app/access_policies/course_access_policy.rb
```diff
@@ -1,13 +1,13 @@
...
-    case action
+    case action.to_sym
     when :index
       !requestor.is_anonymous?
     when :read, :task_plans
       requestor.is_human? && \
       (UserIsCourseStudent[user: requestor, course: course] || \
        UserIsCourseTeacher[user: requestor, course: course])
-    when :export, :roster
+    when :export, :roster, :update, :create
```
jpslav: period and course access policies are different.  Teachers cannot yet create courses on their own.  The periods controller needs to map to its own access policy.

app/controllers/api/v1/periods_controller.rb
```diff
@@ -0,0 +1,56 @@
+class Api::V1::PeriodsController < Api::V1::ApiController
+  before_filter :find_period_and_course
+
+  before_filter -> {
+    OSU::AccessPolicy.require_action_allowed!(:add_period, current_human_user, @course)
```
jpslav: this before_filter will only ever be called in the `create` action, so to me it is clearer to put this line as the first line of the create action.  Ditto for the `update` action.

jpslav: Make this change, then we can merge.  And the period deletion and error responses can be in a subsequent PR.

PR #714, opened 2015-10-14
Update robots.txt

Dantemss:
Phil made this branch but never PR'd.


jpslav:
Hmmm... don't think we want to ban spiders from the home page or other home-page-like pages.

Dantemss:
Could just disallow the api route then...

PR #715, opened 2015-10-15
Concept coach book import

nathanstitt:
Does not create any tasks.   Should it do so?


Dantemss:
https://github.com/openstax/tutor-server/blob/master/app/subsystems/content/routines/import_book_part.rb does not expect pages outside of chapters (preface)

Just have to handle that case. Probably add a switch statement in the "else" section.

nathanstitt:
Looks like the derived copy doesn't have the error.

jpslav:
Thanks @nathanstitt for taking the lead on this!

PR #718, opened 2015-10-15
Self Enrollment

Dantemss:
https://github.com/openstax/napkin-notes/blob/master/jp/self_enrollment.md

- Added `is_concept_coach` flag to courses
- Added dummy CC assistant for CC courses
- EnrollmentChange model/migration
- EnrollmentChange create endpoint/routine
- EnrollmentChange approve endpoint/routine
- Specs


jpslav:
Haha @Dantemss sneaks in the removal of course assistants ;-)

Dantemss:
Since you said we might need them later... I will leave the course assistants in.

app/representers/api/v1/enrollment/period_with_course_representer.rb
```diff
@@ -0,0 +1,38 @@
+module Api::V1::Enrollment
+  class PeriodWithCourseRepresenter < Roar::Decorator
```
jpslav: Since period and course are peers here, can this be `PeriodAndCourseRepresenter`?

app/access_policies/enrollment_change_access_policy.rb
```diff
@@ -0,0 +1,14 @@
+class EnrollmentChangeAccessPolicy
+  def self.action_allowed?(action, requestor, enrollment_change)
+    return false if requestor.is_anonymous? || !requestor.is_human?
+
+    case action
+    when :create
+      true # A (normal) user is always allowed to create EnrollmentChanges for themselves
+    when :approve
+      requestor == enrollment_change.profile || requestor.to_model == enrollment_change.profile
```
jpslav: Can there be an `EnrollmentChange` PORO, and then can this not be aware of `profile`?

app/subsystems/course_membership/approve_enrollment_change.rb
```diff
@@ -0,0 +1,39 @@
...
+    fatal_error(code: :already_rejected,
+                message: 'The given enrollment change request has already been rejected') \
+      if enrollment_change.rejected?
+
+    enrollment = enrollment_change.enrollment
+    if enrollment.nil?
+      # New student
+      profile = enrollment_change.profile
+      strategy = ::User::Strategies::Direct::User.new(profile)
+      user = ::User::User.new(strategy: strategy)
```
jpslav: Here's another reason to have an `EnrollmentChange` PORO -- would let us replace these lines with `user = enrollment_change.user`

app/subsystems/course_membership/approve_enrollment_change.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    enrollment_change.approve_by(approved_by).save!
+
+    # Mark other pending EnrollmentChange records as rejected
+    CourseMembership::Models::EnrollmentChange.where(
+      user_profile_id: enrollment_change.user_profile_id,
+      status: CourseMembership::Models::EnrollmentChange.statuses[:pending],
+      course_membership_period_id: enrollment_change.period.course.periods.map(&:id)
+    ).update_all(status: CourseMembership::Models::EnrollmentChange.statuses[:rejected])
```
jpslav: Not that I'm opposed to this code, but what's the reason for it?

app/subsystems/course_membership/models/enrollment_change.rb
```diff
@@ -0,0 +1,39 @@
...
+  enum status: [ :pending, :approved, :rejected ]
+
+  validates :profile, presence: true
+  validates :enrollment, uniqueness: { allow_nil: true }
+  validates :period, presence: true
+  validate :same_profile, :different_period
+
+  def from_period
+    enrollment.try(:period)
+  end
```
jpslav: Why does `from_period` give back `period` but the comment next to `belongs_to :period` says "to"?

app/subsystems/course_membership/approve_enrollment_change.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    outputs[:enrollment_change] = enrollment_change
```
jpslav: A PORO would also let all of these more public routines not have to return what should be a private model

app/access_policies/enrollment_change_access_policy.rb
Dantemss: Do you want a `can_approve?(user)` method or something?

app/representers/api/v1/enrollment/period_with_course_representer.rb
Dantemss: I had it named like that originally, but changed to `PeriodWithCourseRepresenter` because it is passed a period (not a course...) but we can discuss.

app/subsystems/course_membership/approve_enrollment_change.rb
Dantemss: I guess :)

app/subsystems/course_membership/approve_enrollment_change.rb
Dantemss: Old enrollment change requests will refer to the wrong `enrollment` object so I want to prevent them from being approved (they will most likely blow up).

app/subsystems/course_membership/models/enrollment_change.rb
Dantemss: `from_period` comes from the `enrollment` object
`to_period` comes from the `period_id` field

app/subsystems/course_membership/approve_enrollment_change.rb
Dantemss: Fine I'll do it :)

app/access_policies/enrollment_change_access_policy.rb
Dantemss: But I thought access policies were made to extract out those `can_x?` methods...

app/subsystems/course_membership/create_enrollment_change.rb
```diff
@@ -0,0 +1,44 @@
...
+      user_profile_id: user.id, enrollment: enrollment, period: period.to_model,
+      requires_enrollee_approval: requires_enrollee_approval
+    )
+    transfer_errors_from(enrollment_change_model, {type: :verbatim}, true)
+
+    strategy = CourseMembership::Strategies::Direct::EnrollmentChange.new(enrollment_change_model)
+    enrollment_change = CourseMembership::EnrollmentChange.new(strategy: strategy)
+
+    run(:approve, enrollment_change: enrollment_change, approved_by: user) \
+      unless requires_enrollee_approval
```
jpslav: If approval is not required, I don't think that we should need to run an `approve` routine (and shouldn't need to mark that a user approved when they actually didn't).  More that we want a `ProcessEnrollmentChange` routine that returns immediately if the change is not approved.  

So on these lines we can just say:

```ruby
# Will be a no-op if not approved
run(:process_enrollment_change, enrollment_change: enrollment_change)
```

```ruby
class CourseMembership::ProcessEnrollmentChange
  ...
  def exec(enrollment_change:)
    # return if not approved, otherwise make the change happen
```

At the end of `ApproveEnrollmentChange`, call `ProcessEnrollmentChange`.  Otherwise `ApproveEnrollmentChange` just sets who approved it and saves it.

app/subsystems/course_membership/get_period.rb
```diff
@@ -3,7 +3,8 @@ class CourseMembership::GetPeriod
 
   protected
   def exec(id:)
-    period = CourseMembership::Models::Period.find(id)
-    outputs.period = CourseMembership::Period.new(period)
+    model = CourseMembership::Models::Period.find(id)
+    strategy = CourseMembership::Strategies::Direct::Period.new(model)
+    outputs.period = CourseMembership::Period.new(strategy: strategy)
```
jpslav: Not necessarily for this PR, but it would be cool if we could skip the strategy line in this common pattern:

```ruby
model = CourseMembership::Models::Period.find(id)
outputs.period = CourseMembership::Period.new(strategizable: model)
```

where the `wrapped_by CourseMembership::Strategies::Direct::Period` command would in effect provide the following method dynamically:

```ruby
# in the Period model
def strategize!
  CourseMembership::Strategies::Direct::Period.new(self)
end
```

which could be called in the PORO initializer:

```ruby
class CourseMembership::Period
  def initialize(strategy:, strategizable:)
    @strategy = strategy.nil? ? @strategizable.strategize! : @strategy
  end
```

app/controllers/api/v1/enrollment_changes_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+      EnrollmentChanges indicate that a user has requested to change their status in a course.
+      The changes currently need to be approved by the requesting user to be effective.
+    EOS
+  end
+
+  api :POST, '/enrollment_changes',
+             'Creates a new EnrollmentChange request or updates the current one'
+  description <<-EOS
+    Creates a new EnrollmentChange object, indicating the user's intention to enroll in a course
+    or to switch periods.
```
jpslav: Can you extend the description to describe the possible output error codes?

app/controllers/api/v1/enrollment_changes_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    else
+      render_api_errors(:already_enrolled)
+    end
+  end
+
+  api :PUT, '/enrollment_changes/:enrollment_change_id/approve',
+            'Approves an EnrollmentChange request'
+  description <<-EOS
+    Approves an EnrollmentChange object, causing the user's enrollment status to update.
+
```
jpslav: Ditto on adding description of possible error codes.

app/representers/api/v1/enrollment/period_with_course_representer.rb
jpslav: Works for me

app/subsystems/course_membership/approve_enrollment_change.rb
```diff
@@ -0,0 +1,38 @@
...
+    if enrollment.nil?
+      # New student
+      run(:add_student, user: enrollment_change.user, period: enrollment_change.to_period)
+    else
+      # Existing student
+      student = enrollment.student
+      run(:add_enrollment, student: student, period: enrollment_change.to_period)
+    end
+
+    enrollment_change_model.approve_by(approved_by).save!
```
jpslav: See other discussion about actually processing the change; in any case, the processing should only happen after approval (below this line).

jpslav:
Looks good Dante.

app/subsystems/course_membership/get_period.rb
Dantemss: Why do we introduce more layers of abstraction only to bypass them after anyway?

app/controllers/api/v1/enrollment_changes_controller.rb
Dantemss: Sure

app/controllers/api/v1/enrollment_changes_controller.rb
Dantemss: Done

lib/entity/class_methods.rb
```diff
@@ -9,6 +9,10 @@ module ClassMethods
     # Required for automatic wrapping
     def wrapped_by(entity_class = nil, &block)
       Entity._wrap_class_procs[self.name] = block || lambda { |instance| entity_class }
+
+      define_method :wrap do
+        Entity._wrap(self)
+      end
```
jpslav: Neat.

PR #719, opened 2015-10-16
WIP: Provide "metadata" for courses

nathanstitt:
The goal of this is to provide the FE information to style courses with.  Right now it's just using the first word of the course name, which is far from ideal.  This adds a `config/book-metadata-mapping.yml` file which contains a book UUID -> type mapping.

This isn't ideal either but I think it's the "simplest thing that could possibly work".  I chatted with the content folks in an attempt to find built-in field on the book's JSON that we get the information from, and we came to the conclusion that there really wasn't anything that would work other than just the UUID.  All the other fields can and do vary to much between books.

Two ways I could see to improve this would be to either:
  * Read the YAML file when the book is imported and store the information on the books model.  That's not really much better to me because it's storing information in the db over and over for no real reason.

  * Alter the ecosystem admin screen to include a drop-down to set the type from a pre-vetted list.  The demo scripts could supply it along with their yaml.  **I think this is the better way to go**, but wanted to run it by others before I did that.

### DIFF
--- a/app/controllers/api/v1/courses_controller.rb
+++ b/app/controllers/api/v1/courses_controller.rb
@@ -17,7 +17,7 @@ class Api::V1::CoursesController < Api::V1::ApiController
   def index
     OSU::AccessPolicy.require_action_allowed!(:index, current_api_user, Entity::Course)
     courses_info = CollectCourseInfo[user: current_human_user,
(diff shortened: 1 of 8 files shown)


nathanstitt:
Closing in favor of new "Course Catalog" subsystem.

PR #720, opened 2015-10-19
Ecosystem import job

karenc:
- Use background job for admin ecosystem importing textbook

  Use `Marshal.dump` when passing in the tag_generator to
  `FetchAndImportBookAndCreateEcosystem`.  An instance of the concept
  coach tag generator cannot be used in background jobs:

  ```Unsupported argument type: ConceptCoach::TagGenerator```

- Store the book url with the ecosystem import background job

- Display incomplete ecosystem import jobs on admin ecosystems page

  Display job id (with link to the job details page), job status and link
  to the book in archive.

- Change ecosystems index page to use tabs

  "List of Books" and "Incomplete Imports".

  Add note to "Incomplete Imports" table to say that the table does not
  automatically refresh.

### DIFF
--- a/app/controllers/admin/ecosystems_controller.rb
+++ b/app/controllers/admin/ecosystems_controller.rb
@@ -1,6 +1,9 @@
 class Admin::EcosystemsController < Admin::BaseController
   def index
     @ecosystems = Content::ListEcosystems[]
+    @incomplete_jobs = Lev::BackgroundJob.incomplete.select do |job|
+      job.respond_to?(:ecosystem_import_url)
+    end
   end
 
   def import
--- a/app/controllers/admin/ecosystems_controller.rb
+++ b/app/controllers/admin/ecosystems_controller.rb
@@ -18,6 +18,9 @@ class Admin::EcosystemsController < Admin::BaseController
         book_cnx_id: params[:cnx_id],
         tag_generator: Marshal.dump(ConceptCoach::TagGenerator.new(params[:cc_tag]))
       )
+      job = Lev::BackgroundJob.find(job_id)
+      import_url = OpenStax::Cnx::V1.url_for(params[:cnx_id])
+      job.save(ecosystem_import_url: import_url)
       flash[:notice] = 'Ecosystem import job queued.'
     end
     redirect_to admin_ecosystems_path
--- a/app/controllers/admin/ecosystems_controller.rb
+++ b/app/controllers/admin/ecosystems_controller.rb
@@ -14,11 +14,11 @@ class Admin::EcosystemsController < Admin::BaseController
     archive_url = params[:archive_url].present? ? params[:archive_url] : @default_archive_url
 
     OpenStax::Cnx::V1.with_archive_url(url: archive_url) do
-      ecosystem = FetchAndImportBookAndCreateEcosystem[
+      job_id = FetchAndImportBookAndCreateEcosystem.perform_later(
         book_cnx_id: params[:cnx_id],
-        tag_generator: ConceptCoach::TagGenerator.new(params[:cc_tag])
-      ]
-      flash[:notice] = "Ecosystem \"#{ecosystem.title}\" imported."
+        tag_generator: Marshal.dump(ConceptCoach::TagGenerator.new(params[:cc_tag]))
+      )
+      flash[:notice] = 'Ecosystem import job queued.'
     end
     redirect_to admin_ecosystems_path
   end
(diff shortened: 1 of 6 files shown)


app/views/manager/ecosystems/_incomplete.html.erb
```diff
@@ -0,0 +1,19 @@
+<h2>Incomplete Ecosystem Imports</h2>
+
```
jpslav: Can you add a note that says that this table does not automatically refresh (possible that some might expect it to)

app/controllers/admin/ecosystems_controller.rb
```diff
@@ -14,11 +17,14 @@ def import_ecosystem
     archive_url = params[:archive_url].present? ? params[:archive_url] : @default_archive_url
 
     OpenStax::Cnx::V1.with_archive_url(url: archive_url) do
-      ecosystem = FetchAndImportBookAndCreateEcosystem[
+      job_id = FetchAndImportBookAndCreateEcosystem.perform_later(
         book_cnx_id: params[:cnx_id],
-        tag_generator: ConceptCoach::TagGenerator.new(params[:cc_tag])
```
jpslav: I know why you are passing in the `cc_tag` here instead of the `tag_generator`, but I would very much like to keep the knowledge of concept coach stuff at this level of the code and not inside the Content subsystem.

We can do this by marhsalling the `tag_generator` to a string and then unmarshalling it inside the routine:

```ruby
# EcosystemsController
job_id = FetchAndImportBookAndCreateEcosystem.perform_later(
  ...
  tag_generator: Marshal.dump(ConceptCoach::TagGenerator.new(params[:cc_tag]))
)
```

```ruby
# FetchAndImportBookAndCreateEcosystem
def exec(..., tag_generator: nil)
  ...
  tag_generator = Marshal.load(tag_generator) if tag_generator.is_a?(String)
```

We should make marshalling and unmarshalling standard procedure for any routine that we might call in a background job that takes non-primitive, non-activerecord arguments.

app/routines/fetch_and_import_book_and_create_ecosystem.rb
```diff
@@ -8,10 +8,11 @@ class FetchAndImportBookAndCreateEcosystem
   protected
 
   # Returns a Content::Ecosystem containing a book obtained from the given CNX id
-  def exec(book_cnx_id:, ecosystem_title: nil, exercise_uids: nil, tag_generator: nil)
+  def exec(book_cnx_id:, ecosystem_title: nil, exercise_uids: nil, tag_generator: nil, cc_tag: nil)
     cnx_book = OpenStax::Cnx::V1.book(id: book_cnx_id)
     eco_title ||= "#{cnx_book.title} (#{cnx_book.uuid}@#{cnx_book.version}) - #{Time.now.utc}"
     outputs[:ecosystem] = Content::Ecosystem.create!(title: eco_title)
+    tag_generator ||= ConceptCoach::TagGenerator.new(cc_tag) unless cc_tag.nil?
```
jpslav: See comment above.  Also, don't need `||=` here.

jpslav:
Thanks @karenc! Looks great.  Just a few comments.  Oh, and does Lev's automatic reporting of errors to the background job handle the "get any errors out to the admins" issue?

app/controllers/admin/ecosystems_controller.rb
Dantemss: Wait... why is this needed at all? Is ActiveJob unable to automatically serialize non-ActiveRecord stuff?

karenc:
Added note to say the incomplete imports table does not automatically refresh.

Also changed the ecosystems index page to use tabs:

ecosystems1

ecosystems2

app/controllers/admin/ecosystems_controller.rb
karenc: Yeah I got this error: `Unsupported argument type: ConceptCoach::TagGenerator`

karenc:
The error isn't really clear... I put in an invalid uuid for the cnx book id and the error on the job details page is "exception - Can't convert NilClass into String.".

@jpslav Do you have an example book that has errors in the content?

May be I can add a link to papertrail (all systems) that searches for the error message on the job details page?

jpslav:
@karenc I think if the routine actually added an error instead of just raising an exception (which it does in places), the message will be more meaningful.  I don't have an example of erroring book off hand.  If we really need it I bet we could find an problematic version of a book, but I think we can pass for the moment.

app/controllers/admin/ecosystems_controller.rb
jpslav: I don't think so @Dantemss -- ActiveJob only supports serialization of ActiveRecords, and only then because they support GlobalID.

PR #722, opened 2015-10-19
Adds stubbed endpoint for getting a concept coach task

jpslav:
* Hit `/api/cc/tasks/CNX_COLLECTION_UUID/CNX_MODULE_UUID` and you'll get back a `Task` with normal task JSON.  
*The task will have three exercises each with fake content.  
* The UUIDs don't need to be real (can hit `/api/cc/tasks/foo/bar` if you want) for now, but later will need to be real UUIDs without version info.  
* If you hit the same endpoint twice, the first time the CC task will be generated anew and returned.  The second time you'll get back the task you got the first time (it is lazily created on the first call).  
* Different users calling the same endpoint will get different tasks.  
* This stub stores the task ID in the session now for a given user/UUID combination, so it will likely not survive logouts (can't remember exactly how brutal we are to the session on logout).  
* Some errors will be returned (e.g. 403 if not logged in, 422 if don't provide right args), but not all errors implemented yet (e.g. we don't check that the UUIDs match an ecosystem/course/student, etc).
* The task JSON is what you've seen before (with a different type).  In the real implementation there will be some new fields, and possibly some fields removed (TBD).

cc @openstax/tutor-fe-dev

### DIFF
--- a/app/controllers/api/v1/cc/tasks_controller.rb
+++ b/app/controllers/api/v1/cc/tasks_controller.rb
@@ -1,0 +1,75 @@
+class Api::V1::Cc::TasksController < Api::V1::ApiController
+
+  resource_description do
+    api_versions "v1"
+    short_description 'Represents a concept coach task'
+    description <<-EOS
+      see `/api/tasks`
+    EOS
+  end
+
+  ###############################################################
+  # show
+  ###############################################################
+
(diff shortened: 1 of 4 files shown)


jpslav:
@karenc can you review - you may end up doing some of the work to move this from stub to real implementation.

philschatz:
Thanks!
cc @nathanstitt since he's been working on pulling the exercise list and @pandafulmanda since she's been working on rendering the task

karenc:
It looks good to me.  Should I merge it?

PR #723, opened 2015-10-20
moves external classes to /lib, resolves #494

### LINKED ISSUE #494: Move configurable classes from app to lib

jpslav:


jpslav:
Used explicit listing of `require` calls to handle load order dependencies.

jpslav:
@Dantemss please review.

Dantemss:
Eww can I try to fix this mess of requires instead? :)

Dantemss:
IMO the `requires.rb` should be renamed to the "fake gem" names (biglearn.rb etc)
The `requires.rb` should only require the "fake_gem_name/v1"
The `v1.rb's` should require the other files as appropriate.

jpslav:
That would be fine.

jpslav:
How's that @Dantemss?

Dantemss:
Not sure I like the requires in a .each but it's way better I think

jpslav:
Fixed.

PR #724, opened 2015-10-20
Add existing "whole course" task plans to new periods

joemsak:


app/subsystems/tasks/assign_coursewide_tasking_plans.rb
```diff
@@ -0,0 +1,44 @@
+module Tasks
+  class AssignCoursewideTaskingPlans
```
jpslav: I think the routine name should refer to `TaskPlans` instead of `TaskingPlans`.  You can't assign a tasking plan.

Also, might be better to be more explicit in the name talking about where assigning:

`AssignCoursewideTaskPlansToNewPeriod`

joemsak: I felt like the name was already long enough :/

app/subsystems/tasks/assign_coursewide_tasking_plans.rb
```diff
@@ -0,0 +1,44 @@
...
+    def task_plans_with_matching_tasking_plans_across_periods
+      task_plans_across_periods.reject do |task_plan|
+        target_ids = task_plan.tasking_plans.flat_map(&:target_id).uniq.sort
+        period_ids = existing_period_ids.uniq.sort
+        due_dates = task_plan.tasking_plans.flat_map(&:due_at).uniq
+        open_dates = task_plan.tasking_plans.flat_map(&:opens_at).uniq
+
+        target_ids != period_ids || due_dates.size != 1 || open_dates.size != 1
+      end
+    end
```
jpslav: Someone once told me that private methods clear things up.  See below.  Also, don't need the uniq or flat map on the IDs.

```ruby
    def task_plans_with_matching_tasking_plans_across_periods
      task_plans_across_periods.select do |task_plan|
        assigned_to_all_existing_periods?(task_plan) &&
        tasking_plans_have_same_dates(task_plan)
      end
    end

    def assigned_to_all_existing_periods?(task_plan)
      (tasking_plans.collect(&:id) - existing_period_ids).empty?
    end

    def tasking_plans_have_same_dates?(task_plan)
      tasking_plans.collect(&:due_at).uniq.one? &&
      tasking_plans.collect(&:opens_at).uniq.one?
    end
```

jpslav:
Looks good.

PR #731, opened 2015-10-22
Add remove teacher API

karenc:


app/controllers/api/v1/courses_controller.rb
```diff
@@ -64,6 +64,15 @@ def dashboard
     respond_with data, represent_with: Api::V1::Courses::DashboardRepresenter
   end
 
+  api :DELETE, '/courses/:course_id/teachers/:role_id', 'Removes a teacher from a course'
+  description <<-EOS
+  EOS
+  def remove_teacher
+    course = Entity::Course.find(params[:id])
```
jpslav: Can you make a teachers controller (similar to students controller) to handle deletion (and later addition, might also soon handle teacher self registration)?  Then this remove_teacher becomes the normal `destroy` action in that controller.

app/access_policies/course_access_policy.rb
```diff
@@ -9,6 +9,10 @@ def self.action_allowed?(action, requestor, course)
        UserIsCourseTeacher[user: requestor, course: course])
     when :export, :roster, :add_period
```
jpslav: Coincidentally, I think this `:roster` action is no longer used (moved to students controller).  Can you move?

app/controllers/api/v1/courses_controller.rb
karenc: Yes.

app/access_policies/course_access_policy.rb
karenc: We still use this permission in students controller: https://github.com/openstax/tutor-server/blob/bb2b4c9240059ad50300f136d61dcb4c813e47fa/app/controllers/api/v1/students_controller.rb#L21

What would you like me to do?

jpslav:
Perfect, thanks @karenc!

PR #732, opened 2015-10-22
Restart rails when files are modified

nathanstitt:
Since hot-reloading is broken, this works okish, it takes about 8 seconds to restart after a files is modified  my machine.  It runs Rails in a sub-process though, which means that `debugger` doesn't work.  `remote-pry` does though but that's kind of a pain.  When I want a debug shell, I usually just run rails normally.

To use, bundle and then run `guard`.  Type `q` and hit enter to exit.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -209,6 +209,9 @@ end
 group :development do
   # Trace AR queries
   gem 'active_record_query_trace'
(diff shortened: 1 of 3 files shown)


nathanstitt:
Closing because we've now discovered the root cause of reload failures

PR #735, opened 2015-10-23
API to delete periods

joemsak:


app/subsystems/course_membership/delete_period.rb
```diff
@@ -0,0 +1,11 @@
+module CourseMembership
+  class DeletePeriod
+    lev_routine express_output: :deleted
+
+    protected
+    def exec(period:)
+      outputs.deleted = period.to_model.destroy
+      outputs.errors = period.to_model.errors
```
jpslav: `transfer_errors_from(outputs.deleted, {verbatim: true}, true)` handles moving errors over for you.

app/controllers/api/v1/periods_controller.rb
```diff
@@ -34,6 +34,19 @@ def update
                                  responder: ResponderWithPutContent
   end
 
+  api :DELETE, '/periods/:id',
+               'Deletes a period for authorized teachers'
+  def destroy
+    OSU::AccessPolicy.require_action_allowed!(:destroy, current_human_user, @period)
+    delete_period = CourseMembership::DeletePeriod.call(period: @period).outputs
```
jpslav: hmm... I think `delete_period` is the Lev outputs object right? (not the actual period)  So maybe the variable here shouldn't be `delete_period` (which sounds like a verb / command), but something like `is_deleted`?

You could simplify and say:

```ruby
is_deleted = CourseMembership::DeletePeriod[period: @period]

if is_deleted
  ...
else
  ...
end
```

Actually looking at the `render json` statement below, you should instead do:

```ruby
result = CourseMembership::DeletePeriod.call(period: @period)

if result.errors.any?
  render_api_errors(result.errors)
else
  head :no_content
end
```

app/subsystems/course_membership/delete_period.rb
jpslav: Oops misread as `deleted` being the object, not a flag.  Should be:

```ruby
transfer_errors_from(period.to_model, {verbatim: true}, true)
```

app/subsystems/course_membership/period.rb
```diff
@@ -3,7 +3,8 @@ class Period < Entity
 
     wraps CourseMembership::Models::Period
 
-    exposes :course, :name, :student_roles, :teacher_roles, :enrollment_code
+    exposes :course, :name, :student_roles, :teacher_roles, :enrollment_code, :valid?,
+      :errors
```
jpslav: Can remove errors based on the approach above.

jpslav:
Looking good.  Review complete.  Liking the simpleness of the setup routine calls.

joemsak:
@jpslav awaiting green CI

PR #736, opened 2015-10-23
WIP: Update PG gem version

nathanstitt:
Test to see if it fixes compilation error when deploying to linux host.


nathanstitt:
I don't think this is needed any longer

PR #737, opened 2015-10-23
Add API to undrop students from a course

karenc:


app/controllers/api/v1/students_controller.rb
```diff
@@ -96,6 +96,21 @@ def destroy
...
+    Undrop a student from their course.
+  EOS
+  def undrop
+    OSU::AccessPolicy.require_action_allowed!(:destroy, current_api_user, @student)
+    result = ActivateStudent.call(student: @student)
+
+    if result.errors.any?
+      render_api_errors(result.errors)
+    else
+      head :no_content
```
jpslav: While this response is acceptable for normal PUTs, our FE folks prefer to get the updated record back after a PUT or PATCH.  So you can probably copy the content in lines 78-80 and put them here.

app/routines/activate_student.rb
```diff
@@ -0,0 +1,8 @@
+class ActivateStudent
+  lev_routine express_output: :student
```
jpslav: Thanks for making a routine.

app/routines/activate_student.rb
```diff
@@ -0,0 +1,8 @@
...
+
+  def exec(student:)
+    student.activate.save!
```
jpslav: if you want to use the error reporting mechanisms from lev, you'll want to use `save` instead of `save!` so you get errors back instead of exceptions.  

You should also add `transfer_errors_from(student, {type: :verbatim}, true)` after the save call so that errors are passed back non-nested.

app/controllers/api/v1/students_controller.rb
```diff
@@ -96,6 +96,21 @@ def destroy
...
+  api :PUT, '/students/:student_id/undrop', 'Undrop a student from their course'
+  description <<-EOS
+    Undrop a student from their course.
+  EOS
+  def undrop
+    OSU::AccessPolicy.require_action_allowed!(:destroy, current_api_user, @student)
+    result = ActivateStudent.call(student: @student)
+
+    if result.errors.any?
+      render_api_errors(result.errors)
```
jpslav: I am starting to worry a bit about information leakage to end users in our use of render_api_errors (too much data they don't need to see or shouldn't see).  But this isn't your problem -- you are following the conventions used elsewhere.  Just mentioning it as I start to think about it :-)

karenc:
Ready for review again.

PR #738, opened 2015-10-26
Allow Content Analyst to import ecosystems

karenc:
- Add content analyst console

- Move admin ecosystems controller to Manager::EcosystemsAction

  This allows sharing of code between the admin ecosystems page, the
  content analyst ecosystems page and the customer service ecosystems
  page.

- Move admin ecosystems index template to manager/ecosystems

  Allow sharing between admin, customer service and content analyst
  consoles.

- Move admin ecosystems import template to manager/ecosystems

  So that it can be shared between the admin and content analyst consoles.

- Add ecosystems page for content analysts

- Add spec for content analyst managing ecosystems


app/views/manager/ecosystems/_listing.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      <th>Book UUID</th>
+      <th>Book Version</th>
+      <th>Link to Archive</th>
+      <th>Import Date</th>
+    </tr>
+  </thead>
+  <tbody>
+  <% ecosystems.each do |ecosystem| %>
+    <tr>
+      <td><%= ecosystem.books.first.title %></td>
```
joemsak: looks like this blows up for me when there are no books on the ecosystem... may wish to add some sort of composition method that guards from empty books?

```ruby
XX::Ecosystem.def primary_book(attribute)
  return nil if books.empty?
  books.first.send(attribute)
end
```

karenc: Thanks, I'll fix it

joemsak: FWIW dante says there can't be an ecosystem without books
my point being it might be specifically my db that has corrupted data
however it's still worth taking a look at it from a "law of demeter" pov

karenc: We decided that it doesn't need to be fixed because an ecosystem should always have at least one book.

joemsak: :+1:

karenc:
A link to the content analyst console will need to be added to FE menu dropdown.

philschatz:
Can you add a comment to openstax/tutor-js#823 when this is ready to be merged?

jpslav:
Looks good @karenc -- can merge after rebase.

karenc:
Rebased.

PR #741, opened 2015-10-26
Support fetching cc exercises via CORS

nathanstitt:
Since IE 11 supports CORS I'd suggest this is the way to go unless we really, really need to support IE 10.

One braindead thing that IE 11 does is that it doesn't treat requests to the same host but different ports as a cross-site request.  That means you can't test by just changing the ports on localhost :(

I've setup a proper test page at: http://nathan.stitt.org/ox/

### DIFF
--- a/app/controllers/api/v1/cc/tasks_controller.rb
+++ b/app/controllers/api/v1/cc/tasks_controller.rb
@@ -1,5 +1,8 @@
 class Api::V1::Cc::TasksController < Api::V1::ApiController
 
+  after_filter :set_cors_headers
+  skip_before_action :verify_authenticity_token, only: :show
+
   resource_description do
     api_versions "v1"
     short_description 'Represents a concept coach task'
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/cc/tasks_controller.rb
```diff
@@ -1,5 +1,8 @@
 class Api::V1::Cc::TasksController < Api::V1::ApiController
 
+  before_filter :set_cors_headers
+  skip_before_action :verify_authenticity_token, only: :show
```
jpslav: Dante points out that you don't need to do this here, as it is [already taken care of in openstax_api](https://goo.gl/ICmlKi) (skipped when token provided).

jpslav: Please check the other things going on in that linked openstax_api file to see if it changes what all you need to do in this file.

app/controllers/api/v1/cc/tasks_controller.rb
```diff
@@ -72,4 +81,20 @@ def create_fake_concept_coach_task
     task
   end
 
+  def set_cors_headers
+    headers['Access-Control-Allow-Origin']   = validated_cors_origin
+    headers['Access-Control-Allow-Methods']  = 'GET, OPTIONS' # PUT/POST will be added when tasks are created
+    headers['Access-Control-Request-Method'] = '*'
+    headers['Access-Control-Allow-Credentials'] = 'true'
```
Dantemss: If we're doing the token thing, this header is unnecessary (and it's scary in general...)

nathanstitt:
I think there's some confusion by this PR.  Since we're using CORS, we're using cookies instead of  access tokens.  Therefore, we do need the `Access-Control-Allow-Credentials` header since that's what tells the browser that it's ok to send cookies to the server.  Likewise skipping the `verify_authenticity_token` callback is needed `session_user?` is true in `openstax_api`.

I would suggest we could use the `set_cors_preflight_headers` call in `openstax_api`, except it allows all request methods and wildcards the allowed address.  I was under the impression that we explicitly did not want to allow any host access and wanted to restrict to *.cnx.org?  I've done so with the new `cc_origins` configuration in this PR.   I'm not sure where that method is used at but we might revisit it's usage.

nathanstitt:
Closing since we're using tokens for accessing resources

PR #742, opened 2015-10-26
Magic Map round 2

Dantemss:
- Removed magic map
- Pages now have a UUID tag, which is used to pull exercises
- Pages inherit LO's from imported Exercises


Dantemss:
@jpslav What do you think? (I need to regenerate basically all cassettes for the tests to pass, since the page UUID is now included in all requests to Exercises)

app/subsystems/content/import_book.rb
```diff
@@ -1,6 +1,6 @@
 class Content::ImportBook
 
-  MAX_TAG_QUERY_SIZE = 64
+  MAX_TAG_QUERY_SIZE = 32
```
jpslav: Since this number keeps changing, should it be configurable?

app/subsystems/content/routines/import_exercises.rb
```diff
@@ -33,15 +35,22 @@ def exec(ecosystem:, page:, query_hash:)
...
-                               tagging_class: Content::Models::ExerciseTag,
-                               save: false).outputs.taggings
+      exercise.exercise_tags = run(:tag, exercise, relevant_tags,
+                                   tagging_class: Content::Models::ExerciseTag,
+                                   save: false).outputs.taggings
 
-      exercise.exercise_tags = outputs[:taggings]
       outputs[:exercises] << exercise
+
+      new_lo_tags = relevant_tags.select{ |tag| tag.lo? && !tag.pages.include?(exercise_page) }
```
jpslav: Is `&& !tag.pages.include?(exercise_page)` here so that we don't retag a page with tags it already has?  Should this be a function of the tagging routine (to ignore duplicates)?

app/subsystems/content/import_book.rb
Dantemss: It's weird to configure it though... it depends on the tag string length... I'm not even sure what the exact URL limit is...

app/subsystems/content/routines/import_exercises.rb
Dantemss: I could move it, although this is the only place in the code that attempts to re-tag an existing object.

app/subsystems/content/import_book.rb
jpslav: can the number be the URL limit and we compute the query limit based on the tags we have?

Dantemss: OK

app/subsystems/content/routines/import_exercises.rb
Dantemss: Oops seems like the tagging routine already checks for duplicates.

PR #744, opened 2015-10-27
DISCUSS Get OAuth token when have session user

jpslav:
When talking about using JSONP to let CC make API requests from a different domain, I wondered if we might just want to modify Tutor to let the CC FE get an OAuth token for a session user (letting us just use JSONP on this one token route).

This PR implements such an endpoint, and is WIP so we can discuss it (may or may not be what we want); interested in hearing what @openstax/tutor-fe-dev think, as well as @Dantemss.  I am a little nervous about messing around in OAuth-land, as I'm no security guru.

When you have a logged in user, you can POST to the `/oauth/token` endpoint with a body of `{grant_type: 'session'}` and you'll get back an OAuth token.  I have tested this in a spec as well as by using the DHC chrome plugin (when using DHC I had to use a Content-Type of "application/x-www-form-urlencoded" because I couldn't get the JSON POST syntax to work for some reason).   Here's what DHC looks like:

image

Let me know what y'all think.

### DIFF
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -17,7 +17,6 @@ require 'course_guide_methods'
 require 'type_verification'
 require 'tagger'
 require 'concept_coach/tag_generator'
-require 'doorkeeper_session_grant'
 
 %w(
   biglearn
@@ -32,3 +31,5 @@ COPYRIGHT_HOLDER = "Rice University"
 
 # Initialize the Rails application.
 Rails.application.initialize!
+
+require 'doorkeeper_session_grant'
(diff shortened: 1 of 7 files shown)


Dantemss:
Why not just use the OAuth implicit flow? It's designed for use by Javascript etc.

jpslav:
Yeah I guess that is better.  For some reason I had it in my head that the implicit flow wouldn't work for how we authenticate, but rereading https://www.digitalocean.com/community/tutorials/an-introduction-to-oauth-2#grant-type-implicit I guess it does.

jpslav:
Closing because most of this made it into https://github.com/openstax/tutor-server/pull/765

PR #746, opened 2015-10-27
Admin ecosystem comments

joemsak:
Must merge #738 first

* Add comment field
* Put it where the Book UUID currently is
* Move Link to Archive to the right of Import Date so that Book Version and Date are next to each other.
* Add button and modal dialog to show Book UUID all the way to the right


app/controllers/manager/ecosystems_actions.rb
```diff
@@ -0,0 +1,46 @@
...
+  def import
+    @default_archive_url = OpenStax::Cnx::V1.archive_url_base
+    import_ecosystem if request.post?
+  end
+
+  protected
+
+  def import_ecosystem
+    OpenStax::Cnx::V1.with_archive_url(url: archive_url) do
+      job.save(ecosystem_import_url: import_url)
```
jpslav: `job` does not convey enough information IMO.  Also, since this module is imported elsewhere, it isn't very specific and runs the risk of colliding with an existing variable name.

Something like 

```ruby
job = run_import_in_background
job.save(...)
...
```

would read better I think

app/controllers/manager/ecosystems_actions.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+    redirect_to ecosystems_path
+  end
+
+  def archive_url
+    if params[:archive_url].present?
+      params[:archive_url]
+    else
+      @default_archive_url
+    end
```
jpslav:
```ruby
params[:archive_url] || @default_archive_url
```

app/routines/fetch_and_import_book_and_create_ecosystem.rb
```diff
@@ -8,11 +8,15 @@ class FetchAndImportBookAndCreateEcosystem
   protected
 
   # Returns a Content::Ecosystem containing a book obtained from the given CNX id
-  def exec(book_cnx_id:, ecosystem_title: nil, exercise_uids: nil, tag_generator: nil)
+  def exec(book_cnx_id:, ecosystem_title: nil, exercise_uids: nil, tag_generator: nil,
+           comments: nil)
     cnx_book = OpenStax::Cnx::V1.book(id: book_cnx_id)
-    eco_title ||= "#{cnx_book.title} (#{cnx_book.uuid}@#{cnx_book.version}) - #{Time.now.utc}"
+    eco_title ||= "#{cnx_book.title} \
+                    (#{cnx_book.uuid}@#{cnx_book.version}) - #{Time.current.utc}"
```
jpslav: I think this is going to put a bunch of spaces in after the slash (from the leading blank characters on the next line).

Dante and I have been going with a max line width of 100 characters, so I'd say just leave the original.

jpslav:
Looks good.  Couple of comments.  Will need rebasing after karen's PR is rebased.

app/routines/fetch_and_import_book_and_create_ecosystem.rb
joemsak: okay. i'll doible check but im pretty sure i split screen at 120 now

jpslav:
#738 was merged; rebase away

app/controllers/manager/ecosystems_actions.rb
joemsak: oh ya ;-P i got distracted by the original ternary that was there

app/routines/fetch_and_import_book_and_create_ecosystem.rb
joemsak: Ah okay my line is set at 90.... fair enough

joemsak:
@jpslav awaiting CI

joemsak:
@jpslav CI passing after a little turbulence

PR #748, opened 2015-10-27
adding 2d matrix helper for cell reference

joemsak:
Fixes bug in ExportXlsx not supporting double-letter columns

### DIFF
--- a/app/subsystems/tasks/performance_report/export_xlsx.rb
+++ b/app/subsystems/tasks/performance_report/export_xlsx.rb
@@ -86,13 +86,15 @@ module Tasks
       def add_late_comments(sheet, data, row)
         data.each_with_index do |d, col|
           if d && d.late
-            start_col = ('A'..'Z').to_a[non_data_headings.size]
-            offset_row = row + 4 # offset 4 header rows
+            column = col + non_data_headings.size
+            row_offset  = row + 4 # there are 4 rows of headings
 
-            sheet.add_comment(ref: "#{(start_col..'Z').to_a[col]}#{offset_row}", # 'D5', etc...
+            sheet.add_comment(
+              ref: TwoDMatrixHelper.find_cell(row: row_offset, column: column),
               text: "Homework was worked #{time_ago_in_words(d.last_worked_at)} late",
               author: 'OpenStax',
-              visible: false)
+              visible: false
+            )
           end
         end
       end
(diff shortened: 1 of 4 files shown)


spec/lib/two_d_matrix_helper_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(cell_reference).to eq('AA5')
+
+      cell_reference = described_class.find_cell(row: 69, column: 27)
+      expect(cell_reference).to eq('AB69')
+
+      cell_reference = described_class.find_cell(row: 420, column: 57)
+      expect(cell_reference).to eq('BF420')
+    end
+
+    it 'finds insane cells'
```
Dantemss: Insanity!

Dantemss:
Please move branch head to 0e3b0cd so we can merge to master.

app/subsystems/tasks/performance_report/export_xlsx.rb
```diff
@@ -86,13 +86,15 @@ def lateness_styles(data, worksheet)
...
         data.each_with_index do |d, col|
           if d && d.late
-            start_col = ('A'..'Z').to_a[non_data_headings.size]
-            offset_row = row + 4 # offset 4 header rows
+            column = col + non_data_headings.size
+            row_offset  = row + 4 # there are 4 rows of headings
 
-            sheet.add_comment(ref: "#{(start_col..'Z').to_a[col]}#{offset_row}", # 'D5', etc...
+            sheet.add_comment(
+              ref: TwoDMatrixHelper.find_cell(row: row_offset, column: column),
```
jpslav: FWIW, I think this can be

```ruby
ref: sheet.rows[row_offset].cells[column]
```

per http://www.rubydoc.info/github/randym/axlsx/Axlsx/Comments#add_comment-instance_method

jpslav: Actually looking at the docs I bet it could even be:

```ruby
ref: sheet[row_offset][column]
```

Dantemss:
So... we don't need the TwoDMatrixHelper anymore?

joemsak:
wow. never saw this. how do you see it on the docs about add_comment? i had
to go to worksheet docs to see there are rows and cells methods even then
not entirely sure i would have realized it

PR #749, opened 2015-10-28
Delete ecosystems

karenc:


karenc:
Ready for review.

Dantemss:
Would be nice to have this on QA :) (list of ecosystems is getting pretty big)

PR #750, opened 2015-10-28
Use stored ecosystems import job ids instead of filtering all bg jobs

karenc:
It was making the ecosystems page very slow to load because we were
looping through all the background jobs to look for incomplete ecosystem
import jobs.

### DIFF
--- a/app/controllers/manager/ecosystems_actions.rb
+++ b/app/controllers/manager/ecosystems_actions.rb
@@ -2,8 +2,9 @@ module Manager::EcosystemsActions
 
   def index
     @ecosystems = Content::ListEcosystems[]
-    @incomplete_jobs = Lev::BackgroundJob.incomplete.select do |job|
-      job.respond_to?(:ecosystem_import_url)
+    Content::Models::EcosystemJob.update_status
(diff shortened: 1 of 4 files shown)


karenc:
@joemsak my attempt to fix the slow loading time on the ecosystems page.

jpslav:
I have a feeling the slowness should be fixed in lev.  And I do want to still follow through with a more complete job status in activerecord for tracking all jobs (delayed_job deletes completed ones).  Are you able to replicate the slowness or is it just in production?

karenc:
Both @joemsak and I are able to replicate the slowness in development.

karenc:
I cleared everything in redis by doing `redis-cli flushall` because I was getting some errors when I tried to do `Lev::BackgroundJob.all`.  The slowness has gone away for me for now... (so I won't be able to debug it)

karenc:
Ok I'm able to replicate this again.  It was as @jpslav said, writing to redis when `find` is called.  I'm going to submit a different fix in [lev](https://github.com/lml/lev) background job for this.

karenc:
See https://github.com/lml/lev/pull/42

PR #751, opened 2015-10-28
Salesforce integration

jpslav:
Adds Salesforce integration.  In the admin menu, a user can elect to be the SF user Tutor acts as in API calls.  In that menu an admin can also trigger an import of courses from SF, which will also write information (or errors) back to Salesforce for each SF record processed.

## Tasks for a later PR

* routine for updating statistics
* call routines in whenever


jpslav:
@nathanstitt and/or @Dantemss, can you review?

app/controllers/admin/salesforce_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      ).outputs
+
+      flash[:notice] = "Of #{outputs.num_failures + outputs.num_successes} candidate records in Salesforce, " +
+        "#{outputs.num_successes} were successfully imported and #{outputs.num_failures} failed."
+
+      redirect_to admin_salesforce_path
+    end
+
+  end
+end
```
nathanstitt: I was under the impression that we're supposed to use PORO wrappers rather than the model directly?  Or maybe it's ok in an admin controller.

app/handlers/admin/catalog_offering_update.rb
```diff
@@ -21,7 +9,6 @@ def authorized?
   end
 
   def handle
-    attributes = offering_params.as_json
-    run(:update_offering, params[:id], attributes)
+    run(:update_offering, params[:id], params[:offering].permit!)
   end
```
nathanstitt: Interesting, I didn't know about `permit!`

app/subsystems/content/routines/import_book_part.rb
```diff
@@ -54,7 +54,7 @@ def exec(cnx_book_part:, book:, chapter_tracker: nil, save: true)
         # skip all the pages at the book level
         next if cnx_book_part.is_root && part.is_a?(OpenStax::Cnx::V1::Page)
 
-        raise "Unexpected class #{part.class}" unless part.is_a?(OpenStax::Cnx::V1::BookPart) rescue debugger
+        raise "Unexpected class #{part.class}" unless part.is_a?(OpenStax::Cnx::V1::BookPart)
 
         outs = run(:import_book_part,
                    cnx_book_part: part,
```
nathanstitt: What was that?  I've never seen `rescue debugger` Sounds like it shouldn't have been there though.

app/subsystems/school_district/get_school.rb
```diff
@@ -3,9 +3,11 @@ class GetSchool
     lev_routine express_output: :school
 
     protected
-    def exec(id:)
+    def exec(id: nil, name: nil)
       if !id.blank? && id != 0 # webforms weirdness
         outputs.school = Models::School.find(id)
+      elsif !name.blank?
+        outputs.school = Models::School.where(name: name).first
```
nathanstitt: could use `.present?` here, it's the opposite of `.blank?`

db/migrate/20151031032144_create_salesforce_attached_record.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.string :tutor_gid
+      t.string :salesforce_class_name
+      t.string :salesforce_id
+
+      t.index :tutor_gid
+      t.index [:salesforce_id, :salesforce_class_name, :tutor_gid],
+              unique: true, name: 'salesforce_attached_record_tutor_gid'
+    end
+  end
+end
```
nathanstitt: Shouldn't some of the columns on the above tables have `null: false` specified?  Likewise, I don't see any validations on the model(s).

spec/controllers/admin/salesforce_controller_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+          name: 'Bobby Thomas',
+        },
+        credentials: {
+          oauth_token: 'oauth_token',
+          refresh_token: 'refresh_token',
+          instance_url: 'http://blah.com/'
+        }
+      })
+    }
+  end
```
nathanstitt: This seems like it might be useful on other controllers, we should remember it's here in case you spot someone else doing the same thing.

app/routines/import_salesforce_courses.rb
```diff
@@ -0,0 +1,101 @@
...
+      "Created course '#{course.name}' (#{course.id}) based on Salesforce record " +
+      "#{candidate.id} using offering '#{offering.identifier}' (#{offering.id}) " +
+      "and ecosystem '#{offering.ecosystem.title}'."
+    }
+  end
+
+  def assume_success(candidate)
+    candidate.error = nil
+    outputs.num_successes += 1
+  end
```
nathanstitt: It seems like it's initializing a candidate as successful.  Should these operation be part of the PORO's initialize method so they'd default to that state?

app/controllers/admin/salesforce_controller.rb
jpslav: You're right that in theory we are.  Admin controller is maybe ok to do this, but probably not.   I'll look into alternatives.

app/subsystems/content/routines/import_book_part.rb
jpslav: Yeah just something unrelated to this PR that I came across and decided to fix -- definitely shouldn't have been there :-)

app/subsystems/school_district/get_school.rb
jpslav: good point - was it you who just taught me a week or two ago that `present?` checks `nil?` and not `blank?`.  I used to think it was just not `nil?`.

jpslav: or rather, it just checks not `blank?` which is already not `nil?`

jpslav: I might just leave it to match the style of the other code above and the other routines in that SS.

app/routines/import_salesforce_courses.rb
jpslav: Sorry, I know this "assuming" business is a little odd. 

This PORO is an ActiveForce "model".  ActiveForce is a gem that make Salesforce records looks ActiveRecord-ish.  In general if we load candidates (Salesforce "ClassSize" records), we don't want to nil out their `error` fields (we might want to read what was loaded from SF).  And `num_successes` here is just a counter in the routine so it can report back to the form that 32 courses were imported, 3 failed, etc.  

In my first implementation, I nilled out the error at the end of the `create_course_for_candidate` method, as if to say "we made it to the end so there must not have been any errors".  Assuming success and doing it at the beginning was just a way to get as much logging code out of the main flow of the routine as possible.

Actually, I've just written all of that but I wonder if I really need to assume it.  I bet I can just `record_success` at the end of the main method and do the same things (and not need to retract a `num_successes` increment when an error is counted).

db/migrate/20151031032144_create_salesforce_attached_record.rb
jpslav: Good catch.  Added.

PR #756, opened 2015-10-29
Fix ecosystem import for admins when archive url is not specified

karenc:
When the archive url is not specified, we should use the default archive
url.  The error I got was:

```
Addressable::URI::InvalidURIError (Absolute URI missing hierarchical segment: 'http:'):
  lib/openstax/cnx/v1.rb:26:in `set_archive_url_base'
  lib/openstax/cnx/v1.rb:42:in `with_archive_url'
  app/controllers/manager/ecosystems_actions.rb:32:in `import_ecosystem'
  app/controllers/manager/ecosystems_actions.rb:12:in `import'
```


app/controllers/manager/ecosystems_actions.rb
```diff
@@ -15,7 +15,11 @@ def import
   protected
 
   def import_ecosystem
-    url = params[:archive_url] || @default_archive_url
```
karenc: @joemsak I saw that you changed this part of the code last.  Unfortunately `params[:archive_url] || @default_archive_url` doesn't work when `params[:archive_url]` is an empty string.  So I've reverted it back to how I've written it.  Do you have another way of doing this that you prefer?

joemsak: Karen  

good call… thanks. i had actually just moved it to a private method to improve scannablity of the original method and JP had suggested this change so i did that without realizing your case…. so… you know who to blame ;-P

karenc: Haha... ok.  I've added a test for this case so we won't break it in the future :)

joemsak: FWIW you can shorten with

```ruby
url = if #...
          # ...
        else
           #...
        end
```

karenc: What about `url = params[:archive_url].present? ? params[:archive_url] : @default_archive_url`?

joemsak: Me, personally, I don't like a ternary so long? I tend to opt for more LOC if it means more immediate clarity

joemsak: I did have it as this in private method, before JP's feedback:

```ruby
def archive_url
  if params[:archive_url].blank?
    @default_archive_url
  else
    params[:archive_url]
  end
```

karenc:
Ready for review again.

PR #759, opened 2015-10-29
Concept Coach Task

Dantemss:
- Removed CC assistant
- Added CCTask migration/model
- Added CCTask get/create routines and domain routine
- Real get CC task endpoint
- Existing specs
- Made CC spaced practice ignore assignments that have not been started
- New specs


app/controllers/api/v1/cc/tasks_controller.rb
```diff
@@ -12,64 +12,45 @@ class Api::V1::Cc::TasksController < Api::V1::ApiController
...
-  def create_fake_concept_coach_task
-    task =  Tasks::BuildTask[
-              task_type: :concept_coach,
-              title: 'Dummy task title',
-              description: 'Dummy task description',
-              opens_at: 1000.days.ago,
-              feedback_at: 1000.days.ago]
+    # If page_models.size > 1, the user is in 2 courses with the same CC book (not allowed)
+    page_model = page_models.order(:created_at).last
+    return head(:unprocessable_entity) if page_model.blank?
```
jpslav: Can you document this error and what it means in the apipie docs? (no CC for this user and this book)

app/controllers/api/v1/cc/tasks_controller.rb
```diff
@@ -12,64 +12,45 @@ class Api::V1::Cc::TasksController < Api::V1::ApiController
...
 
-  def create_fake_concept_coach_task
-    task =  Tasks::BuildTask[
-              task_type: :concept_coach,
-              title: 'Dummy task title',
-              description: 'Dummy task description',
-              opens_at: 1000.days.ago,
-              feedback_at: 1000.days.ago]
+    # If page_models.size > 1, the user is in 2 courses with the same CC book (not allowed)
+    page_model = page_models.order(:created_at).last
```
jpslav: If `page_models` guaranteed to have 0 or 1 entries, why the `.order(:created_at).last`?

jpslav:
We should maybe discuss this with @kevinburleigh75, but I'm fairly sure we don't want to pick the spaced practice problems when the CC task is created.  The reason being that the CC widget will request and retrieve the CC task whenever a student is logged in and they open a module in the book.  So if a student just clicks all over the book to check it out), the spaced practice history may be all messed up and the spaced practice exercises selected will definitely not reflect what they've actually read.

I believe the better course of action would be to populate the task with placeholders and then populate the placeholders after the first CC question is answered (likely in a background job).

app/routines/get_concept_coach.rb
```diff
@@ -0,0 +1,109 @@
...
+    translations: { outputs: { type: :verbatim } },
+    as: :create_cc_task
+
+  uses_routine Tasks::CreateTasking,
+    translations: { outputs: { type: :verbatim } },
+    as: :create_tasking
+
+  uses_routine AddSpyInfo, as: :add_spy_info
+
+  uses_routine GetCourseEcosystem, as: :get_course_ecosystem
```
jpslav: Unused

app/routines/get_concept_coach.rb
```diff
@@ -0,0 +1,109 @@
...
+  uses_routine GetHistory, as: :get_history
+
+  protected
+
+  def exec(role:, page:)
```
jpslav: The assumption in this signature is that someone else has done the work to verify that this role is allowed to have a CC for this page.  Would it make sense to move the logic from the controller that figures this out into this routine, and make the routine have this signature:

```ruby
def exec(user:, cnx_book_id:, cnx_page_id:)
```

This way there's no assumption about what was done before.  Also this gets a lot of code out of the controller.  

If the code does move here...

* I might recommend breaking it up into more private methods. 
* You'll have to replace the controller's `return head(:unprocessable_entity) if page_model.blank?` with a fatal_error, and then check for that error in the controller.

app/controllers/api/v1/cc/tasks_controller.rb
jpslav: See you've already done this in the other PR

app/controllers/api/v1/cc/tasks_controller.rb
Dantemss: It's only guaranteed by the fact that the enrollment process doesn't allow it (or won't allow it). There's no guarantee on the database etc.

app/routines/get_concept_coach.rb
Dantemss: OK guess I'll combine the PR's because this will make error handling a nightmare...

Dantemss:
Going to use https://github.com/openstax/tutor-server/pull/771. Closing this one.

PR #760, opened 2015-10-29
Prevent homework spaced practice pages from being used for further SP

Dantemss:
This issue does not happen in readings because we read the page_ids from the teacher settings.


Dantemss:
I tested that this at least did not break the HW spaced practice in the demo script :)

db/schema.rb
```diff
@@ -244,8 +244,8 @@
     t.string   "timezone",                    default: "Central Time (US & Canada)", null: false
     t.datetime "created_at",                                                         null: false
     t.datetime "updated_at",                                                         null: false
-    t.string   "catalog_offering_identifier"
     t.boolean  "is_concept_coach",                                                   null: false
+    t.string   "catalog_offering_identifier"
```
jpslav: wonder why this field keeps moving back and forth in the schema...

PR #763, opened 2015-11-02
Display student data in the correct period in performance reports

karenc:
Specifically after a student has moved periods.

### DIFF
--- a/app/subsystems/tasks/get_performance_report.rb
+++ b/app/subsystems/tasks/get_performance_report.rb
@@ -16,8 +16,8 @@ module Tasks
     private
 
     def get_performance_report_for_teacher(course)
+      taskings = get_taskings(course)
       course.periods.collect do |period|
-        taskings = get_taskings(period)
         tasking_plans = sort_tasking_plans(taskings, course, period)
         task_plan_indices = tasking_plans.each_with_index
                                          .each_with_object({}) { |(tasking_plan, index), hash|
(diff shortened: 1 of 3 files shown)


app/subsystems/tasks/get_performance_report.rb
```diff
@@ -30,6 +30,9 @@ def get_performance_report_for_teacher(course)
         task_plan_results = Hash.new{|h, key|h[key] = []}
 
         student_data = sorted_student_data.collect do |student_role, student_taskings|
+                         # skip if student is no longer in the current period
+                         next if student_role.student.period != period
```
jpslav: Can we add the `student` association to the preloading below to head off any N+1 query issues?  Don't need to add `period` because can replace this line with:

```ruby
next if student_role.student.period_id != period.id
```

jpslav:
Review complete -- looking good @karenc

app/subsystems/tasks/get_performance_report.rb
karenc: I don't think it works because `period_id` is not on `CourseMembership::Models::Student`.  What we have is the `enrollments` and `period` is the latest enrollment.

karenc:
Added preloading for student's period:

```diff
diff --git a/app/subsystems/tasks/get_performance_report.rb b/app/subsystems/tasks/get_performance_report.rb
index 935bfc7..0d2a764 100644
--- a/app/subsystems/tasks/get_performance_report.rb
+++ b/app/subsystems/tasks/get_performance_report.rb
@@ -82,7 +82,8 @@ module Tasks
       task_types = Models::Task.task_types.values_at(:reading, :homework, :external)
       # Return reading, homework and external tasks for a student
       course.taskings.preload(task: {task: {task_plan: {tasking_plans: :target} }},
-                              role: {profile: :account})
+                              role: [{student: {enrollments: :period}},
+                                     {profile: :account}])
                      .joins(task: :task)
                      .where(task: {task: {task_type: task_types}})
     end
```

karenc:
Performance report in this pull request is now making less queries than master because we're getting everything we need for the course instead of per period and we're preloading everything :D

app/subsystems/tasks/get_performance_report.rb
jpslav: You are right!

PR #764, opened 2015-11-02
Teacher gets in class with reg code

joemsak:
* Adds `teacher_access_token` to `CourseProfile::Models::Profile`

Depends on #768


app/controllers/courses_controller.rb
```diff
@@ -0,0 +1,15 @@
+class CoursesController < ApplicationController
+  def access
+    course = Entity::Course.find(params[:id])
+
+    if course.teacher_access_token == params[:access_token]
+      AddUserAsCourseTeacher[course: course, user: current_user]
+      redirect_to dashboard_path
+    else
+      flash[:error] = "You are trying to join a class as a teacher, but the information you provided is either out of date or does not correspond to an existing course."
+      raise InvalidTeacherAccessToken
```
jpslav: Raising an exception is fine I think, since it is fairly exceptional if an access URL is no good (similar to a SecurityTransgression in a way).  I kind of wish there was a way the flash error could be encoded as a specific message on the exception when it is registered (instead of having the flash message show up above the generic 404 message).

app/controllers/courses_controller.rb
```diff
@@ -0,0 +1,15 @@
+class CoursesController < ApplicationController
+  def access
+    course = Entity::Course.find(params[:id])
+
+    if course.teacher_access_token == params[:access_token]
```
jpslav: Don't need the course ID in the URL.  tutor.openstax.org/access/787GFA45...  should be sufficient (since the codes are unique).  ?

app/controllers/courses_controller.rb
joemsak: I totally agree about the flash message issue... I realized a shortcoming in `RescueFrom` that I need to have a real status code and the friendly messages map off that. Perhaps they should map off of the name of the exception, and use a standard one from their status code as fallback?

app/controllers/courses_controller.rb
joemsak: Ah yes, good point, was just a habit of defining a `member` route on `resources :courses`

app/subsystems/course_profile/routines/create_course_profile.rb
```diff
@@ -3,6 +3,8 @@ class CourseProfile::Routines::CreateCourseProfile
 
   protected
   def exec(attrs = {})
-    outputs[:profile] = CourseProfile::Models::Profile.create(attrs)
+    profile = CourseProfile::Models::Profile.new(attrs)
+    GenerateToken.apply!(record: profile, attribute: :teacher_access_token)
```
jpslav: In our approach, routines are used in part to get cross-model logic / extraneous logic outside of the model.  Models are left to manage / validate their internal state (only with respect to primitives -- messing with / validating associations is not something generally done except in certain limited circumstances).  In this case, the token _is_ an internal primitive, and only knowledge of the model is required.  So I recommend doing it internal to the model like is done here: https://github.com/openstax/tutor-server/blob/master/app/subsystems/course_membership/models/student.rb#L38-L43

This way you can validate that it is present before save, etc.

If you want to abstract the code, you can make it a module so that you can say

```ruby
class Blah < ActiveRecord::Base
  unique_token :teacher_access_token, length: 20
```

which will put in the validations, the before_validation, etc.

joemsak: Okay, yeah. I debated putting it in a callback because of a discussion we had about callbacks before, and figured, "okay this is part of the steps to create a course profile, so it goes in that routine"

Looks like there's a chance to add these other instances of token generation to my new utility class. Also I like the idea of the abstraction you gave me.

jpslav: Yep, callbacks are cool when they work on internal state.

app/controllers/courses_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      AddUserAsCourseTeacher[course: course, user: current_user]
+      redirect_to dashboard_path
+    rescue
+      flash[:error] = "You are trying to join a class as a teacher, but the information you provided is either out of date or does not correspond to an existing course."
+      raise InvalidTeacherAccessToken
+    end
+  end
+
+  private
+  def find_course
```
jpslav: name more explicitly so not confusing for future non-token actions, e.g. `find_course_by_token`

spec/features/teachers/access_course_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe 'Teachers' do
+  let(:course) { CreateCourse[name: 'Access me'] }
+  let(:teacher) { FactoryGirl.create(:user) }
```
jpslav: can you rename this variable `:user` -- otherwise, readers (like me) might think this user is already a teacher from the get-go.

jpslav:
Looks good.  Two small naming requests and integration of `unique_token` and this will be good to go.

jpslav:
There's also a leftover, unused `unique_token_mode` method from the other PR -- can you remove in this PR?

joemsak:
Pending CI @jpslav

jpslav:
Will merge after rebase @joemsak -- thanks, looks good.

joemsak:
Thank YOU @jpslav

PR #765, opened 2015-11-02
Fetch access token via CORS request

nathanstitt:
As an alternative to https://github.com/openstax/tutor-server/pull/741, this PR implements an `/auth/status` endpoint that can be access via CORS from whitelisted origins.  If user is logged in, it responds with a token and the user's information. 

The token can then be used to perform additional requests against the site from any origin.

An example request using jQuery:

```coffee
  $.ajax('http://localhost:3001/auth/status.json', {
    crossDomain: true
    xhrFields: { withCredentials: true }
  }).then (resp) ->
    console.log resp
    if resp.current_user
      $.ajax('http://localhost:3001/api/cc/tasks/1/1.json', {
        beforeSend: (xhr) ->
          xhr.setRequestHeader("Authorization", "Bearer " + resp.access_token)
        success: (data) -> console.log(data)
        error: (jqXHR, textStatus, errorThrown) -> console.warn(textStatus)
      })

```
Note: The above won't actually work without https://github.com/openstax/openstax_api/pull/37, which fixes a bug with the header for token based access


nathanstitt:
The only thing I can think that’s a drawback to this approach is that it stores the token in the DB and doesn’t currently expire it

We probably should expire them and write a sweeper to clear out old ones.

app/controllers/auth_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    headers['Access-Control-Allow-Headers']  = 'Origin, X-Requested-With, Content-Type, Accept, Authorization'
+  end
+
+  def strategy
+    @strategy ||= server.token_request 'session'
+  end
+
+  def authorize_response
+    @authorize_response ||= strategy.authorize
+  end
```
jpslav: Is this used?

nathanstitt: Nope, I've removed it

jpslav:
Nathan is the man!

PR #766, opened 2015-11-02
Allow teachers to change the opens_at and due_at

Dantemss:
For assignments past the open date

QA will need to test that this change did not break publishing etc.

### DIFF
--- a/app/controllers/api/v1/task_plans_controller.rb
+++ b/app/controllers/api/v1/task_plans_controller.rb
@@ -82,7 +82,7 @@ class Api::V1::TaskPlansController < Api::V1::ApiController
         OSU::AccessPolicy.require_action_allowed!(:create, current_api_user, task_plan)
         uuid = distribute_or_update_tasks(task_plan)
 
-        if task_plan.save
+        if task_plan.errors.empty?
           respond_with task_plan, represent_with: Api::V1::TaskPlanRepresenter,
                                   status: uuid.nil? ? :ok : :accepted,
                                   location: nil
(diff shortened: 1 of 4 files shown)


Dantemss:
Need to update this to change the feedback_at (= due_at if task_type == :homework, = opens_at otherwise)

app/routines/propagate_task_plan_updates.rb
```diff
@@ -6,6 +6,20 @@ class PropagateTaskPlanUpdates
 
   def exec(task_plan:)
     task_plan.tasks.update_all(title: task_plan.title, description: task_plan.description)
+
+    # For now we only handle tasking_plans that point to periods
+    task_plan.tasking_plans.each do |tasking_plan|
+      period = tasking_plan.target
+      next unless period.is_a?(CourseMembership::Models::Period)
```
jpslav: would recommend raising an error here so we notify our future selves when we start assigning to non-periods.

app/controllers/api/v1/task_plans_controller.rb
```diff
@@ -317,21 +317,23 @@ def destroy
   protected
 
   # Distributes or updates distributed tasks for the given task_plan
-  # Returns the job uuid, if any
+  # Returns the job uuid, if any, or nil if the request was completed inline
   def distribute_or_update_tasks(task_plan)
+    task_plan.publish_last_requested_at = Time.now \
+      if !task_plan.tasks_past_open? && task_plan.is_publish_requested?
+    task_plan.save
+    return unless task_plan.errors.empty?
```
jpslav: Instead of pulling the `publish_last_requested_at` assignment and the save out of the conditional below, can we just say

```ruby
return unless task_plan.valid?
```

?

jpslav:
Review complete

app/controllers/api/v1/task_plans_controller.rb
Dantemss: No... I think these changes require the save to always happen before the `PropagateTaskPlanUpdates` so we can see the updated `tasking_plans` (not sure why the save matters but it seems to...)

app/routines/propagate_task_plan_updates.rb
Dantemss: This will cause the teacher to be unable to update the plan at all when this happens. I don't think we need it on the hotfix branch but I'll change it on this PR for master.

PR #767, opened 2015-11-03
Student-specified ID

Dantemss:
- Added student-specified ID
- Added student-specified ID to representer
- Added student-specified ID to scores routine and export
- Let students set the student-specified ID during enrollment
- Specs


Dantemss:
It's done, but the student ID cannot be edited. Have to decide where the edit text box lives, what the API uses to refer to the student object, etc.

jpslav:
Looks good to me.  Only question is whether `student_identifier` needs a uniqueness constraint within some context (e.g. course).

Also needs a rebase.

PR #768, opened 2015-11-03
Add unique tokenable abstraction from #764

joemsak:
* Includes `UniqueTokenable` mixin for `Tutor::Subsystem::BaseModel`
* Adds `unique_token` method for all subsystem base models

* Current usage: `unique_token :field_name, mode: :urlsafe_base64`
  * Mode supports all `SecureRandom` class methods and use of `Babbler` through `:memorable` mode

* More options to come, but this is to get this work separated from #764


lib/tutor/subsystems/base_model.rb
```diff
@@ -23,6 +23,8 @@ def self.table_name
         end
       end
 
+      include UniqueTokenable
```
jpslav: stick it into `ActiveRecord::Base` at the bottom of `unique_tokenable.rb` so it applies to all ActiveRecords.

lib/unique_tokenable.rb
```diff
@@ -0,0 +1,18 @@
+require 'generate_token'
+
+module UniqueTokenable
+  def self.included(base)
+    base.extend ClassMethods
+  end
+
+  module ClassMethods
+    def unique_token(token_field, options = {})
+      before_validation -> { generate_unique_token(self, token_field, options) }
```
jpslav: also need the validation of uniqueness here

joemsak: Does this work?
https://github.com/openstax/tutor-server/pull/768/files#diff-ea9fe1b6eb87851a912758d76e2bbeddR11

lib/generate_token/token_generator.rb
```diff
@@ -0,0 +1,17 @@
+class TokenGenerator
+  @@token_generators = []
+
+  def self.inherited(base)
+    @@token_generators << base
+  end
+
+  def self.selected_generator(mode)
+    if generator = @@token_generators.select { |g| g.handles_mode?(mode) }.first
```
jpslav: since the `first` only allows one mode to be handled by one generator, can `@@token_generators` be a hash mapping mode to generator?  

```ruby 
def self.inherited(base)
  base.handled_modes.each do |mode||
    # freak out if mode already in token_generators ?
    @@token_generators[mode] = base
  end
end
```

?

lib/unique_tokenable.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+  end
+
+  private
+  def generate_unique_token(record, field, options)
+    GenerateToken.apply(record: record, attribute: field, mode: options[:mode])
```
jpslav: Can we get rid of the `GenerateToken` class and just move its `apply` method here?

lib/generate_token/token_generator.rb
```diff
@@ -0,0 +1,17 @@
+class TokenGenerator
+  @@token_generators = []
+
+  def self.inherited(base)
+    @@token_generators << base
+  end
+
+  def self.selected_generator(mode)
```
jpslav: "selected" sounds kind of like there's some user input here.  Since we probably don't want named arguments if this is eventually going to be a gem, can this method be `self.generator_for_mode(mode)`?

lib/generate_token.rb
```diff
@@ -0,0 +1,18 @@
+require 'generate_token/token_generator'
+require 'generate_token/secure_random_token_generator'
+require 'generate_token/babbler_token_generator'
+
+class GenerateToken
+  def self.apply(record:, attribute:, mode: nil)
+    @mode = mode || :hex
+
+    begin
+      record[attribute] = token_generator.generate_with(@mode)
```
jpslav: If the generator has already been selected with `@mode` why does it need to be passed in again?

joemsak: In the case of SecureRandom, the 'mode' is the method delegated to said class

lib/generate_token/secure_random_token_generator.rb
```diff
@@ -0,0 +1,10 @@
+class SecureRandomTokenGenerator < TokenGenerator
```
jpslav: We'll likely need a way to specify options, e.g. the length of the SecureRandom call (or the number of Babbler words)

Something similar to the following maybe...

```ruby
unique_token :blah, mode: {hex: {length: 10}}
```

joemsak: Right, this is what I meant by "I'll add full options in a future PR" :)

jpslav: ah

lib/unique_tokenable.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  module ClassMethods
+    def unique_token(token_field, options = {})
+      before_validation -> { generate_unique_token(self, token_field, options) }
+    end
+  end
+
+  private
+  def generate_unique_token(record, field, options)
```
jpslav: This method needs to not set the value if it exists already.  Otherwise it'll be reset all the time (whenever validation is run).  Also, it is nice to not set it if set before validation (e.g. set to "blah" in a spec).

lib/unique_tokenable.rb
jpslav: This link doesn't take me anywhere

lib/unique_tokenable.rb
```diff
@@ -0,0 +1,36 @@
...
+  def self.included(base)
+    base.extend ClassMethods
+  end
+
+  module ClassMethods
+    def unique_token(token_field, options = {})
+      @unique_token_mode = options[:mode] ||= :hex
+      before_validation -> { generate_unique_token(self, token_field, options) }
+      validates token_field, uniqueness: true
+    end
```
jpslav: don't think we want to save off one mode per class, b/c there could easily be two unique tokens (e.g. a teacher access token and a student access token).  Can we not save off a `@unique_token_mode` and instead just pass it to the `generate_unique_token` method?  Also, I don't think that `self` needs to be passed, because the private method has access to it as well via `self.class`.

```ruby
def unique_token(token_field, options = {})
  mode = options[:mode] || :hex
  before_validation -> { generate_unique_token(mode, token_field, options) }
  validates token_field, uniqueness: true
end
```

lib/unique_tokenable.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def unique_token_mode
+      @unique_token_mode
+    end
+  end
+
+  private
+  def generate_unique_token(record, field, options)
+    return unless record.send(field).blank?
```
jpslav: `record` is `self` here?

joemsak: oh yeah.... vesitigial from when it was in the other, now deleted, class. good eye JP

PR #771, opened 2015-11-03
Concept Coach Tasks with error handling

Dantemss:
- Removed CC assistant
- Added CCTask migration/model
- Added CCTask get/create routines and domain routine
- Real get CC task endpoint
- Made CC spaced practice ignore assignments that have not been started
- Added error code for invalid book
- Added error code for invalid page
- Added error code for user that is not a CC student
- Specs


Dantemss:
jpslav commented 15 hours ago
We should maybe discuss this with @kevinburleigh75, but I'm fairly sure we don't want to pick the spaced practice problems when the CC task is created. The reason being that the CC widget will request and retrieve the CC task whenever a student is logged in and they open a module in the book. So if a student just clicks all over the book to check it out), the spaced practice history may be all messed up and the spaced practice exercises selected will definitely not reflect what they've actually read.

I believe the better course of action would be to populate the task with placeholders and then populate the placeholders after the first CC question is answered (likely in a background job).

PR #774, opened 2015-11-04
Start and end handlers for iframe login

nathanstitt:
This allows Tutor to set the return point once login completes and then relay the access token back to the iframe once the completion page is loaded

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -67,12 +67,7 @@ class AuthController < ApplicationController
   end
 
   def validated_cors_origin
-    origin = request.headers["HTTP_ORIGIN"]
-    return '' if origin.blank?
-    Rails.application.secrets.cc_origins.each do | host |
-      return origin if origin.match(%r{^#{host}})
-    end
-    '' # an empty string will disallow any access
+    OpenStax::Api.configuration.validate_cors_origin[ request ] ? request.headers["HTTP_ORIGIN"] : ''
   end
 
 end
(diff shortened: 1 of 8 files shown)


app/controllers/auth_controller.rb
```diff
@@ -1,20 +1,21 @@
...
+  before_filter :allow_iframe_access, only: [:iframe_start, :iframe_finish]
+
+  #before_filter :forbid_access
+  # these should always return 200 response regardless of login status
+  skip_before_filter :authenticate_user!, only: [:status, :cors_preflight_check, :iframe_start, :iframe_finish]
+
+
+
+  # Since these endpoints are loaded from foreign sites via cors or iframe, CRSF tokens can't be used
+  skip_before_action :verify_authenticity_token
```
jpslav: Can you explicitly add the `only: blah` here -- I like to skip explicitly so others don't get burned down the line if they add a new route.  You can also do

IFRAME_OR_CORS_ACTIONS = [:status, :cors_preflight_check, :iframe_start, :iframe_finish]
skip_before_filter :authenticate_user!, only: IFRAME_OR_CORS_ACTIONS
etc

nathanstitt: Yep, I just added them explicitly.  There's only two that use the complete list so isn't worth adding a new constant for them quite yet (imo).

jpslav:
Thanks @nathanstitt, feel free to merge when CI passes.

nathanstitt:
Merging.  They'll be a follow up shortly to display T&C,  and stub account choosing for dev

PR #778, opened 2015-11-05
Restrict CC routine/endpoint to CC courses only

Dantemss:


jpslav:
Can merge after rebase

PR #779, opened 2015-11-05
Use stub logins in dev mode and display terms & conditions when needed

nathanstitt:
This moves the contracts filter method from the webview controller into application and adds it to the auth controller as well.

It also modifies the auth controller to display the stub authentication inside the iframe for easier FE dev use.

*Fixes bug discovered pre-demo with unset origins configuration.  https://github.com/openstax/tutor-deployment/pull/171 sets the configuration for them as well.


nathanstitt:
@jpslav ready for review

nathanstitt:
Actually hold off for a bit on review, there's a few small changes I want to make

jpslav:
@Dantemss can you review this?

config/initializers/openstax_api.rb
```diff
@@ -1,13 +1,18 @@
+VALID_IFRAME_ORIGINS = Rails.application.secrets.cc_origins ?
+  Rails.application.secrets.cc_origins.map{|origin| Regexp.new("^#{origin}") } : []
```
Dantemss: Not sure if it matters, but Ruby/Rails tend to complain about `^` because ruby uses `\A` for beginning of strings in regex. (`^` is multiline)

app/controllers/auth_controller.rb
```diff
@@ -25,24 +27,21 @@ def cors_preflight_check # the other CORS headers are set by the before_filter
...
-  def iframe_start
+  def iframe
     if current_user.is_anonymous?
-      session[:accounts_return_to] = after_iframe_authentication_url
-      redirect_to openstax_accounts.login_url #(host_only:false)
+      redirect_to_login_url
     else
       @status = user_status_update
-      render action: :iframe_finish
+      @iframe_origin = stubbed_auth? ? session[:parent] : @status[:endpoints][:accounts_iframe]
```
Dantemss: Shouldn't this be CNX instead of Accounts?

config/initializers/openstax_api.rb
```diff
@@ -1,13 +1,18 @@
+VALID_IFRAME_ORIGINS = Rails.application.secrets.cc_origins ?
+  Rails.application.secrets.cc_origins.map{|origin| Regexp.new("\A#{origin}") } : []
```
Dantemss: I think you might need to write `\\A` here.

app/controllers/auth_controller.rb
nathanstitt: No, because at this point in the login flow tutor is being framed by accounts.

 The flow is that an iframe is opened at accounts.   The frame is then singnaled to start a login using service tutor, and then redirects to the tutor auth/iframe, which redirects back a signed request

After login completes it is redirected back to tutor, who then needs to cummunicate back to the accounts iframe the login token and other info.   Accounts receives the message and relays it back to the parent

jpslav:
@Dantemss or @nathanstitt to merge when they are ready

PR #783, opened 2015-11-06
Fix #769 More Teacher join token followups

### LINKED ISSUE #769: Don't rely on flash messages for the teacher access code error page
Requires an update to how friendly message registration is handled in RescueFrom, see openstax/rescue_from#20

joemsak:
* Add the `teacher_join_token` to the roster page JSON
* Move the roster endpoint from `/api/courses/:course_id/students` to `/api/courses/:course_id/roster`
* Handle the case where a user clicks the link but they are already a teacher in the course
* Integrate rescue_from updates for better error messaging.

Depends on merge of #751


joemsak:
@jpslav awaiting CI

app/controllers/courses_controller.rb
```diff
@@ -1,20 +1,12 @@
...
-  private
-  def find_course_by_join_token
-    profile = GetCourseProfile[attrs: { teacher_join_token: params[:join_token] }]
-    Entity::Course.find(profile.course_id)
+  def join
+    handle_with(CoursesJoin, success: -> { redirect_to dashboard_path },
+                             failure: -> {
+                               @handler_result.errors.each do |e|
+                                 CoursesJoin.handle_error(e)
+                               end
```
jpslav: Why not just raise the exceptions in the handler directly? Raising the exception will by definition bypass anything else in this request, so you could say:

```ruby
handle_with(CoursesJoin, success: -> { redirect_to dashboard_path },
                         failure: -> { raise ThisWillNeverHappen })
```

or

```ruby
handle_with(CoursesJoin, completed: -> { redirect_to dashboard_path })
```

combined with:

```ruby
def handle
  course = find_course_by_join_token
  run(:add_teacher, course: course, user: caller)

  # convert errors to exceptions to use quickie-way of generating error messages without making explicit templates
  raise handled_exceptions[errors.first.try(:code)] if errors.any?
end
```

These errors are on the border of what I'd consider worthy of using exception machinery.  They are kind of like user input error ("shouldn't use that code any more" or "you already used that code").  Consider this alternative that doesn't add any code to the handler: 

```ruby
handle_with(CoursesJoin, 
            success: -> { redirect_to dashboard_path },
            failure: -> { 
              message = case @handler_result.errors.first.code
              when :profile_not_found
                "This join code is not valid"  # could later be replaced with render of explicit, specific error page
              when :user_is_already_teacher_of_course
                "You are already a teacher in the course"
              end
              render 'shared/some_generic_error_page', message: message
            })
```

The PR's approach is 25 LOC.  The above is 11 LOC plus a reusable error template that skips the exception machinery.

app/handlers/courses_join.rb
```diff
@@ -0,0 +1,37 @@
...
+    run(:add_teacher, course: course, user: caller)
+  end
+
+  private
+  def find_course_by_join_token
+    profile = run(:get_course_profile, attrs: {
+      teacher_join_token: params[:join_token]
+    }).outputs.profile
+
+    Entity::Course.find(profile.course_id)
```
jpslav: Until the course profile stuff gets PORO'd, can say:

```ruby
profile.course
```

Also FWIW, the code could be:

```ruby
uses_routine GetCourseProfile, translations: { outputs: { type: :verbatim } }
...
run(:get_course_profile, attrs: { teacher_join_token: params[:join_token] ])
outputs.profile.course
```

app/subsystems/course_profile/get_profile.rb
```diff
@@ -18,6 +18,10 @@ def get_profile_by_course(course)
   end
 
   def get_profile_by_attrs(attrs)
-    CourseProfile::Models::Profile.find_by(attrs)
+    if profile = CourseProfile::Models::Profile.find_by(attrs)
+      profile
+    else
+      fatal_error(code: :profile_not_found)
+    end
   end
```
jpslav:
```ruby
CourseProfile::Models::Profile.find_by(attrs) || fatal_error(code: :profile_not_found)
```

app/handlers/courses_join.rb
joemsak: GetCourseProfile returns a `{}` with `course_id:` and other keys in it, and no `course` key

jpslav: good point :-)

app/controllers/courses_controller.rb
joemsak: It's not about number of LOC to me, it's about specifically avoiding the `switch case` code in your latter example. It's not a maintainable approach, it would make the controller action messier as cases are added, and leaks the `error#code` outside of the handler as well, whereas my approach contains it all.

I could consider your first example with the combination, however.

joemsak: Also, I would prefer to make use of the rescue from automatic error handling/error template, otherwise we're duplicating effort and confusing future devs on what the pattern is

jpslav:
Holding pattern until FE has a corresponding PR.

philschatz:
https://github.com/openstax/tutor-js/pull/831 is good to go and waiting on this getting reviewed :smile:

jpslav:
Oh already reviewed and asked Joe to coordinate with Darren so you can go
ahead and merge both.

PR #785, opened 2015-11-06
CC Teacher Dashboard

Dantemss:
- Cleaned up GetDashboard a bit
- Added required CC info to dashboard
- Specs
- Make it fast

### DIFF
--- a/app/representers/api/v1/courses/dashboard_representer.rb
+++ b/app/representers/api/v1/courses/dashboard_representer.rb
@@ -181,6 +181,11 @@ module Api::V1::Courses
                  readable: true,
                  writeable: false,
                  decorator: Teacher
+
+      collection :periods,
+                 readable: true,
+                 writeable: false,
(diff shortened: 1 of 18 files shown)


Dantemss:
I think this is fast enough. We should test on QA.

app/representers/api/v1/courses/dashboard_representer.rb
```diff
@@ -181,6 +181,11 @@ class Course < Roar::Decorator
                  readable: true,
                  writeable: false,
                  decorator: Teacher
+
+      collection :periods,
```
jpslav: Can this be in a `CcCourse < Course` representer and then have the `course` property switch between the representers depending on if it is a CC course?  (don't think we want CC periods to be included in non-CC course dashboard results)

Dantemss: They won't be because GetDashboard won't return them for non-CC courses.

PR #794, opened 2015-11-09
Fix #793 Show error on admin page for account taken, instead of raised exception

### LINKED ISSUE #793: An ActiveRecord::RecordInvalid occurred in users#create:
This should not be an exception but instead an error that flows up to the admin form that caused it (probably `create` plus `transfer_errors_from` instead of `create!`)

An ActiveRecord::RecordInvalid occurred in users#create:

 Validation failed: Account has already been taken
 lib/entity.rb:44:in `block (2 levels) in exposes'

See email

joemsak:
See #793

### DIFF
--- a/app/subsystems/user/create_user.rb
+++ b/app/subsystems/user/create_user.rb
@@ -20,18 +20,13 @@ module User
         full_name: full_name, title: title
       )
 
-      user = ::User::User.create(
+      outputs.user = ::User::User.create(
         exchange_read_identifier: (exchange_identifiers || new_identifiers).read,
         exchange_write_identifier: (exchange_identifiers || new_identifiers).write,
         account_id: account_id
       )
 
-      if user.to_model.valid?
-        outputs.user = user
-      else
-        fatal_error(code: :could_not_create_user,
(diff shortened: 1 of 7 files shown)


jpslav:
Looks good.  Merge when CI passes.

app/subsystems/user/create_user.rb
```diff
@@ -20,11 +20,18 @@ def exec(account_id: nil, exchange_identifiers: nil,
...
         exchange_write_identifier: (exchange_identifiers || new_identifiers).write,
         account_id: account_id
       )
+
+      if user.to_model.valid?
+        outputs.user = user
+      else
+        fatal_error(code: :could_not_create_user,
+                    message: user.to_model.errors.first.join(' ').capitalize)
+      end
```
jpslav: weren't you considering calling `transfer_errors_from` instead of this block of code?

joemsak: ah okay have to use `user.to_model` but yes

joemsak:
@jpslav awaiting CI

PR #799, opened 2015-11-10
CC score report

Dantemss:
- Added CC scores to the score endpoint
- Specs


jpslav:
Any thoughts about ways to separate the CC code from the non-CC code?  Kind of hesitant to let them live as quasi-parallel code in the same routine.  

Kind of like they should be two separate routines that use a common helper object for `get_student_data` and `exercise_counts`.

Dantemss:
OK

PR #800, opened 2015-11-10
Fix #788 - prevent trying to read a nil student roster file

### LINKED ISSUE #788: A NoMethodError occurred in courses#students:
A NoMethodError occurred in courses#students:

 undefined method `read' for nil:NilClass
 app/controllers/admin/courses_controller.rb:37:in `students'

See email for more detail

joemsak:
Moved the `courses#students` code off into the `Admin::CoursesStudents` handler

### DIFF
--- a/app/controllers/admin/courses_controller.rb
+++ b/app/controllers/admin/courses_controller.rb
@@ -1,5 +1,6 @@
 class Admin::CoursesController < Admin::BaseController
   include Manager::CourseDetails
+  include Lev::HandleWith
 
   before_action :get_schools, only: [:new, :edit]
 
@@ -29,30 +30,16 @@ class Admin::CoursesController < Admin::BaseController
   end
 
   def students
-    # Upload a csv file with columns: first_name, last_name, username, password
-    period = CourseMembership::Models::Period.find(params[:course][:period])
-    roster_file = params[:student_roster]
-
-    begin
-      csv_reader = CSV.new(roster_file.read, headers: true)
-      users_attributes = []
-      errors = []
-      csv_reader.each do |row|
-        users_attributes << row
-        errors << "On line #{csv_reader.lineno}, username is missing." unless row['username'].present?
-        errors << "On line #{csv_reader.lineno}, password is missing." unless row['password'].present?
-      end
-      if errors.present?
-        flash[:error] = ['Error uploading student roster'] + errors
-      else
-        add_students(period, users_attributes)
-        flash[:notice] = 'Student roster has been uploaded.'
-      end
-    rescue CSV::MalformedCSVError => e
-      flash[:error] = e.message
-    end
-
-    redirect_to edit_admin_course_path(params[:id], anchor: 'roster')
+    handle_with(Admin::CoursesStudents, success: -> {
+      add_students(@handler_result.outputs.period,
+                   @handler_result.outputs.users_attributes)
+      flash[:notice] = 'Student roster has been uploaded.'
+      redirect_to edit_admin_course_path(params[:id], anchor: 'roster')
+    },
+    failure: -> {
+      flash[:error] = @handler_result.errors.collect(&:message).compact
+      redirect_to edit_admin_course_path(params[:id], anchor: 'roster')
+    })
   end
 
   def set_ecosystem
--- a/app/controllers/admin/courses_controller.rb
+++ b/app/controllers/admin/courses_controller.rb
@@ -31,15 +31,15 @@ class Admin::CoursesController < Admin::BaseController
 
   def students
     handle_with(Admin::CoursesStudents, success: -> {
-      add_students(@handler_result.outputs.period,
-                   @handler_result.outputs.users_attributes)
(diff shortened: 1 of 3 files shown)


app/controllers/admin/courses_controller.rb
```diff
@@ -29,30 +30,16 @@ def update
...
-        flash[:notice] = 'Student roster has been uploaded.'
-      end
-    rescue CSV::MalformedCSVError => e
-      flash[:error] = e.message
-    end
-
-    redirect_to edit_admin_course_path(params[:id], anchor: 'roster')
+    handle_with(Admin::CoursesStudents, success: -> {
+      add_students(@handler_result.outputs.period,
+                   @handler_result.outputs.users_attributes)
```
jpslav: `add_students` and `find_or_create_user` are helper methods for the code that got moved into the controller, so might as well move them into the handler too -- since they also just wrap routines, you can call the routines directly in the handler.  Since all handlers are routines, you can use `uses_routine` and call these other routines like you would in any other routine.

app/handlers/admin/courses_students.rb
```diff
@@ -0,0 +1,41 @@
+class Admin::CoursesStudents
+  lev_handler
+
+  protected
+  def authorized?; true; end
+
+  def handle
+    outputs.period = CourseMembership::Models::Period.find(params[:course][:period])
+    outputs.users_attributes = []
```
jpslav: will no longer need to output these when move the private methods in here as discussed above.

app/handlers/admin/courses_students.rb
```diff
@@ -0,0 +1,41 @@
...
+    rescue CSV::MalformedCSVError => e
+      nonfatal_error(code: :malformed_csv, message: e.message)
+    end
+
+    if errors.any?
+      errors.insert(0, Lev::Error.new(code: :error_uploading,
+                                      message: 'Error uploading student roster'))
+      fatal_error(code: :fatal_errors)
+    end
+  end
```
jpslav: I do get what you're doing.  The missing cell errors are really fatal errors, but you're accumulating them with non-fatal errors so that the routine finds all of the problems before bailing.  Since the different codes for the different errors are not being used, I would probably rather `nonfatal_error` not be used since they are fatal, just on a delay.

Instead, I'd recommend just accumulating the error messages and sending them out in one fatal error like so:


```ruby
begin
  csv_reader = CSV.new(params[:student_roster].read, headers: true)

  user_attributes = []
  parse_errors = []

  csv_reader.each do |row|
    parse_errors << "On line #{csv_reader.lineno}, username is missing." if row['username'].blank?
    parse_errors << "On line #{csv_reader.lineno}, password is missing." if row['password'].blank?

    next if parse_errors.any?

    users_attributes << row
  end

  fatal_error(code: :parse_error, message: parse_errors.join("\n"))
rescue CSV::MalformedCSVError => e
  fatal_error(code: :malformed_csv, message: e.message)
end
```

I see how you're prepending an error in order to get the "Error uploading student roster" text to show up in the browser first.  I think this is just an error message / UI thing (it isn't really an error), so it should go in the controller:

```ruby
failure: -> {
  flash[:error] = "Error uploading student roster: " + @handler_result.errors.collect(&:message).compact
  redirect_to edit_admin_course_path(params[:id], anchor: 'roster')
})
```

(incidentally, this isn't really an "upload" error which implies that the file didn't make it to the server, but rather a parsing error.)

If you really want to pass extra data about the different parse errors out, `Lev::Error` accepts a `data` attribute, which is intended to be parse by some code that understands the code, e.g. for a `code` of `parse_error` data could be: `[{username_missing: {line: 23}}, ...]`, but that might be too much work for this admin code.

jpslav: Really, the function of handlers is to prep form data in cases where you're not dealing with just a simple ActiveRecord CRUD operation, and during that prep to do AR-like validation and type casting.  Ideally, we'd be able to supplement our normal `paramify` calls like:

```ruby
paramify do
  attribute :student_roster, type: CSV
  validates :student_roster, presence: true
  validates :student_roster, fields: {username: :required}, {password: :required}
end
```

This would take care of all of the input checking you're doing by hand and leave you with a `paramify_params.student_roster` CSV object.  While this is too much work in this case, it would be the most ideal solution in the context of how Lev is intended to work.

joemsak: Love it, but not so sure I can verify the 'type' on student_roster correctly... CSV is not a type to cast to, and File isn't good because in tests it's Rack::Test::UploadedFile, also it says "unable to cast to type File" anyway :)

joemsak: Hmmm can't really get the validation bit to work without too much effort, sticking to original suggestions :)

joemsak:
@jpslav awaiting CI

jpslav:
Yeah, sorry didn't mean should try to make CSV validator, just that in an ideal world it would be nice :-)  Way too much work for too little gain tho.

PR #803, opened 2015-11-11
protect against missing cc origins

jpslav:
Protects against "undefined method `any?' for nil:NilClass" when `cc_origins` secret not set.

Probably not going to happen much more b/c of recent updates to deployment scripts, but hey...

@Dantemss can review/merge or close if not something we want.


nathanstitt:
@jpslav I also included that as part of #779 (https://github.com/openstax/tutor-server/blob/cc-tnc/config/initializers/openstax_api.rb#L1-L2)   @Dantemss also points out that we should use `\A` vs `^` since `\A` matches the start of string and `^` also matches after a new line.

Dantemss:
Closing since Nathan seems to have done it.

PR #808, opened 2015-11-12
Logout support for concept coach

nathanstitt:
I see three ways we can accomplish logout for concept coach

1) Via a special CORS request to a controller with the appropriate OPTIONS and headers.  Our existing auth one is already setup to do so.  Thats what this PR implements.

2) Extend the existing `sessions` controller's `delete` action to be reachable via CORS. Since that lives in the accounts gem, I'm not sure we want to do so for all uses.

3) Use an iframe proxy method.  That seems a bit complex but might be more secure.

After playing with #1 and #2 I'm afraid they're not going to work because CORS will not follow redirects.  

We could work around this by logging out of tutor and then replying with a `next_url` parameter that CC would then load on accounts, or just loading the logout inside the iframe.  That'd require opening CORS up on accounts though. 

Given that, I think it's probably better to just logout inside the iframe. A nice side effect of this is that the login page is then immediately shown so a new session can be started from there.

https://github.com/openstax/accounts/pull/210 goes along with this.

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -34,7 +34,7 @@ class AuthController < ApplicationController
       redirect_to_login_url
     else
       @status = user_status_update
-      @iframe_origin = stubbed_auth? ? session[:parent] : @status[:endpoints][:accounts_iframe]
+      @iframe_origin = stubbed_auth? ? session[:parent] || '*' : @status[:endpoints][:accounts_iframe]
     end
   end
 
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -39,15 +39,18 @@ class AuthController < ApplicationController
   end
 
   def logout
-    if current_user.is_anonymous?
-      render status: :forbidden, text: 'You must be logged in to logout'
-    else
-      sign_out!
-      render json: { logout: true }
-    end
+    sign_out!
+    redirect_to stubbed_auth? ? authenticate_via_iframe_url :
+                  OpenStax::Utilities.generate_url(
+                    OpenStax::Accounts.configuration.openstax_accounts_url,
+                    "remote/iframe", parent: params[:parent]
+                  )
   end
 
   private
+
+  def accounts_url_to(page)
+  end
 
   def stubbed_auth?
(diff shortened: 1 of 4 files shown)


app/controllers/auth_controller.rb
```diff
@@ -38,6 +38,15 @@ def iframe
     end
   end
 
+  def logout
+    if current_user.is_anonymous?
+      render status: :forbidden, text: 'You must be logged in to logout'
+    else
+      sign_out!
+      render json: { logout: true }
+    end
```
jpslav: `sign_out!` just always sets the user to be signed out -- doesn't care if already signed out.  So, unless there's a need on the FE to get a forbidden status, this can just be 

```ruby
def logout
  sign_out!
  head :no_content # not :ok b/c that looks to freak out ajax calls, or can put your existing render here
end
```

jpslav:
I think logout isn't a particularly sensitive route, right?  (so maybe extra security of other options not needed)

jpslav:
@Dantemss your opinion appreciated :-)

nathanstitt:
@jpslav yes I don't see logout being particularly sensitive.  I may have missed some possible vectors though that @Dantemss might know about

Dantemss:
I think I'm OK with the logout route being open not like you can steal any info from a user after they get logged out.

Dantemss:
Could set logout to use GET in Tutor even, if that's more convenient.

nathanstitt:
Yeah, I've since discovered some problems with logging out this way.  It only clears the tutor session, not the underlying one on accounts.  This means that when you attempt to login again with a different account you're immediately logged into the first one since accounts just redirects back. 

And since CORS doesn't follow redirects it's impossible to follow the same flow that a "normal" logout does.  

I'm reworking using the logout by iframe option.

jpslav:
I have never wanted logout to be a GET because browser optimizers that preload links could/would mistakenly log you out.  

Nathan, by "reworking" does that mean this PR is back in WIP?

nathanstitt:
@jpslav  Sorry that comment was from before when I was working on the iframe logout method.  This and https://github.com/openstax/accounts/pull/210 are both ready for review.

PR #814, opened 2015-11-17
Include teacher names as part of enrollment request

nathanstitt:
Will enable concept coach to display the teacher name as part of the enrollment confirmation


jpslav:
possible to add a spec?

PR #815, opened 2015-11-17
Cache teacher dashboard

Dantemss:
- Cache the teacher dashboard stats in Redis
- Specs


Dantemss:
First request takes about the same as before. Second request about 390ms (previously 525ms).

Dantemss:
Cache expires when anyone in the period answers any exercise.
Cache duration is the default for now (90 minutes, set in application.rb). So it will always expire after 90 minutes (we could make it last forever instead if that would be better).

Dantemss:
The only thing this PR does not do is warm the cache (so the first request to the dashboard will be a little slow but not toooo slow).

jpslav:
@Dantemss I think the cache expiration can probably be longer.

PR #817, opened 2015-11-17
CC Spaced Practice fix and history spy info

Dantemss:
- Added Task history to the task spy info
- Made the spaced practice algorithm not use all random-ago
- Fixed CC spaced practice
- CC spaced practice specs


Dantemss:
Kathi is fine with showing the complete history, but said it would be good if we had an easy way to find the current task in it.

app/routines/add_spy_info.rb
```diff
@@ -18,6 +18,10 @@ def values_from(value)
     case value
     when Content::Models::Ecosystem, Content::Ecosystem
       { ecosystem_id: value.id, ecosystem_title: value.title }
+    when Array
+      { history: value.map do |task|
+        { task_id: task.id, task_title: task.title }
+      end }
```
jpslav: @Dantemss how can we not just check for `Array`?  (this implementation means we won't be able to pass in any other arrays of other things later).

Maybe

```ruby
run(:add_spy_info, to: outputs.task, from: [ecosystem, {history: history.tasks}])
```

And then have a `when Hash` and treat the keys differently...

Dantemss: Hm ok

PR #818, opened 2015-11-17
Redirect teacher to course page on subsequent join clicks

joemsak:


app/controllers/courses_controller.rb
```diff
@@ -2,6 +2,9 @@ class CoursesController < ApplicationController
   include Lev::HandleWith
 
   def join
-    handle_with(CoursesJoin, complete: -> { redirect_to dashboard_path })
+    handle_with(CoursesJoin, complete: -> {
+      course = @handler_result.outputs.course
+      redirect_to "/courses/#{course.id}"
```
jpslav: good catch on correcting this redirect (to specific course instead of more generic dashboard which could present teacher with multiple courses if she has them), but would you mind making a specific route in routes.rb so we can use a helper and restrict future changes to one place?

```ruby
# routes.rb
get '/courses/:id', to: 'webview#index', as: :course_dashboard
```

```ruby
redirect_to course_dashboard_path(course.id)
```

app/handlers/courses_join.rb
```diff
@@ -1,7 +1,6 @@
 class CoursesJoin
   lev_handler
 
-  uses_routine AddUserAsCourseTeacher, as: :add_teacher
```
jpslav: Leave this way so the transactions are nested.

jpslav: If you want to get the `course` variable in this routine's output, use verbatim output translations.

jpslav: I imagine maybe you changed the style b/c you didn't want errors to bubble up.  You can just ignore the exception (see the lev readme):

```ruby
uses_routine AddUserAsCourseTeacher, as: :add_teacher,
             ignored_errors: [:user_is_already_teacher_of_course]
```

joemsak: Funny thing is I read about that when I was last working on this routine in the last PR, and I forgot about it ;-)

jpslav:
Looks good. Please merge I am afk

PR #823, opened 2015-11-19
Redirect logging out CC users to cc.openstax.org/logout

jpslav:
Uses updates in https://github.com/openstax/accounts-rails/pull/43 to redirect logging out CC users to a variant of cc.openstax.org/logout, which will in turn redirect the Accounts logout URL to complete the signout.  By redirecting via the cc.openstax.org site, Accounts will redirect back to cc.openstax.org when it completes the logout process (which we want).

Once https://github.com/openstax/accounts-rails/pull/43 is merged, we can update this PR's gemfile to point to the new accounts-rails version.

### DIFF
--- a/lib/logout_redirect_chooser.rb
+++ b/lib/logout_redirect_chooser.rb
@@ -7,27 +7,12 @@ class LogoutRedirectChooser
     @url = request_url
   end
 
-  def choose(default: nil)
+  def choose(default:)
     if url_indicates_concept_coach?
-      cc_redirect_url
+      URI.join(default, "?cc=1").to_s
     else
       default
     end
-  end
-
-  def cc_redirect_url
-    subdomain = uri.host.split('.')[0]
-
-    suffix = case subdomain
-             when /localhost/
-               "-localhost"
-             when /-(\w+)/
-               "-" + $1
-             else
-               ""
-             end
-
-    "http://cc.openstax.org/logout#{suffix}"
   end
 
   private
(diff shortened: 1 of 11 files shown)


jpslav:
@Dantemss please review

jpslav:
The CI failures don't fail when I just run that one spec... I'll have to look into it, but still should be GTG to review.

lib/logout_redirect_chooser.rb
```diff
@@ -0,0 +1,52 @@
...
+  def query_hash
+    uri.query.nil? ? {} : URI::decode_www_form(uri.query).to_h
+  end
+
+  def uri
+    @uri ||= URI.parse(@url)
+  end
+
+  def to_bool(boolean_thing)
+    ActiveRecord::Type::Boolean.new.type_cast_from_user(boolean_thing)
```
Dantemss: `!!boolean_thing`?

Dantemss:
Looks good, just the 1 comment.

lib/logout_redirect_chooser.rb
Dantemss: Ah nvm this is parsing the url params... I see.

lib/logout_redirect_chooser.rb
```diff
@@ -0,0 +1,52 @@
...
+               ""
+             end
+
+    "http://cc.openstax.org/logout#{suffix}"
+  end
+
+  private
+
+  def url_indicates_concept_coach?
+    @url.match(/\/ConceptCoach\//i) ||
```
Dantemss: Actually do we have any URL that has `ConceptCoach` in it?

jpslav: we will (for branding purposes), tho I don't know if it will actually make it into the logout URL.

PR #827, opened 2015-11-21
More score fixes

Dantemss:
- Stop CC scores from blowing up if a student didn't work a certain page in the period
- Handle CC tasks in the export
- Remove due date row from CC export
- Improved handling of offsets and styles in the export


Dantemss:
FE has to be aware that this PR changes the averages to be between 0 and 1 like all the other scores for BOTH CC and non-CC (currently it's out of 100).

app/subsystems/tasks/performance_report/export_xlsx.rb
```diff
@@ -73,23 +74,20 @@ def gather_averages(data_headings)
...
 
       def add_late_comments(sheet, data, row)
         data.each_with_index do |d, col|
           if d && d.late
             column = col + non_data_headings.size
-            row_offset  = row + 4 # there are 4 rows of headings
 
             sheet.add_comment(
-              ref: TwoDMatrixHelper.find_cell(row: row_offset, column: column),
+              ref: TwoDMatrixHelper.find_cell(row: row.row_index + 1, column: column),
```
jpslav: Can you check to see if https://github.com/openstax/tutor-server/pull/748#discussion_r43190110 works?  If so, we can remove the helper.

jpslav:
@philschatz / @openstax/tutor-fe-dev can comment on if we need to block this until FE updated for the change from [0,100] to [0,1].

@Dantemss otherwise it looks good - you can merge when they are ready (or you can try what I mentioned in my one comment).

app/subsystems/tasks/performance_report/export_xlsx.rb
Dantemss: I'll check.

PR #829, opened 2015-11-21
Update Salesforce course stats / add global settings

jpslav:
DONE:

* Added a cron job for updating Salesforce course stats nightly
* Updated `Salesforce::AttachedRecord` to load all remote data in one API query instead of N.
* Initial SF import and subsequent updates now update the "num sections" field.
* Add a global settings feature for efficiently storing/setting/retrieving site-wide settings (see screenshot below, uses https://github.com/huacnlee/rails-settings-cached and https://github.com/accessd/rails-settings-ui.  Put a "import real SF courses" setting here so we can control what happens from the cron job that imports from SF. (screenshot below)

image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -149,6 +149,10 @@ gem 'omniauth-salesforce'
 # Fork that supports Ruby >= 2.1
 gem 'active_force', git: 'https://github.com/maddijoyce/active_force', ref: '9695896f5'
 
+# Global settings
+gem 'rails-settings-cached', '~> 0.4.0'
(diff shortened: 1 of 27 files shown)


jpslav:
Dependent on https://github.com/openstax/tutor-deployment/pull/176

jpslav:
@Dantemss please review.

jpslav:
Actually, this is mergeable before the tutor-deployment PR is merged, it just will never work on real data on real production until that PR is merged.

PR #830, opened 2015-11-23
Make UUID on ecosystem list more accessible

jpslav:
The UUID was in a popover that went away whenever clicked on (so UUID couldn't be copied).  @gregfitch reported this.  

This PR changes the table to always show the UUID:

image

If @gregfitch just wants the popover to stick, all we have to do is remove the `trigger: 'focus'` attribute from the source.  @gregfitch it is up to you.  If you like the new look, feel free to merge this PR.  If you want it the other way, let me know.

### DIFF
--- a/app/views/manager/ecosystems/_listing.html.erb
+++ b/app/views/manager/ecosystems/_listing.html.erb
@@ -1,10 +1,12 @@
 <table class="table table-striped">
   <thead>
     <tr>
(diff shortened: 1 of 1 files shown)


gregfitch:
No, the current design is really nice -- just need the popover to stick around so we can get just the UUID.

jpslav:
OK, I reverted to the simple fix

PR #834, opened 2015-11-24
Attempts to fix duplicate CC tasks

Dantemss:
- Adds a (required) reference to a role in concept_coach_task and a unique index.
- Does a select for update when trying to read the concept_coach_task before creation.

### DIFF
--- a/app/subsystems/tasks/get_concept_coach_task.rb
+++ b/app/subsystems/tasks/get_concept_coach_task.rb
@@ -5,7 +5,8 @@ class Tasks::GetConceptCoachTask
 
   def exec(role:, page:)
     outputs[:entity_task] = Tasks::Models::ConceptCoachTask
-      .where(content_page_id: page.id, entity_role_id: role.id)
+      .joins(:page)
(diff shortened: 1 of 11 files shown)


app/subsystems/tasks/get_concept_coach_task.rb
```diff
@@ -5,8 +5,7 @@ class Tasks::GetConceptCoachTask
 
   def exec(role:, page:)
     outputs[:entity_task] = Tasks::Models::ConceptCoachTask
-      .joins(task: :taskings)
-      .where(content_page_id: page.id, task: { taskings: { entity_role_id: role.id } })
-      .take.try(:task)
+      .where(content_page_id: page.id, entity_role_id: role.id)
```
jpslav: Um... just realized... does the way we get CC's mean that completed CC's don't survive ecosystem updates? (where the content page ID changes)

jpslav:
Reviewed.  1 comment.

Dantemss:
Each course has (hopefully) only one book. And each role is only in 1 course. So by the role in that case.

PR #837, opened 2015-11-25
Allow Concept Coach teachers to request task results

nathanstitt:
This is needed so a CC teacher can review a student's answers from the student scores report.

Per feedback, I'm now loading the course from the concept coach task in the access check (if needed)


nathanstitt:
I've also attempted to get the FE tasks review component to work and don't think it's going to be possible with the way things are currently setup.

That's the "Review" button that should appear at the top of each column.

That button and the api endpoint that it loads (`api/plans/<task_id>/stats`), depends on a `task_id`, which a concept coach course doesn't have.

I think a new endpoint that accepts `course_id` & `page_id` could be added, but it's format would be different than what the current review component expects.

app/access_policies/task_access_policy.rb
```diff
@@ -6,10 +6,11 @@ def self.action_allowed?(action, requestor, task)
         (
           DoesTaskingExist[task_component: task, user: requestor] &&
           task.past_open?
-        ) ||
-        (
+        ) || (
           (course = task.task_plan.try(:owner)).is_a?(Entity::Course) &&
           UserIsCourseTeacher[user: requestor, course: course]
+        ) || (
+          CourseMembership::IsConceptCoachTaskTeacher[task: task, user: requestor]
```
jpslav: `CourseMembership::IsConceptCoachTaskTeacher` should ideally not call `UserIsCourseTeacher` (bad dependency arrow direction from SS up to top-level routines).  Instead of making a new routine here, since we just need a variant of line 10, can we do the following?

```ruby
class TaskAccessPolicy
  def self.action_allowed?(action, requestor, task)
    case action
    when :read
      requestor.is_human? && (
        (
          DoesTaskingExist[task_component: task, user: requestor] &&
          task.past_open?
        ) || (
          (course = get_entity_course(task)) &&
          UserIsCourseTeacher[user: requestor, course: course]
        ) 
      )
    else
      false
    end
  end

  def self.get_entity_course(task)
    course = task.task_plan.try(:owner)) ||  # normal course
             task.concept_coach_task.task.taskings.first.try(:period).try(:course) # cc course
    course.is_a?(Entity::Course) ? course : nil
  end
end
```

jpslav:
Yeah @nathanstitt it is my bad that the functionality you're looking for got lost in the shuffle.  I said the below to Dante a while back but I was supposed to make a ticket for it but did not:

> [... the next thing to do] will be to create a "teacher reviews all work from one page to all of his students" endpoint for CC -- like the one we have in normal tutor, except we don't have plans.
>
> hopefully reusing a lot of what we have, e.g. CalculateTaskPlanStats doesn't really depend on the plan, it immediately gets plan.tasks -- so I think that routine might be able to be refactored as just taking a list of tasks, and from normal tutor or CC we can get those list of tasks differently (the former from plan.tasks, the latter from whatever)

I made a ticket for this in PT.

nathanstitt:
Thanks @jpslav.  I've removed the prototype routine and am now performing the check in the access as you suggested.

PR #839, opened 2015-12-01
add webview_url to course JSON

jpslav:


philschatz:
Thanks! that was quick!

PR #841, opened 2015-12-01
Include comments in ecosystem's representation

nathanstitt:
The QA view needs to display the ecosystem comments on the book selection menu so they can tell the different books apart.


jpslav:
Looks good.  Merge when CI passes.

PR #843, opened 2015-12-01
Enrollments and EnrollmentChanges now use the paranoia gem

Dantemss:
- Enrollment and EnrollmentChange are now soft deleted
- Spec to make sure it works for the Roster

### DIFF
--- a/app/subsystems/course_membership/models/enrollment.rb
+++ b/app/subsystems/course_membership/models/enrollment.rb
@@ -1,8 +1,10 @@
 class CourseMembership::Models::Enrollment < Tutor::SubSystems::BaseModel
-  belongs_to :period
+  acts_as_paranoid
+
+  belongs_to :period, -> { with_deleted }
   belongs_to :student
 
-  has_one :enrollment_change, dependent: :destroy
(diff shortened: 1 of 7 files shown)


jpslav:
Was it this one?

```
A Module::DelegationError occurred in students#index:

  CourseMembership::Models::Student#course_membership_period_id delegated to latest_enrollment.course_membership_period_id, but latest_enrollment is nil: #<CourseMembership::Models::Student id: 297, entity_course_id: 14, entity_role_id: 317, deidentifier: "8f28fb37", inactive_at: "2015-11-23 16:55:59", created_at: "2015-11-23 15:39:11", updated_at: "2015-11-23 16:55:59", student_identifier: nil>
  app/subsystems/course_membership/models/student.rb:15:in `rescue in course_membership_period_id'
```

jpslav:
Looks good.

jpslav:
Also, should be able to add spec that period can in fact be dropped when all students are dropped from it.

Dantemss:
Should be good now. The period model already had a spec to make sure it could only be deleted with no active students.

PR #844, opened 2015-12-01
Force CC to re-open iframe if terms were skipped

nathanstitt:
*Some* people think they're soooo sneaky and will deliberately close CC down and then re-open it rather than agreeing to the terms and conditions.

This changes the status check endpoint to not return user information until all the terms are agreed to.  CC will then re-open the iframe where Tutor will be able to display the terms.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,18 +6,10 @@ class ApplicationController < ActionController::Base
 
   before_filter :authenticate_user!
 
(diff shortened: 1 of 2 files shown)


nathanstitt:
Ok, I *think* this will work.  Unfortunatnly my accounts/tutor link isn't working and I can't seem to test locally :(

jpslav:
Funny NS

PR #845, opened 2015-12-01
CC task stats

Dantemss:
- Refactor CalculateTaskPlanStats to take an array of tasks instead of a plan
- Refactor CalculateTaskPlanStats specs
- Added page UUID to CC score API
- Add stats endpoint for CC
- Specs for new endpoint


app/controllers/api/v1/cc/tasks_controller.rb
```diff
@@ -41,4 +41,34 @@ def show
...
+
+  api :GET, 'cc/tasks/:course_id/:cnx_page_id/stats',
+            'Gets the Concept Coach Task stats for the given CNX page'
+  description <<-EOS
+    The `cnx_page_id` should not contain version information.
+
+    #{json_schema(Api::V1::ConceptCoachStatsRepresenter, include: :readable)}
+  EOS
+  def stats
+    course = Entity::Course.find(params[:course_id])
```
jpslav: do we need to check that this course is a CC course? or if called on a non-CC course will it just return emptiness?

Dantemss: Hm I could make it a 403 for non-cc courses. But should be empty.

jpslav:
Nice, @Dantemss!

PR #846, opened 2015-12-01
Import CC courses

nathanstitt:
Extends the demo configuration to support globing, and then uses that to create 7 concept coach courses using the newly created content.  Note that content only has 6 books currently available, so I'm re-using the current concept-coach derived to get 7. 

When ran as:  `rake demo[small/*]` it will create:

- 8 books
- 8 courses (2 courses have >1 teacher)
- 4 sections/course
- 10 students/section
- 5 assignments

It's named "small" because we eventually want to create:
- 8 books
- 16 courses (2 courses have 4 teachers)
- 4 sections/course
- 200 students/section
- 30 assignments

The current "small" setup I've implemented brings my computer to an absolute halt though, so we may not be able to implement the second variant.


jpslav:
Looks good to me.  Merge after CI.

PR #849, opened 2015-12-02
Content Analyst (and Customer Support) can now see incomplete/failed jobs

Dantemss:
Plus specs (that the controllers work, not exactly testing the links themselves)


app/controllers/content_analyst/jobs_controller.rb
```diff
@@ -0,0 +1,7 @@
+module ContentAnalyst
+  class JobsController < BaseController
+    include Manager::JobActions
+
+    self.job_url_proc = ->(job) { content_analyst_job_path(job.id) }
```
jpslav: Is this used?

PR #852, opened 2015-12-02
Fix uniqueness exceptions for creating CC tasks

Dantemss:
When testing locally, it seems that making the transaction serializable does it... Hopefully won't affect performance too badly?


jpslav:
Hey! transaction_isolation! :-)  cc @joemsak @kevinburleigh75

PR #853, opened 2015-12-03
Allow admins to create ConceptCoach courses

Dantemss:
- Allow admin to set course offering and is_concept_coach in the course edit screen
- Make CourseProfile::Profile belong_to an offering via its primary key, not via a string identifier (change indexes to match)
- Replace a catalog offering's "identifier" string with "salesforce_book_name" and "appearance_code" strings.  "salesforce_book_name" should be globally unique and indexed.
- Change `https://github.com/openstax/tutor-server/blob/master/app/routines/import_salesforce_courses.rb#L52` to `find_by(salesforce_book_name: ...)`
- Change admin create / edit course forms to let the appearance_code be set (with help message to say if left blank will defer to the offering if available)
- Fixed specs


nathanstitt:
The accompanying FE PR also adds support for a "default" code that will be used if the `appearance_code` is blank or null.  If you don't know what a course should be set to you can either send "default" or just omit the code.

jpslav:
Believe will be omitted

Dantemss:
@jpslav ready for review

app/subsystems/course_profile/get_profile.rb
```diff
@@ -4,12 +4,7 @@ class CourseProfile::GetProfile
   protected
   def exec(course: nil, attrs: {})
     profile = course ? get_profile_by_course(course) : get_profile_by_attrs(attrs)
-
-    outputs.profile = {
-      course_id: profile.entity_course_id,
-      catalog_offering_identifier: profile.catalog_offering_identifier,
-      name: profile.name
-    }
+    outputs.profile = profile
```
jpslav: Ok doing this for now, but just means asap we have to PORO course profiles.

PR #856, opened 2015-12-03
Set a random student identifier for demo imported students

nathanstitt:
Otherwise it's hard for the FE to see student id without running concept coach themselves

### DIFF
--- a/lib/tasks/demo/content.rb
+++ b/lib/tasks/demo/content.rb
@@ -72,7 +72,7 @@ class DemoContent < DemoBase
             user = get_student_user(initials) ||
                    new_user(username: student_info.username, name: student_info.name)
             log("    #{initials} #{student_info.username} (#{student_info.name})")
-            run(:add_student, period: period, user: user) \
+            run(:add_student, period: period, user: user, student_identifier: "#{rand(100000)+10000}") \
                unless run(:is_student, user: user, course: course).outputs.user_is_course_student
           end
 


lib/tasks/demo/content.rb
```diff
@@ -72,7 +72,7 @@ def exec(book: :all, print_logs: true, random_seed: nil, version: :defined)
             user = get_student_user(initials) ||
                    new_user(username: student_info.username, name: student_info.name)
             log("    #{initials} #{student_info.username} (#{student_info.name})")
-            run(:add_student, period: period, user: user) \
+            run(:add_student, period: period, user: user, student_identifier: "#{rand(100000)+10000}") \
```
jpslav: is that enough variability to get unique IDs for the ludicrous example where there are 20k students?  (not that uniqueness is enforced, but we might like them to be unique).

nathanstitt: Well, that's a 100,000 range. So given a student population of 20K I guess there's a 1/5 chance that someone will have a duplicate.

I'm thinking that's probably fairly real-life like.  I'm sure that somewhere there's two schools that are using the same ranges and will have duplicate students ids between them.

Not that I pre-thought about that - I just figured "big enough" :grinning:

jpslav: I think the assumption is that within one real course it is very unlikely
to have duplicate ids. In our demo data we will have 20k kids in one course
so there will be dupes. But since it is just informational I'm good with it
as is.

Dantemss: https://en.wikipedia.org/wiki/Birthday_problem I think you will have a 50% chance of at least one collision with only ~300 students.

PR #857, opened 2015-12-03
Let configure webview URL base for demo classes

jpslav:
Previously, the webview URL for demo courses pointed to archive.  This PR makes demo webview_urls use a configurable base, and if not present likely defaults to 'https://cnx.org/contents/'.

@nathanstitt please review.


nathanstitt:
Looks good to me.

PR #858, opened 2015-12-03
Improve SF cron logging; notify for SF cron exceptions

jpslav:
Salesforce cron jobs are getting called but not doing anything.  This PR adds more logging to help figure out what's going on.  It also adds our exception processing machinery to the cron job calls; previously, any exceptions were not being rescued and notified.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -10,9 +10,9 @@ every 1.day, at: '3:00 AM' do
 end
 
 every 1.hour do
-  runner "ImportSalesforceCourses.call"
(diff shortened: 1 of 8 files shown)


config/schedule.rb
```diff
@@ -10,9 +10,9 @@
 end
 
 every 1.hour do
-  runner "ImportSalesforceCourses.call"
+  runner "OpenStax::RescueFrom.this{ ImportSalesforceCourses.call }"
```
jpslav: @joemsak can you check this out and the related code below?

PR #864, opened 2015-12-04
WIP: Upgrading to Lev 7.0

joemsak:
On here for being able to use CI and for anyone to see the kinds of differences introduced


jpslav:
Joe, please change the minimum code required to integrate with new lev, otherwise we'll be at risk of every file changing completely and it'll be impossible to review.  So don't do any code style updates or move methods around in the file, please.

app/routines/calculate_task_stats.rb
```diff
@@ -1,10 +1,14 @@
...
+  lev_routine outputs: { stats: :_self },
+              uses: { name: Role::GetUsersForRoles, as: :get_users_for_roles }
 
   protected
+  def exec(tasks:, details: false)
+    @tasks = tasks
+    @details = details
+
+    set(stats: generate_period_stat_data)
+  end
```
joemsak: @jpslav can it be okay that I moved exec to the top here? I found it really awkward to find it at the bottom, it's not consistent with how our routines usually look

app/routines/distribute_tasks.rb
```diff
@@ -1,28 +1,7 @@
...
-        task_step.number = index + 1
-      end
-
-      entity_task.task.update_step_counts
-    end
-
-    Entity::Task.import! entity_tasks, recursive: true
-    entity_tasks.each(&:clear_association_cache)
-  end
-
```
joemsak: @jpslav same question here regarding the order of functions in routines, keeping the consistency of exec at the top. It's just very awkward to me that the first place the routine starts wouldn't be at the top of the code

jpslav: I agree on this -- better to have a consistent flow (but still want to change it later)

Issue #739, opened 2015-10-26
Don't inactivate already-inactive students


jpslav:
If someone calls the "drop student" API (students#destroy) on an already-dropped student, the `inactive_at` timestamp will be updated, which removes information about when the real drop occurred.  

Make the `InactivateStudent` routine error when already inactive.  Make sure that the drop endpoint reflects this error (add a spec for the routine and for the controller).

PR #867, opened 2015-12-04
Don't drop dropped students

Dantemss:
- Return errors when trying to inactivate inactive students or activate active students
- Specs


Dantemss:
https://github.com/openstax/tutor-server/issues/739

PR #872, opened 2015-12-07
Attempt to fix studentless roles

Dantemss:
Needs a lot of testing on QA since I was never able to reproduce the issue.


Dantemss:
I believe the issue was being caused by Lev raising an `ActiveRecord::Rollback`, which would be caught by Lev's transaction and not rollback the transaction in the controller.

PR #876, opened 2015-12-09
Handle CC content updates in teacher views

Dantemss:
- Handle content updates properly in the CC teacher dashboard and scores view
- Fixed existing specs


Dantemss:
Well writing specs for those content updates is a task in itself... how about we just merge this and test? :D Manual testing works...

PR #878, opened 2015-12-09
Dont send emails for 403 on performance reports

Dantemss:
This controller seems to be missing access specs?


jpslav:
https://github.com/openstax/tutor-server/issues/880

PR #884, opened 2015-12-10
Update the enrollment code migration to use the new save callback

joemsak:


jpslav:
Looks good.  Merge after CI.

PR #886, opened 2015-12-10
Handle some spaced practice corner cases

### LINKED ISSUE #877: Reading/Homework Spaced Practice Algorithm needs changes
There are two simple changes that should be made ASAP:

* after excluding previously-worked exercises, the number of SPEs assigned should be `min(k-ago map value, remaining exercises)`
* even if one k-ago slot cannot be satisfied (because we ran out of exercises), we should continue on to the next slot

Changing [this](https://github.com/openstax/tutor-server/blob/master/app/subsystems/tasks/assistants/homework_assistant.rb#L145) and [this](https://github.com/openstax/tutor-server/blob/master/app/subsystems/tasks/assistants/i_reading_assistant.rb#L167) from 
```ruby
break if candidate_exercises.size < number
```
to
```ruby
number = [candidate_exercises.size, number].min
next if number == 0
```
should do the trick.  Perhaps a new variable should be created instead of changing `number`.

The more complicated adjustment is:
* after excluding previously-worked exercises and drawing from the remaining set, if more exercises are needed we should re-consider previously-worked exercises (though not those already chosen in for the current assignment)

Pseudocode for all of this:
```
chosen_exercises = []

if chosen_exercises.count < number
  chosen_exercises = (candidate_exercises - worked_exercises).shuffle.first(number)

if chosen_exercises.count < number
  chosen_exercises.append (candidate_exercises - chosen_exercises).shuffle.first(number-chosen_exercises.count)

next if chosen_exercises.none?

add_steps_for(chosen_exercises)
```

Dantemss:
Closes #877.


Dantemss:
@jpslav @kevinburleigh75 ready for review

Dantemss:
@jpslav @kevinburleigh75 ready for review and merge... again

jpslav:
@kevinburleigh75 this is all you :-)

app/routines/add_spy_info.rb
```diff
@@ -16,14 +16,16 @@ def exec(to:, from:, save: false)
 
   def values_from(value)
     case value
-    when Content::Models::Ecosystem, Content::Ecosystem
-      { ecosystem_id: value.id, ecosystem_title: value.title }
-    when Array
-      { history: value.map do |task|
```
jpslav: is it ok to lose the `history` key in these changes?

app/routines/get_concept_coach.rb
```diff
@@ -40,18 +40,41 @@ def exec(user:, cnx_book_id:, cnx_page_id:)
...
+                                                                                 .compact
+      forbidden_random_ks.sort.reverse.each do |forbidden_random_k|
+        # Subtract 1 from k_ago because this history does not include the current task (0-ago)
+        random_tasks.delete_at(forbidden_random_k - 1)
+      end
+    end
+
+    spaced_exercises = \
+      Tasks::Models::ConceptCoachTask::SPACED_EXERCISES_MAP.flat_map do |k_ago, num_requested|
+      if k_ago.nil?
```
jpslav: where does this `flat_map` block end?  Can you fix the indentation?

app/routines/get_concept_coach.rb
```diff
@@ -40,18 +40,41 @@ def exec(user:, cnx_book_id:, cnx_page_id:)
...
-    spaced_exercises = spaced_tasks.empty? ? \
-      [] : Tasks::Models::ConceptCoachTask::SPACED_EXERCISES_MAP.flat_map do |k_ago, ex_count|
-      # Subtract 1 from k_ago because this history does not include the current task (0-ago)
-      spaced_task = k_ago.nil? ? spaced_tasks.sample : spaced_tasks[k_ago - 1]
+    spaced_practice_status = []
 
-      # If no k_ago task, pick a random one
-      spaced_task ||= spaced_tasks.sample
+    # Prepare eligible random-ago tasks, but only if we have 4 or more tasks
+    if spaced_tasks.size >= 4
```
kevinburleigh75: If I understand GetHistory correctly, the history includes the current task.  In that case, the inequality should be strict, since random-ago should only happen once there are at least four _past_ events.

app/routines/get_concept_coach.rb
kevinburleigh75: why would `k_ago` be `nil`?

app/routines/get_concept_coach.rb
```diff
@@ -40,18 +40,41 @@ def exec(user:, cnx_book_id:, cnx_page_id:)
...
+      Tasks::Models::ConceptCoachTask::SPACED_EXERCISES_MAP.flat_map do |k_ago, num_requested|
+      if k_ago.nil?
+        # Do not do random-ago if less than 4 past tasks in history
+        next if spaced_tasks.size < 4
+        spaced_task = random_tasks.sample
+        k_ago_str = 'random'
+      else
+        spaced_task = spaced_tasks[k_ago - 1]
+        k_ago_str = k_ago.to_s
+      end
```
kevinburleigh75: I think it would be cleaner to separate the control flow from the assignment logic, and use multiple return values (perhaps keeping the intermediate values in the `if-else-end` branches, but they seem simple enough to eliminate in this case):
```ruby
## skip random-ago if event history is too short
next if k_ago.nil? && spaced_tasks < 4

## select a task an a moniker
spaced_task, k_ago_str = 
  if k_ago.nil?
    random_tasks.sample, 'random'
  else
    spaced_tasks[k_ago-1], k_ago.to_s
  end
```

app/routines/get_concept_coach.rb
```diff
@@ -40,18 +40,41 @@ def exec(user:, cnx_book_id:, cnx_page_id:)
...
+
+      # Skip if no k_ago task
+      if spaced_task.nil?
+        spaced_practice_status << "Not enough tasks in history to fill the #{k_ago_str}-ago slot"
```
kevinburleigh75: perhaps we should have a similar message when skipping random-ago above?  just a thought...

app/routines/get_concept_coach.rb
```diff
@@ -63,35 +86,41 @@ def exec(user:, cnx_book_id:, cnx_page_id:)
       # Map the spaced page to exercises in the current ecosystem
```
kevinburleigh75: I thought we had decided that this wasn't necessary - did that change?

app/routines/get_concept_coach.rb
```diff
@@ -63,35 +86,41 @@ def exec(user:, cnx_book_id:, cnx_page_id:)
...
+      num_repeated_exercises = [repeated_candidate_exercises.size, num_req_repeated_exercises].min
 
-          next if chosen_exercise.nil?
-        end
+      chosen_exercises = []
+      # Randomize and grab the required numbers of exercises
+      chosen_exercises = candidate_exercises.shuffle.first(num_candidate_exercises) \
+        unless num_candidate_exercises == 0
+      chosen_exercises += repeated_candidate_exercises.shuffle.first(num_repeated_exercises) \
+        unless num_repeated_exercises == 0
```
kevinburleigh75: I believe both of these `unless` clauses are unnecessary, as is the initialization of `chosen_exercises`.

app/subsystems/tasks/assistants/homework_assistant.rb
```diff
@@ -134,30 +149,46 @@ def add_spaced_practice_exercise_steps!(task:, taskee:)
...
+      num_repeated_exercises = [repeated_candidate_exercises.size, num_req_repeated_exercises].min
 
-        step = add_exercise_step(task: task, exercise: chosen_exercise)
-        step.group_type = :spaced_practice_group
+      chosen_exercises = []
+      # Randomize and grab the required numbers of exercises
+      chosen_exercises = candidate_exercises.shuffle.first(num_candidate_exercises) \
+        unless num_candidate_exercises == 0
+      chosen_exercises += repeated_candidate_exercises.shuffle.first(num_repeated_exercises) \
+        unless num_repeated_exercises == 0
```
kevinburleigh75: See comments for `get_concept_coach.rb`.

app/subsystems/tasks/assistants/i_reading_assistant.rb
```diff
@@ -156,30 +171,46 @@ def add_spaced_practice_exercise_steps!(task:, taskee:)
...
+      num_repeated_exercises = [repeated_candidate_exercises.size, num_req_repeated_exercises].min
 
-        step = add_exercise_step(task: task, exercise: chosen_exercise)
-        step.group_type = :spaced_practice_group
+      chosen_exercises = []
+      # Randomize and grab the required numbers of exercises
+      chosen_exercises = candidate_exercises.shuffle.first(num_candidate_exercises) \
+        unless num_candidate_exercises == 0
+      chosen_exercises += repeated_candidate_exercises.shuffle.first(num_repeated_exercises) \
+        unless num_repeated_exercises == 0
```
kevinburleigh75: See comments for `get_concept_coach.rb`.

app/subsystems/tasks/models/concept_coach_task.rb
```diff
@@ -1,8 +1,8 @@
 module Tasks::Models
   class ConceptCoachTask < Tutor::SubSystems::BaseModel
 
-    CORE_EXERCISES_COUNT = 4
-    SPACED_EXERCISES_MAP = [[2, 1], [4, 1], [nil, 1]]
+    CORE_EXERCISES_COUNT = 3
+    SPACED_EXERCISES_MAP = [[2, 1], [nil, 1]]
```
kevinburleigh75: We should avoid using `nil` as a special value - would `:random` work instead?

app/routines/add_spy_info.rb
Dantemss: History is now hardcoded in the methods that actually used this feature (they just pass in a hash instead).

app/routines/get_concept_coach.rb
Dantemss: OK and true

app/routines/get_concept_coach.rb
Dantemss: Spoke with Kevin. Will clarify comment.

app/routines/get_concept_coach.rb
Dantemss: I think it will also work for random-ago.

app/routines/get_concept_coach.rb
Dantemss: I'll do this.

app/routines/get_concept_coach.rb
Dantemss: I'll redo this part.

app/subsystems/tasks/models/concept_coach_task.rb
Dantemss: Sure I'll change to :random

app/routines/add_spy_info.rb
Dantemss: And I just noticed that I missed one of them.

app/routines/get_concept_coach.rb
Dantemss: Oh I see what you mean now.

PR #887, opened 2015-12-11
WIP When import real SF data, don't import fake

jpslav:
We probably don't want this before going live.


jpslav:
I actually think this behavior is not desired.

Dantemss:
Is this OBE?

jpslav:
Yeah I don't think we want this, seeing as how when real courses were imported the team also wanted to import fake.

PR #894, opened 2015-12-16
Admin bulk update course ecosystems

karenc:
- Show ecosystem for courses on courses admin page
- Allow multiple extra fields to be added on admin courses page
- Add bulk set ecosystem functionality to admin courses page
- Add spec for bulk updating course ecosystem
- Add bulk course ecosystem jobs tabs to admin courses index page

course-ecosystem-bulk-update


Dantemss:
The failing spec is in master so don't worry about it.

PR #896, opened 2015-12-16
Add validated return_to's on login URLs

jpslav:
URLs with hosts of `cnx.org`, `openstax.org`, and `localhost` are now accepted as `return_to` arguments on the login URL.  After login, Tutor will redirect to those URLs.  Any other `return_to` values will be ignored.

This is a backup way to login for CC, CNX.

### DIFF
--- a/config/initializers/openstax_accounts.rb
+++ b/config/initializers/openstax_accounts.rb
@@ -5,11 +5,12 @@ secrets = Rails.application.secrets['openstax']['accounts']
 # By default, stub unless in the production environment
 stub = secrets['stub'].nil? ? !Rails.env.production? : secrets['stub']
 
-APPROVED_HOST_REGEXES = [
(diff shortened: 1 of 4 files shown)


config/initializers/openstax_accounts.rb
```diff
@@ -5,6 +5,12 @@
 # By default, stub unless in the production environment
 stub = secrets['stub'].nil? ? !Rails.env.production? : secrets['stub']
 
+APPROVED_HOST_REGEXES = [
+  /.*openstax\.org$/,
+  /.*cnx\.org$/,
```
Dantemss: I think .* can be left out. Also, ruby usually says to use \z for some reason.

jpslav: For a good reason! http://stackoverflow.com/a/577675

PR #897, opened 2015-12-16
Blow up on invalid TaskPlan content

Dantemss:
- TaskPlans now fail to validate if all page_ids and exercise_ids don't belong to the TaskPlan's ecosystem
- TaskPlans now fail to validate if there are no page_ids (reading) or exercise_ids (homework)
- Specs


jpslav:
Looks good to me.  Merge after CI.

PR #900, opened 2015-12-17
Highlight real production admin views

jpslav:
Puts a red border on the admin, customer service, and customer analyst nav bars for the real production deployment, as a visual hint to people to not go deleting stuff :-)

image


jpslav:
cc @gregfitch this is for you :-)  (and prior us's)

PR #904, opened 2015-12-18
add katex css for front end maths

pandafulmanda:
for use with openstax/tutor-js#908


philschatz:
The `tutor-js` side of this Pull Request was closed

PR #909, opened 2015-12-20
Add throttled endpoint for FE logging

jpslav:
Adds an endpoint that the FE can use to submit log messages to the server. 

```
POST /api/log/entry

{
  level: "info",
  message: "tutor-js wishes tutor-server a Merry Christmas"
}
```

Both `level` and `message` are required.  

`level` can be one of `"unknown"`,  `"fatal"`, `"error"`, `"warn"`, `"info"`, or `"debug"` -- these match the default Rails logging levels.  By default, `"debug"` is not logged in production environments.  The `level` field is case insensitive.

`message` is restricted to 1000 characters; messages beyond that length are truncated.

This endpoint is throttled to 50 requests per day per IP.

cc @nathanstitt since we were recently talking about a situation where FE logging to the server would have been handy and since we also talked about another potential use for throttling.

### DIFF
--- a/app/controllers/api/v1/log_controller.rb
+++ b/app/controllers/api/v1/log_controller.rb
@@ -43,7 +43,7 @@ class Api::V1::LogController < Api::V1::ApiController
     else
       message = entry_params.message[0..MAX_LOG_LENGTH]
 
-      Rails.logger.log(level, message)
+      Rails.logger.log(level, "(ext) #{message}")
       head :created
     end
   end
--- a/app/controllers/api/v1/log_controller.rb
+++ b/app/controllers/api/v1/log_controller.rb
@@ -1,0 +1,51 @@
+class Api::V1::LogController < Api::V1::ApiController
+
+  LOG_LEVELS = {
+    "unknown" => Logger::UNKNOWN,
+    "fatal" => Logger::FATAL,
+    "error" => Logger::ERROR,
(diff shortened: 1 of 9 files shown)


jpslav:
@nathanstitt can you review?

app/controllers/api/v1/log_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      render_api_errors(errors)
+    else
+      message = entry_params.message[0..MAX_LOG_LENGTH]
+
+      Rails.logger.log(level, "(ext) #{message}")
+      head :created
+    end
+  end
+
+end
```
nathanstitt: Should this endpoint be callable by an anonymous user?  If so, we need a `skip_before_filter :authenticate_user!`   If not, it might be helpful to log the user's id.

nathanstitt:
Looks good @jpslav

jpslav:
Good catch thanks! I do separately want to log user ID on all log messages.

app/controllers/api/v1/log_controller.rb
jpslav: Actually, on closer inspection. the API endpoints don't have the `authenticate_user!` before_filter.  They just check for logged-in user on an as-needed basis.  The before_filter is in `ApplicationController` which is not what the API controllers inherit from.  I didn't remember that this was the case.

PR #911, opened 2015-12-22
Enrollment prevalidate endpoint

nathanstitt:
Adds an `/enrollment_changes/prevalidate` endpoint that can be called by an anonymous user.  It replies with either a `{response: true}` or a 422 "invalid_enrollment_code" header if invalid.

### DIFF
--- a/app/subsystems/course_membership/validate_enrollment_parameters.rb
+++ b/app/subsystems/course_membership/validate_enrollment_parameters.rb
@@ -1,0 +1,16 @@
+# Checks if an enrollment code is valid for a course
+class CourseMembership::ValidateEnrollmentParameters
+
+  lev_routine express_output: :is_valid
+
+  protected
+
+  def exec(book_uuid:, enrollment_code: )
+    ecosystem = Content::Ecosystem.find_by_book_uuid(book_uuid)
+    if ecosystem
+      period = CourseMembership::Models::Period.find_by(enrollment_code: enrollment_code)
+    end
+    outputs[:is_valid] = !!(ecosystem && period && period.course.ecosystems.exists?(id: ecosystem.id))
+  end
+
+end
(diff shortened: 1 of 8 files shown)


nathanstitt:
Ok, FE PR is completed.  Ready for review

philschatz:
https://github.com/openstax/concept-coach/pull/100 looks fine. waiting on BE :ok_hand: .

app/subsystems/course_membership/validate_enrollment_parameters.rb
```diff
@@ -0,0 +1,16 @@
...
+  lev_routine express_output: :is_valid
+
+  protected
+
+  def exec(book_uuid:, enrollment_code: )
+    ecosystem = Content::Ecosystem.find_by_book_uuid(book_uuid)
+    if ecosystem
+      period = CourseMembership::Models::Period.find_by(enrollment_code: enrollment_code)
+    end
+    outputs[:is_valid] = !!(ecosystem && period && period.course.ecosystems.exists?(id: ecosystem.id))
```
jpslav: I'm a little concerned that because the code earlier is returning the `first` ecosystem that matches a book UUID that that this check here might not work (if two courses use the same book and have different ecosystems, or if one has an earlier version of an ecosystem that another doesn't have).  

Does this work instead?

```ruby
def exec(book_uuid:, enrollment_code:)
  period = CourseMembership::Models::Period.find_by(enrollment_code: enrollment_code)
  period.course.ecosystems.any?{|ecosystem| ecosystem.books.any?{|book| book.uuid == uuid}}
end
```

(just checking that the course has a book with the given UUID, instead of finding an ecosystem for the UUID and matching it)

nathanstitt: @jpslav, Great point.  There's a bug report waiting to happen there as soon as multiple courses are setup with the same book.

I've reworked along the lines you suggested.

PR #912, opened 2015-12-22
Create performance report export with valid worksheet names

karenc:
Valid characters for Excel sheet names:

 - does not exceed 31 characters
 - does not contain : \ / ? * [ ]

### DIFF
--- a/app/subsystems/tasks/performance_report/export_xlsx.rb
+++ b/app/subsystems/tasks/performance_report/export_xlsx.rb
@@ -32,9 +32,16 @@ module Tasks
         end
       end
 
+      def worksheet_name(name)
+        # worksheet names cannot be longer than 31 characters
+        name = name[0..30]
+        # worksheet names cannot contain characters :\/?*[]
+        name.gsub(/[:\\\/\[\]*?]/, '-')
+      end
+
       def create_data_worksheets(performance_report, package)
         performance_report.each do |report|
-          package.workbook.add_worksheet(name: report[:period][:name]) do |sheet|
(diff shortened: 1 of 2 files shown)


jpslav:
wunderbar!

philschatz:
Should FE (and `tutor-server`) Error when the teacher tries to create/edit a Period name?


/cc @darrenmason, @nathanstitt?

Fredasaurus:
Maybe I need context. Why would we create a scenario where something is editable, or looks to be, and then error?

jpslav:
I think he's saying should we restrict the valid period names to be those that won't give errors when we export to Excel.  

The context is that Excel restricts what characters can be in worksheet names, and we use Period names for worksheet names.  This pull request (merged) replaces invalid Excel worksheet name characters with hyphens.  So using the exception example from this morning, a period name of "101 - 7:30AM  - Waldo" would lead to a worksheet name of "101 - 7-30AM  - Waldo"

jpslav:
I'm in favor of not restricting period name characters to what Excel can handle, and just doing a search and replace like this PR does.

philschatz:
thanks @jpslav! and it would cut off the period name after 31 characters (which might be a bit confusing). Would cutting it off at 31 characters introduce a new bug? ie does Excel allow multiple worksheets with the same name?

Dantemss:
I don't think we ever intended period names to be that long...

jpslav:
Yes we could be subject to another bug if people make period names that are the same for the first 31 characters (or that become the same when we substitute out illegal characters).  I doubt it will happen (I think our users will see rendering issues that encourage them to make the names shorter), but I will make an issue nonetheless.

PR #915, opened 2015-12-29
[bugfix] Don't extract url if it's a fragment identifier 

nathanstitt:
Fixes bug when links in content were only a fragment identifier

   `Absolute URI missing hierarchical segment: 'https:#Figure_19_02_Const_volt'`


nathanstitt:
I attempted to update the spec at `spec/lib/openstax/cnx/v1/fragment/interactive_spec` with a page containing fragment href.  That worked, but then some of the other features in the spec didn't get tested.  I decided to leave it as-is unless we want to update.

PR #918, opened 2016-01-05
Removed the http status code based on job status from `api/jobs#show`

Dantemss:
This fixes the chain 500 errors in the teacher dashboard.


app/controllers/api/v1/jobs_controller.rb
```diff
@@ -28,21 +28,6 @@ def index
...
-    respond_with job, represent_with: Api::V1::JobRepresenter, status: code
-  end
-
-  private
-  def http_status_code(status)
-    case status
-    when Lev::BackgroundJob::STATE_SUCCEEDED
-      200
-    when Lev::BackgroundJob::STATE_FAILED
-      500
```
philschatz: Could we just change this to a non-500 status?

Dantemss: The representer already has a status field that returns a string.

Issue #855, opened 2015-12-03
Make bad URLs on content import not be so bad


jpslav:
Users often input a bad URL or base URL / UUID combination to the ecosystem import page, then the background job starts and fails over and over with `OpenURI::HTTPError` 404 exceptions that get emailed out to devs.

Options (maybe not mutually-exclusive): 

1. Make sure ecosystem import URL reachable before start background import job (do a head request on the URL, if no good, give an error back to the import page and don't start the background job).
2. Make these exceptions not send notifications (as long as they still halt the background job, so just a openstax_rescue_from config) -- if don't want to block notifications for all `OpenURI::HTTPError`, we'd need to wrap the error in something content import-specific and block that.
3. Make it easier to kill jobs that are retrying to no good end.

PR #925, opened 2016-01-07
Retry only some exceptions

Dantemss:
- Updated Lev
- Made lev background jobs retry only some exceptions
- Specs


Dantemss:
Closes https://github.com/openstax/tutor-server/issues/855 since those jobs won't be retried.

Issue #898, opened 2015-12-16
"Pretty" 500 pages not getting compiled CSS


jpslav:
In https://github.com/openstax/tutor-server/pull/847 I made nice looking error pages for those errors that are staticly served. (normal exception error pages already look nice).

For some reason on production, the error pages are not getting linked to the precompiled `application.css`:

![image](https://cloud.githubusercontent.com/assets/1001691/11857295/bb9554a0-a40b-11e5-8b5a-57a12cfa4ee3.png)

(visible at https://tutor.openstax.org/500.html)

The CSS should have a hash digest on it but doesn't:

```html
<!DOCTYPE html>
<html class="no-js" lang="en">
  <head>
    <meta charset="utf-8">
    <title>500 error - OpenStax</title>
    <link rel="stylesheet" media="all" href="/stylesheets/application.css" digest="true" />
  </head>
  <body>
    <div class='static-error'>
      <div class='no-css-message'>To see this page properly styled, precompile assets</div>
      <div class='logo'></div>
      <section id="error">
        <h1>An unexpected error occurred!</h1>
        This is a 500 error.  Please <a href="https://openstaxtutor.zendesk.com/">contact support</a> if you need assistance or <a href="/">return to the home page</a>.
      </section>
    </div>
  </body>
</html>
```

PR #927, opened 2016-01-08
Fix error pages

Dantemss:
Should work properly in both production and dev now.

Closes https://github.com/openstax/tutor-server/issues/898


jpslav:
:+1:

Issue #917, opened 2016-01-04
More worksheet name checks


jpslav:
Follow on to https://github.com/openstax/tutor-server/pull/912, @philschatz pointed out that we could have another bug if a teacher had multiple period names that were the same in the first 31 characters.  I think this is unlikely, but nonetheless it is possible.

As a low priority task, update the exporter to ensure that the worksheet names (after illegal characters are removed) are unique.

PR #932, opened 2016-01-12
Solve collisions between period names that start with 31 identical characters

Dantemss:
- With specs


Dantemss:
Closes https://github.com/openstax/tutor-server/issues/917

Issue #369, opened 2015-06-11
Add fallbacks for BigLearn RealClient


Dantemss:
Maybe retry 1-2 times but then get the questions locally if it fails?

Dantemss:
Send an exception email to devs if we fallback to a local query.

PR #936, opened 2016-01-13
Added a fallback if we fail to talk to Biglearn and specs

Dantemss:
- If we fail to contact Biglearn 3 times, resort to getting random local exercises and notify developers
- Specs


Dantemss:
Closes #369

PR #939, opened 2016-01-15
Removed old migration-like tasks

Dantemss:
Since hopefully all servers should be migrated now (confirm that demo has been migrated or that we don't care before merging)


jpslav:
@Dantemss I leave this to you to check with @cnuber about status of demo, then you can merge.

Dantemss:
@cnuber said anything important is on par or newer than Production, so I'll merge this.

Issue #938, opened 2016-01-15
Upgrade Lev to version with Jobba


jpslav:
https://github.com/lml/lev/pull/49 uses Jobba, which should give us *much* better performance for monitoring background jobs.

Upgrade Tutor's Lev to the version on this branch.  I'll leave the branch unmerged in case any changes are required.  When Tutor is happy with the branch, we can merge the lev branch and release 7.0.0.  

Note the required changes to the Lev configuration (see the [Lev README](https://github.com/lml/lev/tree/jobba#calling-routines-as-activejobs)).  

You'll need to update [all of the code](https://github.com/openstax/tutor-server/search?utf8=%E2%9C%93&q=BackgroundJob) that pulls background job info.

Most routines in Tutor shouldn't need updating as Lev already handles the default state changes, etc.  At least one example that does is https://github.com/openstax/tutor-server/blob/master/app/subsystems/tasks/export_performance_report.rb#L36, which should change to 

```ruby
status.save(url: export.url)
```

Note that `job` inside routines is no longer -- use `status` instead.  

There are a couple of `job.status` and `job.errors` that you should check on (admin views)

PR #941, opened 2016-01-15
Updated Lev to the newest version and added Jobba

Dantemss:
- Updated Lev to the newest version and added Jobba
- Specs


Dantemss:
Closes #938

Dantemss:
I think the nil status might just be jobs interfering with each other in Redis, but I have no clue what causes the nil data.

Dantemss:
The 2 that I think might be bugs in Lev/Jobba are:

1.

```
Failure/Error: visit admin_courses_path
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./app/controllers/admin/courses_controller.rb:15:in `block in index'
```

This line says: `job.data['course_ecosystem']`

So `job.data` can be nil?

2.

```
Failure/Error: ImportSalesforceCourses[include_real_salesforce_data: true]
     NoMethodError:
       undefined method `started!' for nil:NilClass
```

Simply calling the routine causes a `NoMethodError` because the job status is nil again.

app/views/manager/ecosystems/_incomplete.html.erb
```diff
@@ -10,8 +10,8 @@
   <% jobs.each do |job| %>
     <tr>
       <td><%= link_to(job.id, job_path_proc.call(job)) %></td>
-      <td><%= job.status %></td>
-      <td><%= link_to(job.ecosystem_import_url, job.ecosystem_import_url, target: '_blank') %></td>
+      <td><%= job.state %></td>
```
jpslav: You might want this to be `job.state.name` otherwise I think you'll be printing out the `Jobba::State` object.

Dantemss:
Hm wait this breaks the accounts dev login... the search routine probably has stuff in the initialize method too.

jpslav:
Hmm... probably want to still allow initialize method to be used instead of moving code out of initialize again.

app/views/manager/ecosystems/_incomplete.html.erb
Dantemss: Oh missed this one.

jpslav:
@Dantemss do we need changes to tutor-deployment for the secrets use of `jobba` key?

Dantemss:
Here's the deployment PR: https://github.com/openstax/tutor-deployment/pull/202

cnuber:
Merged the deployment counterpart, please merge

jpslav:
thanks folks!

Issue #945, opened 2016-01-22
Spec failure on Master (either OS X or timezone-based)


Dantemss:
Does not happen for Karen/on Travis.

```
Dantes-MacBook-Pro:tutor-server dante$ bundle exec ruby -S rspec spec/lib/export_data_spec.rb 
100 / 108
F

Failures:

  1) ExportData with book exports data as a xlsx file
     Failure/Error: expect(data['First Completed At'].beginning_of_minute).to eq(
       
       expected: 2016-01-22 16:22:00.000000000 +0000
            got: 2016-01-22 10:22:00.000000000 +0000
       
       (compared using ==)
       
       Diff:
       @@ -1,2 +1,2 @@
       -Fri, 22 Jan 2016 16:22:00 UTC +00:00
       +2016-01-22 10:22:00 UTC
       
     # ./spec/lib/export_data_spec.rb:80:in `block (3 levels) in <top (required)>'

Finished in 20.72 seconds (files took 10.02 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/lib/export_data_spec.rb:62 # ExportData with book exports data as a xlsx file
```

PR #948, opened 2016-01-22
Change research export excel file dates to strings in iso8601 format

karenc:
The dates now look like this: "2016-01-22T20:03:34Z".

The failed test:

```
$ TZ=America/Chicago rspec spec/lib/export_data_spec.rb
100 / 108
F

Failures:

  1) ExportData with book exports data as a xlsx file
     Failure/Error: expect(data['First Completed At'].beginning_of_minute).to eq(

       expected: 2016-01-22 17:23:00.000000000 +0000
            got: 2016-01-22 11:23:00.000000000 +0000

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -Fri, 22 Jan 2016 17:23:00 UTC +00:00
       +2016-01-22 11:23:00 UTC

     # ./spec/lib/export_data_spec.rb:80:in `block (3 levels) in <top (required)>'

Finished in 31.97 seconds (files took 5.9 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/lib/export_data_spec.rb:62 # ExportData with book exports data as a xlsx file

Coverage report generated for RSpec to /home/karen/tutor-server/coverage. 3318 / 4371 LOC (75.91%) covered.
[Coveralls] Outside the CI environment, not sending data.
```

The reason for this is that the time is converted to the local / system
timezone in excel, but when it was read back out, the timezone was set to UTC.

This commit changes the date exported to a string so excel doesn't parse
it.

### DIFF
--- a/lib/export_data.rb
+++ b/lib/export_data.rb
@@ -106,10 +106,10 @@ class ExportData
           step.number,
           type,
           step.group_name,
-          [step.first_completed_at, style: date],
-          [step.last_completed_at, style: date],
-          [step.task.opens_at, style: date],
-          [step.task.due_at, style: date],
+          [step.first_completed_at.try(:iso8601)],
+          [step.last_completed_at.try(:iso8601)],
+          [step.task.opens_at.try(:iso8601)],
+          [step.task.due_at.try(:iso8601)],
           url
         ]
 
--- a/spec/lib/export_data_spec.rb
+++ b/spec/lib/export_data_spec.rb
@@ -77,14 +77,10 @@ RSpec.describe ExportData do
       expect(data['Step ID']).to eq(step.id)
       expect(data['Step Type']).to eq('Reading')
       expect(data['Group']).to eq(step.group_name)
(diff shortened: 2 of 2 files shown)


Dantemss:
So while this probably fixes the spec, I'm wondering if relying on Excel's local times is good at all... Won't the researchers read some weird times too if the server timezone differs from their own?

karenc:
I think the server timezone is set to UTC.  If we use a string instead, it'll look like `2016-01-22 18:04:31 UTC`.  Is it ok?

Dantemss:
Maybe we can just write the format Phil wants with `strftime`?

karenc:
What format does Phil want?

Dantemss:
ISO 8601... You already have the format written somewhere in there I think

karenc:
Ah yes...  Ok, let me change this.

karenc:
The dates now look like this: "2016-01-22T20:03:34Z".

Dantemss:
Looks good.

Dantemss:
Solves #945

Issue #740, opened 2015-10-26
Update CourseMembership::Models::Student (inactive_at) and User::Models::Profile (deleted_at) to use the Paranoia gem


Dantemss:

Dantemss:
PR is made, not sure when it will be merged, but I will close this to remove it from my list...

PR #950, opened 2016-01-23
Convert all soft-deleted models to use the paranoia gem

Dantemss:
Low priority. Will need a lot of testing once it is done. Solves https://github.com/openstax/tutor-server/issues/740.


Dantemss:
This is done... not sure when to merge or if we wanna turn this into a refactoring of the course membership subsystem... There are so many routines there (and in the domain) that do almost the same thing with slightly different assumptions...

Dantemss:
Merge this during Summer.

PR #954, opened 2016-01-27
Stats menu in Tutor

Dantemss:
- Added course stats menu
- Controller specs


Dantemss:
Merging so we can test since this was high priority...

PR #955, opened 2016-01-27
Optimized the UpdateSalesforceStats routine to be faster

Dantemss:
- Optimized the UpdateSalesforceStats routine to be faster
- Fix specs


Dantemss:
We can make this run more often, if desired. Also, last I checked we are nowhere close to the daily API limit.

Dantemss:
Merge this during Summer.

Issue #861, opened 2015-12-04
Populate salesforce_book_name dropdowns in Offering from SF


jpslav:
Make an ActiveForce object to query to Book objects from SF to get the list of valid names.  So we don't incur an API hit on each page load with the offering, store them in Redis, run the query once a day, and add a button on the SF admin page to trigger an update manually.  This should not change existing values, just the options in the dropdown.

Dantemss:
Solved in https://github.com/openstax/tutor-server/pull/960

PR #960, opened 2016-01-30
Salesforce book names

Dantemss:
Closes https://github.com/openstax/tutor-server/issues/861

- Populate book names from SF
- Results are cached
- Cache is regenerated every day, when we update the Salesforce stats
- (Some) specs

### DIFF
--- a/app/views/admin/catalog_offerings/_edit_row.html.erb
+++ b/app/views/admin/catalog_offerings/_edit_row.html.erb
@@ -9,11 +9,12 @@
 <div class='offering-row'>
   <div class="form-group">
     <%= form.label :salesforce_book_name %>
-    <%= form.text_field :salesforce_book_name,  {class: 'form-control'} %>
+    <%= form.select :salesforce_book_name, @salesforce_book_names,
+                    { include_blank: true }, {class: 'form-control'} %>
   </div>
   <div class="form-group">
     <%= form.label :appearance_code %>
-    <%= form.text_field :appearance_code,  {class: 'form-control'} %>
+    <%= form.text_field :appearance_code, {class: 'form-control'} %>
   </div>
   <div class="form-group">
     <%= form.label :description %>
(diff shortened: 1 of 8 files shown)


Dantemss:
As a side note, we won't be able to create/edit course offerings in dev, but maybe we don't care...

config/secrets.yml.example
```diff
@@ -32,6 +32,7 @@ development:
       cache: <%= ENV["REDIS_NAMESPACES_CACHE"] || 'cache' %>
       settings: <%= ENV["REDIS_NAMESPACES_SETTINGS"] || 'settings' %>
       jobba: <%= ENV["REDIS_NAMESPACES_LEV"] || 'jobba' %>
+      active_force: <%= ENV["REDIS_NAMESPACES_ACTIVE_FORCE"] || 'active_force' %>
```
jpslav: Need a corresponding tutor-deployment PR?

app/views/admin/catalog_offerings/_edit_row.html.erb
```diff
@@ -9,11 +9,12 @@
 <div class='offering-row'>
   <div class="form-group">
     <%= form.label :salesforce_book_name %>
-    <%= form.text_field :salesforce_book_name,  {class: 'form-control'} %>
+    <%= form.select :salesforce_book_name, @salesforce_book_names,
+                    { include_blank: true }, {class: 'form-control'} %>
```
jpslav: To enable course offering creation and editing in dev/QA, how about:

```ruby
<% if is_real_production_site? %>
  <%= form.text_field :salesforce_book_name,  {class: 'form-control'} %>
<% else %>
  <%= form.select :salesforce_book_name, @salesforce_book_names,
                  { include_blank: true }, {class: 'form-control'} %>
<% end %>
```

config/secrets.yml.example
Dantemss: I thought there was but apparently not... yeah we need one.

app/views/admin/catalog_offerings/_edit_row.html.erb
Dantemss: Sure, I can do that.

Issue #951, opened 2016-01-25
Optimize N+1 queries in lib/export_data.rb and convert the output to CSV


Dantemss:
The export takes so long mostly because it is doing one or more queries per `TaskStep` in the DB.

Dantemss:
Convert the output to CSV since that's what the researchers use...

PR #963, opened 2016-02-02
Optimize research export and make it output CSV

Dantemss:
Closes https://github.com/openstax/tutor-server/issues/951

- Optimize out N+1 queries in research export
- Make its output CSV
- Updated spec


Dantemss:
Tested on QA and it finishes in ~5 minutes (650,000 rows), using about 320MB.

PR #968, opened 2016-02-05
Concept Coach stats page

Dantemss:
- Added Concept Coach stats page
- Specs

dummy data
(dummy data)


nathanstitt:
Looks awesome @Dantemss, you might consider slapping a JS table sorter onto it.  I've used the [StupidTablePlugin](https://github.com/joequery/Stupid-Table-Plugin) before with good results, it's very small and works well enough.  If you'd like I can help with it.

Dantemss:
I was thinking of somehow collapsing the books instead or in addition to sorting. What do you think?

Dantemss:
I ended up going with https://github.com/jmosbech/StickyTableHeaders and Bootstrap collapse... Order is fixed because it's TOC order, though we could potentially sort the books themselves if necessary.

PR #978, opened 2016-02-09
Added total_average and attempted_average to the performance report

Dantemss:
- Added total_average and attempted_average to the performance report
- Updated specs

Hopefully the correct API this time...

### DIFF
--- a/app/representers/api/v1/performance_report_representer.rb
+++ b/app/representers/api/v1/performance_report_representer.rb
@@ -111,7 +111,11 @@ module Api::V1
                writeable: false,
                getter: ->(*) { DateTimeUtilities.to_api_s(due_at) }
 
-      property :average,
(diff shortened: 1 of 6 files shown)


Dantemss:
@jpslav btw I opted to calculate the averages here because we have to do it anyway for the export

Dantemss:
@darrenmason
This requires FE changes on:
https://github.com/openstax/tutor-js/blob/master/src/components/scores/table-hs.cjsx#L93
https://github.com/openstax/tutor-js/blob/master/src/components/scores/table-hs.cjsx#L96
https://github.com/openstax/tutor-js/blob/master/src/components/scores/table-cc.cjsx#L72
https://github.com/openstax/tutor-js/blob/master/src/components/scores/table-cc.cjsx#L75

PR #986, opened 2016-02-16
Include all tags with exercises

nathanstitt:
Indicate which ones should be normally displayed via a new 'is_visible' property

Needs to be merged & deployed along with FE PR https://github.com/openstax/tutor-js/pull/990, otherwise teachers will see all tags on the HW builder.

### DIFF
--- a/app/representers/api/v1/exercise_representer.rb
+++ b/app/representers/api/v1/exercise_representer.rb
@@ -32,6 +32,7 @@ module Api::V1
                readable: true,
                writeable: false,
                decorator: TagRepresenter,
+               getter: ->(*) { (tags + tags.flat_map(&:teks_tags)).uniq },
                schema_info: { required: true,
                               description: 'Tags for this exercise' }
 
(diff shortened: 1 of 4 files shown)


app/representers/api/v1/exercise_representer.rb
```diff
@@ -31,9 +31,6 @@ class ExerciseRepresenter < Roar::Decorator
     collection :tags,
                readable: true,
                writeable: false,
-               getter: ->(*) {
-                 (tags + tags.flat_map(&:teks_tags)).select{ |tag| tag.visible? }.uniq
-               },
```
jpslav: @nathanstitt do we still need the getter so we can include the `teks_tags`?

```ruby
getter: ->(*) { (tags + tags.flat_map(&:teks_tags)).uniq
```

?

nathanstitt: Yes, we do.  Thanks for pointing that out!

jpslav:
Looks good.  @philschatz and @nathanstitt can merge this at their discretion (still says WIP so I won't merge)

nathanstitt:
Sorry! forgot to remove the WIP prefix

PR #992, opened 2016-02-19
Background job changes

karenc:
- Fix 500 error when viewing background job without custom data

  The error was:

  ```
Failure/Error: Unable to find matching line from backtrace
NoMethodError:
  undefined method `each' for nil:NilClass
  # ./app/views/manager/jobs/show.html.erb:25:in `_app_views_manager_jobs_show_html_erb__1808779094055715127_97159720'
  ```

- Rename jobs index page from "Queued jobs" to "Jobs"

    It is confusing because "queued" is a state that not all of the jobs are
    in.

- Change admin jobs feature specs to look up jobs by id …

  `Jobba.all.to_a.last` was not always returning the job we expected (the
  job that was most recently created).

- Show more information on admin background jobs page

### DIFF
--- a/app/controllers/manager/job_actions.rb
+++ b/app/controllers/manager/job_actions.rb
@@ -13,7 +13,7 @@ module Manager::JobActions
 
   def show
     @job = Jobba.find(params[:id])
-    @custom_fields = @job.data
+    @custom_fields = @job.data || {}
     render 'manager/jobs/show'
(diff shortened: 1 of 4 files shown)


karenc:
Ready for review

karenc:
Before:

jobs_before

After:

jobs_after

PR #994, opened 2016-02-25
Course exercise listings and exclusions

jpslav:
`GET /courses/:course_id/exercises`

^^ Would be just like the existing `/ecosystems/:ecosystem_id/exercises(/:pool_types)` endpoint, but we'd also add a `is_excluded` boolean to each exercise (or a nested object of course exercise options).

For excluding/including, phil and I decided to go with:

`PUT /courses/:course_id/exercises/:exercise_id` with body of `{id: 123, is_excluded: false}`

This PR also needs to maintain a BL pool of excluded exercises by course.  When the teacher excludes or unexcludes a new exercise, we need to make a new pool and send to BL.  Somewhere in Tutor we need to store that BL pool ID attached to the course so that we can use it in BL API calls (see KB for new APIs)


app/controllers/api/v1/cc/course_exercises_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  api :GET, '/courses/:course_id/exercises',
+            "Returns exercises for a given course, filtered by the page_ids param and optionally an array of pool_types"
+  description <<-EOS
+    Returns a list of exercises in the specified course associated with the pages with the given ID's and pool types.
+
+    #{json_schema(Api::V1::ExerciseSearchRepresenter, include: :readable)}
+  EOS
+  def index
```
jpslav: This started as a copy of `GET /api/ecosystems/:id/exercises` with an `is_excluded` added to the search results, but today I changed it to have `GetExercises` instead of what was in the ecosystems controller.  Was thinking this was more direct / less code in the controller.

app/controllers/api/v1/cc/course_exercises_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+        page_ids: params[:page_ids],
+        pool_types: params[:pool_types]
+      },
+      with: :course_exercise_options
+    ]
+
+    exercise_hashes = exercises.collect do |_, exercise|
+      Api::V1::ExerciseRepresenter.new(exercise).to_hash.tap do |hash|
+        hash['pool_types'] = exercise['pool_types']
+        hash['course_exercise_options'] = exercise['course_exercise_options']
```
jpslav: This really needs to be `CourseExerciseRepresenter.new(<right hand side>)` or similar -- right now the right hand side is an object.

spec/controllers/api/v1/course_exercises_controller_spec.rb
```diff
@@ -0,0 +1,221 @@
+require 'rails_helper'
+
+RSpec.describe Api::V1::CourseExercisesController, type: :controller, api: true, version: :v1 do
```
jpslav: this is only a verbatim copy of the `EcosystemsController` spec.

app/controllers/api/v1/cc/course_exercises_controller.rb
```diff
@@ -0,0 +1,63 @@
+class Api::V1::CourseExercisesController < Api::V1::ApiController
```
nathanstitt: This file is located in the `api/v1/cc` directory.  When `/api/courses/1/exercises` is loaded "for real", it causes a `ActionController::RoutingError (uninitialized constant Api::V1::CourseExercisesController)` exception because the router can't find the controller.

When I move the file it get's further but then throws an `SecurityTransgression` even though I'm logged in as a teacher.  Debugging that now

app/queries/get_exercises.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    @course_exercise_options ||=
+      CourseContent::Models::ExerciseOptions.where(entity_course_id: @by[:course].id).all.each_with_object do |eo, hash|
+        hash[eo.exercise_uid] = eo
+      end
+    end
+
+    @result[exercise.uid]['course_exercise_options'] = @course_exercise_options[exercise.uid]
+  end
+
```
nathanstitt: this `end` is extra and causes a `SyntaxError`

app/queries/get_exercises.rb
```diff
@@ -0,0 +1,78 @@
...
+  def pages
+    @pages ||= @by[:page_ids] ? ecosystem.pages_by_ids(@by[:page_ids]) : ecosystem.pages
+  end
+
+  def pools
+    if !@pools
+      # Default to all pool types
+      pool_types = [by[:pool_types]].flatten.compact.uniq
+      pool_types = Content::Pool.pool_types if pool_types.empty?
```
nathanstitt: `NameError (undefined local variable or method `by' for #<GetExercises:0x007fcdbc3a1998>):`

app/queries/get_exercises.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+    end
+
+    result
+  end
+
+  def ecosystem
+    raise "Either :ecosystem or :course must be provided" unless @by[:ecosystem] || @by[:course]
+    @ecosystem ||= @by[:ecosystem] || GetCourseEcosystems[course: @by[:course]].first
+  end
```
nathanstitt: `uninitialized constant GetExercises::GetCourseEcosystems`.  And here I'm stumped. I don't see a `GetCourseEcosystems` or anything similar.  Is that still TODO?

nathanstitt:
I think I've almost got this working locally, but am stumped by the missing `GetCourseEcosystems`  Is that missing, or does it exist but the name's wrong?

app/controllers/api/v1/cc/course_exercises_controller.rb
nathanstitt: Had to add:
```ruby
    when :exercises
      requestor.is_human? && \
      (UserIsCourseStudent[user: requestor, course: course] || \
```
to `course_access_policy`

app/queries/get_exercises.rb
nathanstitt: Should be `::CourseContent::GetCourseEcosystems`

app/queries/get_exercises.rb
```diff
@@ -0,0 +1,78 @@
...
+  def self.[](by:, with: nil)
+    new.run(by: by, with: with)
+  end
+
+  def run(by:, with:)
+    @by = by
+    @with = [with].flatten.compact
+
+    pools.each do |pool|
+      pool.exercises(preload_tags: true).each do |exercise|
```
nathanstitt: `pools` is a nested array.  adding a `.flatten` , i.e. ` pools.flatten.each do |pool|`

app/queries/get_exercises.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def record_exercise(exercise)
+    @result[exercise.uid] ||= exercise
+  end
+
+  def record_pool_type(exercise, pool_type)
+    @result[exercise.uid]['pool_types'] ||= []
+    @result[exercise.uid]['pool_types'] << pool_type
+  end
```
nathanstitt: `@result` is nil, probably needs to be set in `run` ?

jpslav:
Hey @nathanstitt this is WIP in the truest since of the word, and is not expected to work at all.  Has been passed to @Dantemss for completion.

jpslav:
@Dantemss -- #999 will probably be useful as you work on this.

nathanstitt:
Yep, was just hoping I could make it work with a bit of polish but kept falling further down the rabbit hole :)  @Dantemss Let me know if you'd like me to push up my fixes

Dantemss:
@nathanstitt I rewrote a lot of it so most likely your fixes are old now :)

Dantemss:
Uh getting there... the only change to JP's description so far is that the exercise id is now at the end of the url `PATCH /courses/:course_id/exercises/:exercise_id`

Dantemss:
Oh and... actually using the exclusion lists :)

nathanstitt:
Given re-introduction of the 'save' button, it'll be best if this would accept an array of exclusions.  That way the FE can send all the newly included/excluded exercises at once.

Dantemss:
I guess that means the patch route changes to `PATCH /courses/:course_id/exercises` and the id goes in the body. Body is now an array of `{id: blah, is_excluded: true/false}`

Dantemss:
Hmm... practice tasks are not really associated with a course. But I guess since they now respect the exclusion pools, maybe they should be?

app/controllers/api/v1/course_exercises_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+        exercise_representation['is_excluded'] = true
+      elsif !exercise_params.is_excluded.nil? # false
+        excluded_exercise = CourseContent::Models::ExcludedExercise.find_by(
+          entity_course_id: @course.id, exercise_number: exercise.number
+        ).try(:destroy)
+        exercise_representation['is_excluded'] = false
+      end
+
+      exercise_representation
+    end
```
jpslav: Why have a full-blown exercise representation that we parse and return?  Why not just have the JSON be the `is_excluded` field and not much else?  (The "exercise options" approach).  This seems like a lot of work that we don't necessarily need here and looks to align with a future where teachers edit exercises for a course (but that doesn't work b/c exercises are shared between courses).  If this direction is because the route is "PUT /courses/42/exercises" then we could always have a separate restful endpoint dedicated to course exercise options or even excluded exercises (I know we talked about some stuff like this).  Maybe we can discuss.

app/routines/get_ecosystem_exercises_from_biglearn.rb
```diff
@@ -9,17 +9,28 @@ class GetEcosystemExercisesFromBiglearn
...
+    excluded_pools = [course_profile.try(:biglearn_excluded_pool_uuid)].compact.map do |uuid|
+      OpenStax::Biglearn::V1::Pool.new(uuid: uuid)
+    end
+
     attempts = 0
     begin
       urls = OpenStax::Biglearn::V1.get_projection_exercises(
         role:              role,
         pools:             biglearn_pools,
+        excluded_pools:    excluded_pools,
```
jpslav: I'm pretty sure the BL PR for this does not include changes to the real client (just the fake) - can you address? (coordinating with KB)

jpslav:
Lookin good @Dantemss -- review done.

app/controllers/api/v1/course_exercises_controller.rb
Dantemss: I mostly added all the exercise info because the FE might have to re-render everything... @nathanstitt maybe you don't need it?

Dantemss: Let me ask this in a better way, @nathastitt: do you need the full exercise with all fields in the response from the `PATCH` route?

app/routines/get_ecosystem_exercises_from_biglearn.rb
Dantemss: I talked to Kevin and I'll go ahead and do this.

Dantemss:
Actually Kevin is now making updates to the Biglearn API so we can specify if the exercise versions in the pool should be compared (admin case) or not (teacher case)

nathanstitt:
Is this still being worked, or ready for merge?  The FE is complete so we can potentially merge whenever

Dantemss:
The Biglearn part (and the Biglearn real client) still need to be finished

app/controllers/api/v1/course_exercises_controller.rb
nathanstitt: No, just the `is_excluded` is sufficient.  We'll merge whatever you send back with what we had previously.  Which makes conceptual sense since it's a `PATCH`

Dantemss:
Renamed the parameter on method calls to Biglearn clients to `pool_exclusions` to account for the fact that they are now no longer arrays of `OpenStax::Biglearn::V1::Pool`. The API call still uses `excluded_pools`.

Dantemss:
RealClient specs with exclusion pools have to wait for the Biglearn API to be in dev (for the VCR cassettes).

Dantemss:
Uh apparently this PR is missing exclusion in some places that I added in https://github.com/openstax/tutor-server/pull/1019. We can merge the PR's or keep them separated for review.

Dantemss:
Merge #1020 before this.

Dantemss:
This is ready. Needs to be deployed at the same time as https://github.com/openstax/biglearn-platform/pull/82

Dantemss:
Now it's really ready :)

Dantemss:
I won't say anything anymore other than I hope that Travis succeeds.

jpslav:
Woohoo!

PR #997, opened 2016-02-25
More improvements for old jobs

Dantemss:
- Changed jobs controllers to return 404's on #show when the jobs don't exist
- Return publish information directly on TaskPlans so FE doesn't have to make extra requests for the jobs.

### DIFF
--- a/app/representers/api/v1/courses/dashboard_representer.rb
+++ b/app/representers/api/v1/courses/dashboard_representer.rb
@@ -146,7 +146,7 @@ module Api::V1::Courses
     collection :plans,
                readable: true,
                writeable: false,
-               decorator: TaskPlanRepresenter
+               decorator: ::Api::V1::TaskPlanRepresenter
 
     collection :tasks,
                readable: true,
--- a/app/representers/api/v1/courses/dashboard_representer.rb
+++ b/app/representers/api/v1/courses/dashboard_representer.rb
@@ -4,7 +4,7 @@ module Api::V1::Courses
 
     include ::Roar::JSON
 
-    class Base < Roar::Decorator
+    class TaskBase < Roar::Decorator
       include Roar::JSON
       include Representable::Coercion
 
@@ -17,49 +17,7 @@ module Api::V1::Courses
                type: String,
                readable: true,
                writeable: false
-    end
 
-    class Plan < Base
-
-      property :is_trouble,
-               readable: true,
-               writeable: false,
-               schema_info: { type: 'boolean' }
-
-      property :type,
-               type: String,
-               readable: true,
-               writeable: false
-
-      property :is_publish_requested,
-               readable: true,
-               writeable: true,
-               schema_info: { type: 'boolean' }
-
-      property :published_at,
-               type: String,
-               readable: true,
-               writeable: false,
-               getter: ->(*) { DateTimeUtilities.to_api_s(published_at) }
-
-      property :publish_last_requested_at,
-               type: String,
-               readable: true,
-               writeable: false,
-               getter: ->(*) { DateTimeUtilities.to_api_s(publish_last_requested_at) }
-
-      property :publish_job_uuid,
-               type: String,
-               readable: true,
-               writeable: false
-
-      collection :tasking_plans,
-                 readable: true,
-                 writeable: false,
-                 decorator: Api::V1::TaskingPlanRepresenter
-    end
-
-    class TaskBase < Base
       property :opens_at,
                type: String,
                readable: true,
@@ -188,7 +146,7 @@ module Api::V1::Courses
     collection :plans,
                readable: true,
                writeable: false,
-               decorator: Plan
+               decorator: TaskPlanRepresenter
 
     collection :tasks,
                readable: true,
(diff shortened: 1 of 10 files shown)


Dantemss:
Although this task requires an associated FE PR, this particular BE PR should be backwards-compatible so you can merge any time.

Dantemss:
Actually scratch that, wait for the FE PR :)

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -36,24 +42,25 @@ class TaskPlanRepresenter < Roar::Decorator
...
 
-    property :publish_job_uuid,
-             type: String,
+    property :publish_job,
+             decorator: JobRepresenter,
              readable: true,
-             writeable: false
+             writeable: false,
+             getter: ->(*) { Jobba.find(publish_job_uuid) },
+             if: ->(*) { !publish_job_uuid.blank? }
```
jpslav: What do you think about leaving the existing `publish_blah` fields (for backwards compatibility) and being more aggressive about new ways to represent jobs on this Plan?

e.g.

```ruby
collection :jobs,
           readable: true,
           writeable: false
           decorator: Api::V1::JobRepresenter,
           getter: ->(*) { Jobba.where(job_arg: "gid://.../TaskPlan/42").to_a }
```

We'd want all of the information currently in the plan representer for the publish job to be duplicated in these job entries.  Most of the timestamp stuff is available (tho likely with different names); we'd probably need to add in the `publish_job_url` as `job_url` or something.

We would need to modify `DistributeTasks` to set the job name and args:

```ruby
  def exec(task_plan, publish_time = Time.now, protect_unopened_tasks = false)
    status.set_job_name("DistributeTasks")
    status.set_job_args(task_plan: task_plan.to_global_id.to_s)
```

After this, we could change task plan destroy to use a background job and by setting its job args the job info would automatically popup in the `jobs` collection above (then FE could check to see when a destroy is ongoing -- @pandafulmanda asked for this last fall).

What do you think?  With these changes we get away from needing to separately track job UUIDs in activerecords and wouldn't need one-off `last_requested_at` fields, etc.

app/representers/api/v1/courses/dashboard_representer.rb
```diff
@@ -17,49 +17,7 @@ class Base < Roar::Decorator
...
-
-      property :is_trouble,
-               readable: true,
-               writeable: false,
-               schema_info: { type: 'boolean' }
-
-      property :type,
-               type: String,
-               readable: true,
-               writeable: false
```
jpslav: Did we lose `type` with the switch away from this dashboard-specific representer to the main `TaskPlanRepresenter`?

app/representers/api/v1/task_plan_representer.rb
jpslav: We'd also want some script that goes through all publish UUIDs that we have and calls `status.set_job_name` and `status.set_job_args` for them so that our new data structure has all the data of the old data structure.

Dantemss: The issue with this is that it will break for old jobs, since those don't have the job_arg set. Would need to migrate old jobs somehow?

Dantemss: Oh I guess you mentioned that :)

Dantemss: How do we tell which one is the most recent publish? Are they sorted? Also... maybe the FE only cares about the most recent one?

app/representers/api/v1/courses/dashboard_representer.rb
Dantemss: The TaskPlanRepresenter has type: https://github.com/openstax/tutor-server/blob/master/app/representers/api/v1/task_plan_representer.rb#L18

app/representers/api/v1/task_plan_representer.rb
Dantemss: Personally I'd rather have the FE have to know less about jobs, instead of even more... (for example, how to find the destroy job in an array)

pandafulmanda: Does this mean that `DELETE` is running as a job now? or not yet?  Also, what if the `DELETE` returns the corresponding job id?  If jobs does become an array on the `TaskPlanRepresenter`, would we be told the job's purpose -- i.e. publish vs. delete?

Dantemss: Oh right I guess delete is still not a job. I wonder if we should just fix the speed problem instead of making it a job? No reason delete should take 10+ seconds (it's mostly just running validations... which maybe we can skip since we are deleting?)

pandafulmanda:
By the way openstax/tutor-js#1004 has been updated to work with this.  If there will be further changes to `TaskPlanRepresenter`, let me know when those are settled so I can update accordingly

pandafulmanda:
@jpslav @Dantemss is this likely to be merged as is, or will there be updates/changes to this for this sprint?

jpslav:
I believe @pandafulmanda and @Dantemss are good with this PR.  I leave to @philschatz to merge with the FE one when it is time.

PR #999, opened 2016-02-26
add excluded_pools to biglearn fake_client

kevinburleigh75:
Updates to the real_client are still needed.


Dantemss:
Oh btw I merged because I shall put this code to use :) Thanks, @kevinburleigh75

PR #1001, opened 2016-02-29
WIP Course-specific excluded exercises

Dantemss:
- Added CourseContent::ExcludedExercise model to keep track of course-specific excluded exercises

# TODO

- New or adapted API route to send questions to FE with exclusion info
- New API route so FE can exclude questions
- Update spaced practice/personalized algorithms to not pick excluded exercises
- Update personalized/spaced practice algorithms to skip steps without errors if no questions available (might need FE update?)


Dantemss:
Wait... JP had a WIP PR... guess I'll merge my stuff with his.

PR #1004, opened 2016-03-01
CC spreadsheet export changes

jpslav:
Changes the concept coach exported spreadsheet to the new format.

### TODO

* [x] Add comments on data.
* [x] More specs
* [x] Make sure actually works at integration level
* [x] Make other spreadsheet code use the new helper
* [x] Add documentation to `XlsxHelper`

### Things We Can't Do

* The tool we use to generate XLSX files does not have an ability to format text within a cell comment, so we cannot use bold text like in the mocked up file.
* [Excel has a known issue](http://answers.microsoft.com/en-us/office/forum/office_2007-excel/autofit-row-height-doesnt-work/ca2a86b8-585a-46fd-a8a7-377e76ac5528?msgId=38709b42-daa4-4739-ae8c-ab8dcf2e2f0c&auth=1) where merged cells with wrapped text don't grow their height to show the wrapped text.

### DIFF
--- a/app/subsystems/tasks/performance_report/export_cc_xlsx.rb
+++ b/app/subsystems/tasks/performance_report/export_cc_xlsx.rb
@@ -9,6 +9,11 @@ module Tasks
         filename
       end
 
+      # So can be called like other exporters
+      def self.[](profile:, report:, filename:)
+        call(course_name: profile.name, report: report, filename: filename)
+      end
+
       def initialize(course_name:, report:, filepath:)
         @course_name = course_name
         @report = report
@@ -46,11 +51,12 @@ module Tasks
           @pct_L = s.add_style border: {edges: [:left], :color => '000000', :style => :thin}, num_fmt: Axlsx::NUM_FMT_PERCENT
           @right_R = s.add_style border: {edges: [:right], :color => '000000', :style => :thin}, alignment: {horizontal: :right}
           @date_R = s.add_style border: {edges: [:right], :color => '000000', :style => :thin}, num_fmt: 14
-          @average_style = s.add_style b: true, border: { edges: [:top], :color => '000000', :style => :medium}
-          @average_R = s.add_style b: true, border: { edges: [:top, :right], :color => '000000', :style => :medium}, border_right: {style: :thin}
+          @average_style = s.add_style b: true, border: { edges: [:top], :color => '000000', :style => :medium}, format_code: "#.0"
+          @average_R = s.add_style b: true, border: { edges: [:top, :right], :color => '000000', :style => :medium}, border_right: {style: :thin}, format_code: "#.0"
           @average_pct = s.add_style b: true,
                                      border: { edges: [:top], :color => '000000', :style => :medium},
                                      num_fmt: Axlsx::NUM_FMT_PERCENT
+
         end
       end
 
@@ -163,7 +169,7 @@ module Tasks
                   correct_count,
                   {
                     style: @normal_L,
-                    comment: "Correct: #{(correct_pct*100).round(0)} Completed: #{(completed_pct*100).round(0)}"
+                    comment: "Correct: #{(correct_pct*100).round(0)}% Completed: #{(completed_pct*100).round(0)}%"
                   }
                 ])
                 student_columns.push(completed_count)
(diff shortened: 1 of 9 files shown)


jpslav:
@Dantemss please review

app/subsystems/tasks/performance_report/export_cc_xlsx.rb
```diff
@@ -0,0 +1,276 @@
+module Tasks
+  module PerformanceReport
+    class ExportCcXlsx
+
+      def self.call(course_name:, report:, filename:)
```
Dantemss: You don't like routines anymore? XD

We might want a transaction here so we get 1 consistent snapshot of the DB.

app/subsystems/tasks/performance_report/export_cc_xlsx.rb
```diff
@@ -0,0 +1,276 @@
...
+        end
+
+        @helper.add_row(sheet, data_heading_columns)
+
+        # STUDENT DATA
+
+        first_student_row = sheet.rows.count + 1
+
+        students = report[:students].deep_dup
+        students.sort_by!{|student| student[:last_name]}
```
Dantemss: Hm... Couldn't have just used `sort_by` instead of `deep_dup` and `sort_by!`?

lib/xlsx_helper.rb
```diff
@@ -0,0 +1,150 @@
...
+    suffix = suffix.gsub(INVALID_WORKSHEET_NAME_CHARS_REGEX, '-')
+
+    # worksheet names cannot be longer than 31 characters so we truncate any
+    # names that will be 31 characters or longer when the suffix is included.
+    # we also add a counter to avoid collisions between truncated names.
+
+    if name.size <= 31 - suffix.size
+      "#{name}#{suffix}"
+    else
+      raise(IllegalArgument, "`suffix` is limited to 27 characters") if suffix.size > 27
```
Dantemss: I think this check should happen at the top of this method.

lib/xlsx_helper.rb
```diff
@@ -0,0 +1,150 @@
...
+      sheet.merge_cells sheet.rows.last.cells[merge_range] unless merge_range.size == 1
+    end
+
+    optioned_values.comments.each_with_index do |comment, index|
+      next if comment.nil?
+      sheet.add_comment(ref: row.cells[index], text: comment, author: "", visible: false)
+    end
+  end
+
+  def self.cell_ref(row:, column:)
```
Dantemss: I think we found that Axlsx has a built-in method that does this?

Dantemss:
Looks good overall, but hard to picture the excel file without seeing it :)

philschatz:
For what it's worth, GitHub now allows uploading arbitrary files to Pull Requests :smile:

app/subsystems/tasks/performance_report/export_cc_xlsx.rb
jpslav: The `report` is already the result of a transaction outside of this file that just writes the XLSX file and I don't think there's any DB stuff going on this PR, with the one `course_profile.name` call.

app/subsystems/tasks/performance_report/export_cc_xlsx.rb
jpslav: Indeed!  since I'm just changing the order and not mutating anything lower than this, I don't actually need the deep copy.

lib/xlsx_helper.rb
jpslav: The 27 character limit is actually only true when `name.size <= 31 - suffix.size`.  When that isn't true, the suffix can be up to 31 characters long.  `suffix` is only supplied by coders, so I think it could be here, but I see your point and I think it'd be ok and clearer to just require that `suffix` size always be less than or equal to 27 characters.

lib/xlsx_helper.rb
jpslav: [Ah you're right](http://www.rubydoc.info/github/randym/axlsx/Axlsx#col_ref-class_method) but is it as amazing as my interpolated string??? :-)

Dantemss: Well your string won't work past column `26*26` :)

jpslav: Man..... you and @kevinburleigh75 are always right.  Argh!

jpslav: Changed :-)

PR #1011, opened 2016-03-07
Simple System Notifications

nathanstitt:
Sets up an admin screen for storing system notifications messages to the redis store.  Notices are stored as a JSON encoded string and then served from the `/api/notifications` endpoint.

### DIFF
--- a/app/controllers/api/v1/notifications_controller.rb
+++ b/app/controllers/api/v1/notifications_controller.rb
@@ -6,7 +6,8 @@ class Api::V1::NotificationsController < Api::V1::ApiController
   EOS
 
   def index
-    # NO SECURITY!  Should it at least check if the user is logged in?
+    # Note: this endpoint is unsecured by design.
+    # Notifications are intended to be viewable by anyone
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/notifications_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  api :GET, '/notifications', 'Request current system notifications'
+  description <<-EOS
+    #{json_schema(Api::V1::NotificationRepresenter, include: :readable)}
+  EOS
+
+  def index
+    # NO SECURITY!  Should it at least check if the user is logged in?
+    render json: Settings::Notifications.raw
+  end
```
nathanstitt: Note the comment. Should we check logged in status? Keep in mind that each client is going to hit this endpoint every X (currently 5) minutes.

jpslav: probably no security risk?  if going to be used on home page, we don't want to check logged in status.  Otherwise we can.

PR #1012, opened 2016-03-08
Add student info to course JSON

jpslav:


jpslav:
@Dantemss or @nathanstitt to review

nathanstitt:
:+1:  Looks good to me!

app/routines/collect_course_info.rb
```diff
@@ -98,4 +106,17 @@ def set_ecosystem_book(info, entity_course)
...
+
+    student_role_ids = info.roles.collect{|rr| rr[:id] if rr[:type] == 'student'}.compact
+
+    return nil if student_role_ids.empty?
+
+    info.students = entity_course.students
+                                 .select{ |student|
+                                   student_role_ids.include?(student.entity_role_id)
+                                 }
+  end
```
nathanstitt: This is fine for now since it's only returning a single student, which CC can assume will always be the currently logged in user.  However if it ever returns multiple students, then CC will need to know the id of the currently logged in user so it can search the `students` array for the current user.

PR #1015, opened 2016-03-10
Admin delete empty courses

Dantemss:
- Allow admins to delete courses that have no teachers, periods or students


jpslav:
Nice @Dantemss

PR #1020, opened 2016-03-16
Admin exclude exercises

Dantemss:
Merge this before #994

- Combined the Settings and GlobalSettings modules into one
- Add DB settings to track excluded exercise numbers, versions and pool uuid
- Remove admin excluded exercises from HW builder, Question Library, Practice Widget, Try Another, Spaced Practice, Personalized, Concept Coach
- Specs


app/routines/choose_exercises.rb
```diff
@@ -0,0 +1,36 @@
...
+    repeated_exercises = []
+
+    # Partition exercises into new exercises and the repeated exercises
+    exercises.each do |ex|
+      if all_worked_exercise_numbers.include?(ex.number)
+        repeated_exercises << ex
+      else
+        new_exercises << ex
+      end
+    end
```
jpslav: Concerned at all about the efficiency of the `include?` call inside the loop?  Maybe isn't a worry, but I wonder if doing differences of arrays would be faster, a la

```ruby
> history = [1,2,3,4]
 => [1, 2, 3, 4] 
> exercises = [3,4,5,6]
 => [3, 4, 5, 6] 
> new = exercises - history
 => [5, 6] 
2.2.3 :005 > repeat = exercises - new
 => [3, 4] 
```

In any case if you keep it as is I think you can use the built-in `partition` method (http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-partition)

Dantemss: Right... I just checked the docs and `include?` is `O(N)` (I thought it was better because of some ruby array magic, but guess not). I'll definitely try to improve this chunk of code.

app/routines/get_ecosystem_exercises_from_biglearn.rb
```diff
@@ -36,7 +49,13 @@ def exec(ecosystem:, role:, pools:, count:, difficulty: 0.5, allow_repetitions:
...
         retry
       else
-        numbers = pools.flat_map{ |pl| pl.exercises.map(&:number) }.uniq.sample(count)
+        pool_exercises = pools.flat_map(&:exercises)
+        candidate_exercises = run(:filter, exercises: pool_exercises,
+                                           course: course).outputs.exercises
+        history = run(:get_history, role: role, type: :all).outputs
+        chosen_exercises = run(:choose, exercises: candidate_exercises, count: count,
+                                        history: history, allow_repeats: false).outputs.exercises
+        numbers = chosen_exercises.map(&:number).uniq
```
jpslav: Is there a spec for this block of code?

app/routines/get_exercises.rb
```diff
@@ -21,6 +21,8 @@ def exec(ecosystem: nil, course: nil, page_ids: nil, pool_types: nil)
                                                        page_ids: page_ids,
                                                        pool_types: pool_types]
 
+    admin_excluded_uids = Settings::Exercises.excluded_uids.split(',').map(&:strip)
+
     excluded_exercise_numbers = CourseContent::Models::ExcludedExercise
                                   .where(entity_course_id: course.id)
                                   .pluck(:exercise_number) unless course.nil?
```
jpslav: Above, this code looked like `course.excluded_exercises.pluck(:exercise_number)` -- can it be that here too?

app/routines/choose_exercises.rb
Dantemss: I think maybe making `all_worked_exercise_numbers` a Set would work. `Set#include?` is `O(1)`.

app/routines/reset_practice_widget.rb
```diff
@@ -35,7 +37,8 @@ def exec(role:, exercise_source:, page_ids: nil, chapter_ids: nil, randomize: tr
       ecosystem = Content::Ecosystem.find_by_exercise_ids(exercises.first.id) if exercises.any?
     when :local
       ecosystem, pools = get_ecosystem_and_pools(page_ids, chapter_ids, role)
-      exercises = get_local_exercises(ecosystem, count, role, pools, randomize: randomize)
+      exercises = get_local_exercises(ecosystem: ecosystem, count: count,
+                                      role: role, pools: pools, randomize: randomize)
```
jpslav: thanks for making named args

app/routines/send_tasked_exercise_answer_to_exchange.rb
```diff
@@ -6,7 +6,7 @@ class SendTaskedExerciseAnswerToExchange
 
   def exec(tasked_exercise:)
     url = tasked_exercise.url
-    roles = tasked_exercise.task_step.task.taskings.collect{ |t| t.role }
+    roles = tasked_exercise.task_step.task.taskings.map(&:role)
```
jpslav: The `collect` is dead. Long live the `map`!

config/initializers/rails_settings_ui.rb
```diff
@@ -8,9 +8,50 @@
...
+        excluded_number, excluded_version = excluded_uid.strip.split('@')
+        OpenStax::Biglearn::V1::Exercise.new(question_id: excluded_number,
+                                             version: excluded_version.to_i,
+                                             tags: [])
+      end
+
+      excluded_pool = OpenStax::Biglearn::V1::Pool.new(exercises: excluded_exercises)
+      excluded_pool = OpenStax::Biglearn::V1.add_pools([excluded_pool]).first
+      Settings::Exercises.excluded_pool_uuid = excluded_pool.uuid
+    end
```
jpslav: nice!

config/initializers/rails_settings_ui.rb
```diff
@@ -8,9 +8,50 @@
...
+      field = if setting_value.to_s.size > 36
+        text_area_tag("settings[#{setting_name}]", setting_value.to_s, options.merge(rows: 10))
+      else
+        text_field_tag("settings[#{setting_name}]", setting_value.to_s, options)
+      end
+
+      help_block_content = I18n.t("settings.attributes.#{setting_name}.help_block", default: '')
+      field + (help_block_content.presence && content_tag(:span, help_block_content, class: 'help-block'))
+    end
+  end
```
jpslav: nicer!

lib/settings.rb
```diff
@@ -1,16 +1,11 @@
...
+    mattr_accessor :store
+  end
 
+  module Redis
+    mattr_accessor :store
   end
 end
 
-require_relative 'settings/notifications'
+Dir[File.join(__dir__, 'settings', '*.rb')].each{ |file| require file }
```
jpslav: nice to break it apart

jpslav:
Review complete

app/routines/get_exercises.rb
Dantemss: Right... and I forgot I could use the FilterExcludedExercises here too (without a course param at least).

app/routines/reset_practice_widget.rb
Dantemss: np :)

app/routines/send_tasked_exercise_answer_to_exchange.rb
Dantemss: :+1:

app/routines/get_ecosystem_exercises_from_biglearn.rb
Dantemss: Not sure how good this is, but it's here: https://github.com/openstax/tutor-server/blob/master/spec/routines/get_ecosystem_exercises_from_biglearn_spec.rb#L40

PR #1023, opened 2016-03-17
tasks with immediate feedback

nathanstitt:
Is safe to be merged before FE code, just won't do anything until that's completed

This adds a `is_feedback_immediate` field on `task_plans`.  The distribute tasks routine then checks that and sets the `feedback_at` on `tasks`  to equal either the `opens_at` or `due_at`

### DIFF
--- a/app/representers/api/v1/task_plan_representer.rb
+++ b/app/representers/api/v1/task_plan_representer.rb
@@ -45,7 +45,6 @@ module Api::V1
     property :is_feedback_immediate,
              readable: true,
              writeable: true,
-             getter: ->(*) { is_feedback_immediate? },
              schema_info: { type: 'boolean' }
 
     property :publish_last_requested_at,
--- a/app/representers/api/v1/task_plan_representer.rb
+++ b/app/representers/api/v1/task_plan_representer.rb
@@ -42,6 +42,12 @@ module Api::V1
              getter: ->(*) { is_publish_requested? },
              schema_info: { type: 'boolean' }
 
+    property :is_feedback_immediate,
+             readable: true,
+             writeable: true,
+             getter: ->(*) { is_feedback_immediate? },
+             schema_info: { type: 'boolean' }
+
     property :publish_last_requested_at,
              type: String,
              readable: true,
--- a/app/representers/api/v1/task_representer.rb
+++ b/app/representers/api/v1/task_representer.rb
@@ -74,7 +74,6 @@ module Api::V1
              as: :is_feedback_available,
              writeable: false,
              readable: true,
-             getter: ->(*) { feedback_available? },
              schema_info: { type: 'boolean', description: "If the feedback should be shown for the task" }
 
     property :spy,
--- a/app/representers/api/v1/task_representer.rb
+++ b/app/representers/api/v1/task_representer.rb
@@ -70,6 +70,13 @@ module Api::V1
                  description: "The steps which this Task is composed of"
                }
 
+    property :feedback_available?,
+             as: :is_feedback_available,
+             writeable: false,
+             readable: true,
+             getter: ->(*) { feedback_available? },
+             schema_info: { type: 'boolean', description: "If the feedback should be shown for the task" }
+
     property :spy,
              type: Object,
              readable: true,
--- a/app/routines/distribute_tasks.rb
+++ b/app/routines/distribute_tasks.rb
@@ -65,9 +65,10 @@ class DistributeTasks
       entity_task.taskings << tasking
 
       task = entity_task.task
+
       task.opens_at = opens_ats[ii]
       task.due_at = due_ats[ii] || (task.opens_at + 1.week)
-      task.feedback_at ||= task.due_at
+      task.feedback_at = task_plan.is_feedback_immediate? ? task.opens_at : task.feedback_at
     end
 
     save(entity_tasks)
--- a/app/routines/distribute_tasks.rb
+++ b/app/routines/distribute_tasks.rb
@@ -68,7 +68,7 @@ class DistributeTasks
 
       task.opens_at = opens_ats[ii]
       task.due_at = due_ats[ii] || (task.opens_at + 1.week)
-      task.feedback_at = task_plan.is_feedback_immediate? ? task.opens_at : task.feedback_at
+      task.feedback_at ||= task_plan.is_feedback_immediate? ? task.opens_at : task.feedback_at
     end
 
     save(entity_tasks)
--- a/app/routines/distribute_tasks.rb
+++ b/app/routines/distribute_tasks.rb
@@ -68,7 +68,7 @@ class DistributeTasks
 
       task.opens_at = opens_ats[ii]
       task.due_at = due_ats[ii] || (task.opens_at + 1.week)
-      task.feedback_at ||= task_plan.is_feedback_immediate? ? task.opens_at : task.feedback_at
+      task.feedback_at ||= task_plan.is_feedback_immediate? ? task.opens_at : task.due_at
     end
 
     save(entity_tasks)
--- a/app/routines/propagate_task_plan_updates.rb
+++ b/app/routines/propagate_task_plan_updates.rb
@@ -8,14 +8,12 @@ class PropagateTaskPlanUpdates
 
     # For now we only handle tasking_plans that point to periods
     task_plan.tasking_plans.each do |tasking_plan|
-
       period = tasking_plan.target
       raise 'Cannot propagate plan changes for plan not assigned to a period' \
         unless period.is_a?(CourseMembership::Models::Period)
 
-          task_plan.assistant.update_tasks_for_plan(tasking_plan: tasking_plan, where: {
-                                                    taskings: { course_membership_period_id: period.id }
-                                                  })
+      task_plan.tasks.joins(:taskings).where(taskings: { course_membership_period_id: period.id })
+                     .update_all( task_plan.assistant.updated_attributes_for(tasking_plan:tasking_plan) )
     end
 
     task_plan.tasks.reset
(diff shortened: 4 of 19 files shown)


app/routines/distribute_tasks.rb
```diff
@@ -65,9 +65,10 @@ def exec(task_plan, publish_time = Time.now, protect_unopened_tasks = false)
       entity_task.taskings << tasking
 
       task = entity_task.task
+
       task.opens_at = opens_ats[ii]
       task.due_at = due_ats[ii] || (task.opens_at + 1.week)
-      task.feedback_at ||= task.due_at
+      task.feedback_at = task_plan.is_feedback_immediate? ? task.opens_at : task.feedback_at
     end
```
nathanstitt: *Note:*  The old code only set feedback_at if it was `nil`.  I'm not sure why it didn't update the `feedback_at` date when distributing as it seems like that's something that should always be set.  Maybe I'm missing something though?  At any rate the specs still all pass

app/representers/api/v1/task_representer.rb
```diff
@@ -70,6 +70,13 @@ class TaskRepresenter < Roar::Decorator
                  description: "The steps which this Task is composed of"
                }
 
+    property :feedback_available?,
+             as: :is_feedback_available,
+             writeable: false,
+             readable: true,
+             getter: ->(*) { feedback_available? },
```
jpslav: I think you don't need this `getter` (I think the first argument to `property` is sufficient)?

If that is indeed the case, then the `property` on line 45 can become 

```ruby
property :is_feedback_immediate,
         as: :is_feedback_immediate
```

without the `getter`

app/routines/distribute_tasks.rb
jpslav: I'm guessing this was just a fallback to make sure something was in there.  Looks like `feedback_at` is set a number of other places like `ExtraAssignmentAssistant` and `IReadingAssistant`.  Should this still be `||=` to allow those pieces of code to set `feedback_at`?  That seems slightly problematic given that the `PropagateTaskPlanUpdates` would be overriding it, but it also doesn't seem good to let the assistants set it when it is just overridden.  Seems like we need one way or the other.

app/representers/api/v1/task_representer.rb
nathanstitt: Yes, that's not needed,  thanks - I've now removed it.

app/routines/propagate_task_plan_updates.rb
```diff
@@ -13,8 +13,11 @@ def exec(task_plan:)
       raise 'Cannot propagate plan changes for plan not assigned to a period' \
         unless period.is_a?(CourseMembership::Models::Period)
 
-      feedback_at = task_plan.type == 'homework' ? tasking_plan.due_at : Time.now
-
+      feedback_at = if task_plan.type == 'homework'
+                      task_plan.is_feedback_immediate? ? tasking_plan.opens_at : tasking_plan.due_at
```
jpslav: Should the setting of the feedback_at call back to the assistant?  So we don't end up with "if this type, do this, if that type do that".... just wondering.

jpslav: or should all non-homework plans have `is_feedback_immediate == true`? and then no type-specific checks here

app/routines/distribute_tasks.rb
nathanstitt: Yes, that makes sense. I guess I find it a bit unclear where the assistant's responsibility lies vs the distribution routine.  It seems to me that it would make more sense if assistant built tasks, and the `distribute_tasks` only created taskings and marked the `task_plan` as published.  I may not be completely understanding the routine though - plus that cleanup is probably better left for another PR that's dedicated to just that.

For now I've restored the `||=`

nathanstitt:
I've added a `GenericAssistant` base class for assistants, which contains an initializer that sets `task_plan` & `taskees` (all the assistants shared that code). 

It also has a `update_tasks_for_plan` method that contains most of the code from `PropagateTaskPlanUpdates`   The `HomeworkAssistant` overrides that method to adjust the `feedback_at` attribute, which removes the nasty `if task_plan.type == 'homework'` conditional.

app/routines/distribute_tasks.rb
Dantemss: If the assistant set feedback_at, the DistributeTaskPlans would keep it as is. Otherwise, it would set it to the due date. You can get rid of this behavior, but keep in mind you have to account for readings and CC (instant) and homeworks (due date) somewhere.

app/routines/distribute_tasks.rb
```diff
@@ -65,9 +65,10 @@ def exec(task_plan, publish_time = Time.now, protect_unopened_tasks = false)
       entity_task.taskings << tasking
 
       task = entity_task.task
+
       task.opens_at = opens_ats[ii]
       task.due_at = due_ats[ii] || (task.opens_at + 1.week)
-      task.feedback_at ||= task.due_at
+      task.feedback_at ||= task_plan.is_feedback_immediate? ? task.opens_at : task.feedback_at
```
jpslav: I don't think the last clause should be `task.feedback_at` since that is what comes before the `||=`, i.e. `blah ||= blah` doesn't make sense.

app/routines/propagate_task_plan_updates.rb
```diff
@@ -5,20 +5,17 @@ class PropagateTaskPlanUpdates
...
 
-      feedback_at = task_plan.type == 'homework' ? tasking_plan.due_at : Time.now
-
-      task_plan.tasks.joins(:taskings).where(taskings: { course_membership_period_id: period.id })
-                     .update_all(opens_at: tasking_plan.opens_at,
-                                 due_at: tasking_plan.due_at,
-                                 feedback_at: feedback_at)
+          task_plan.assistant.update_tasks_for_plan(tasking_plan: tasking_plan, where: {
+                                                    taskings: { course_membership_period_id: period.id }
+                                                  })
```
jpslav: I think this got too complicated with passing `where` clauses, attributes, `reverse_merge` and so forth.  

What about just delegating to a single method on the assistant instead of delegating all updating to it?

```ruby
feedback_at: task_plan.assistant.feedback_at_for(tasking_plan: tasking_plan)
```


which returns `Time.now` for all assistants except the HW assistant.

Otherwise we are on a trajectory to have Assistants take over all work of `PropagateTaskPlanUpdates`, which is not necessarily a bad thing except for the fact that Assistants aren't routines and so don't tie into the single rolled back transaction (and also this would be a big shift).

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -42,6 +42,12 @@ class TaskPlanRepresenter < Roar::Decorator
              getter: ->(*) { is_publish_requested? },
              schema_info: { type: 'boolean' }
 
+    property :is_feedback_immediate,
+             readable: true,
+             writeable: true,
+             getter: ->(*) { is_feedback_immediate? },
```
jpslav: recommend `property :is_feedback_immediate?, as: :is_feedback_immediate`, without the `getter`, to make it similar to the code you have in `TaskRepresenter`

app/routines/distribute_tasks.rb
jpslav: @nathanstitt have you accounted for @Dantemss's comment?

app/routines/propagate_task_plan_updates.rb
nathanstitt: Yes, I see your point here regarding it getting complicated.

My thoughts were that the assistants would be best suited for determining how a task_plan should be updated, since they configured it initially.  However they probably shouldn't be firing the actual model updates since that belongs in a routine.

I think we can take your suggestion a bit further and accomplish both of our goals.  How about instead of a `feedback_at_for` method, it's a `updated_attributes_for(tasking_plan: tasking_plan)`

That way the assistant could be the "source of truth" for what should be updated on a task, but the routine is in charge of implementing the change.

jpslav: love it

app/routines/propagate_task_plan_updates.rb
```diff
@@ -5,7 +5,6 @@ class PropagateTaskPlanUpdates
   protected
 
   def exec(task_plan:)
-    task_plan.tasks.update_all(title: task_plan.title, description: task_plan.description)
```
nathanstitt: **Note**: I'm unsure if I should have removed this.  It seems like it's a duplicated update since the task is also updated in the loop below, but maybe there's a corner case I'm unaware of?  Right now I'm returning the updated `title` and `description` from the assistant's `updated_attributes_for` method and applying it there.  That saves us an update by combining them.

nathanstitt:
@jpslav ready for another review - I left a note above about a bit of code I removed that would benefit from another look see.

app/routines/propagate_task_plan_updates.rb
jpslav: I like this.

app/routines/distribute_tasks.rb
```diff
@@ -65,9 +65,10 @@ def exec(task_plan, publish_time = Time.now, protect_unopened_tasks = false)
       entity_task.taskings << tasking
 
       task = entity_task.task
+
       task.opens_at = opens_ats[ii]
       task.due_at = due_ats[ii] || (task.opens_at + 1.week)
```
jpslav: Unrelated to this PR, but @Dantemss any idea why we have an arbitrary "1 week later" default due date??

app/routines/distribute_tasks.rb
```diff
@@ -68,7 +68,7 @@ def exec(task_plan, publish_time = Time.now, protect_unopened_tasks = false)
 
       task.opens_at = opens_ats[ii]
       task.due_at = due_ats[ii] || (task.opens_at + 1.week)
-      task.feedback_at ||= task_plan.is_feedback_immediate? ? task.opens_at : task.feedback_at
+      task.feedback_at ||= task_plan.is_feedback_immediate? ? task.opens_at : task.due_at
```
jpslav: I think this line is fine, but we may want to revisit duplication here (with propagate routine)

PR #1024, opened 2016-03-18
maintenance HTML page

jpslav:
Adds a "We're in maintenace" page to tutor-server.  When assets are precompiled during a normal deploy, this page will be autogenerated in the /public directory.  After this is done once, for future deploys, ansible can copy this file plus all of /public/assets into some place where nginx can serve this file during downtime.  Need the assets folder to get the CSS and logo file.

### DIFF
--- a/app/assets/html/maintenance.html.erb
+++ b/app/assets/html/maintenance.html.erb
@@ -1,0 +1,6 @@
+<%=
+  require 'error_page_builder'
+  ErrorPageBuilder.build(view: self,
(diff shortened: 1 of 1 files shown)


jpslav:
@Dantemss / @cnuber please review

jpslav:
@cnuber if possible nginx should respond with a 503 and set "retry after" headers.  https://yoast.com/http-503-site-maintenance-seo/

PR #1025, opened 2016-03-18
Potentially made the TaskedExercise migration way faster

Dantemss:


nathanstitt:
Hm, so if it's just a matter of copying the content over from `Content::Exercises` to `Tasks::Models::TaskedExercise` then why not just do it in SQL?

```sql
update tasks_tasked_exercises 
   set content = content_exercises.content
from content_exercises 
where content_exercises.id = tasks_tasked_exercises.content_exercise_id;
```

Issue #1000, opened 2016-02-29
Limit free responses to 10,000 characters


Dantemss:
Because... Moby Dick

/cc @heathersee

philschatz:
There should maybe also be a FE PR for this (I guess it's optional though) /cc @openstax/tutor-fe

philschatz:
# :tada: :balloon: :sparkles: #1000 !!!! :sparkles: :balloon: :tada:

Dantemss:
This is necessary because exceeding Excel's limit of 32,767 characters breaks the research export (in Excel).

PR #1026, opened 2016-03-18
Limit free responses to 10,000 characters

darrenmason:
free response char limit and spec

regarding: #1000


jpslav:
@Dantemss do we need to do anything to trim existing free responses or did you already deal with Moby Dick?

jpslav:
Also @Dantemss do we need a DB constraint or is this enough?

Dantemss:
Probably don't need DB constraint. Could make a migration to truncate all free responses to 10k or just do it manually.

PR #1034, opened 2016-03-28
Fixed practice widget having not enough exercises

Dantemss:
- Send email if Biglearn returns less than the expected number of exercises
- Use local exercises if BL doesn't return requested qty
- Returns api_error if no exercises can be found by BL or locally

### DIFF
--- a/app/routines/get_ecosystem_exercises_from_biglearn.rb
+++ b/app/routines/get_ecosystem_exercises_from_biglearn.rb
@@ -70,16 +70,24 @@ class GetEcosystemExercisesFromBiglearn
       end
     end
 
+    biglearn_count = numbers.size
+
+    Rails.logger.warn {
+      "Biglearn returned less exercises than requested. " +
+      "[pools: #{pools.inspect}, role: #{role.id}, needed: #{count}, got: #{biglearn_count}]"
+    } if biglearn_count < count
+
     exercises = ecosystem.exercises_by_numbers(numbers)
     fatal_error(code: :missing_local_exercises,
                 message: "Biglearn returned more exercises than were " +
                          "present locally. [pools: #{biglearn_pools.map(&:uuid)}, " +
                          "role: #{role.id}, requested: #{count}, " +
-                         "from biglearn: #{numbers.size}, " +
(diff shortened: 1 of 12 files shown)


app/routines/get_ecosystem_exercises_from_biglearn.rb
```diff
@@ -70,16 +70,24 @@ def exec(ecosystem:, role:, pools:, count:, difficulty: 0.5, allow_repetitions:
       end
     end
 
+    biglearn_count = numbers.size
+
+    Rails.logger.warn {
```
Dantemss: Developer email could replace this.

Dantemss:
To handle 0 exercises, modify PracticesController to not blow up if ResetPracticeWidget errors out.

app/routines/get_ecosystem_exercises_from_biglearn.rb
nathanstitt: @Dantemss  do you think the email should be from:
  1) it's own mailer
  2) `OpenStax::RescueFrom`. I suspect this is the way to go, except we don't really have an exception to be rescued from.   Maybe we'd need to add a `log_error` method to `RescueFrom`
  3) `ExceptionNotifier`

Dantemss: I think this decision is up to @jpslav

jpslav: I don't think we should use `RescueFrom` because we don't want to use an exception that breaks us out of this code (and `RescueFrom` is very exception-centric).

Almost seems like we need a `WarningMailer` or something that is used by a helper method that logs and emails:

```ruby
warn_log_and_email {
  # message here gets logged to warn and gets emailed to devs
} if biglearn_count < count
```

?

nathanstitt:
I think this is now doing the right thing?

Dantemss:
Seems good. Might need a tutor-js PR to display a nice message for the 422 error (if it does not handle that already).

jpslav:
Thanks @nathanstitt and @Dantemss!

PR #1035, opened 2016-03-28
Sped up (?) the content map (performance forecast) by using less ruby and more SQL

Dantemss:
Should make the performance forecast about twice as fast for Eastman...

### DIFF
--- a/app/routines/get_student_guide.rb
+++ b/app/routines/get_student_guide.rb
@@ -25,7 +25,7 @@ class GetStudentGuide
   end
 
   def get_tasked_exercises_from_completed_exercise_task_steps(completed_exercise_task_steps)
-    tasked_exercise_ids = completed_exercise_task_steps.collect{ |ts| ts.tasked_id }
+    tasked_exercise_ids = completed_exercise_task_steps.map(&:tasked_id)
     Tasks::Models::TaskedExercise.where(id: tasked_exercise_ids).preload(
       [{task_step: {task: {taskings: {role: :profile}}}}, :exercise]
     ).to_a.group_by{ |te| te.task_step.id }
--- a/app/routines/get_student_guide.rb
+++ b/app/routines/get_student_guide.rb
@@ -12,12 +12,23 @@ class GetStudentGuide
 
   private
 
+  def get_completed_exercise_task_steps(task_steps)
+    task_steps.select{ |ts| ts.exercise? && ts.completed? }
+  end
+
   def completed_exercise_task_steps_for_role(role)
     get_completed_exercise_task_steps(
       role.taskings
         .preload(task: {task: :task_steps})
         .flat_map{ |tasking| tasking.task.task.task_steps }
     )
+  end
+
+  def get_tasked_exercises_from_completed_exercise_task_steps(completed_exercise_task_steps)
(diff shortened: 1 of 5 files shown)


Dantemss:
Oh and someone should test this really well :)

Dantemss:
I think to speed up the learning guide further, we'd have to not query all TaskedExercises. The main thing, then, is that we would have to find some other way to deduce the chapter/section of each Task. Concept Coach uses its own object and the performance forecast just orders TaskPlans by date and displays the title, so those do not necessarily apply.

Dantemss:
Oh... MAYBE this will also fix the /admin/stats/concept_coach page

app/routines/get_student_guide.rb
```diff
@@ -20,6 +24,13 @@ def completed_exercise_task_steps_for_role(role)
     )
   end
 
+  def get_tasked_exercises_from_completed_exercise_task_steps(completed_exercise_task_steps)
+    tasked_exercise_ids = completed_exercise_task_steps.collect{ |ts| ts.tasked_id }
```
jpslav: What, no `map(&:tasked_id)`?? :-)

app/routines/get_student_guide.rb
```diff
@@ -20,6 +24,13 @@ def completed_exercise_task_steps_for_role(role)
...
+    Tasks::Models::TaskedExercise.where(id: tasked_exercise_ids).preload(
+      [{task_step: {task: {taskings: {role: :profile}}}}, :exercise]
+    ).to_a.group_by{ |te| te.task_step.id }
```
jpslav: Slight optimization? `.group_by(&:task_step_id)`

app/subsystems/content/strategies/generated/map.rb
```diff
@@ -82,89 +81,71 @@ def map_exercises_to_pages(exercises:)
...
+                       exercises: {
+                         id: unmapped_exercise_ids
+                       }
                      }
                    })
-            .preload(tags: {same_value_tags: :pages})
-            .uniq
+            .select{[Content::Models::Page.arel_table[Arel.star],
+                     tags.same_value_tags.exercises.id.as(:unmapped_exercise_id)]}
+            .to_a.group_by(&:unmapped_exercise_id)
```
jpslav: should this code be combined with the `page_to_page_map` code in a helper?  Seems almost identical.

jpslav:
Looks good... I think.  this code is complex :-)

jpslav:
How about we rebase and merge this at the end of the sprint so we can pull it off if we need to? Or do you feel like it is low risk enough to merge now?

app/subsystems/content/strategies/generated/map.rb
Dantemss: It is pretty similar, but there are small differences... I suppose you could parametrize it, but I feel it would be a mess.

app/routines/get_student_guide.rb
Dantemss: Can't do this because the foreign key is in the `TaskStep` (it's polymorphic).

Dantemss:
Sure you can delay merging. I already rebased it.

app/subsystems/content/strategies/generated/map.rb
jpslav: Sounds good.

PR #1040, opened 2016-03-31
Reading processing instructions

Dantemss:
- Added reading feature map fields to Content::Models::Book
- Added reading feature map to Ecosystem manifests
- Fragment splitter now uses the feature map
- Fixed HS iReading specs
- When no embed tag found, search for exercises by context tag
- Exercises with multiple LO's and/or CNXMOD tags in multiple pages now are imported/map into the highest numbered page (!!!)
- CNXMOD tags can now also be used to map exercises to pages across multiple book versions
- Added manifest format-related error messages
- Specs


app/subsystems/content/models/page.rb
```diff
@@ -45,16 +45,32 @@ def ccs
...
+        fragments: fragment_splitter.split_into_fragments(snap_lab_node, 'snap-lab')
+      }
     end
   end
 
   protected
 
   def parser
-    OpenStax::Cnx::V1::Page.new(title: title, content: content)
+    @parser ||= OpenStax::Cnx::V1::Page.new(title: title, content: content)
```
jpslav: good catch

app/subsystems/tasks/assistants/extra_assignment_assistant.rb
```diff
@@ -24,45 +26,39 @@ def initialize(task_plan:, taskees:)
...
-      pages: @ecosystem.pages_by_ids(page_ids),
+      pages: ecosystem.pages_by_ids(page_ids),
       page_id_to_snap_lab_id: page_to_snap_lab
     }
   end
 
   def build_extra_task(pages:, page_id_to_snap_lab_id:)
     task = build_task
 
+    reset_used_exercises
```
jpslav: I'm a little worried that some day down the line we're going to forget to call this reset method in some later assistant... but maybe I'm worried over nothing.

app/subsystems/tasks/assistants/i_reading_assistant.rb
```diff
@@ -65,60 +59,18 @@ def build_ireading_task(pages:, taskee:)
...
       task_type: :reading,
       title:     title,
       description: description,
       feedback_at: Time.now
     ]
-    AddSpyInfo[to: task, from: @ecosystem]
-    return task
-  end
-
-  def task_fragments(task:, fragments:, fragment_title:, page:, related_content:)
```
jpslav: super cool to pull this code up into `FragmentAssistant`!

lib/openstax/cnx/v1/fragment_splitter.rb
```diff
@@ -1,98 +1,116 @@
...
+    def split_into_fragments(root, type = nil)
+      result = [root.dup]
+      type_string = type.to_s
+
+      processing_instructions.each do |processing_instruction|
+        next if processing_instruction.css.blank? ||
+                processing_instruction.fragments.nil? ||
+                (
+                  (!processing_instruction.only.nil? &&
+                   ![processing_instruction.only].flatten.map(&:to_s).include?(type_string)) ||
```
jpslav: `flat_map`?

lib/openstax/cnx/v1/fragment_splitter.rb
```diff
@@ -1,98 +1,116 @@
...
+                  (!processing_instruction.except.nil? &&
+                   [processing_instruction.except].flatten.map(&:to_s).include?(type_string))
+                )
```
jpslav: Why skip the instruction if the `only` matches the type string?  (I understand skipping if `except` matches)

Also I think the nesting can be simplified here.  This has `A || B || (C || D)` which can be `A || B || C || D`.

spec/factories/content/books.rb
```diff
@@ -9,6 +9,22 @@
     content { contents.to_json }
     uuid { SecureRandom.uuid }
     version { Random.rand(1..10) }
+    reading_processing_instructions {
+      [
+        { css: '.ost-reading-discard, .os-teacher, [data-type="glossary"]',
+          fragments: [], except: 'snap-lab' },
```
jpslav: What does "excepting" `snap-lab` do here since `fragments` is `[]`?

jpslav:
Looking good!

lib/openstax/cnx/v1/fragment_splitter.rb
Dantemss: There's a `!` on that line :D and ok I'll simplify the nesting.

lib/openstax/cnx/v1/fragment_splitter.rb
Dantemss: `flat_map` is equivalent to `flatten(1).map` but I really want to `flatten` everything here so I don't think it applies...

spec/factories/content/books.rb
Dantemss: Each instruction removes the matching nodes unless `"node"` is in the `fragments` list. So `fragments: []` is the discard instruction. Skipping it for `snap-lab` prevents the discard instruction from applying to snap-labs.

Dantemss:
I decided to split the try another after reading step stuff into a separate PR so this one can be finished.

Dantemss:
@jpslav feel free to review this whenever. It's only missing specs.

Dantemss:
OK I think I am happy with the number of tests for this now. Merge away when ready.

PR #1046, opened 2016-04-12
Replaced collect with map

Dantemss:
By popular request


app/controllers/api/v1/courses_controller.rb
```diff
@@ -82,7 +82,7 @@ def tasks
     # back an empty list of tasks
     course = Entity::Course.find(params[:id])
     tasks = GetCourseUserTasks[course: course, user: current_human_user]
-    output = Hashie::Mash.new('items' => tasks.collect{|t| t.task})
+    output = Hashie::Mash.new('items' => tasks.collect(&:task))
```
jpslav: missed this one

PR #1047, opened 2016-04-12
Multi part exercise support

jpslav:
## Summary of Key Changes

* When a multipart exercise is assigned, its parts each become separate task steps presented in order in the task (see exception in "TODO" below).
* Each exercise step JSON now has new `is_in_multipart` and `question_id` fields.  See below for documentation.
* The frontend will need to see when steps in sequence have the `is_in_multipart` field set to true to collapse these steps visually into one display.  Alternatively, the FE can find the steps that all share the same URL to recognize that those are parts of a multipart.
* The `question_id` likely doesn't need to be used right now by the FE.  Where it could be used is to report errors on individual exercise parts.  We got robert to make the errata reporting global for one multipart, but with the `question_id` we could conceivably go back to reporting errors on each part (I don't know much about the errata mechanism tho).
* The task review endpoint now has its `exercises` field content broken up so that we can report on multiple questions per exercise.

## New Fields in Step JSON

```ruby
    property :is_in_multipart,
             writeable: false,
             readable: true,
             schema_info: {
               required: true,
               description: "If true, indicates this object is part of a multipart"
             }

    property :question_id,
             type: String,
             writeable: false,
             readable: true,
             schema_info: {
               required: true,
               description: "The ID of the part, present even if there is only one part."
             }
```

The Task review endpoint now has this structure for the `exercises` field:

```ruby
{
  ...
  exercises: [
    {
      content: # the exercise JSON *without* selected answer counts,
      question_stats: [ # new, previously assumed only one question
        {
          question_id: "2932",
          answered_count: 13,
          answer_stats: [ # new, replaces selected counts in `content`
            {
              answer_id: "09022",
              selected_count: 6
            },
            {
              answer_id: "21112",
              selected_count: 7,
            },
            {
              answer_id: "97733",
              selected_count: 0
            }
          ],
          answers: [
            # unchanged from current implementation
          ]
        },
        # more hashes for other question parts
      ],
      average_step_number: 5.2
    }
  ],
  ... 
}
```
## Misc

* Updates `rspec` so that we can use `rspec --only-failures` to just rerun what failed on the last run.

## Questions

1. Is our only way to get test data for the FE to get some multiparts actually on to Exercises?

## TODO

* [x] I'm pretty sure steps for a multipart will only be added sequentially when building the task anew, i.e. if we were inserting a multipart into a task (like to replace a personalized step placeholder) the first part would replace the placeholder and the remaining parts would get tacked on to the end.
* [ ] This PR doesn't yet handle changes to Exchange and BigLearn, but it should be enough for the FE to work with.
* [x] Change `CalculateTaskPlanStats` to appropriately count multipart exercises.  
* [x] Make sure that the scores report counts each part as one exercise (no real reason it shouldn't).
* [x] Migrate existing TEs to populate their new `question_id` field.
* [ ] More specs


jpslav:
@Dantemss if you need a break and want to do an intermediate code review, that'd be welcome.

app/routines/task_exercise.rb
```diff
@@ -1,18 +1,40 @@
...
+    task ||= task_step.try(:task)
+    fatal_error(code: :cannot_get_task) if task.nil?
+
+    parts = exercise.content_as_independent_parts
+
+    outputs[:task_steps] = parts.each_with_index.map do |part, ii|
+
+      # Make sure that all steps after the first exercise part get their own new step
+      task_step = nil if ii > 0
+      task_step ||= Tasks::Models::TaskStep.new
```
Dantemss: Could just make `task` mandatory and remove the `task_step` argument. You can safely ignore the unsaved steps in the FragmentAssistant.

Dantemss: Oh I supppose you'd have to skip that last bit of the assistant where the step is added... hmmm....

Dantemss: Actually looking at the changes you did to the assistant, I think it should be possible to do this.

Dantemss:
Looks good to me, but wait until FE is ready to merge.

Dantemss:
Unless you are confident they can handle the base case (1 part) right now :)

app/routines/task_exercise.rb
jpslav: The code that replaces placeholders needs to pass in a specific step.

app/routines/calculate_task_stats.rb
```diff
@@ -6,36 +6,64 @@ class CalculateTaskStats
...
-  def answer_stats_for_tasked_exercises(tasked_exercises)
-    tasked_exercises.first.answer_ids.each_with_object({}) do |answer_id, hash|
-      hash[answer_id] = tasked_exercises.select{ |te| te.answer_id == answer_id && \
-                                                      te.completed? }.count
+  def assignee_names_for(task)
+    (@names ||= {})[task.id] ||= ->() {
+      roles = task.taskings.map(&:role)
+      users = run(:get_users_for_roles, roles).outputs.users
+      users.map(&:name)
+    }.call
```
Dantemss: Don't think this lambda is doing anything... Remove it?

app/subsystems/content/models/exercise.rb
```diff
@@ -43,7 +43,24 @@ def cnxmods
...
+
+  def content_as_independent_questions
+    @content_as_independent_questions ||= ->() {
+      exercise_json = parsed_json
+      questions = exercise_json['questions']
+      questions.map do |question|
+        exercise_json['questions'] = [question]
+        {id: question['id'], content: exercise_json.to_json}
+      end
+    }.call
```
Dantemss: Same here.

Dantemss: Hm and maybe we should keep these JSON parsing things in the same object (I was keeping them in the `OpenStax::Exercises::V1::Exercise`

spec/routines/task_exercise_spec.rb
```diff
@@ -6,16 +6,71 @@
...
+    expect(task.task_steps[2].tasked.question_id).to eq question_ids[1]
+    expect(task.task_steps[3]).to eq exercise_step
+
+    expect(task.task_steps[1].tasked.is_in_multipart).to be_truthy
+    expect(task.task_steps[2].tasked.is_in_multipart).to be_truthy
+    expect(task.task_steps[3].tasked.is_in_multipart).to be_falsy
+
+    expect(task.task_steps[1].tasked.content).to match("(0)")
+    expect(task.task_steps[2].tasked.content).to match("(1)")
+  end
```
jpslav: nice - thanks!

jpslav:
Ready for simultaneous merge with FE

app/subsystems/content/routines/import_exercises.rb
```diff
@@ -24,7 +24,7 @@ def exec(ecosystem:, page:, query_hash:)
 
     exercise_pages = wrappers.map do |wrapper|
       # Don't import multipart questions for now
-      next if wrapper.is_multipart?
+      # next if wrapper.is_multipart?
```
nathanstitt: I think this can be removed altogether 😁

nathanstitt: This can be removed altogether 😁

PR #1051, opened 2016-04-16
Fix try anothers

Dantemss:
This PR will fix try another issues. However, it no longer (tries to) catch the following tagging errors:

`ost-feature, ost-assessed-feature with 2 or more exercises` -> `1 exercise + try another`

The following will still work:

`ost-feature, ost-assessed-feature with ost-exercise-choice` -> `1 exercise + try another`

I briefly attempted to use nokogiri's xpath on processing instructions instead of CSS to handle the above cases, however that failed because nokogiri's xpath methods behaves differently from the css methods.

In particular, `root.at_css()` gives you a descendant of that `root` node, but `root.at_xpath()` gives you a descendant of a copy of the `root` node.


Dantemss:
Hm don't merge yet I'll try something else.

Dantemss:
Should be good now. I can manually rerun the migration on dev/qa/staging after.

philschatz:
@openstax/test: @Dantemss said this is the Try Another hotfix 😄

jpslav:
@Dantemss let me know if this should be merged or if it is to be a cherry picked hotfix.

Dantemss:
@jpslav I'm gonna go ahead and merge

PR #1057, opened 2016-04-20
Show timezone in course representer and allow timezone to be updated

karenc:


jpslav:
@openstax/tutor-fe heads up -- the below time zones values are what the BE will expect for our neck of the woods:

```
Hawaii
Alaska
Pacific Time (US & Canada)
Arizona
Mountain Time (US & Canada)
Central Time (US & Canada)
Eastern Time (US & Canada)
Indiana (East)
Atlantic Time (Canada)
```

If you want a list of all time zones that will work in the BE, you can run:

```ruby
ActiveSupport::TimeZone.all.map(&:name)
```

philschatz:
Oh, are those the standard timezone strings (that also happen to be english strings)?

jpslav:
Those are the strings that Rails' time zone infrastructure recognizes.  FE can do whatever it wants as long as it sends and understands those.  Or we can use something else and do a conversion in the BE.

PR #1070, opened 2016-04-27
Disallow repeats in practice when drawing locally

jpslav:
When BigLearn doesn't supply enough exercises, the practice widget should no longer be filled with repeats of locally-drawn exercises.


jpslav:
Should maybe be hotfix merge

Dantemss:
`allow_repeats` was supposed to be exercises you have done, but not present in the current task (those are excluded via the `additional_excluded_numbers` and `ChooseExercises` itself should never pick the same exercise twice)

so I'm kind of worried if Greg found a bug in one of those routines?

PR #1071, opened 2016-04-27
add account urls to bootstrap data representer

nathanstitt:


jpslav:
Looking good @nathanstitt

PR #1072, opened 2016-04-30
Reduce massive CC course export time by 98%

jpslav:
Turns out that the vast majority of the time spent exporting a spreadsheet was in the method in our Excel library that saves the file.  In there there is some very inefficient code having to do with recording the authors of comments.  Since our spreadsheets have lots of comments, this was killing our performance.  

This monkey patch of that code takes the time it takes to export the massive tutor-staging CC course from 317 seconds to 5 seconds. 

:-)


Dantemss:
I'll go ahead and merge this. We can remove the monkeypatch once axlsx fixes the issue.

PR #1073, opened 2016-04-30
Get rid of N+1 queries on CC performance report

Dantemss:
Saves 5-10 seconds loading the scores page and the export for large courses.
Tested in the console on staging.


jpslav:
👍

PR #1074, opened 2016-05-02
Short codes

karenc:


app/controllers/short_codes_controller.rb
```diff
@@ -0,0 +1,9 @@
+class ShortCodesController < ApplicationController
+  skip_before_filter :authenticate_user!
+
+  def redirect
+    handle_with(ShortCode::ShortCodeRedirect,
+                success: -> (*) { redirect_to @handler_result.outputs.uri },
+                failure: -> (*) { render text: 'Short code not found', status: 404 })
```
jpslav: Our current 404 approach is to raise a custom exception and then teach `OpenStax::RescueFrom` how to handle it -- it will then display the appropriate 404 page.  

https://github.com/openstax/tutor-server/blob/master/config/initializers/rescue_from.rb#L28-L34

So somewhere you can add a `ShortCodeNotFound` custom exception (`class ShortCodeNotFound < StandardError; end`) and then just `raise ShortCodeNotFound` in this `failure` block.

That should display our pretty looking 404 page (wouldn't hurt to check that I'm not going crazy on this and it actually does do that :-)

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -78,6 +78,11 @@ class TaskPlanRepresenter < Roar::Decorator
              readable: true,
              writeable: true
 
+    property :shareable_url,
+             type: String,
+             readable: true,
+             writeable: false
```
jpslav: Instead of storing the shareable url in the TaskPlan record, I'd rather just compute it dynamically (all the information is already otherwise stored in the DB)

app/subsystems/short_code/create.rb
```diff
@@ -0,0 +1,10 @@
+class ShortCode::Create
+  lev_routine express_output: :short_code
+
+  protected
+
+  def exec(uri)
+    outputs.short_code = ShortCode::Models::ShortCode.create(uri: uri)
```
jpslav: If we're not storing the shareable_url in the TaskPlan, there's no need to return this model out of this routine.  This is good in that it keeps the implementation guts hidden from the rest of Tutor.

app/subsystems/short_code/models/short_code.rb
```diff
@@ -0,0 +1,18 @@
+class ShortCode::Models::ShortCode < Tutor::SubSystems::BaseModel
+  validates :uri, presence: true
+
+  before_save :generate_code
+
+  protected
+
+  def generate_code
+    if self.code.nil?
+      alphanumerics = (0..9).to_a + ('a'..'z').to_a + ('A'..'Z').to_a
```
jpslav: minor, but I think this can be a constant in the class, `ALPHANUMERICS = blah` so it doesn't have to be expanded each time.

config/routes.rb
```diff
@@ -272,5 +272,8 @@
     end
   end
 
+  get '/@:short_code' => 'short_codes#redirect'
+  get '/@:short_code/:optional' => 'short_codes#redirect'
```
jpslav: Should be able to combine these two lines into:

```ruby
get '/@:short_code(/:optional)' => 'short_codes#redirect'
```

And maybe change `optional` to `human_readable` to give it some context for future readers.

See http://guides.rubyonrails.org/routing.html#bound-parameters

app/subsystems/short_code/models/short_code.rb
```diff
@@ -0,0 +1,18 @@
...
+      self.code = (0..5).map { |i| alphanumerics.sample }.join
+      while ShortCode::Models::ShortCode.find_by_code(self.code)
+        self.code = (0..5).map { |i| alphanumerics.sample }.join
```
jpslav: you can leave out the `|i|`

app/representers/api/v1/task_plan_representer.rb
jpslav: To dynamically generate the url, add a `getter: ->(*) { blah }` argument.  Append the title of the task plan as the "optional" part of the URL, properly escaped/formatted for inclusion in the URL (e.g. spaces converted to hyphens, strip out everything except hyphens, periods, and alphanumerics).  Will result in something like `https://.../@239AZ/Read-Ch-4.1-4.3`

app/subsystems/short_code/models/short_code.rb
```diff
@@ -0,0 +1,18 @@
+class ShortCode::Models::ShortCode < Tutor::SubSystems::BaseModel
+  validates :uri, presence: true
```
jpslav: also need to `validates :code, presence: true` -- to make this check happy, the `before_save` below needs to change to `before_validation`

app/subsystems/tasks/models/task_plan.rb
```diff
@@ -29,6 +29,8 @@ class Tasks::Models::TaskPlan < Tutor::SubSystems::BaseModel
 
   before_destroy :not_open_before_destroy, prepend: true
 
+  after_save :populate_shareable_url
```
jpslav: Per the above, if we are not storing the shareable url in the TaskPlan (but rather computing it dynamically), this would become a `before_create` call that just called `ShortCode::Create[self.to_global_id.to_s]`.

But, because we (or at least I :-) like to keep outside knowledge out of our model classes, I don't want TaskPlan to know about (aka have a dependency on) the ShortCode subsystem.  Ideally there'd be a `CreateTaskPlan` routine somewhere where we could add this call to populate the short code.  Unfortunately at the moment, what would be in such a routine is just written inline in the `Api::V1::TaskPlansController`.  So go with that flow and put the `ShortCode::Create` call right before this line: https://github.com/openstax/tutor-server/blob/master/app/controllers/api/v1/task_plans_controller.rb#L85

db/migrate/20160427192800_create_short_code_short_codes.rb
```diff
@@ -0,0 +1,10 @@
+class CreateShortCodeShortCodes < ActiveRecord::Migration
+  def change
+    create_table :short_code_short_codes do |t|
+      t.string :code, null: false, unique: true
```
jpslav: I don't _think_ that `unique: true` is used on this line.  I thought uniqueness constraints were only enforced on the index line below (could be wrong tho, just haven't seen this usage).

db/migrate/20160502155735_add_shareable_url_to_tasks_task_plans.rb
```diff
@@ -0,0 +1,5 @@
+class AddShareableUrlToTasksTaskPlans < ActiveRecord::Migration
+  def change
+    add_column :tasks_task_plans, :shareable_url, :string
```
jpslav: Per above, remove this migration.

jpslav:
Review complete.

app/controllers/short_codes_controller.rb
karenc: It is not showing the 404 page in development :/

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -78,6 +78,13 @@ class TaskPlanRepresenter < Roar::Decorator
              readable: true,
              writeable: true
 
+    property :shareable_url,
+             type: String,
+             readable: true,
+             writeable: false,
+             if: ->(*) { ShortCode::FindShortCode[self.to_global_id.to_s] },
+             getter: ->(*) { "/@#{ShortCode::FindShortCode[self.to_global_id.to_s]}" }
```
jpslav: Cool, Karen -- can you add the end part mentioned in https://github.com/openstax/tutor-server/pull/1074#discussion_r61902900?

jpslav:
Looks great @karenc!  Just one last comment and then it is good to merge.

app/representers/api/v1/task_plan_representer.rb
karenc: It's already been added if you're talking about changing `before_save` to `before_validation`

jpslav: Talking about appending a sanitized version of the title to the end of the `shareable_url`

PR #1067, opened 2016-04-26
Tasking plan open / due times

karenc:

PR #1077, opened 2016-05-05
Open / due times and time zones (alt)

jpslav:
Builds on KC's open / due times PR, mostly just moves logic around.

## Relevant API Info

* Course JSON continues to have a `timezone` field that can be set.  
* Course and Period JSON have `default_open_time` and `default_due_time` that can be read and set.  These fields will always return something (there is a global default for these times for all of Tutor if not set per course or period).  The format of these times is `'HH:MM'` in a 24 hour clock.  When setting, you patch the course or period info.  Bad formats will result in a `422`.

## Key Changes from #1067:

* default times are now strings (we weren't using them as time objects before)
* side effects of changing course time zone have moved out of the course profile model into routines.  
* specifying time zone relies on `use_zone` call in the `TaskPlansController` instead of having the `TaskingPlanRepresenter` look up the relevant time zone.

## TODO

* [x] Standardize `timezone` to `time_zone` (made this a separate issue for later)
* [x] Make sure default times are available on admin settings UI
* [x] Spec for `Tasks::CourseTimeZoneChanged`
* [x] Validate the default times are in the range `00:00` to `23:59`
* [x] Make sure bad defaults give non 200 HTTP status
* [x] Check for all `Time.now` usage to make sure we are doing things in a time-zone aware way


lib/date_time_utilities.rb
```diff
@@ -4,4 +4,27 @@ module DateTimeUtilities
...
   end
+
+  def self.from_string(datetime_string:, time_zone: nil, ignore_existing_zone: true)
+    time_zone ||= Time.zone
+
+    raise(IllegalArgument, "time_zone must not be nil") if time_zone.nil?
+
+    if ignore_existing_zone
+      datetime_string.gsub!(/(\d)T(\d)/,'\1 \2') # handle the 'T' in w3c format
+      datetime_string.gsub!(/[-+]\d\d\d\d/,'')  # remove time zone offsets
```
karenc: I think timezones can be defined like "-07" as well as "-0700".

karenc:
I would suggest you remove the stuff I added (squash all my commits into your first one).

It doesn't really make sense to create some database columns, remove them, add them with a different type.  Same with code.

jpslav:
Actually we can't just remove them because they are already merged to
master in an earlier PR, so some environments may have already had them
deployed (hence the need for the migration)

lib/date_time_utilities.rb
jpslav: Thanks @karenc, I added this case as well as "-07:00" which popped up in some testing.

app/subsystems/tasks/course_time_zone_changed.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    old_time_zone = ActiveSupport::TimeZone[old_time_zone_name]
+    new_time_zone = ActiveSupport::TimeZone[new_time_zone_name]
+
+    # Update task plans
+
+    Tasks::Models::TaskingPlan.joins{task_plan}.where{task_plan.owner == course}.find_each do |tp|
+      tp.opens_at = change_time_zone(tp.opens_at, old_time_zone, new_time_zone)
+      tp.due_at = change_time_zone(tp.due_at, old_time_zone, new_time_zone)
+      tp.save
```
Dantemss: Huh? Is this even needed? I think in Exercises all times are stored in the DB in UTC.

app/subsystems/tasks/course_time_zone_changed.rb
```diff
@@ -0,0 +1,45 @@
...
+    # Update existing tasks in this course
+
+    student_roles = run(:get_course_roles, course: course, types: :student).outputs.roles
+    entity_tasks = run(:get_tasks, roles: student_roles).outputs.tasks
+    tasks = Tasks::Models::Task.where{entity_task_id.in entity_tasks.map(&:id)}
+
+    tasks.each do |task|
+      task.opens_at = change_time_zone(task.opens_at, old_time_zone, new_time_zone)
+      task.due_at = change_time_zone(task.due_at, old_time_zone, new_time_zone)
+      task.save
```
Dantemss: Same here.

config/initializers/settings.rb
```diff
@@ -6,10 +6,10 @@ class Settings::Db::Store < RailsSettings::CachedSettings
...
 Settings::Db.store.defaults[:excluded_pool_uuid] = ''
 Settings::Db.store.defaults[:import_real_salesforce_courses] = false
-Settings::Db.store.defaults[:course_default_open_time] = '00:00'
-Settings::Db.store.defaults[:course_default_due_time] = '00:00'
-Settings::Db.store.defaults[:period_default_open_time] = '00:00'
-Settings::Db.store.defaults[:period_default_due_time] = '00:00'
+Settings::Db.store.defaults[:course_default_open_time] = '00:01'
+Settings::Db.store.defaults[:course_default_due_time] = '07:00'
+Settings::Db.store.defaults[:period_default_open_time] = '00:01'
+Settings::Db.store.defaults[:period_default_due_time] = '07:00'
```
Dantemss: Why not just have 1 default open and 1 default due time?

lib/date_time_utilities.rb
```diff
@@ -4,4 +4,36 @@ module DateTimeUtilities
...
+    end
+  end
+
+  def self.keep_time_change_zone(time, old_timezone, new_timezone)
+    return nil if time.nil?
+    # Often, the time will already be in the specified old_timezone, but if not
+    # get it there.
+    old_time = time.in_time_zone(old_timezone)
+    new_time = time.in_time_zone(new_timezone)
+    old_time.to_datetime.change(offset: new_time.formatted_offset)
```
Dantemss: What is this doing? Unlike what the name says, it seems to be changing the time... Are we trying too hard to undo timezone mistakes?

app/subsystems/tasks/course_time_zone_changed.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  uses_routine CourseMembership::GetCourseRoles, as: :get_course_roles
+  uses_routine Tasks::GetTasks, as: :get_tasks
+
+  protected
+
+  def exec(course:, old_time_zone_name:, new_time_zone_name:)
+    return if old_time_zone_name == new_time_zone_name
+
+    # TODO fail if plans currently publishing?  Could maybe do in the loop
```
Dantemss: If you are concerned about this, I suggest you lock the course's `TaskPlans` (not `TaskingPlans`) using something like `task_plans = TaskPlan.lock.where(owner: course)` then at the end you MUST update something in the `task_plans` to make the lock actually do anything (or even `task_plans.each(&:touch)`). But maybe you just wanna update the open/due dates there too?

Dantemss: Oh my bad, TaskPlans have no dates... touch should be fine.

config/initializers/settings.rb
jpslav: Great point.

app/subsystems/tasks/course_time_zone_changed.rb
jpslav: is it a problem if no task plans are returned by the query (and so there is nothing to update)

lib/date_time_utilities.rb
```diff
@@ -4,4 +4,37 @@ module DateTimeUtilities
...
+      datetime_string.gsub!(/[a-zA-Z]/,'')             # remove any other PST or similar
+      datetime_string.strip!
+    else
+      raise NotYetImplemented
+    end
+
+    original_time_class = Chronic.time_class
+    begin
+      Chronic.time_class = time_zone
+      Chronic.parse(datetime_string)
```
Dantemss: Add `ambiguous_time_range: :none`

Dantemss:
I started converting this to store the open/due dates in a timezone-independent way. Still need to make old specs pass and convert specs I temporarily removed to work with the new way.

jpslav:
@Dantemss maybe the `TimeZone` model should just be in `/app/models` since it is not really dependent on any subsystem?  I don't think it is `course_profile` specific.

Dantemss:
OK all that is left now is to salvage those specs and maybe make some more.

app/controllers/api/v1/periods_controller.rb
```diff
@@ -9,57 +9,37 @@ class Api::V1::PeriodsController < Api::V1::ApiController
...
-      render_api_errors(result.errors)
-    else
-      respond_with result.outputs.period,
-                   represent_with: Api::V1::PeriodRepresenter,
-                   location: nil
+    period_model = CourseMembership::Models::Period.new
+    CourseMembership::Models::Period.transaction do
+      standard_nested_create(period_model, :course, @course, Api::V1::PeriodRepresenter)
+      period = CourseMembership::Period.new(strategy: period_model.wrap)
+      Tasks::AssignCoursewideTaskPlansToNewPeriod[period: period]
```
jpslav: I'd prefer not to pull the calls out of the `CreatePeriod` routine here.  This duplicates the code and also has the potential to miss things like the `name ||= (course.periods.count + 1).ordinalize` line in `CreatePeriod`.  Trying to think of an alternative that lets us just call the create routine and use the representer...

jpslav:
@Dantemss, I (mostly) found the source of the failing specs in https://github.com/openstax/tutor-server/pull/1077/commits/6ddb2cd2461ec9847172d86557cec844b58332c6.  When a profile was being created and its blank `time_zone` association populated from the default value in `belongs_to_time_zone`, the profile was being saved twice.  The first time had all of the values passed to `...Profile.create(blah)`.  The second save was after the `TimeZone` record was persisted and there were no changes to the profile -- this resulted in the dirty state of the profile being erased, which broke the code that was failing its spec.

I found some mention of similar rails problems having to do with autosave and duplicate saves of parent records.  Will send you the link separately so GH doesn't link this issue into the Rails issue :-)  I don't know tho if that issue is exactly the same situation here tho.

Here's the commit with my fix: https://github.com/openstax/tutor-server/pull/1077/commits/4db771bb30f14139ba180d7bc8c38f1b2c6e3c25.  I pushed the initialization of the default into a `before_save` call and did some finessing with the presence and uniqueness validations.  Take a look and let me know what you think.  I wish there were a cleaner solution.

Dantemss:
Huh thanks for debugging. I'll play with it a bit and see if I manage to fix it while keeping the validation.

Dantemss:
I think the error was being caused by the method I was defining colliding with the built-in rails `build` method. It seems using `before_validation` instead of `before_save` just works. I removed the validation stuff from the `belongs_to_time_zone` method because it seems very confusing to me (are you validating the `time_zone` or the fields?)

jpslav:
Ready to merge simultaneously with https://github.com/openstax/tutor-js/pull/1058

PR #1079, opened 2016-05-06
Build task plan sharable code without model

nathanstitt:
The dashboard JSON doesn't include the `sharable_code`, which is of course where the FE needs to read it from.  Traced it back to this, the representer was actually operating on lev outputs, and global id was failing to build the global_id.


jpslav:
@nathanstitt -- can you elaborate on "global id was failing to build the global_id"?

Also note I found some other issues: https://www.pivotaltracker.com/story/show/119087723

app/subsystems/short_code/find_task_plan_short_code.rb
```diff
@@ -0,0 +1,16 @@
...
+  protected
+
+  def exec(task_plan_id)
+    gid = GlobalID::URI::GID.build(
+      app: GlobalID.app, model_name: 'Tasks::Models::TaskPlan', model_id: task_plan_id
+    ).to_s
+    run(:find_short_code, gid)
+  end
+
+end
```
jpslav: I'd like to not have the short code SS know anything specifically about TaskPlans, but because GIDs are a nice abstraction, I would be fine with modifying the existing `ShortCode::FindShortCode` routine to be aware of things that are GID'able, e.g.:

```ruby
class ShortCode::FindShortCode
  lev_routine express_output: :short_code

  protected

  def exec(uri_or_model:)
    uri = case uri_or_model
    when String
      uri_or_model
    else
      uri_or_model.to_global_id.try(:to_s)
    end

    fatal_error(:could_not_get_uri) if uri.nil?

    outputs.short_code = ShortCode::Models::ShortCode.find_by_uri(uri).try(:code)
  end
end
```

And if you're feeling ambitious, you could tackle the other issue about lazily-instantiating the short code by changing

```ruby
outputs.short_code = ShortCode::Models::ShortCode.find_by_uri(uri).try(:code)
```

to 

```ruby
short_code = create_if_missing ? 
  ShortCode::Models::ShortCode.find_or_create_by_uri(uri) : 
  ShortCode::Models::ShortCode.find_by_uri(uri)

outputs.short_code = short_code.try(:code)
```

where `create_if_missing` would be a new method argument with a default value of `false`.  This will also simplify further the `getter` logic in the representer, as you can set `create_if_missing` to `true` so you always get one back.

jpslav: This lazy instantiation would also let us remove the explicit call to create the short code in the TaskPlan controller.  We'd just always lazily make them the first time the JSON needs it.

nathanstitt:
The root issue is that the `global_id` gem builds the id based on the model type. 

Which works fine as long as the representer is operating on an actual instance of `Tasks::Models::TaskPlan`, but fails if the instance is `Lev::Outputs`.  In that case calling `to_global_id` returns nil because it's intercepted by the output's method_missing. 

Even if it wasn't intercepted, it would be something like `gid://Lev::Outputs/1` instead of the correct  `gid://tutor/Tasks::Models::TaskPlan/1`

The short_code generation needs to be told the correct class type.  I considered doing it directly in the representer, but decided to move it to a dedicated routine that would know the correct class type for a task plan model.

Your idea re lazy instantiation is good, but I think the above issue is still relevant.  ShortCodes is going to have to be told the type it should operate on.

Dantemss:
Personally I don't like the fact that we have those routines that return hashes for display (I'd prefer to use representers for that) but oh well...

app/representers/api/v1/task_plan_representer.rb
```diff
@@ -82,8 +82,9 @@ class TaskPlanRepresenter < Roar::Decorator
              type: String,
              readable: true,
              writeable: false,
-             if: ->(*) { ShortCode::FindShortCode[self.to_global_id.to_s] },
-             getter: ->(*) { "/@#{ShortCode::FindShortCode[self.to_global_id.to_s]}/#{self.title.parameterize}" }
+             getter: ->(*) {
+               self.try(:sharable_url) || ShortCode::UrlFor[self]
```
jpslav: all looks good, except `:sharable_url` here is missing an `e`, and can `UrlFor` take an optional `suffix` too instead of it having to know to try to call `model.title`?

nathanstitt: Yep, have done so - good eye spotting the missing in action `e`.  I've added a spec that tests the 'provided shareable_url' case  so it'll catch stuffs like that in the future.

app/routines/get_non_cc_dashboard.rb
```diff
@@ -33,7 +34,8 @@ def load_plans(course)
         publish_last_requested_at: task_plan.publish_last_requested_at,
         publish_job_uuid: task_plan.publish_job_uuid,
         tasking_plans: task_plan.tasking_plans,
-        is_trouble: out[:trouble_plan_ids].include?(task_plan.id)
+        is_trouble: out[:trouble_plan_ids].include?(task_plan.id),
+        shareable_url: run(:get_short_code_url, task_plan).outputs.url
```
jpslav: Need to pass in the suffix here too.

nathanstitt: Well, not really since `suffix` defaults to `model.try(:title)`.  But probably best to be explicit and avoid the `.try` call

PR #1088, opened 2016-05-12
Record explicit late work accepted counts

jpslav:
Previously, when a teacher accepted late work for an assignment, the BE implementation effectively accepted the current late work done by the student and any future late work done by that student.  Assume Johnny had correctly worked 6 of 10 problems at the due date and a week later answered 1 more problem correctly.  The teacher at that point accepts his late work which moves him from a 60% score to a 70% score.  If he later works the rest of the problems he could increase his grade beyond 70%, unbeknownst to the teacher.

This PR adds new `..._accepted_late_...` counts to the performance report JSON:

* completed_accepted_late_step_count
* completed_accepted_late_exercise_count
* correct_accepted_late_exercise_count

These new fields will record the number of correct exercises, completed exercises, and completed steps that were accepted late.  These numbers are normally 0 unless the teacher has accepted late work.  If the teacher accepts late work, they go up.  If the teacher rejects late work, they are reset to 0.  In Johnny's case when the teacher accepted his late work (1 new problem worked correctly), each of these fields would equal 1.

This will of course lead to some changes in the FE code.  For starters the number of correct exercises will no longer toggle between `task.correct_exercise_count` and `task.correct_on_time_exercise_count`.  It'll just always be:

`task.correct_on_time_exercise_count + task.correct_accepted_late_exercise_count`

These changes will also lead to a new state in the FE, namely the one that Johnny is in above.  The teacher has accepted late work, but then Johnny went and did *more* late work.  (To detect that there is unaccepted late work, you'd check to see if the number of completed steps is greater than the sum of the steps completed on time and the steps accepted late.)  In this scenario, late work has been accepted but the flag marker on the cell will need to be orange again to indicate there is unaccepted late work.  I'm pretty sure the UX decision was that we don't need a special "accept more late work" button on the popup, but rather that the teacher could reject late work and then accept it right back, which will have the effect of accepting all of the current late work.

However, now that I think about it, there will probably have to be some UX changes so that teachers can see that they accepted late work but that there is more to accept, otherwise they will never know.


rvnewell:
This should not create a new UI state for the user.  The accepted late scoreis now the current on-time score (to the user).  All behavior and text would remain the same.  There would be a new state in the export:  The first accepted late score is in the on-time score column.  The as-of-yet unaccepted late score is in the Pending Late work column, and a comment appears over the on-time score that displays the on-time score.

rvnewell:
*displays the "original" on-time score

PR #1090, opened 2016-05-16
Tutor and CC scores export updates

jpslav:
Requires #1088 

Tutor and CC scores exports.  

I'm just going to say that building a file in Excel could really use the cascading part of CSS :-)

[tutor_example_export.xlsx](https://github.com/openstax/tutor-server/files/269685/tutor_example_export.xlsx)
[cc_example_export.xlsx](https://github.com/openstax/tutor-server/files/269686/cc_example_export.xlsx)


jpslav:
@rvnewell -- check out the example files in the description.  I know there was a discussion about averages on CC, and for now, I have all CCs included in the averages (regardless of if complete or not).

rvnewell:
Both look great.  We're still waiting on 2 decisions:  1) External assignments and 2) what to include in CC averages.  We should know about external assignments after FAB HS meeting tonight.   External assignments will most likely be one column with no late designation/info.  Will know more after tonight. I'm assuming Kathi will make a call on CC averages very soon (Maybe already made it in the Sprint planning meeting?).

Dantemss:
While you are waiting :) you could fix this line: https://github.com/openstax/tutor-server/pull/1090/files#diff-e993d4d5aa7e32573201c5143ff1fbd8L148

To handle nil (or alternatively migrate that column in openstax_accounts_account to disallow nil values).

jpslav:
@Dantemss fixed the bug you mentioned in https://github.com/openstax/tutor-server/pull/1090/commits/2a35306fa78b8433c7a5d699e3f3da184c6a2d13

Dantemss:
@jpslav I think you might have to also fix https://github.com/openstax/tutor-server/pull/1090/files#diff-a39789b63cd9e432c33b637155c3695bR282?

And I noticed the sort is ALSO happening in both `get_xx_performance_report.rb` routines, since the API output has to be sorted too. I recommend just removing the sort from the export routines and fixing the spec data to be sorted like the output of `get_xx_performance_report`. (see my PR)

Issue #1065, opened 2016-04-22
Handle exercises that fail to map to any pages


Dantemss:
Since we added manifests, it is possible to import exercises that fail to map to any pages in the ecosystem (for example, if the cnxmod tag was removed in the most recent version). In that case, just don't import those exercises.

jpslav:
Closed by #1098

PR #1098, opened 2016-05-19
Prevent the same exercise from being imported multiple times

### LINKED ISSUE #1095: Duplicate imports
Questions tagged to multiple modules seem to be imported multiple times (but they all end up in the same module). Add a check to prevent this.

Dantemss:
Due to splitting calls to Exercises

Closes #1095


Dantemss:
Also skips exercises that map to no pages to close #1065

jpslav:
Closes #1065

PR #1104, opened 2016-05-20
Fix background job exception rescue

### LINKED ISSUE #1092: Deleting a task that is publishing can make publishing fail over and over
Since it can't find the record anymore.

Dantemss:
OK this *seems* to successfully report the error and fail the job.

Closes #1092 by making it fail only once.


lib/active_job/provider_job_id_backport.rb
```diff
@@ -1,21 +0,0 @@
-# https://github.com/rails/rails/commit/dd8e859829bfcfd8cb0287ce804055b827a35af1
-# Remove when upgrading to Rails 5
```
jpslav: Why are we getting rid of this file?

jpslav:
Ready to merge when we answer my one question

lib/active_job/provider_job_id_backport.rb
Dantemss: I changed the rescue to be in the Delayed::Job instead of the ActiveJob so the id is present there already, making this monkey patch unnecessary. Also, it didn't work, which is what caused it to not fail the jobs properly :).

Dantemss:
@jpslav answered

lib/active_job/provider_job_id_backport.rb
jpslav: love the "also, it didn't work" :-)

PR #1109, opened 2016-05-25
Upgrade dependencies

dak:
Upgrades eventmachine and libv8 to resolve issues with OS X 10.11.

eventmachine was fixed in 1.0.9, but there doesn't appear to be any breaking changes in the newest 1.2.0.1 either, despite the minor upgrade.


Dantemss:
Somehow travis did not start... Let me try poking it.

PR #1107, opened 2016-05-23
Editing assignments

Dantemss:
IRREVERSIBLE MIGRATION

BACKUP YOUR DB IF INTEND TO USE IT TO TEST OLD CODE IN THE FUTURE

- Changed task_plan and tasking_plan allowed edits to match latest decisions
- Removed Entity::Task
- Added acts_as_paranoid to task_plan, tasking_plan, tasks, taskings, task_steps and taskeds.
- Added delete/undelete assignment API for teachers
- Add hide API for students
- Specs

PR #1112, opened 2016-05-27
Archive periods

Dantemss:
Depends on https://github.com/openstax/tutor-server/pull/1107

- Restore period API
- Specs for deleting/restoring
- Made students in archived periods not count as active
- Added archived periods to the course roster


app/routines/collect_course_info.rb
```diff
@@ -24,8 +24,6 @@ def collect_courses(courses: nil, user: nil)
   end
 
   def collect_course_info(courses, user, with)
-    with = enforce_order(of: :roles, before: :students, within: with)
```
jpslav: Why did this go away?

jpslav: Maybe because getting course periods now takes the roles...

jpslav:
Looks good. Does this need to simul-merge with a tutor-js PR?

app/routines/collect_course_info.rb
Dantemss: I changed the methods so enforcing the order of the method calls was not needed.

app/routines/collect_course_info.rb
```diff
@@ -79,47 +73,34 @@ def set_teacher_names(info, entity_course)
...
-    else
-      student_roles = run(:get_course_roles, course: entity_course, user: user, types: :student)
-                        .outputs.roles
-      student_period_ids = student_roles.map{ |role| role.student.period.id }.uniq
-      info.periods = periods.select{ |period| student_period_ids.include?(period.id) }
-    end
+    @roles ||= run(:get_course_roles, course: entity_course, user: user).outputs.roles
+
+    info.periods = run(:get_course_periods, course: entity_course, roles: @roles,
+                                            include_archived: true).outputs.periods
```
nathanstitt: @Dantemss This is including archived periods in the teacher's course JSON.  Is that intended?  FE can filter them out easily enough, but I thought you'd mentioned that archived periods should not sent.

Dantemss: I think you are right. I'll remove.

Dantemss: Actually I think I might not be able to. See slack.

PR #1113, opened 2016-05-31
Provide Exercise preview and context fields to FE

Dantemss:
This PR adds a few different fields to Exercises:

- `preview` field to be used in HW builder/question library cards (replaces iframes with static divs)
- `context` field to be used in the HW builder/question library AND in homeworks themselves (provides context for questions that have a `context-cnxfeature` tag)
- `has_interactive` boolean
- `has_video` boolean


jpslav:
Looks good.

PR #1117, opened 2016-06-02
capture shortId in imports, more data in Toc representer

jpslav:
Makes it so that https://github.com/openstax/tutor-js/pull/1078 can show CNX URLs with short IDs instead of long UUIDs.  Also lets that PR have non-hard-coded CNX webview URLs.


jpslav:
Added specs

PR #1120, opened 2016-06-03
Combine exercises that require context with the step before when possible

Dantemss:
- Actually send the exercise context to FE when students are working tasks
- Combine exercise context with the previous step when possible
- cnxfeature is now a new tag type
- Fixes a bug that prevented readings from picking up exercises by the cnxfeature tags


Dantemss:
Uh needs an FE PR to display the context but honestly go ahead and merge or else readings with cnxfeature tags will be broken anyway.

PR #1121, opened 2016-06-06
Limit export to HWs, readings, CCs

jpslav:
Types outside of these three blow up in the export, and we actually only want these types in the export anyway.

### DIFF
--- a/app/subsystems/tasks/performance_report/export_xlsx.rb
+++ b/app/subsystems/tasks/performance_report/export_xlsx.rb
@@ -29,7 +29,7 @@ module Tasks
         @helper.standard_package_settings(@package)
 
         setup_styles
-        exclude_non_due_tasks
+        remove_excluded_tasks
         write_data_worksheets
         make_first_sheet_active
         save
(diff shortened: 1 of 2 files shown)


app/subsystems/tasks/performance_report/export_xlsx.rb
```diff
@@ -29,27 +29,33 @@ def create
...
-          period_report[:data_headings].slice!(num_tasks-not_due_count..-1)
+          excluded_indices = []
+
+          period_report[:data_headings].each_with_index do |heading, ii|
+            if heading[:due_at] > Time.now || !%w(homework reading concept_coach).include?(heading[:type])
+              excluded_indices.push(ii)
+            end
+          end
+
+          period_report[:data_headings].reject!.with_index {|heading, ii| excluded_indices.include?(ii) }
```
Dantemss: In general I recommend assigning this to another variable and using `reject` instead of `reject!` to avoid possibly weird behavior if a transaction is retried or something like that. But it's probably fine in this case.

Dantemss:
Seems good.

PR #1127, opened 2016-06-09
DB dump and restore rake tasks

jpslav:
Tasks for dumping and restoring databases (for development purposes).

Dumped filenames start with the timestamp of the data's latest migration file.

When restoring dumps, this code will freak out if the data being restored is at a migration beyond what the current checked out code has. You can restore any dump generated from an older migration.  The restore task runs migrations as the last step to bring the data up to date.

```
rake db:dump['physics']  makes a filename ending with `_physics.dump`
rake db:dump             makes a filename ending with `_tutor.dump`
rake db:restore['phys']  will find a file matching first `phys` exactly,
                         then /\d+_phys(.dump)?/, then /.*phys.*/
rake db:restore          defaults the name argument to 'tutor'
```

To share these, upload dumps to owncloud (let's all use a consistent directory).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 # Ignore the default SQLite database and yaml_db database files
 /db/*.sqlite3*
 /db/data.yml*
+/db/dumps*
 
 # Ignore all logfiles and tempfiles.
 /log/*.log
--- a/Gemfile
+++ b/Gemfile
@@ -91,9 +91,6 @@ gem 'deep_cloneable'
 # Real time application monitoring
 gem 'newrelic_rpm'
 
-# YAML database backups
-gem 'yaml_db'
-
 # Date validations
 gem 'validates_timeliness'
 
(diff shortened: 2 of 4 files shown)


Dantemss:
The dump includes all tables and migrations so I think checking the migration timestamp is not necessary.

Also the pg_dump instructions say a lot of these options only make sense for the plain text format but idk how true that is.

Dantemss:
Also hopefully all devs understand that these have all the PII.

nathanstitt:
Looks great @jpslav.

jpslav:
@Dantemss I agree the dump includes all migrations, but if you're using a dump from today's master branch but working a hotfix off of something from a month ago, the dump won't work for you (the data will have migrated past what your code support).  That is what the migration timestamp is meant to help with.  That make sense?  

And these dumps won't necessarily have PII.  I'm more thinking about databases that we generated from the demo script or other script (like the large QA database), or databases that we make by hand through Tutor and want to save off so we can restore over and over while testing.  If we do use this on PII databases we'll have to be extremely careful and thoughtful in how we got about it.

Dantemss:
OK makes sense

PR #1132, opened 2016-06-11
Fix readings API performance

Dantemss:
- Cache page fragments and snap labs on import


Dantemss:
Since I cache fragments and snap labs here, this should also make assignment creation very slightly faster...

PR #1133, opened 2016-06-13
Salesforce changes to accomodate Tutor courses in addition to CC courses

jpslav:
### TODO

* [ ] Split `ImportSalesforceCourse` out of `ImportSalesforceCourses` so that (1) running course creation for one SF record still runs in a transaction (and is able to `run` other subroutines) and (2) any individual blowups don't roll back prior successes.
* [ ] Use (and be able to set) `is_normally_college` on offerings.


app/routines/import_salesforce_courses.rb
```diff
@@ -1,29 +1,37 @@
 class ImportSalesforceCourses
-  lev_routine
```
Dantemss: If you want the lev boilerplate stuff, you could do `lev_routine transaction: :no_transaction` right?

app/subsystems/course_membership/models/enrollment.rb
```diff
@@ -19,6 +19,15 @@ class CourseMembership::Models::Enrollment < Tutor::SubSystems::BaseModel
     }.outer}.where(newer_enrollment: {id: nil})
   }
 
+  scope :latest_with_deleted, -> {
```
Dantemss: Couldn't you just chain `.latest.with_deleted` instead?

app/routines/import_salesforce_course.rb
```diff
@@ -0,0 +1,95 @@
...
+    success!(candidate)
+  end
+
+  def success!(candidate)
+    candidate.error = nil
+  end
+
+  def error!(candidate, message)
+    candidate.error = message
+    log { "Error! candidate: #{candidate.id}; message: #{message}." }
```
Dantemss: Might wanna use `Rails.logger.error` or `Rails.logger.fatal` here because I think we usually don't even log `info`.

app/routines/update_salesforce_course_stats.rb
```diff
@@ -0,0 +1,294 @@
...
+# and period data.  Finds periods in courses that don't yet have an assigned
+# SF object and finds or makes one for them.
+#
+# Notes:
+#   1. In this file, "ar" stands for AttachedRecord, not ActiveRecord
+#   2. This file cannot be a Lev routine because it can create SF objects
+#      and we don't want a transaction that rolls back local objects because
+#      the SF objects cannot be rolled back.
+
+class UpdateSalesforceCourseStats
```
Dantemss: Same comment about `transaction: :no_transaction`.

Dantemss:
I like all the specs :)

app/routines/import_salesforce_courses.rb
jpslav: Yes, I think so.  Still might.

jpslav: Actually, remember why I didn't do this.  The only danger is someone changing in the future to use a `run(...)` call in this code (e.g. to run `ImportSalesforceCourse`) , which they could do when it is a routine.  If one of those subroutines has a higher isolation, the `no_transaction` here would get overridden (which is what I'm particularly trying to avoid).

app/routines/update_salesforce_course_stats.rb
jpslav: Ditto above.

app/subsystems/course_membership/models/enrollment.rb
jpslav: Tried it, didn't work for some reason.  When all specs are done at end, I can try again, maybe I was just doing something wrong.

jpslav:
More specs coming today!

app/routines/import_salesforce_course.rb
jpslav: We log info in production

app/subsystems/course_membership/models/enrollment.rb
jpslav: Oops I misconstrued your comment.  Yes, that makes sense.

PR #1140, opened 2016-06-14
Moved Hacks::AnswerExercise to lib/tasks/demo

Dantemss:
Also namespaced the Demo stuff.


lib/tasks/demo/demo_base.rb
```diff
@@ -481,7 +481,7 @@ def work_step(step, response)
     response = (response == '0' ? false : true) if response.is_a?(String)
 
     if step.tasked.exercise?
-      Hacks::AnswerExercise.call(task_step: step, is_correct: response)
+      AnswerExercise.call(task_step: step, is_correct: response)
```
jpslav: How is this `require`'d now that it is in `/spec`?

jpslav: I'm guessing the demo spec passes because it loads `/spec/support`, but that a straight demo run would blow up?

Dantemss: Hmm I suppose you are right... Wonder if AnswerExercise should live in the demo stuff instead?

jpslav: there's always the "hacks" ss ;-)

wherever you want is fine with me.

PR #1144, opened 2016-06-16
Require user sign-in to use short codes

nathanstitt:
Currently when a non-logged in user attempts to use a short code for an assignment, a `SecurityTransgression` is raised.

This is because the short-code routines eventually call `Tasks::GetRedirectUrl`, which needs to know if the user is a teacher or student, it calls `ChooseCourseRole` which raises `invalid_role` because the anonymous user doesn't have any role.

It seems to me that the redirected url will require login anyway, so it's easiest just to require logging in before the redirect occurs.  However this isn't going to work if there are uses for short-codes that do not require logging in.  If there are, then we'll have to figure out a different solution.

I haven't fixed up the specs yet until we can discuss.

### DIFF
--- a/app/controllers/short_codes_controller.rb
+++ b/app/controllers/short_codes_controller.rb
@@ -1,5 +1,4 @@
 class ShortCodesController < ApplicationController
-  skip_before_filter :authenticate_user!
 
   def redirect
     handle_with(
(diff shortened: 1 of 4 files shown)


jpslav:
Beautimous.  Thanks @nathanstitt

PR #1145, opened 2016-06-17
Update representable to 3.0.0

Dantemss:
Should improve overall API performance due to caching, but I expect more gains on big API's like performance report/forecast.

Also removes an unused representer and commented out controller method.


jpslav:
Looks good @Dantemss -- please release the gems and update the Gemfile, then merge at your discretion.

PR #1148, opened 2016-06-17
Fix plan settings after due date

Dantemss:
- Added a RailsCollectionSetter to properly populate collection records from representers without throwing errors


Dantemss:
Amanda said she'll convince everyone to keep the due-date-must-be-in-the-future validation during Monday scrum.

Dantemss:
And btw this fixes the issue without Amanda's PR... but Amanda's PR should save some DB accesses.

PR #1152, opened 2016-06-23
Speed up teacher dashboard

Dantemss:
- Allow FE to specify time range for dashboard tasks and plans
- Let FE know if a plan is done publishing so they don't have to query the job

# FE PR needs to:

- Specify time range for the teacher dashboard (time zones are assumed to be course time zone so maybe give the range 1 extra day)
- Change the logic for determining if a plan is published or not to use the is_draft, is_publishing, is_published fields
- Load the plan's shareable_url from the plan stats instead of the dashboard


app/controllers/api/v1/task_plans_controller.rb
```diff
@@ -358,20 +359,24 @@ def update_task_plan!(task_plan)
   # Distributes or updates distributed tasks for the given task_plan
   # Returns the job uuid, if any, or nil if the request was completed inline
   def distribute_or_update_tasks(task_plan)
-    task_plan.publish_last_requested_at = Time.current \
-      unless task_plan.tasks_past_open? || task_plan.is_draft?
+    should_publish_or_update = task_plan.is_publish_requested || task_plan.is_published?
+    should_publish = should_publish_or_update && !task_plan.tasks_past_open?
+    # should_update = should_publish_or_update && task_plan.tasks_past_open?
+
```
nathanstitt: ♻️ Commented code here and on line 376 below could be removed

db/schema.rb
```diff
@@ -778,6 +780,7 @@
   add_index "tasks_tasks", ["due_at_ntz", "opens_at_ntz"], name: "index_tasks_tasks_on_due_at_ntz_and_opens_at_ntz", using: :btree
   add_index "tasks_tasks", ["hidden_at"], name: "index_tasks_tasks_on_hidden_at", using: :btree
   add_index "tasks_tasks", ["last_worked_at"], name: "index_tasks_tasks_on_last_worked_at", using: :btree
+  add_index "tasks_tasks", ["opens_at_ntz"], name: "index_tasks_tasks_on_opens_at_ntz", using: :btree
   add_index "tasks_tasks", ["task_type"], name: "index_tasks_tasks_on_task_type", using: :btree
   add_index "tasks_tasks", ["tasks_task_plan_id"], name: "index_tasks_tasks_on_tasks_task_plan_id", using: :btree
   add_index "tasks_tasks", ["time_zone_id"], name: "index_tasks_tasks_on_time_zone_id", using: :btree
```
nathanstitt: Unrelated to the current PR, but so many indexes - can any of them be removed?  Doing so would speed up task creation.  

Wondering if the standalone indexes on `deleted_at`, `time_zone_id`, `task_type`, `time_zone_id` are actually used for querying.  Not sure we're every loading tasks using just those columns so it seems like they might be un-used.

db/migrate/20160623014118_add_first_and_last_published_at_to_task_plans.rb
```diff
@@ -0,0 +1,14 @@
+class AddFirstAndLastPublishedAtToTaskPlans < ActiveRecord::Migration
+  def change
+    rename_column :tasks_task_plans, :published_at, :first_published_at
+    add_column :tasks_task_plans, :last_published_at, :datetime
+
+    reversible do |dir|
+      dir.up do
+        Tasks::Models::TaskPlan.update_all(
+          'last_published_at = GREATEST(first_published_at, publish_last_requested_at)'
+        )
```
nathanstitt: Cool!  Didn't know about `GREATEST`

db/schema.rb
Dantemss: I think the only unused one might be the time_zone_id but even then I'm not 100% sure it's unused.

db/migrate/20160623014118_add_first_and_last_published_at_to_task_plans.rb
Dantemss: Me neither... had to google :)

lib/belongs_to_time_zone.rb
```diff
@@ -40,27 +40,17 @@ def define_time_zone_methods(args, options)
...
             # Drop any time_zones given, then write the result to the DB
             define_method("#{method_name}=") do |value|
-              datetime = case value
-              when String
-                DateTime.parse(value) rescue nil
-              else
-                value.try(:to_datetime)
-              end
+              datetime = value.is_a?(String) ? DateTimeUtilities.from_s(value) :
+                                               value.try(:to_datetime)
```
nathanstitt: Something I just did on the FE was to create a `DateHelper.to_date( datething )` that encodes the above pattern to convert a string to date unless it already was.  You might consider adding the above snippet to `DateTimeUtilities.to_datetime( datething )` and move this to there.

app/controllers/api/v1/task_plans_controller.rb
Dantemss: True, but I thought leaving this here improved the clarity of what the code is trying to do (commenting it out was more of an optimization, really).

nathanstitt:
Looks great to me @Dantemss.  I like how you beefed up the spec coverage considerably.

PR #1153, opened 2016-06-23
Speed up serialization

Dantemss:
- Added Oj gem for fast JSON parsing
- Converted all serialized fields to use Oj for serialization


db/migrate/20160623231430_convert_serialized_fields_to_json.rb
```diff
@@ -0,0 +1,64 @@
...
+          attributes[field_name] = new_value
+        end
+
+        model.update_columns attributes
+      end
+
+      fields.each{ |field_name, default| remove_column table_name, "#{field_name}_old" }
+    end
+  end
+
```
nathanstitt: Will this need to run while access to the system is on-going?  If so, I think we could reduce down-time by locking the table for updates, adding new column, convert values, then remove old and rename new to old.  That would allow the existing code to still query the column while conversion is on-going.

Things will still fail though since old code will be expecting YAML.  Best to run during maintenance if possible.

app/subsystems/content/models/map.rb
```diff
@@ -79,7 +80,7 @@ def create_page_id_to_page_id_map
     # but for now we don't handle that case
     # since it's hard to figure out what to do for the dashboard/scores
     page_id_to_pages_map.each do |page_id, pages|
-      page_id_to_page_id_map[page_id] = pages.size == 1 ? pages.first.id : nil
+      page_id_to_page_id_map[page_id.to_s] = pages.size == 1 ? pages.first.id : nil
     end
```
nathanstitt: I started to suggesting using `stringify_keys` on `page_id_to_page_id_map`, but then decided I liked the explicit `.to_s` you added better.  Left this useless comment just in case you didn't know about `stringify_keys` 😁

lib/json_serialize.rb
```diff
@@ -0,0 +1,41 @@
...
+        send options[:setter_method], typecast_value
+      end
+
+      protected options[:typecast_method]
+
+    end
+  end
+end
+
+ActiveRecord::Base.extend JsonSerialize::ClassMethods
```
nathanstitt: I like this and seems to be pretty solid to me.  We might want @jpslav to chime in though since it seems like a fairly large change and serialization isn't something we want to mess up.

app/subsystems/content/models/map.rb
Dantemss: :) I know about it hehe

db/migrate/20160623231430_convert_serialized_fields_to_json.rb
Dantemss: Yes, server goes down for migrations... or should.

lib/json_serialize.rb
Dantemss: Hmm...

lib/json_serialize.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      serialize attribute, JSON
+
+      send(options[:typecast], options[:typecast_method]) if options[:typecast].present?
+
+      define_method(options[:typecast_method]) do
+        return unless has_attribute?(attribute)
+
+        value = send attribute
+        return if options[:array] ? value.to_a.all?{ |val| val.is_a? type } : value.is_a?(type)
```
jpslav: I wonder if you're going to need some parens here around the ternary expression

lib/json_serialize.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def json_serialize(attribute, type, options = {})
+      options[:typecast] = :after_initialize unless options.has_key?(:typecast)
+      options[:setter_method] ||= "#{attribute}=".to_sym
+      options[:typecast_method] ||= "typecast_#{attribute}".to_sym
+      options[:conversion_method] ||= DEFAULT_CONVERSION_METHODS[type]
+
+      serialize attribute, JSON
+
+      send(options[:typecast], options[:typecast_method]) if options[:typecast].present?
```
jpslav: ok to cause typecasting to happen after initialize by default instead of doing it lazily on demand?

lib/json_serialize.rb
```diff
@@ -0,0 +1,41 @@
+module JsonSerialize
+  module ClassMethods
+    DEFAULT_CONVERSION_METHODS = {
+      Hash => :to_h,
+      Array => :to_a,
+      String => :to_s,
+      Integer => :to_i
```
jpslav: Add `Float => :to_f`?

jpslav:
Review complete.

lib/json_serialize.rb
Dantemss: Seems to work fine...

lib/json_serialize.rb
Dantemss: Hm not sure... I think I ran into some problem trying to do it lazily which is why I put it in after_initialize.

lib/json_serialize.rb
Dantemss: Sure.

PR #1157, opened 2016-06-28
Don't blow up when importing exercises with no context

Dantemss:
Also improves several error messages

### DIFF
--- a/app/routines/get_ecosystem_exercises_from_biglearn.rb
+++ b/app/routines/get_ecosystem_exercises_from_biglearn.rb
@@ -74,18 +74,18 @@ class GetEcosystemExercisesFromBiglearn
 
     WarningMailer.log_and_deliver({
       message: "Biglearn returned less exercises than requested.",
-      details: {
-        pools: pools.inspect, role: role.id, needed: count, got: biglearn_count
-      }
+      details: { pools: pools.map(&:uuid), role: role.id,
+                 requested: count, got: biglearn_count, exercise_numbers: numbers }
     }) if biglearn_count < count
 
     exercises = ecosystem.exercises_by_numbers(numbers)
     fatal_error(code: :missing_local_exercises,
                 message: "Biglearn returned more exercises than were " +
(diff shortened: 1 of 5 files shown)


app/routines/get_ecosystem_exercises_from_biglearn.rb
```diff
@@ -74,18 +74,18 @@ def exec(ecosystem:, role:, pools:, count:, difficulty: 0.5, allow_repetitions:
...
-                         "present locally. [pools: #{biglearn_pools.map(&:uuid)}, " +
-                         "role: #{role.id}, requested: #{count}, " +
-                         "from biglearn: #{biglearn_count}, " +
-                         "local found: #{exercises.size}] biglearn question ids: #{numbers}") \
+                         "present locally. [Pools: #{biglearn_pools.map(&:uuid)}, " +
+                         "Role: #{role.id}, Requested: #{count}, " +
+                         "Got (Biglearn): #{biglearn_count}, " +
+                         "Got (local): #{exercises.size}, " +
+                         "Exercise numbers: #{numbers}]") \
       if exercises.size < biglearn_count
```
nathanstitt: FWIW, you could use a HEREDOC with `strip_heredoc` as:
```ruby
 message: <<-EOM.strip_heredoc
             Biglearn returned more exercises than were
             present locally. [pools: #{biglearn_pools.map(&:uuid)}
             present locally. [Pools: #{biglearn_pools.map(&:uuid)}
             role: #{role.id}, requested: #{count}
EOM
```

Rails 2.3 has the neat "squiggly heredoc"  `<<~` operator which does the leading space stripping as well.  Now if spellcheck would stop auto-correcting "heredoc" to "heretic" all would be well 😁

nathanstitt:
looks great @Dantemss - my email thanks you as well

app/routines/get_ecosystem_exercises_from_biglearn.rb
Dantemss: Yes :) too lazy

PR #1161, opened 2016-06-28
Prevent potentially assigning the wrong open/due dates…

Dantemss:
… to students that join a period after an assignment is published.


Dantemss:
This is a bug that I noticed while looking for performance improvements. There is a chance that a student being added to a period would receive the wrong open/due dates on their assignments.

PR #1166, opened 2016-06-29
Fix AP Bio mapping

Dantemss:
- Removed APLO from mapping tag types
- Exercise to page mapping now simply follows the page to page mapping


Dantemss:
Removed the `aplo` tags from page mappings. This fixes AP Bio pages that mapped to multiple pages (due to aplos appearing in multiple pages).
Changed exercise mapping to simply follow the result of the page mapping. This fixes exercises that had no LO (and no CNXMOD tag due to very old ecosystems) being unable to map to any new pages.

PR #1172, opened 2016-07-01
Distribute Tasks Performance

Dantemss:
- Updated activerecord-import for performance gains (and to remove autosave associations)
- Disabled uniqueness validations during import (DB constraints enforce uniqueness)


Dantemss:
Would be nice if someone could check that DB constraints for all Content and Task models are really present.

PR #1175, opened 2016-07-05
Test url_node exists before accessing it when adding iframe classes

nathanstitt:
Importing physics content was failing because there's a block that doesn't have any iframe or a tags in the fragment

### DIFF
--- a/lib/openstax/cnx/v1/fragment/embedded.rb
+++ b/lib/openstax/cnx/v1/fragment/embedded.rb
@@ -28,11 +28,11 @@ module OpenStax::Cnx::V1
 
       url_node = node.at_css(TAGGED_URL_CSS) || node.css(UNTAGGED_URL_CSS).last
 
-      if url_node && url_node.name == 'iframe'
-        @width  = url_node['width']
-        @height = url_node['height']
-        @url    = url_node['src'] || url_node['href']
+      @width = url_node.try(:[], 'width') || default_width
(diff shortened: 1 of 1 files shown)


lib/openstax/cnx/v1/fragment/embedded.rb
```diff
@@ -28,12 +28,11 @@ def initialize(node:, title: nil, labels: nil)
 
       url_node = node.at_css(TAGGED_URL_CSS) || node.css(UNTAGGED_URL_CSS).last
 
-      @width = url_node.try(:[], 'width') || default_width
-      @height = url_node.try(:[], 'height') || default_height
+      if url_node && url_node.name == 'iframe'
+        @width  = url_node['width']
+        @height = url_node['height']
```
jpslav: `url_node['width']` is only being used if the node exists and the name is iframe (whereas before it was only checking existence of the node) -- just checking that this is what you were intending.

nathanstitt: Hm, that was my initial thought for what the code should be doing.  Only iframe's seem like they should have a width set.  But then I see the previous version (https://github.com/openstax/tutor-server/pull/1171/files#diff-092a8923ab4fe824e2f3fb7c58824a46R33) had the same semantics.

Given that I don't fully understand the fragment parsing code, it's probably better to put it back to the original behavior.  Making the change now

PR #1177, opened 2016-07-06
Undropped students get assignments from when they were dropped

jpslav:
I wrote a test to show two things:

1. that new students added to a course get previously published assignments
2. that undropped students get assignments published while they were dropped.

The first one was working, the second one was not.  This PR adds a call to reassign published task plans for students that were undropped.


kathi-fletcher:
Kudos for these tests @jpslav

PR #1185, opened 2016-07-08
Local query version of Biglearn client

jpslav:


lib/openstax/biglearn/v1.rb
```diff
@@ -85,28 +86,61 @@ def self.configuration
...
+  def self.client_name
+    Settings::Db.store.biglearn_client
   end
 
   private
 
   def self.client
+    @a_local_query_client ||= local_query_client
+    @a_real_client ||= real_client
+    @a_fake_client ||= fake_client
```
kevinburleigh75: Do we really want to be forcing all clients to exist/be instantiated here?

lib/openstax/biglearn/v1.rb
```diff
@@ -85,28 +86,61 @@ def self.configuration
...
+    case client_name
+    when :local_query
+      @a_local_query_client
+    when :real
+      @a_real_client
+    when :fake
+      @a_fake_client
+    else
+      @a_real_client
+    end
```
kevinburleigh75: can `client_name` ever be not set? if not, else should really be an exception

lib/openstax/biglearn/v1.rb
jpslav: good point

lib/openstax/biglearn/v1.rb
jpslav: It should always be set unless something bad happens.  So you're probably right

PR #1186, opened 2016-07-11
Adds a completion rate to the `data_headings` of performance reports.

pwolfert:


nathanstitt:
@pwolfert I think you can probably fix the specs by adding `completion_rate:` to the spec's `data_headings` and `data` sections https://github.com/openstax/tutor-server/blob/master/spec/controllers/api/v1/performance_reports_controller_spec.rb#L74-L88

That's what travis is trying to say but it's hard to read.  You can re-run the failing spec locally by running `bundle exec spec spec/controllers/api/v1/performance_reports_controller_spec.rb` from the tutor-server directory

Dantemss:
Yeah ignore the other differences, they aren't real... just add the `completion_rate` to the specs

app/subsystems/tasks/performance_report_methods.rb
```diff
@@ -2,13 +2,20 @@ module Tasks
   module PerformanceReportMethods
     protected
 
+    def completion_fraction(tasks)
+      completed_count = tasks.reduce(0) do |count, task|
+        count + (task.status == 'completed' ? 1 : 0)
+      end
```
jpslav: Is `task` a real `Tasks::Models::Task`? If so, replace with `completed_count = tasks.count(&:completed?)`

jpslav:
Looks good @pwolfert -- only one comment

PR #1187, opened 2016-07-12
Make specs more resilient to ordering errors

jpslav:
Fixes some irregular errors due to ordering of specs.  Even with fixes that make the specs less reliant on things like `SomeRecord.first`, I was still getting intermittent and varying spec failures on full runs.  After reading around, decided to make more use of `DatabaseCleaner`, which resolved the issues.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -67,20 +67,24 @@ RSpec.configure do |config|
     DatabaseCleaner.clean_with(:truncation)
   end
 
-  config.before(:each) do
+  config.before(:all) do
     DatabaseCleaner.strategy = :transaction
   end
 
-  config.before(:each, :js => true) do
+  config.before(:all, js: true) do
     DatabaseCleaner.strategy = :truncation
   end
 
-  config.before(:each, type: :request) do
+  config.before(:all, type: :request) do
     DatabaseCleaner.strategy = :truncation
   end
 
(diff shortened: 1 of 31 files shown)


spec/rails_helper.rb
```diff
@@ -60,14 +60,39 @@
...
+
+  config.before(:each, type: :request) do
+    DatabaseCleaner.strategy = :truncation
+  end
+
+  config.before(:each, truncation: true) do
+    DatabaseCleaner.strategy = :truncation
+  end
+
+  config.before(:each) do
```
Dantemss: Maybe `before(:all)`? If that exists. `before(:each)` is too late.

spec/rails_helper.rb
```diff
@@ -60,14 +60,39 @@
...
+    DatabaseCleaner.start
+  end
+
+  # https://github.com/DatabaseCleaner/database_cleaner#rspec-with-capybara-example says:
+  #   "It's also recommended to use append_after to ensure DatabaseCleaner.clean
+  #    runs after the after-test cleanup capybara/rspec installs."
+  config.append_after(:each) do
```
Dantemss: Same comment as above.

Dantemss:
This PR wraps all RSpec `describe` blocks in transactions that rollback at the end so we no longer need to explicitly call `DatabaseCleaner`.

PR #1188, opened 2016-07-12
Verify the presence of role associations

nathanstitt:
Somehow roles are getting their associations nulled, but retaining the
target_type. i.e. a role will say it's :target_type is :student, but
role.student will be null

### DIFF
--- a/app/subsystems/role/get_user_roles.rb
+++ b/app/subsystems/role/get_user_roles.rb
@@ -13,7 +13,7 @@ class Role::GetUserRoles
 
     if role_types != :any
       role_types = [role_types].flatten.map(&:to_s)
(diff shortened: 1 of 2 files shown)


nathanstitt:
i'm not super happy with this approach. I think it'd be better to figure out how the associations are getting removed without the `target_type` updating.

An alternate approach would be to remove the `target_type` and only check for the associations presence.  That'd be more accurate, but slower since it'd require a DB query.

Dantemss:
I think we should rethink how roles work later... but not right now

nathanstitt:
I think we can close this PR.  It turns out the underlying bug that it attempts to fix was due to a content error with old courses on the qa server.

PR #1199, opened 2016-07-19
Biglearn local query client preload pools, exercises and tasked exercises

Dantemss:
Includes https://github.com/openstax/tutor-server/pull/1196


Dantemss:
I might want to make the pool exercises into a real association in the future so I don't have to do this roundabout way of preloading...

Or maybe I should backport Rails 5's association `or` and use that...

PR #1206, opened 2016-07-20
Fix specs leaking models (and a dummy_models table)

Dantemss:


Dantemss:
Closing because https://github.com/openstax/tutor-server/pull/1187 has a better solution now.

PR #1209, opened 2016-07-21
Spring fixes

Dantemss:
- Removed some old spring binstubs
- Added spring explicitly to the Gemfile
- Made the Guardfile load spring from the bin folder
- Updated parallel_tests


nathanstitt:
looks good to me @Dantemss. I was able to run `guard` as well as `bin/rake parallel:spec` without errors.

PR #1211, opened 2016-07-21
Add additional appearance_codes with admin dropdown

nathanstitt:
Not sure if `secrets.yml` is the best place to put the mapping, since there's nothing secret about it at all.  We could also just add it as a constant somewhere, maybe `Content::APPEARANCE_CODES` ?


Dantemss:
Hm shouldn't the Concept Coach biology courses (including ludicrous and mini) have appearance code `college_biology`?

nathanstitt:
yes, I think you're correct. I'll update them to be so.  Right now it doesn't really matter since the FE will use identical styling.  At some point they're going to have different book icons and stuff though

Dantemss:
Hm and yes I would suggest putting the list of appearance codes somewhere else... secrets.yml comes from the deployment script and is spread there among like 20 files for different environments so it's a nightmare to update...

nathanstitt:
Yeah, @jpslav suggested adding it to the rails-settings-ui config values.  Will do that as well once I get back to it

PR #1212, opened 2016-07-21
Reorganize demo script config files; Add college demo scripts

Dantemss:
Currently testing this locally...


Dantemss:
I gzipped this up and ran it on tutor-demo btw.

PR #1217, opened 2016-07-28
SF bug fix + reliability improvements

jpslav:
Fixes one Salesforce stats update bug and adds several changes to make the update code less susceptible to oddities in the Salesforce data.

Suggest reviewing with `?w=0` appended to URL to eliminate whitespace changes (a large section was indented)

### DIFF
--- a/app/handlers/admin/courses_remove_salesforce.rb
+++ b/app/handlers/admin/courses_remove_salesforce.rb
@@ -31,7 +31,12 @@ class Admin::CoursesRemoveSalesforce
 
     # Get rid of all the attached records
     existing_ars.each do |existing_ar|
-      existing_ar.destroy
+      case existing_ar.attached_to_class_name
+      when 'Entity::Course'
+        existing_ar.destroy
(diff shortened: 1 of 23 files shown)


app/handlers/admin/courses_remove_salesforce.rb
```diff
@@ -29,18 +29,20 @@ def handle
...
     existing_ars.each do |existing_ar|
       existing_ar.destroy
       transfer_errors_from(existing_ar, {type: :verbatim}, true)
     end
 
-    # If that all went well, zero the stats on the SF object
-    sf_object.reset_stats
-    fatal_error(code: :could_not_clear_salesforce_stats) if !sf_object.save
+    # All have the same SF object, so just pull from one
+    sf_object = existing_ars.first.salesforce_object
```
Dantemss: Might be better to do this before destroying them to be safe?

jpslav: had it there but it worked below too, so I moved it to keep it together.  I think it is ok.

PR #1219, opened 2016-07-28
Restore period error message

Dantemss:
Return a proper error message (instead of a 500 error) if a deleted period being restored has a name that conflicts with another period


Dantemss:
You can merge this whenever, but an FE PR is required for a pretty error message. This will stop the emails, though :)

PR #1227, opened 2016-08-04
Dont blow up salesforce import if an OsAncillary has a null term_year

Dantemss:
This just returns nil instead which is fine for the comparison it is used in.


Dantemss:
Closing because Denver would rather be informed of this.

PR #1230, opened 2016-08-09
Add `ui_settings` to user profile

nathanstitt:
Needs to be merge along with: https://github.com/openstax/tutor-js/issues/1262

Also renames bootstrap data keys to have uniform names for urls: `tutor_api_url` and `accounts_api_url`

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -23,4 +23,27 @@ class Api::V1::UsersController < Api::V1::ApiController
     end
   end
 
+  api :POST, '/ui-settings', 'Save settings about the currently logged in user'
+  description <<-EOS
+    Saves the user interface settigs for the current user.
+    The front-end of the application uses this endpoint to save various non-critical settings
+    related to how the interface should behave based on actions the user has performed
+    Returns header forbidden (403) if the user is not logged in or api_errors if the update fails.
+  EOS
+  def ui_settings
+    profile = current_human_user.to_model
+
+    # sure seems like this *should* work, but it raises:
+    #   NoMethodError: undefined method `user_models_profile_url
+    # standard_update(profile, Api::V1::UiSettingsRepresenter)
+
+    OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, profile)
+    consume!(profile, represent_with: Api::V1::UiSettingsRepresenter)
+    if profile.save
+      head :accepted
+    else
+      render_api_errors(model.errors)
+    end
+  end
+
 end
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -31,19 +31,7 @@ class Api::V1::UsersController < Api::V1::ApiController
     Returns header forbidden (403) if the user is not logged in or api_errors if the update fails.
   EOS
   def ui_settings
-    profile = current_human_user.to_model
-
-    # sure seems like this *should* work, but it raises:
-    #   NoMethodError: undefined method `user_models_profile_url
-    # standard_update(profile, Api::V1::UiSettingsRepresenter)
-
-    OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, profile)
-    consume!(profile, represent_with: Api::V1::UiSettingsRepresenter)
-    if profile.save
-      head :accepted
-    else
-      render_api_errors(model.errors)
-    end
+    standard_update(current_human_user.to_model, Api::V1::UiSettingsRepresenter, location: nil)
   end
 
 end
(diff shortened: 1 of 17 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -23,4 +23,27 @@ def show
...
+    related to how the interface should behave based on actions the user has performed
+    Returns header forbidden (403) if the user is not logged in or api_errors if the update fails.
+  EOS
+  def ui_settings
+    profile = current_human_user.to_model
+
+    # sure seems like this *should* work, but it raises:
+    #   NoMethodError: undefined method `user_models_profile_url
+    # standard_update(profile, Api::V1::UiSettingsRepresenter)
+
```
nathanstitt: any thoughts on why this doesn't work? 

Note: I only see one other use of `standard_update` :
https://github.com/openstax/tutor-server/blob/master/app/controllers/api/v1/task_steps_controller.rb#L28 and it has a comment saying "Modified standard_update code"
 
maybe it just isn't tested very well?

jpslav: Try `standard_update(profile, Api::V1::UiSettingsRepresenter, location: nil)` to prevent `respond_with` from trying to guess a URL for the profile model

app/controllers/api/v1/users_controller.rb
```diff
@@ -23,4 +23,27 @@ def show
     end
   end
 
+  api :POST, '/ui-settings', 'Save settings about the currently logged in user'
```
jpslav: should be `api :PUT`

app/subsystems/user/models/profile.rb
```diff
@@ -19,9 +19,12 @@ class Profile < Tutor::SubSystems::BaseModel
       has_one :customer_service, dependent: :destroy, inverse_of: :profile
       has_one :content_analyst, dependent: :destroy, inverse_of: :profile
 
+      json_serialize :ui_settings, Hash
```
jpslav: Do we need to do special stuff with this on the BE?  Can we just have it be text and have all serialization/deserialization be in the FE?

app/controllers/api/v1/users_controller.rb
```diff
@@ -23,4 +23,27 @@ def show
...
+    # standard_update(profile, Api::V1::UiSettingsRepresenter)
+
+    OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, profile)
+    consume!(profile, represent_with: Api::V1::UiSettingsRepresenter)
+    if profile.save
+      head :accepted
+    else
+      render_api_errors(model.errors)
+    end
+  end
```
jpslav: Any risk of multiple browser windows (or tabs) being open and having one preference setting override an earlier one?  (changed preferences in one tab, work in another change preferences again, overwrites earlier prefs)

If so, could require updates to send a hash of the previous value which the server could check to make sure the FE is updating from the most up-to-date data.  Maybe not that big of a deal tho.

app/subsystems/user/models/profile.rb
```diff
@@ -19,9 +19,12 @@ class Profile < Tutor::SubSystems::BaseModel
...
+
       validates :account, presence: true, uniqueness: true
       validates :exchange_read_identifier, presence: true
       validates :exchange_write_identifier, presence: true
+      validates :ui_settings, max_json_length: 1000
```
jpslav: If don't do serialization on BE, can just use standard length validator.

app/subsystems/user/models/profile.rb
nathanstitt: I hadn't really considered doing that.  An earlier version of this had the column as `jsonb` but then @Dantemss showed me his faster `json_serialize` helper. 

I'm conflicted as to whether a plain text column makes sense.  

On one hand the BE should never need to read the column values (and could always parse it itself if needed),  but on the other hand having it stored as JSON is a sanity check to make sure the FE isn't just spewing junk into it which would prevent the BE from ever parsing it if needed, for instance if someday research wants to know how many students have "x" flag toggled.

Do you see any value in the extra check to make sure the column is valid JSON?  If not, then I'd say we could just do a text field.

app/controllers/api/v1/users_controller.rb
nathanstitt: Yeah, I like this idea.  Will add that concept

app/subsystems/user/models/profile.rb
jpslav: Fine to leave it this way.  Will an error be kicked out if not valid JSON?

nathanstitt: Yes, it raises a `JSON::ParserError` exception

PR #1231, opened 2016-08-10
Use `unscoped` to update all records in migration

nathanstitt:
Otherwise they'll fail the later `set not null`

This is going in my examples of  "default_scope is dangerous" 😁


Dantemss:
Yeah I'm not too happy paranoia's defaults either...

PR #1236, opened 2016-08-12
Display performance forecast for a section as soon as any task is open

Dantemss:
Emergency PR for Soc

### DIFF
--- a/lib/course_guide_methods.rb
+++ b/lib/course_guide_methods.rb
@@ -42,7 +42,7 @@ module CourseGuideMethods
                          completed_exercises_count_by_mapped_core_page_ids)
 
     mapped_core_pages_by_chapter.map do |chapter, mapped_core_pages|
-      mapped_core_page_ids = mapped_core_pages.map(&:id).sort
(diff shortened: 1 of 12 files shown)


Dantemss:
Question: should we get rid of the `UpdateClues` routine? I don't think the new biglearn local query client caches clues, so it's probably not doing us any good.

lib/course_guide_methods.rb
```diff
@@ -2,122 +2,162 @@ module CourseGuideMethods
...
-
-    exercise_id_map = {}
-    exercises.each{ |exercise| exercise_id_map[exercise.id] = exercise_map[exercise] }
-
-    exercise_id_map
+  def self.included(base)
+    base.lev_routine express_output: :course_guide
+    base.uses_routine GetHistory, as: :get_history
+    base.uses_routine GetCourseEcosystem, as: :get_course_ecosystem
+    base.uses_routine GetCourseEcosystemsMap, as: :get_course_ecosystems_map
```
jpslav: fancy

PR #1237, opened 2016-08-12
Removed the UpdateClues routine

Dantemss:
Because the Biglearn local query client does not cache the clues, this routine is no longer needed.

JP says to wait to merge until the new BL is ready.


jpslav:
Can you also remove the rake task, `tutor-server/lib/tasks/openstax/biglearn/update_clues.rake`?  Then we can merge (this "JP" fellow doesn't know what he's talking about).

Dantemss:
@jpslav done

PR #1238, opened 2016-08-12
Update url rewriting code

Dantemss:
- Relative links default to using cnx.org instead of archive.cnx.org
- Link rewriting now replaces in-book links with the book_location of the matching page in the format `2` or `2.1` instead of the page UUID

### DIFF
--- a/app/subsystems/content/models/book.rb
+++ b/app/subsystems/content/models/book.rb
@@ -24,7 +24,7 @@ class Content::Models::Book < Tutor::SubSystems::BaseModel
   end
 
   def webview_url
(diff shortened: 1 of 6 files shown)


app/subsystems/content/models/book.rb
```diff
@@ -24,7 +24,7 @@ def archive_url
   end
 
   def webview_url
-    archive_url.gsub('archive.','')
+    archive_url.sub(/archive[\.-]?/, '')
```
jpslav: What case does the `-` account for?

PR #1242, opened 2016-08-15
Several improvements on tutor admin console

BryanEliDimas:
**This PR right now is just for @Dantemss to review and to see if it passes the Travis tests.**

Added pagination on tutor admin courses index page.

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -69,6 +69,31 @@ $(document).ready(function() {
     $('a[href="' + tab + '"]').click();
   }
 
+  //========== Courses table ordering =============//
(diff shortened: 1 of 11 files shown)


Dantemss:
`CollectCourseInfo` is not guaranteed to return the results in the same order as the inputs, so that would need to change to preserve the order within each page.

Also, would be nice to get rid of the N+1 queries in `CollectCourseInfo`...

Dantemss:
And even better, we could just NOT use `CollectCourseInfo` here and instead define some getters to get the stuff we care about (teacher names, ecosystem name) in the course model.

Dantemss:
I guess I'll merge this and worry about removing `CollectCourseInfo` later...

PR #1246, opened 2016-08-16
Fix the chapter clues

Dantemss:
The optimized get_clues code I wrote assumed each exercise was only in 1 pool. That caused chapter clues to be computed using 0 exercises, since other pools were overwriting each exercises' pool with the page pools. Fixed that mistake and added a test for it in the specs.

### DIFF
--- a/lib/openstax/biglearn/v1/local_query_client.rb
+++ b/lib/openstax/biglearn/v1/local_query_client.rb
@@ -59,7 +59,6 @@ module OpenStax::Biglearn::V1
 
       hash = {}
       tasked_exercises_by_pool_uuid.each do |uuid, tasked_exercises|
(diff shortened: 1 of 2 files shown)


lib/openstax/biglearn/v1/local_query_client.rb
```diff
@@ -98,12 +97,12 @@ def completed_tasked_exercises_by(pool_uuids:, roles:)
                                                       .select([:id, :number])
                                                       .index_by(&:id)
 
-      pool_uuids_by_exercise_number = {}
+      pool_uuids_by_exercise_number = Hash.new{ |hash, key| hash[key] = [] }
```
jpslav: neat

jpslav:
Looks good.

PR #1248, opened 2016-08-17
Teacher updates student id

nathanstitt:
Goes with https://github.com/openstax/tutor-js/pull/1288, can be merged before it is though

### DIFF
--- a/app/controllers/api/v1/students_controller.rb
+++ b/app/controllers/api/v1/students_controller.rb
@@ -37,18 +37,23 @@ class Api::V1::StudentsController < Api::V1::ApiController
     #{json_schema(Api::V1::StudentRepresenter, include: :writeable)}
   EOS
   def update
+    @period_update_result = nil
     @student.with_lock do
       OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, @student)
-      payload = consume!(Hashie::Mash.new, represent_with: Api::V1::StudentRepresenter)
-      period = CourseMembership::Models::Period.find(payload['course_membership_period_id'])
-      @result = MoveStudent.call(student: @student, period: period)
-      OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, @student)
+      payload = consume!(@student, represent_with: Api::V1::StudentRepresenter)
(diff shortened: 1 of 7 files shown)


app/representers/api/v1/student_teacher_update_representer.rb
```diff
@@ -0,0 +1,16 @@
...
+    property :student_identifier,
+             type: String,
+             writeable: true,
+             readable: true,
+             schema_info: {
+               required: true
+             }
+
+  end
+end
```
jpslav: Why not just put this in the existing `StudentRepresenter`?

app/controllers/api/v1/students_controller.rb
```diff
@@ -37,18 +37,23 @@ def update_self
     #{json_schema(Api::V1::StudentRepresenter, include: :writeable)}
   EOS
   def update
+    @period_update_result = nil
     @student.with_lock do
       OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, @student)
-      payload = consume!(Hashie::Mash.new, represent_with: Api::V1::StudentRepresenter)
-      period = CourseMembership::Models::Period.find(payload['course_membership_period_id'])
-      @result = MoveStudent.call(student: @student, period: period)
-      OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, @student)
```
jpslav: Why was this line here a second time, and why is it gone now?

app/representers/api/v1/student_teacher_update_representer.rb
nathanstitt: While they're identical right now, I thought that the Teacher shouldn't be able to update the same fields the student does.  

For instance in the future we might allow a student to directly edit their name (w/out going through accounts), but we wouldn't want a Teacher to do so.

app/controllers/api/v1/students_controller.rb
nathanstitt: Hm, at first I thought it was an oversight, but now suspect it's checking that teacher also has permissions on the period the student is moving to.  It uses `UserIsCourseTeacher` to verify the user is a teacher on the student's course.  I'm restoring it and commenting to that effect

PR #1251, opened 2016-08-18
Optimize GetHistory and CourseGuideMethods for faster performance forecast

Dantemss:
More ARel...

After these changes the performance forecast takes about the same time it did before the most recent changes.


Dantemss:
Hm hold on a sec, I thought of one more optimization...

Dantemss:
Nah nvm, current code is fine.

PR #1255, opened 2016-08-19
If a page fails to map for some reason, skip it and don't explode

Dantemss:


jpslav:
Do we need to know that it failed to load? (is this change hiding something important from devs and the content team?)

Dantemss:
Hm good question. I will poke production and try to figure out why this is happening there.

Dantemss:
I poked production and it turned out to be some old Content::Models::Map from old code... I recreated them all (never Content::Models::Map code maps intro pages, old code didn't).

Dantemss:
I guess we can close this for now.

PR #1256, opened 2016-08-20
Prevent corrupted exports when there is no assignment data

jpslav:
The main change is adding a `disjoint_range` method that returns a good default value when there is no range to give.  Lots of other changes are because I moved some methods into an included module.


jpslav:
@BryanEliDimas  please review

spec/lib/xlsx_utils_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(dummy.cell_ref(row: 99, column: 2)).to eq('C99')
+      expect(dummy.cell_ref(row: 137, column: 25)).to eq('Z137')
+    end
+
+    it 'finds double-letter cells' do
+      expect(dummy.cell_ref(row: 5, column: 26)).to eq('AA5')
+      expect(dummy.cell_ref(row: 69, column: 27)).to eq('AB69')
+      expect(dummy.cell_ref(row: 420, column: 57)).to eq('BF420')
+    end
+  end
```
BryanEliDimas: I see the "happy path" but I asked myself, what would happen if `row` or `column` is nil? I tested it; raises exception `NoMethodError: undefined method `>=' for nil:NilClass` . Would it be worth it to protect against this?

spec/lib/xlsx_utils_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(dummy.cell_ref(row: 5, column: 0)).to eq('A5')
+      expect(dummy.cell_ref(row: 99, column: 2)).to eq('C99')
+      expect(dummy.cell_ref(row: 137, column: 25)).to eq('Z137')
+    end
+
+    it 'finds double-letter cells' do
+      expect(dummy.cell_ref(row: 5, column: 26)).to eq('AA5')
+      expect(dummy.cell_ref(row: 69, column: 27)).to eq('AB69')
+      expect(dummy.cell_ref(row: 420, column: 57)).to eq('BF420')
+    end
```
BryanEliDimas: I see the "happy path" but I asked myself, what would happen if `row` or `column` is `nil`? I tested it; raises exception `NoMethodError: undefined method>=' for nil:NilClass` . Would it be worth it to protect against this?

spec/lib/xlsx_utils_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  describe '#disjoin_range' do
```
BryanEliDimas: Tiny typo. `#disjoin*t*_range`

spec/subsystems/tasks/performance_report/export_cc_xlsx_spec.rb
```diff
@@ -96,6 +97,26 @@
...
+                                        options: {stringify_formulas: false})
+
+        # Uncomment this to open the file for visual inspection
+        # `open "#{filepath}"` and sleep(0.5)
+
+        expect{ @wb = Roo::Excelx.new(filepath) }.to_not raise_error
+      end
+    end
+
+    it 'does not put in invalid formulas that explode when open in Excel' do
```
BryanEliDimas: If this is not testing anything, might it be worth skipping it? I would use `skip "it doesn't currently explode, leaving this context for ease of future testing"` so we get a nice warning :).

spec/subsystems/tasks/performance_report/export_cc_xlsx_spec.rb
```diff
@@ -96,6 +97,26 @@
...
+      # it doesn't currently explode, leaving this context for ease of future testing
```
BryanEliDimas: If this is not testing anything, might it be worth skipping it? I would use `skip "it doesn't currently explode, leaving this context for ease of future testing"` so we get a nice warning :).

spec/subsystems/tasks/performance_report/export_cc_xlsx_spec.rb
```diff
@@ -236,6 +257,43 @@ def report_with_empty_period
...
+        } ,
+        data_headings: [],
+        students: [
+          {
+            name: "Abby Gail",
+            first_name: "Abby",
+            last_name: "Gail",
+            student_identifier: "SID2",
+            role: nil,
+            data: []
```
BryanEliDimas: Are we sure sure that "no data" means an empty array? What if "no data" means `nil` or `""`?

spec/subsystems/tasks/performance_report/export_cc_xlsx_spec.rb
```diff
@@ -96,6 +96,26 @@
...
+                                        options: {stringify_formulas: false})
+
+        # Uncomment this to open the file for visual inspection
+        # `open "#{filepath}"` and sleep(0.5)
+
+        expect{ @wb = Roo::Excelx.new(filepath) }.to_not raise_error
+      end
+    end
+
+    it 'does not put in invalid formulas that explode when open in Excel' do
```
BryanEliDimas: Looking at the purpose for this PR and this test's purpose, it seems to me like we should really be testing this. Is it or why is it not possible to test this currently?

spec/lib/xlsx_utils_spec.rb
jpslav: Would be reasonable to raise an `IllegalArgument` error or something inside that code so we don't have to track down a nil error in the future.  I yield to you on this after this PR is merged :-)

spec/subsystems/tasks/performance_report/export_cc_xlsx_spec.rb
jpslav: We can't ever really automatically test it because we don't have an emulator for what Excel does to tell if files are valid.  I also don't want to `xit` it b/c I don't want it to show up as a skipped test in our runs, because there is nothing we can ever do to finish it.  We could comment the whole thing out, but eh...

spec/subsystems/tasks/performance_report/export_cc_xlsx_spec.rb
jpslav: don't think `""` will happen, but who knows `nil` could, so I renamed `report_with_empty_data`.

spec/subsystems/tasks/performance_report/export_cc_xlsx_spec.rb
jpslav: `Roo` won't blow up if there's an invalid formula, so I think we just need to leave for manual testing.  I commented out the example and clarified the comment.

spec/lib/xlsx_utils_spec.rb
jpslav: I made this change.

jpslav:
Review comments addressed - thanks @BryanEliDimas!

spec/lib/xlsx_utils_spec.rb
BryanEliDimas: re: Would be reasonable to raise an IllegalArgument.
Sounds good, will do!

spec/subsystems/tasks/performance_report/export_cc_xlsx_spec.rb
BryanEliDimas: Ok, works too.

BryanEliDimas:
re: Review comments addressed.
Done! I'll a guard clause to `#cell_ref` to protect against `nil`. And, it's been my pleasure –and honor– doing my first code review! 😄

spec/lib/xlsx_utils_spec.rb
BryanEliDimas: Awesome! Ok, I just realized.

PR #1259, opened 2016-08-26
Fixed minor inconsistencies with representer use

Dantemss:


app/representers/api/v1/admin/user_search_representer.rb
```diff
@@ -2,7 +2,7 @@ module Api::V1
   module Admin
     class UserSearchRepresenter < OpenStax::Api::V1::AbstractSearchRepresenter
 
-      collection :items, inherit: true, decorator: Api::V1::Admin::UserRepresenter
+      collection :items, inherit: true, extend: Api::V1::Admin::UserRepresenter
```
jpslav: Is `extend` better than `decorator`?  Relatively ancient guidance led to use `decorator` from way back... don't know if still relevant. https://apotonick.wordpress.com/2013/05/04/use-roars-new-decorator-if-you-dislike-extend/

Dantemss: It's just an alias, but the documentation seems to just use `extend`.

PR #1261, opened 2016-08-29
Student Identifier Fixes

Dantemss:
- Changed student_identifier unique index to apply only to records that are not deleted
- Added a check for duplicate student identifiers when undropping students
- Added error messages for duplicate student identifiers when updating and undropping students
- Specs

### DIFF
--- a/app/subsystems/course_membership/activate_student.rb
+++ b/app/subsystems/course_membership/activate_student.rb
@@ -3,7 +3,12 @@ module CourseMembership
     lev_routine express_output: :student
 
     def exec(student:)
-      fatal_error(code: :already_active) unless student.deleted?
+      fatal_error(code: :already_active,
+                  message: 'Student is already active') unless student.deleted?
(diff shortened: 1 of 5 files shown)


Dantemss:
Might need FE... Possibly.

nathanstitt:
yah, I think it will need some FE attention so it can handle the `student_identifier_has_already_been_taken` error on the additional endpoints.  I'll take a look to make sure

jpslav:
Do we want to just not require student IDs to be unique and just flag them for teacher if there are dupes?

jpslav:
Looks good otherwise.

Dantemss:
@jpslav I suggested that but I believe Dr. Grim said that would just create instructor headaches.

nathanstitt:
Yeah, I also updated the course settings FE so it's pretty easy for a teacher to change the student ID of a dropped student if there's duplicate errors when attempt to  un-drop.

PR #1262, opened 2016-08-29
Fix personalized multipart group types, related content and labels

Dantemss:
- TaskExercise now preserves group_type, related_content and labels for multipart exercises
- Specs
- Added a migration to fix existing personalized multipart questions


jpslav:
Looks good.  Thanks @Dantemss

PR #1263, opened 2016-08-30
Create routine to collect import jobs data

BryanEliDimas:
Created `CollectImportJobsData` to clean up the controller but more importantly, to get rid of an N+1 query due to `where` statements inside of the views.

### DIFF
--- a/app/controllers/customer_service/courses_controller.rb
+++ b/app/controllers/customer_service/courses_controller.rb
@@ -5,11 +5,17 @@ class CustomerService::CoursesController < CustomerService::BaseController
 
   def index
     @query = params[:query]
+    courses = SearchCourses.call(query: params[:query], order_by: params[:order_by]).outputs
+    params[:per_page] = courses.total_count if params[:per_page] == "all"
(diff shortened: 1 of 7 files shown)


app/controllers/customer_service/courses_controller.rb
```diff
@@ -5,11 +5,17 @@ class CustomerService::CoursesController < CustomerService::BaseController
...
-    @course_infos = CollectCourseInfo[courses: courses,
-                                      with: [:teacher_names, :ecosystem_book]]
+    @total_courses = courses.try(:total_count)
+    @course_infos = courses.items.preload(
+      [
+        :profile, { teachers: { role: [:role_user, :profile] }, periods_with_deleted: :latest_enrollments_with_deleted }
+      ],
+      [ ecosystems: [:books] ],
+      [ :periods ]
+    ).try(:paginate, params_for_pagination)
```
Dantemss: These 2 `try`s should not be needed?

BryanEliDimas: Fixed and spec still passed.

PR #1268, opened 2016-09-02
DON'T MERGE - Replaced unicorn with puma for evaluation

Dantemss:
Don't merge until we are sure we want puma AND we have updated the deployment script.
- Upgraded to ruby 2.3.1 because ruby < 2.3 has some memory leak bug when lots of threads (puma) are involved
- Added puma/puma_worker_killer
- Removed thin/unicorn/unicorn-worker-killer
- Moved some gems around

Deployment script changes needed:
- Remove installation of unicorn gem (no need to install puma during deployment, it's already in the Gemfile)
- Copy unicorn.rb settings to puma.rb (some DSL changes needed)
- Figure out if 2 processes with 5 threads each is really what we want
- Make sure deployed database.yml has pool == number of puma threads
- Figure out memory limits for puma_worker_killer and rolling restart interval (currently 1 GB and 12 hours)


Dantemss:
Kevin said he wanted about a million tutor workers for himself so I decided to give puma a try :)

Dantemss:
Closing mostly because I believe puma_worker_killer needs improvements to be as good as unicorn_worker_killer...

PR #1269, opened 2016-09-03
Improved thread-safety

Dantemss:
- Improved thread-safety of several modules
- Added comments about thread-unsafe code that only runs during initialization


jpslav:
@Dantemss - at some point do you think you could write up a napkin note of the things we should be doing to maintain the thread safety of our code going forward?

Dantemss:
@jpslav sure

PR #1272, opened 2016-09-09
Display an error when student attempts to join after drop or period is archived

nathanstitt:
Fixes a bug where if a student was dropped or the period was archived, but the student still has an enrollment url and uses it.

In that case we would seemingly accept the url and ask for a student id.  The BE would seem to accept the id, and then redirect to the course's dashboard.  The FE would then have a JS error when it didn't find an active course membership (because dropped/inactive), and would display a blank page.

When student has been dropped:
screen shot 2016-09-16 at 3 02 41 pm

When course is archived:
screen shot 2016-09-16 at 10 05 29 am

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -8,14 +8,7 @@ class CoursesController < ApplicationController
     handle_with(CoursesEnroll,
                 success: -> {},
                 failure: -> {
-                  case @handler_result.errors.map(&:code).first
-                  when :user_is_already_a_course_student
-                    send_to_student_dashboard(notice: "You are already enrolled in this course.")
-                  when :enrollment_code_not_found
-                    enrollment_code_not_found
-                  else
-                    raise StandardError, "Student URL enrollment failed: #{@handler_result.errors}"
-                  end
+                  handle_enrollment_failures(@handler_result.errors.map(&:code).first)
                 })
   end
 
@@ -28,21 +21,31 @@ class CoursesController < ApplicationController
                   )
                 },
                 failure: -> {
-                  case @handler_result.errors.map(&:code).first
-                  when :user_is_already_a_course_student
-                    send_to_student_dashboard(notice: "You are already enrolled in this course.")
-                  when :enrollment_code_not_found
-                    enrollment_code_not_found
-                  when :taken
-                    flash[:error] = "That school-issued ID is already in use."
-                    redirect_to token_enroll_path(params[:enroll][:enrollment_token])
-                  else
-                    raise StandardError, "Student URL enrollment confirmation failed: #{@handler_result.errors}"
-                  end
+                  handle_enrollment_failures(@handler_result.errors.map(&:code).first)
                 })
   end
 
   private
+
+  def handle_enrollment_failures(error_code)
+    case error_code
+    when :user_is_already_a_course_student
+      send_to_student_dashboard(notice: "You are already enrolled in this course.")
+    when :user_is_an_inactive_student
+      send_to_dashboard(notice: "Your membership in the course is inactive.")
+    when :enrollment_code_not_found
+      enrollment_code_not_found
+    when :taken
+      flash[:error] = "That school-issued ID is already in use."
+      redirect_to token_enroll_path(params[:enroll][:enrollment_token])
+    else
+      raise StandardError, "Student URL enrollment failed: #{@handler_result.errors}"
+    end
+  end
+
+  def send_to_dashboard(notice: nil)
+    redirect_to dashboard_path, webview_notice: notice
+  end
 
   def send_to_student_dashboard(notice: nil)
     course = @handler_result.outputs.course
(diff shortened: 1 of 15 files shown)


app/subsystems/course_membership/get_period.rb
```diff
@@ -8,7 +8,7 @@ def exec(id: nil, enrollment_code: nil)
       CourseMembership::Models::Period.with_deleted.preload(:course).find(id)
     elsif enrollment_code.present?
       enrollment_code = enrollment_code.gsub(/-/,' ') # for codes from URLs
-      CourseMembership::Models::Period.find_by(enrollment_code: enrollment_code)
+      CourseMembership::Models::Period.with_deleted.find_by(enrollment_code: enrollment_code)
     else
```
nathanstitt: **Note**  This changes the behavior of `GetPeriod` to include deleted records when it searches by enrollment code.  I did a search of `/app/*` and am pretty sure that only `CoursesEnroll` searches that way.

Dantemss: I think if we do this we need to add an error message that says something like "sorry, but registration is closed for this period" when a period is archived.

app/controllers/courses_controller.rb
```diff
@@ -28,22 +21,32 @@ def confirm_enrollment
...
+  def handle_enrollment_failures(error_code)
+    case error_code
+    when :user_is_already_a_course_student
+      send_to_student_dashboard(notice: "You are already enrolled in this course.")
+    when :user_is_an_inactive_student
+      send_to_dashboard(notice: "Your membership in the course is inactive.  Please contact the instructor for assistance.")
+    when :enrollment_code_not_found
+      enrollment_code_not_found
+    when :taken
+      flash[:error] = "That school-issued ID is already in use."
```
Dantemss: Ask UX, but I think we should also say something like "Have you already registered for this course on a different account?"

app/handlers/courses_enroll.rb
```diff
@@ -14,8 +14,16 @@ def handle
     fatal_error(code: :enrollment_code_not_found) if outputs.period.nil?
     outputs.course = outputs.period.course
 
+    user_test = UserIsCourseStudent.call(
+      course: outputs.period.course, user: caller,
+      include_dropped: true, include_archived: true
+    ).outputs
+
+    fatal_error(code: :user_is_an_inactive_student) \
+               if user_test.is_dropped || user_test.is_archived
```
Dantemss: Can't error out if period is archived. Need to proceed as usual. This will prevent students who took the course last semester from taking it again.

app/routines/add_user_as_period_student.rb
```diff
@@ -14,15 +14,20 @@ def exec(user:, period:, student_identifier: nil, assign_published_period_tasks:
     result = run(UserIsCourseTeacher, user: user, course: course)
 
     unless result.outputs.user_is_course_teacher
-      result = run(UserIsCourseStudent, user: user, course: course, include_dropped: true)
+      user_test = run(UserIsCourseStudent, user: user, course: course,
+                      include_dropped: true, include_archived: true).outputs
```
Dantemss: Same thing about archiving applies here I think.

Dantemss:
Rest looks good, I think. Still waiting for UX?

app/subsystems/course_membership/get_period.rb
nathanstitt: Yes, @rvnewell is working on different wording for the archive vs dropped state.

app/handlers/courses_enroll.rb
nathanstitt: hm, if they're taking the course again they'll be in a different section though, right?  This is only for the case where the student is using an enrollment code for a section that's been archived.  

In that case we cannot redirect them to the course since we did not add them to the section (it was archived).  If we were to do so then the FE would refuse to display it anyway and redirect to the course listing.

nathanstitt:
yes, waiting for UX.  They're now thinking we should display different pages for archived & dropped students that explain what's going on in more depth.

app/handlers/courses_enroll.rb
Dantemss: As per our IM discussion, I believe this change would affect students in an archived period attempting to re-take the course.

nathanstitt:
@Dantemss - I think the code is complete if you wanna take a look.  All that's left is to flesh out the views with whatever @rvnewell comes up with

app/controllers/courses_controller.rb
```diff
@@ -28,30 +23,48 @@ def confirm_enrollment
...
     course = @handler_result.outputs.course
-    redirect_to student_course_dashboard_path(course), webview_notice: notice
+    redirect_to student_course_dashboard_path(course.id), webview_notice: notice
   end
 
   def send_to_teacher_dashboard(notice: nil)
     course = @handler_result.outputs.course
-    redirect_to course_dashboard_path(course), notice: notice
+    redirect_to course_dashboard_path(course.id), notice: notice
   end
```
nathanstitt: Need to switch to `course.id` here because the course is a Hashie::Mash now

PR #1273, opened 2016-09-10
Multi enrollment fix

Dantemss:
Hotfix branch


Dantemss:
This is now in production so might as well merge.

PR #1275, opened 2016-09-12
Fix error on the views for course with no profile

BryanEliDimas:
Bug fix for `undefined method 'school' for nil:NilClass`.


Dantemss:
Hopefully we can eventually merge course profile and entity course.

PR #1274, opened 2016-09-12
Remove multipart exercises from dynamic pools

Dantemss:

PR #1276, opened 2016-09-12
Remove MPQ from consideration for spaced practice

nathanstitt:
Fixes bug where the HW for different students would have a different number of steps


Dantemss:
Might be a good idea to merge this and add the `allow_multipart: false` option to spaced practice/personalized/biglearn client/practice widget for double safety, but it definitely still needs the pool change due to Biglearn.

If Biglearn returns a multi-part and we exclude it, we are left with no assignable personalized exercises when we would have several other perfectly valid exercises.

Dantemss:
Superseded by https://github.com/openstax/tutor-server/pull/1274

PR #1277, opened 2016-09-13
Touch task when tasking_plans are updated, test concurrent locks

nathanstitt:
Adds a new `Tutor.fork_with_connection` helper method and uses it to test that concurrent updates deadlock.

### DIFF
--- a/lib/fork_with_connection.rb
+++ b/lib/fork_with_connection.rb
@@ -35,7 +35,7 @@ module Tutor
 
       ensure
         ActiveRecord::Base.remove_connection
-        Process.exit! exit_status.to_i
+        Process.exit! exit_status.is_a?(Fixnum) ? exit_status : 0
       end
     end
 
--- a/lib/fork_with_connection.rb
+++ b/lib/fork_with_connection.rb
@@ -12,8 +12,14 @@ module Tutor
 
       begin
         ActiveRecord::Base.establish_connection(ar_connection_config)
+        # settings are cached in Redis by rails-settings-cached
+        Rails.cache.reconnect
 
         Jobba.redis.client.reconnect
+
+        # reconnect fake client if Biglearn stubbing is enabled
+        OpenStax::Biglearn::V1.client.store.reconnect \
+          if OpenStax::Biglearn::V1.client.is_a? OpenStax::Biglearn::V1::FakeClient
 
         # This is needed to re-initialize the random number generator after forking (if you want diff random numbers generated in the forks)
         srand
--- a/lib/fork_with_connection.rb
+++ b/lib/fork_with_connection.rb
@@ -1,0 +1,42 @@
+# Logic for forking connections
+
+module Tutor
+  def self.fork_with_connection
+
+    # Store the ActiveRecord connection information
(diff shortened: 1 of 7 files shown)


lib/fork_with_connection.rb
```diff
@@ -0,0 +1,48 @@
+# Logic for forking connections
+
+module Tutor
+  def self.fork_with_connection
+
+    # Store the ActiveRecord connection information
+    ar_connection_config = ActiveRecord::Base.remove_connection
```
Dantemss: Is it OK for different processes to share the same connection config? I assume the connections are still different?

lib/fork_with_connection.rb
```diff
@@ -0,0 +1,48 @@
...
+        exit_status = yield
+
+      rescue => e
+        STDERR.puts "Forked operation failed with exception: #{e}"
+
+        # the op failed, so note it for the Process exit
+        exit_status = 1
+
+      ensure
+        ActiveRecord::Base.remove_connection
```
Dantemss: Probably not necessary, since `Process.exit!` below is supposed to close all active connections.

lib/fork_with_connection.rb
nathanstitt: the `ar_connection_config` is the hash that was used to establish the connection, it's something like: 
```ruby
{:adapter=>"postgresql", :username=>"ox_tutor", :password=>"XXX", :database=>"ox_tutor_test"}
```

lib/fork_with_connection.rb
nathanstitt: hm, yeah maybe.  I'm not sure if ActiveRecord has a hook to cleanly close the connection when the process exits though.

Dantemss: I believe the OS would be responsible for that, since it's a file descriptor. Still, doesn't hurt to close it I guess.

nathanstitt: Oh yeah, it'll definitely be closed.  I'm just not sure if it'll be closed **cleanly**. I'm pretty sure the client is supposed to send something to PG saying "I'm going away", that way the server won't wait until the socket times out.

spec/routines/distribute_tasks_spec.rb
```diff
@@ -21,6 +22,61 @@
...
+        end
+      end
+      failed = []
+      pids.each do |pid|
+        Process.wait(pid)
+        failed << pid if $?.exitstatus == 99
+      end
+      expect(failed).not_to be_empty
+    end
+  end
```
jpslav: is this spec guaranteed to pass?  (if so, how do we know that?)

nathanstitt: @jpslav it's not 100% guaranteed - there is very small change that the 5 requests may somehow all finish sequentially and not have the exception raised.  In my testing I never saw that, usually at least 3 had isolation exceptions.  

If we wanted to increase the chances, we could raise the fork count to 10 or something, but then the spec would run even slower.

Or maybe we just `xit` this and leave it for an example of testing concurrency ?

PR #1279, opened 2016-09-15
When locking task steps, lock the entire task before locking the step

Dantemss:
This is an attempt to prevent a deadlock that happened on Staging. I believe this is caused by `MarkTaskStepCompleted` locking the task step but modifying the entire task (the step touches the entire task and updates step counts). Workaround in that case would be to always lock the task before locking the task step.

### DIFF
--- a/app/routines/mark_task_step_completed.rb
+++ b/app/routines/mark_task_step_completed.rb
@@ -6,20 +6,18 @@ class MarkTaskStepCompleted
 
   def exec(task_step:, completion_time: Time.current)
     # Pessimistic locking to prevent race conditions with the update logic
+    task_step.task.lock!
     task_step.lock!
 
-    # Get the task after the reload so we have its updated state
-    task = task_step.task
-
-    task_step.complete(completion_time: completion_time)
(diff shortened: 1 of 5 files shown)


app/routines/mark_task_step_completed.rb
```diff
@@ -6,20 +6,18 @@ class MarkTaskStepCompleted
 
   def exec(task_step:, completion_time: Time.current)
     # Pessimistic locking to prevent race conditions with the update logic
+    task_step.task.lock!
     task_step.lock!
```
jpslav: Should we perhaps override `TaskStep#lock!` to always lock the task first?  Seems like if we want this here, we'd want it anywhere we lock a task step? (and I see in code below we repeat the task lock call) -- worried we'd forget to do this next time we lock a task step.

Dantemss: Hmmm.... that might be a good thing.

Dantemss: @jpslav done

PR #1281, opened 2016-09-19
Hide archived periods from CC enrollments

nathanstitt:
https://trello.com/c/9XeBgaGp/128-bug-cc-students-archived-students-re-enrolling-get-the-changing-section-message-maybe-on-tutor-also-maybe-dropped-also

If the period is shown to CC then it thinks it's a move and displays the wrong message.  

A different approach we could do is send the period, but flag it as "archived", then fix the FE to ignore it.  I don't see any reason for the FE to know anything about it though, so this just hides it.

### DIFF
--- a/app/subsystems/course_membership/create_enrollment_change.rb
+++ b/app/subsystems/course_membership/create_enrollment_change.rb
@@ -5,11 +5,9 @@ class CourseMembership::CreateEnrollmentChange
   uses_routine GetCourseEcosystem, as: :get_ecosystem
 
   def exec(user:, period:, book_uuid: nil, requires_enrollee_approval: true)
-    student_roles = run(:get_roles, user, 'student').outputs.roles.reject do | r |
-        r.student.period.deleted?
-    end
+    student_roles = run(:get_roles, user, 'student').outputs.roles.reject{ |r| r.student.period.deleted? }
 
-   course = period.course
+    course = period.course
 
     ecosystem = run(:get_ecosystem, course: course).outputs.ecosystem
     # Assumes 1 book in the ecosystem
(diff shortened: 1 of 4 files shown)


app/subsystems/course_membership/create_enrollment_change.rb
```diff
@@ -5,9 +5,11 @@ class CourseMembership::CreateEnrollmentChange
   uses_routine GetCourseEcosystem, as: :get_ecosystem
 
   def exec(user:, period:, book_uuid: nil, requires_enrollee_approval: true)
-    student_roles = run(:get_roles, user, 'student').outputs.roles
+    student_roles = run(:get_roles, user, 'student').outputs.roles.reject do | r |
+        r.student.period.deleted?
```
Dantemss: Indentation :)

Dantemss:
The representer change would be in case the instructor archives the period after the student creates the enrollment change, but before they confirm it. Corner case though I guess...

nathanstitt:
@Dantemss, checking for deletion in the representer's a good idea.  The check will need to be `if: ->(*) { !from_period.deleted? }` though instead of checking the `enrollment` since enrollment's won't be the thing deleted when a period is archived.

I've made the change

Dantemss:
The enrollment gets deleted when the period is archived via the `dependent: :destroy` but yeah details... checking the period works too.

spec/representers/api/v1/enrollment_change_representer_spec.rb
```diff
@@ -37,4 +37,12 @@
           'last_name'  => teacher_role.last_name
         }])
   end
+
+  it 'hides deleted from periods' do
```
jpslav: I'd like to have some of the reasoning why we are hiding deleted periods, like what you have in the PR description so that later we know why we are doing this.  Can you put the reasoning at least here in a comment or expand the example name to indicate why?

nathanstitt: sure, that makes sense

PR #1282, opened 2016-09-19
Only query "new" opportunities

Dantemss:
According to Denver, this is needed to fix the current Salesforce reporting errors.

### DIFF
--- a/app/subsystems/salesforce/renew_os_ancillary.rb
+++ b/app/subsystems/salesforce/renew_os_ancillary.rb
@@ -11,7 +11,8 @@ module Salesforce
       target_opportunity_criteria = {
         contact_id: based_on_opportunity.contact_id,
         book_name: based_on_opportunity.book_name,
-        term_year: renew_for_term_year.to_s
+        term_year: renew_for_term_year.to_s,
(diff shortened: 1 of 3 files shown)


app/subsystems/salesforce/renew_os_ancillary.rb
```diff
@@ -11,7 +11,8 @@ def self.call(based_on:, renew_for_term_year:)
       target_opportunity_criteria = {
         contact_id: based_on_opportunity.contact_id,
         book_name: based_on_opportunity.book_name,
-        term_year: renew_for_term_year.to_s
+        term_year: renew_for_term_year.to_s,
+        new: true
```
jpslav: Is `new` a new field in SF OS Ancillaries?  If so, I think it needs to be included as a `field` in the `OsAncillary` class so that it can be used for queries.

Dantemss: Done.

jpslav: For posterity, I should have said "Opportunity" instead of "OS Ancillary"

PR #1284, opened 2016-09-21
Add the ability to filter research data export

BryanEliDimas:
## Before:
screen shot 2016-09-21 at 10 35 08 am

## After:
screen shot 2016-09-21 at 10 35 12 am

With datepicker
screen shot 2016-09-21 at 10 37 08 am

### DIFF
--- a/app/controllers/admin/research_data_controller.rb
+++ b/app/controllers/admin/research_data_controller.rb
@@ -6,7 +6,10 @@ module Admin
 
     def create
       filename = "export_#{Time.now.utc.strftime("%Y%m%dT%H%M%SZ")}.csv"
-      ExportAndUploadResearchData.perform_later(filename)
+      from_date = Chronic.parse(params[:from])
+      to_date = Chronic.parse(params[:to])
+      date_range = from_date..to_date if from_date && to_date
+      ExportAndUploadResearchData.perform_later(filename: filename, date_range: date_range, include_tutor: params[:tutor], include_concept_coach: params[:concept_coach])
       redirect_to admin_research_data_path,
                   notice: "#{ExportAndUploadResearchData::RESEARCH_FOLDER}/#{filename} should be available in a few minutes in ownCloud (does not refresh automatically)"
     end
--- a/app/controllers/admin/research_data_controller.rb
+++ b/app/controllers/admin/research_data_controller.rb
@@ -6,10 +6,27 @@ module Admin
 
     def create
       filename = "export_#{Time.now.utc.strftime("%Y%m%dT%H%M%SZ")}.csv"
-      from_date = Chronic.parse(params[:from])
-      to_date = Chronic.parse(params[:to])
-      date_range = from_date..to_date if from_date && to_date
-      ExportAndUploadResearchData.perform_later(filename: filename, date_range: date_range, include_tutor: params[:tutor], include_concept_coach: params[:concept_coach])
+      mapping = {
+        "tutor" => Tasks::Models::Task.task_types.values_at(:homework, :reading, :chapter_practice,
+                   :page_practice, :mixed_practice, :external, :event, :extra),
+        "concept_coach" => Tasks::Models::Task.task_types.values_at(:concept_coach)
+      }
+
+      task_types_params = params.fetch(:export_research_data).fetch(:task_types).reject{|i|i.blank?}
+
+      if task_types_params.blank?
+        flash.now[:alert] = "You must select at least one type of Tasks"
+        render :index
+        return
+      end
+
+      task_types = task_types_params.inject([]){ |result, task_types_param|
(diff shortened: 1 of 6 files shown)


BryanEliDimas:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;O0DuVvuZ&#x2F;126-add-the-ability-to-filter-research-data-by-date-range-and-by-concept-coach-only"> Add the ability to filter research data by date range and by Concept Coach only</a>

app/routines/export_and_upload_research_data.rb
```diff
@@ -135,7 +141,7 @@ def is_cc?(course_id)
       CourseProfile::Models::Profile
         .select([:entity_course_id, :is_concept_coach])
         .each_with_object({}) do |profile, hsh|
-          hsh[profile.entity_course_id] = profile.is_concept_coach
+          hsh[profile.entity_course_id] = profile.is_concept_coach.try(:capitalize)
```
BryanEliDimas: Per Grimmaldi's request.

app/routines/export_and_upload_research_data.rb
```diff
@@ -6,10 +6,12 @@ class ExportAndUploadResearchData
 
   lev_routine express_output: :filename
 
-  def exec(filename = nil)
+  def exec(filename: nil, include_concept_coach: false, include_tutor: false, date_range: nil)
     outputs[:filename] = FilenameSanitizer.sanitize(filename) || \
                          "export_#{Time.now.utc.strftime("%Y%m%dT%H%M%SZ")}.csv"
-    create_export_file
+    only_tutor = include_tutor && !include_concept_coach
+    only_concept_coach = include_concept_coach && !include_tutor
```
jpslav: This means we can only do one of tutor or CC, but the checkboxes on the form imply that we can do both.  Can you change to make it less "only"? (only use the `where` clause if one of them is set)

spec/routines/export_and_upload_research_data_spec.rb
```diff
@@ -101,5 +101,64 @@
...
+        expect_any_instance_of(described_class).to receive(:upload_export_file) do |routine|
+          filepath = routine.send :filepath
+          expect(File.exists?(filepath)).to be true
+          expect(filepath.ends_with? '.csv').to be true
+
+          rows = CSV.read(filepath)
+
+          expect(Tasks::Models::TaskStep.count).to eq 3
+          expect(rows.count - 1).to eq(1)
+        end
```
jpslav: To reduce the duplicated test code, can you add a helper method that does the common CSV checks, e.g.:

```ruby
def with_export_rows(&block)
  expect_any_instance_of(described_class).to receive(:upload_export_file) do |routine|
    filepath = routine.send :filepath
    expect(File.exists?(filepath)).to be true
    expect(filepath.ends_with? '.csv').to be true
    rows = CSV.read(filepath)
    block.call(rows)
  end
end
```

and also make the existing spec at the top use the new helper?

jpslav: actually you could include the `capture_stdout` line too, if you pass in the options to `call` as the first argument of `with_export_rows`

app/controllers/admin/research_data_controller.rb
```diff
@@ -6,7 +6,10 @@ def index
 
     def create
       filename = "export_#{Time.now.utc.strftime("%Y%m%dT%H%M%SZ")}.csv"
-      ExportAndUploadResearchData.perform_later(filename)
+      from_date = Chronic.parse(params[:from])
+      to_date = Chronic.parse(params[:to])
```
BryanEliDimas: Isn't there already a default? The default is to include all because I only pull records within a date range _if_ they provide a date range (`params[:from]` and `params[:to]`. https://github.com/openstax/tutor-server/pull/1284/files#diff-a670158333c81f17a6559527e1d27ea4R55

jpslav: Right but as is it doesn't allow for just one of the times be blank (to do `> some time` without a `< some other time`).

BryanEliDimas: That is true. Dante also gave me feedback on this PR and I'm cleaning it up.

BryanEliDimas:
@jpslav I made the changes you suggested. It sorta turned into a complete refactoring.
It should be ready to merge now unless there is a scenario I have not covered in the tests.

app/views/manager/courses/_index.html.erb
```diff
@@ -5,6 +5,8 @@
 <% failed_jobs ||= [] %>
 <% total_count ||= course_infos.try(:count) %>
 
+<div>Showing <%= total_count %> total</div>
```
BryanEliDimas: Dante's request. Absolutely harmless so I just included it in this PR.

PR #1287, opened 2016-09-22
Dashboard counts again...

Dantemss:
- Added a failing spec for task counts being out of sync


Dantemss:
This time I made the spec before the fix so hopefully you can see it fail, then be fixed...

PR #1288, opened 2016-09-23
warn about not creating duplicate accounts when student id is taken

nathanstitt:
Updates the error message to warn students not to create a new account.
screen shot 2016-09-27 at 11 20 04 am

### DIFF
--- a/app/assets/stylesheets/enroll.scss
+++ b/app/assets/stylesheets/enroll.scss
@@ -55,7 +55,10 @@
       }
     }
   }
-
+  .enrollment-flash {
+    max-width: 800px;
+    margin: 20px auto;
+  }
   span.help {
     display: block;
     font-size: 12px;
--- a/app/assets/stylesheets/enroll.scss
+++ b/app/assets/stylesheets/enroll.scss
@@ -1,4 +1,9 @@
+// the enrollment page displays the flash itself
+.courses-enroll .application-flash { display: none; }
+
 #enroll {
+  margin-top: 20px;
+
   &.well {
     margin-bottom: 0px;
   }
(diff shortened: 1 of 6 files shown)


Dantemss:
@Fredasaurus I think we have to mention they have to Logout and try to reset the password or something
If they reach this point, they already created a duplicate account.

Fredasaurus:
The alert and page messaging, "You are joining...." (you are not) are still in conflict with each other. I'm not sure if the plan is to change that now though.

The alert doesn't relate well in terms of proximity. Should it be inline with the actual location of the error? Also the terminology doesn't relate well. "Skip this step" does not imply that you are creating a new account.

Dantemss:
The reason "Skip this step" doesn't imply they are creating a new account is that they are already in the new account at this point. They already crossed the new account creation bridge. They need to logout and backtrack.

Fredasaurus:
That's even more confusing. I don't think we can reasonably expect a user to know this.

Dantemss:
Hence why I suggested to ask them to logout and try to login into their previous account.

Fredasaurus:
Yeah, I feel the appropriate course of action is to review you account info.

nathanstitt:
@kathi-fletcher - your comment on the Trello card indicated you wanted the wording shown in the screenshot, but @Fredasaurus has a good point about maybe we should just instruct them to review their account info.

Should I change the wording?

nathanstitt:
@jpslav & @Dantemss this is updated with the new wording and is ready for review & merge

Fredasaurus:
Not perfect, but this is an improvement over existing.

PR #1289, opened 2016-09-23
Script for helping to find dupe usage

jpslav:


lib/scripts/find_dupes.rb
```diff
@@ -0,0 +1,42 @@
...
+                 ee.course_membership_period_id,
+                 completed_steps_count,
+                 latest_worked_at,
+                 ee.deleted? ? "Deleted" : "Active",
+                 ee.created_at ]
+
+      end
+    end
+
+    CSV.open('/tmp/teacher_accounts.csv','w+') do |csv|
```
Dantemss: I meant the Rails tmp folder, but this is the system tmp folder. Is that OK with you?

PR #1290, opened 2016-09-28
Record user tours views

nathanstitt:
Saves viewing statistics for users viewing tours.

Is pretty bare-bones but should give us a a place to start from.

The "Tour" record is auto-created so the FE can supply an arbitrary id like "teacher-dashboard"


app/subsystems/user/models/tour_view.rb
```diff
@@ -0,0 +1,11 @@
+module User
+  module Models
+    class TourView < Tutor::SubSystems::BaseModel
+
+      belongs_to :tour
```
Dantemss: Could add `inverse_of: :tour_views` here after defining it in `User::Models::Tour`.

app/subsystems/user/models/tour_view.rb
```diff
@@ -0,0 +1,11 @@
...
+
+      belongs_to :profile, -> { with_deleted }, inverse_of: :tour_views
+
+    end
```
Dantemss: Could add presence validations for tour and profile and a uniqueness validation for profile scoped to `:user_tour_id`.

app/subsystems/user/record_tour_view.rb
```diff
@@ -0,0 +1,26 @@
+module User
+
+
+  class RecordTourView
+    lev_routine express_output: :view
+
+    protected
+
+    def exec(user:, tour_identifier: )
```
Dantemss: 2 find_or_create race conditions here but not sure if we can do much until postgres 9.5...

app/subsystems/user/models/tour_view.rb
Dantemss: And validate presence of `view_count`.

db/migrate/20160923180438_create_user_tours.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def change
+    create_table :user_tours do |t|
+      t.text :identifier, null: false, index: { unique: true }
+      t.timestamps null: false
+    end
+
+    create_table :user_tour_views do |t|
+      t.integer :view_count, null: false, default: 0
+      t.belongs_to :user_profile, foreign_key: true, null: false, index: { unique: false }
```
Dantemss: You probably want to get rid of this index and add a unique index on `[:user_profile_id, :user_tour_id]`

db/migrate/20160923180438_create_user_tours.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.belongs_to :user_tour,    foreign_key: true, null: false, index: { unique: false }
```
Dantemss: I think you can just say `index: true` here

spec/factories/tours.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :tour do
+    
+  end
+
+end
```
Dantemss: I think this empty factory can be removed (seems like you moved the model into the User subsystem).

nathanstitt:
Good points all around @Dantemss 😁

PR #1291, opened 2016-09-29
Add ability to export excluded exercise stats dashboard to csv

BryanEliDimas:
**_Before:_**
before ee stats

**_After:_**
after ee stats

### DIFF
--- a/app/routines/get_excluded_exercises.rb
+++ b/app/routines/get_excluded_exercises.rb
@@ -1,6 +1,6 @@
 class GetExcludedExercises
   owncloud_secrets = Rails.application.secrets['owncloud']
-  EE_STATS_FOLDER = owncloud_secrets['excluded_exercises_stats'] || "excluded_exercises_stats"
+  EE_STATS_FOLDER = owncloud_secrets['excluded_exercises_stats_folder'] || "excluded_exercises_stats"
   WEBDAV_BASE_URL = "#{owncloud_secrets['base_url']}/remote.php/webdav/#{EE_STATS_FOLDER}"
 
   lev_routine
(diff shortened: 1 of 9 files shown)


BryanEliDimas:
[Add the ability to export Excluded Exercise Stats dashboard to CSV](https://trello.com/c/x2CA4qPl/177-add-the-ability-to-export-excluded-exercise-stats-dashboard-to-csv)

BryanEliDimas:
@Dantemss Please review.

app/routines/get_excluded_exercises.rb
```diff
@@ -0,0 +1,243 @@
+class GetExcludedExercises
+  owncloud_secrets = Rails.application.secrets['owncloud']
+  EE_STATS_FOLDER = owncloud_secrets['excluded_exercises_stats'] || "excluded_exercises_stats"
```
BryanEliDimas: We have to create this `secrets` variable, pointing a folder in ownCloud which we have to create.

spec/routines/get_excluded_exercises_spec.rb
```diff
@@ -0,0 +1,203 @@
+require 'rails_helper'
+require 'webmock/rspec'
+
+describe GetExcludedExercises, type: :routine do
+  before(:each) do
+    WebMock.disable_net_connect!
+    stub_request(:put, /remote.php/).to_return(status: 200)
+  end
+
+  context "with data" do
```
Dantemss: You could make this spec run faster by putting all this stuff (except `outputs`) in a `before(:all)` block so it's done only once. You would have to use `@instance_variables` though...

app/routines/get_excluded_exercises.rb
```diff
@@ -0,0 +1,245 @@
...
+    end
+
+    def webdav_url(fname)
+      Addressable::URI.escape "#{WEBDAV_BASE_URL}/#{fname}"
+    end
+
+    def remove_exported_files
+      File.delete(filepath_by_course) if File.exists?(filepath_by_course)
+      File.delete(filepath_by_exercise) if File.exists?(filepath_by_exercise)
+    end
```
Dantemss: I usually refrain from indenting protected stuff because it messes up the indentation at the end (and plus `protected` is a method in `BasicObject`, not a code block like `begin`).

But I suppose you will find people who prefer it this way...

BryanEliDimas:
@Dantemss I made the changes you suggested.

PR #1294, opened 2016-10-05
Fixes and improvements to admin dashboard

BryanEliDimas:
Changes requested by Kajal and Greg (plus the last one by Dante)
- Change default `order`ing by `id`!
- Allow changing the order `on click`
- Fix `course`s "card" styling on smaller than huge screens
- Fix the "ability to change `results per page`" feature. Weird bug: if there was a search query in the search form, and then you changed the results per page, all except letter characters got escaped essentially breaking the search form.

### DIFF
--- a/app/assets/javascripts/manager.js
+++ b/app/assets/javascripts/manager.js
@@ -70,9 +70,8 @@ $(document).ready(function() {
 
     var setPerPage = function(val){
       var paramsObj = locationSearchInJSON();
-      paramsObj.per_page = val;
-
-      window.location.search = $.param(paramsObj);
+      paramsObj.per_page = Number(val);
+      window.location.search = decodeURIComponent($.param(paramsObj));
     }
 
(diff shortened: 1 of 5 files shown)


Dantemss:
The controller explodes if SearchCourses returns an error, for example, when the search string is invalid, like: http://localhost:3001/admin/courses?query=%22invalid;string%22

Would be nice to return an error message instead (you can use flash[:error] or flash.alert).

BryanEliDimas:
@Dantemss It explodes meaning it returns 0 results?  
screen shot 2016-10-06 at 1 54 14 pm

BryanEliDimas:
@Dantemss I made the change you suggested, in the other PR I was working at the same time. https://github.com/openstax/tutor-server/pull/1291/commits/16176739ee0cba7f3e8148be70eb4fa83723bd54#diff-2e3655641532311d8e8cc6f66d9803ffR14

PR #1296, opened 2016-10-06
Spike course start and end dates

Dantemss:
- Add start and end dates to courses and make them required
- End must come after start
- Course profiles have an `active?` method to determine if they are active


Dantemss:
MOST of this PR deals with old specs that assumed a certain signature for the `CreateCourse` routine. I changed all of them to use the `entity_course` factory instead so I could change `CreateCourse` to make the start and end dates mandatory.

app/routines/get_excluded_exercises.rb
```diff
@@ -193,11 +193,11 @@ def owncloud_csv_by_exercise
     end
 
     def filename_by_course
-      "excluded_exercises_stats_by_course_#{Time.now.utc.strftime("%Y%m%dT%H%M%SZ")}.csv"
+      "excluded_exercises_stats_by_course_#{Time.current.utc.strftime("%Y%m%dT%H%M%SZ")}.csv"
     end
 
     def filename_by_exercise
-      "excluded_exercises_stats_by_exercise_#{Time.now.utc.strftime("%Y%m%dT%H%M%SZ")}.csv"
+      "excluded_exercises_stats_by_exercise_#{Time.current.utc.strftime("%Y%m%dT%H%M%SZ")}.csv"
```
BryanEliDimas: `Time.current` is like calling Time.now but with the time zone (`Time.zone.now`), correct? Doesn't it defeat the point to call `#.utc` on that? Feel free to not make this modification in your PR because I am fixing (extracting out) these two lines for getting the current time anyway.

Dantemss: You are right... Even `Time.current` is useless I suppose, since the server's timezone is not what we usually want.

Dantemss:
Closing in favor of https://github.com/openstax/tutor-server/pull/1301

PR #1297, opened 2016-10-07
Create rake task to export CSV with info to match with consent forms

BryanEliDimas:
Phil Grimaldi's two-part request. The other part of his request is [here](https://github.com/openstax/accounts/pull/354) in Accounts.

### DIFF
--- a/app/routines/export_users_info_to_match_with_consent_forms.rb
+++ b/app/routines/export_users_info_to_match_with_consent_forms.rb
@@ -1,0 +1,55 @@
+class ExportUsersInfoToMatchWithConsentForms
+  lev_routine express_output: :filename
+
+  def exec(delete_after: true)
+    outputs[:filename] = filename
+    outputs[:info] = info
+    generate_csv
+    remove_exported_files if delete_after
+  end
+
+  def info
+    output_users = []
+    User::Models::Profile.with_deleted.joins(:account).find_each do |user|
+      output_users << Hashie::Mash.new({
+        user_id: user.account.openstax_uid,
+        student_identifiers: user.roles.map(&:student).compact.map(&:student_identifier),
+        name: user.name,
+        username: user.username
+      })
+    end
+    output_users
+  end
+
+  def generate_csv
+    CSV.open(filename, 'w') do |file|
+      file.add_row ([
+        "User ID",
+        "Student Identifiers",
+        "Name",
+        "Username"
+      ])
+
+      info.each do |hashie|
+        file.add_row([
+          hashie.user_id,
+          hashie.student_identifiers.join(", "),
+          hashie.name,
+          hashie.username
+        ])
+      end
+    end
+  end
+
+  def filename
+    File.join exports_folder, "users_info_to_match_with_consent_forms.csv"
+  end
+
+  def exports_folder
+    File.join 'tmp', 'exports'
+  end
+
+  def remove_exported_files
+    File.delete(filename) if File.exist?(filename)
+  end
+end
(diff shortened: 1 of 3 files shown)


BryanEliDimas:
[Create CSV exports for Dr. Grimaldi to match against signed consent forms](https://trello.com/c/dfMfkdhf/187-create-csv-exports-for-dr-grimaldi-to-match-against-signed-consent-forms)

app/routines/export_users_info_to_match_with_consent_forms.rb
```diff
@@ -0,0 +1,54 @@
...
+    File.join exports_folder, "users_info_to_match_with_consent_forms.csv"
+  end
+
+  def exports_folder
+    File.join 'tmp', 'exports'
+  end
+
+  def remove_exported_files
+      File.delete(filename) if File.exist?(filename)
+    end
```
Dantemss: Indentation

app/routines/export_users_info_to_match_with_consent_forms.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      info.each do |hashie|
+        file.add_row([
+          hashie.user_id,
+          hashie.student_identifiers.join(", "),
+          hashie.name,
+          hashie.username
+        ])
+      end
+    end
```
Dantemss: I suggest printing the export file location at the end so we know where to find it.

spec/routines/export_users_info_to_match_with_consent_forms_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    let!(:user_1){ FactoryGirl.create :user, username: "TonyStark", first_name: "Tony", last_name: "Stark" }
+    let!(:some_other_user){ FactoryGirl.create :user }
+    let!(:period){ FactoryGirl.create :course_membership_period, course: course }
+
+    before(:each) do
+      role_1 = AddUserAsPeriodStudent[period: period, user: user_1, student_identifier: "333999"]
+      @student_1 = role_1.student
+    end
+
+    let!(:outputs){ described_class.call(as_csv: false).outputs }
```
Dantemss: Seems weird to me that this switch exists just for this test? I suggest removing `as_csv` and just letting it remove the CSV file after the test is done.

spec/routines/export_users_info_to_match_with_consent_forms_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    it "includes name" do
+      expect(first_output.name).to eq "Tony Stark"
+    end
+
+    it "includes username" do
+      expect(first_output.username).to eq "TonyStark"
+    end
+
+    context "as csv file with student information" do
+      it "includes includes all the information it should" do
```
Dantemss: includes includes

spec/routines/export_users_info_to_match_with_consent_forms_spec.rb
BryanEliDimas: You're right. Much better. Generating a CSV is the sole responsibility of this routine anyway.

spec/routines/export_users_info_to_match_with_consent_forms_spec.rb
BryanEliDimas: Fixed.

app/routines/export_users_info_to_match_with_consent_forms.rb
BryanEliDimas: I didn't `puts` inside this method because I'd have to `capture_stdout{ ... }` in the specs and I didn't like it.
Instead, I added an `express_output: :filename` and then I `puts ExportUsersInfoToMatchWithConsentForms[...]` in the rake file.

Dantemss:
Don't think we ever delete users but I guess it doesn't hurt?

PR #1298, opened 2016-10-07
Replace deprecated method File.exists?

BryanEliDimas:
Replaced with `File.exist?` @Dantemss

### DIFF
--- a/app/routines/get_excluded_exercises.rb
+++ b/app/routines/get_excluded_exercises.rb
@@ -239,7 +239,7 @@ class GetExcludedExercises
     end
 
     def remove_exported_files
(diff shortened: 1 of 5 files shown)


Dantemss:
Could you take a look at the spec failure? I think the specs are stepping on each other and deleting the file too soon?

BryanEliDimas:
@Dantemss Fixed the spec failure by stubbing out the call to the routine `GetExcludedExercises` in [stats_controller_spec](https://github.com/openstax/tutor-server/pull/1298/files#diff-b84865e52490a1679e002a400d36e042R92).

PR #1300, opened 2016-10-10
Fix: Show all courses in admin courses controller

BryanEliDimas:
There was a bug in [manager.js](https://github.com/openstax/tutor-server/compare/fix_infinity_on_admin_courses_index?expand=1#diff-93b8e3f1c5f6bf274f1b956aad113aa1R100): it expected a `Number`, so when the `per_page` param was set to `"all"` the page broke. This PR fixes that.

I also added more test coverage.

### DIFF
--- a/app/assets/javascripts/manager.js
+++ b/app/assets/javascripts/manager.js
@@ -97,7 +97,7 @@ $(document).ready(function() {
 
     var setPerPage = function(val){
       var paramsObj = locationSearchInJSON();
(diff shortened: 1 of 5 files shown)


BryanEliDimas:
[BUG: Showing all courses in admin courses dashboard is broken](https://trello.com/c/g8IBnqmd/191-bug-showing-all-courses-in-admin-courses-dashboard-is-broken)

PR #1301, opened 2016-10-12
Spike new and clone course with dates

Dantemss:
# Part 1
- Self-service new course and clone course
- Courses have required start and end dates and become inactive once those pass, blocking new teacher assignments and practice
- Specs

# Part 2
- Integration with accounts-rails faculty verification
- Demo script sets teachers as verified
- is_available flag in catalog offerings
- API to return all available catalog offerings
- Option to copy QL
- Auto-creating a certain number of sections
- Api to return uncloned task plans from original course (`GET  /api/courses/:cloned_course_id?clone_status=unused_source`) -- suggestions accepted
- More specs


config/routes.rb
```diff
@@ -80,13 +80,12 @@
...
 
-    resources :courses, only: [:show, :update] do
+    resources :courses, only: [:create, :show, :update] do
       member do
         get 'dashboard(/role/:role_id)', action: :dashboard
         get 'cc/dashboard(/role/:role_id)', action: :cc_dashboard
-        get 'plans'
-        get 'tasks'
         get 'roster'
+        post 'clone'
```
nathanstitt: We'd discussed adding an endpoint for the FE to load the course cloning setup info.  The FE somehow needs to know: the list of `course_offering_id`: `course name`  pairs, as well as the year and terms that should be offered as options.  The year & terms we can probably auto-calculate to match the BE though, but if we're already making a request we might as well include them.

app/routines/clone_course.rb
```diff
@@ -0,0 +1,35 @@
...
+      time_zone: course.profile.time_zone
+    }.merge(attributes)
+
+    run(:create_course, **attrs)
+
+    run(:add_teacher, course: outputs.course, user: teacher_user)
+
+  end
+
+end
```
Dantemss: Sounds reasonable

config/routes.rb
Dantemss: What do you mean by `course_offering_id: course name` pairs? Is that the default course name?

nathanstitt: Yes, what I'd meant here is the default course name.   I've further discovered that we'll also need a bit more info, which I'd put in the trello card as:

> BE -- add endpoint for retrieving a lightweight (only id, title & type) listing of all **uncloned** past assignments for a given course.  The FE can pass in the current course id and BE can return all unused assignments for other course(s) with the same offering id.  When the FE clones an assignment, it will send the id of past assignment so it can then be marked used

app/controllers/api/v1/courses_controller.rb
```diff
@@ -120,15 +138,54 @@ def cc_dashboard
...
+    They will default to the given course's attributes if ommitted
+    #{json_schema(Api::V1::CourseRepresenter, include: :writeable)}
+  EOS
+  def clone
+    OSU::AccessPolicy.require_action_allowed!(:clone, current_api_user, @course)
+
+    attributes = consumed(Api::V1::CourseCloneRepresenter)
+                   .merge(course: @course, teacher_user: current_human_user)
+
+    course = CloneCourse[attributes]
```
jpslav: Can we have named args instead of a hash?  Hard to know if things are missing / what's in the attrs.

app/controllers/admin/courses_controller.rb
```diff
@@ -31,14 +31,21 @@ def index
...
   def create
     handle_with(Admin::CoursesCreate,
-                complete: ->(*) {
-                  flash[:notice] = 'The course has been created.'
+                success: ->(*) {
+                  flash.notice = 'The course has been created.'
                   redirect_to admin_courses_path
+                },
+                failure: ->(*) {
+                  flash[:error] = @handler_result.errors.full_messages
```
Dantemss: Sure

db/migrate/20161028232036_drop_entity_courses.rb
```diff
@@ -0,0 +1,72 @@
...
+    remove_foreign_key :course_membership_students, :entity_courses
+    remove_foreign_key :course_membership_teachers, :entity_courses
+    remove_foreign_key :course_profile_courses, :entity_courses
+    remove_foreign_key :tasks_course_assistants, :entity_courses
+    remove_foreign_key :tasks_performance_report_exports, :entity_courses
+
+    # Keep the old course IDs so we don't have to
+    # change all the columns that used to point to entity_course_id
+    CourseProfile::Models::Course.update_all('id = -id')
+    CourseProfile::Models::Course.update_all('id = entity_course_id')
```
Dantemss: I couldn't change it directly due to the uniqueness constraint

app/controllers/api/v1/courses_controller.rb
Dantemss: Hm I'll think about this...

lib/term_year.rb
```diff
@@ -0,0 +1,53 @@
+TermYear = Struct.new(:term, :year) do
+  LEGACY_TERM_STARTS_AT = DateTime.parse('July 1st, 2015')
+  LEGACY_TERM_ENDS_AT   = DateTime.parse('July 1st, 2017')
```
Dantemss: Hm?

lib/term_year.rb
```diff
@@ -0,0 +1,53 @@
...
+    legacy: ->(year) { LEGACY_TERM_STARTS_AT },
+    demo:   ->(year) { DateTime.new(year   ) },  # January 1st of the year
+    spring: ->(year) { DateTime.new(year   ) },  # January 1st of the year
+    summer: ->(year) { DateTime.new(year, 5) },  # May 1st of the year
+    fall:   ->(year) { DateTime.new(year, 7) }   # July 1st of the year
+  }
+
+  TERM_END_DATES = {
+    legacy: ->(year) { LEGACY_TERM_ENDS_AT    },
+    demo:   ->(year) { DateTime.new(year + 1) }, # January 1st of next year
```
Dantemss: You shouldn't be demoing during Christmas :) but fine

db/migrate/20161028232036_drop_entity_courses.rb
```diff
@@ -0,0 +1,72 @@
...
+              name: 'index_t_performance_report_exports_on_c_p_course_id'
+
+    # Change other columns that reference courses in a polymorphic way
+    Tasks::Models::TaskPlan.where(owner_type: 'Entity::Course')
+                           .update_all(owner_type: 'CourseProfile::Models::Course')
+    Tasks::Models::TaskingPlan.where(target_type: 'Entity::Course')
+                              .update_all(target_type: 'CourseProfile::Models::Course')
+    Salesforce::Models::AttachedRecord.where{tutor_gid.like '%Entity::Course%'}.update_all(
+      "tutor_gid = replace(tutor_gid, 'Entity::Course', 'CourseProfile::Models::Course')"
+    )
```
Dantemss: Argh! and... no idea.

app/controllers/admin/courses_controller.rb
Dantemss: Actually `flash.error` is not a real method in flash, I think it's something we invented.

app/controllers/api/v1/courses_controller.rb
Dantemss: Do you mean changing it here or in the routine itself?

PR #1302, opened 2016-10-17
Spike Trial Course Content

Dantemss:
- Added teacher_student_role_id to periods and made it autopopulate and mandatory
- The teacher_student_role receives assignments but does not show up in task QL, performance report and performance forecast.
- teacher_student_role_id shows up in the PeriodRepresenter
- Student deidentifier moved to Entity::Role and renamed research_identifier
- Added a way to differentiate demo courses/assignments/students from real ones (is_trial in the course record)
- is_trial is set to true during course creation through the API
- Added a way to transform a demo course into a real one (cloning)
- Create routine that populates demo content and call it from CreateCourse
- Specs


app/routines/import_salesforce_course.rb
```diff
@@ -62,6 +62,7 @@ def exec(candidate:, log_prefix: nil)
       starts_at: candidate.try(:course_start_date),
       school: school,
       catalog_offering: offering,
+      is_trial: true,
```
jpslav: if we want this disabled for now, should this be `false`?  While I think the goal is to not approve more courses from SF, it could still happen and I don't think we want the trial to be what they see.

app/routines/individualize_tasking_plans.rb
```diff
@@ -20,15 +20,18 @@ def exec(task_plan)
         user = ::User::User.new(strategy: strategy)
         Role::GetDefaultUserRole[user]
       when CourseProfile::Models::Course
-        CourseMembership::GetCourseRoles.call(course: target, types: :student).outputs.roles
+        CourseMembership::GetCourseRoles.call(course: target, types: :student).outputs.roles +
+        target.periods.map(&:teacher_student_role)
```
jpslav: `CourseMembership::GetCourseRoles` already takes a `types` array.  Can that array arg be `[:student, :teacher_student]`?

app/routines/populate_trial_course_content.rb
```diff
@@ -0,0 +1,199 @@
+class PopulateTrialCourseContent
+
+  STUDENT_INFO = [
+    { username: 'trialstudent1', title: 'Trial', first_name: 'Student', last_name: 'One'   },
+    { username: 'trialstudent2', title: 'Trial', first_name: 'Student', last_name: 'Two'   },
+    { username: 'trialstudent3', title: 'Trial', first_name: 'Student', last_name: 'Three' },
+    { username: 'trialstudent4', title: 'Trial', first_name: 'Student', last_name: 'Four'  },
+    { username: 'trialstudent5', title: 'Trial', first_name: 'Student', last_name: 'Five'  },
+    { username: 'trialstudent6', title: 'Trial', first_name: 'Student', last_name: 'Six'   }
```
jpslav: I think `title` is not required so I'd leave it out unless there's a specific reason you know of to keep it.

app/routines/populate_trial_course_content.rb
```diff
@@ -0,0 +1,199 @@
...
+        OpenStax::Accounts::Account.find_or_create_by(
+          username: student_info[:username]
+        ) do |acc|
+          acc.title = student_info[:title]
+          acc.first_name = student_info[:first_name]
+          acc.last_name = student_info[:last_name]
+        end.tap do |acc|
+          raise "Someone took the trial username #{acc.username}!" if acc.valid_openstax_uid?
+        end
+      end
```
jpslav: where do we say "this is only a local user"?

app/routines/populate_trial_course_content.rb
jpslav: Is this going to have a problem when it is run the second time and we get username collisions?

app/routines/populate_trial_course_content.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    book = ecosystem.books.first
+    return if book.nil?
+
+    # Use only chapters that have some homework exercises
+    candidate_chapters = book.chapters.select do |chapter|
+      chapter.pages.any?{ |page| page.homework_core_pool.content_exercise_ids.any? }
+    end
+    trial_chapters = candidate_chapters[0..NUM_CHAPTERS-1]
+    return if trial_chapters.blank?
```
jpslav: Am I reading this correctly that we are not saying "for Physics, use Chs. 3, 17, 18, 24"?  (ok if that's how it is now, just want to check)

app/subsystems/course_membership/models/period.rb
```diff
@@ -50,4 +56,10 @@ def enrollment_code_for_url
     enrollment_code.gsub(/ /,'-')
   end
 
+  protected
+
+  def build_teacher_student_role
+    self.teacher_student_role ||= Entity::Role.new(role_type: :teacher_student)
```
jpslav: does get saved, right?

db/migrate/20161101220241_add_is_demo_to_courses.rb
```diff
@@ -0,0 +1,7 @@
+class AddIsDemoToCourses < ActiveRecord::Migration
+  def change
+    add_column :course_profile_courses, :is_trial, :boolean
```
jpslav: default needed?

app/routines/import_salesforce_course.rb
Dantemss: Yes you are right... I forgot that both SF and the instructor could make new courses and disabled it only for instructor-created courses

app/routines/individualize_tasking_plans.rb
Dantemss: Hmm sure I guess I can change `GetCourseRoles` to tell it how to look for teacher_student_roles.

app/routines/populate_trial_course_content.rb
Dantemss: OK I can remove trial...
It does a find_or_create so the second time it will reuse the student accounts.

app/routines/populate_trial_course_content.rb
Dantemss: We don't want anyone claiming those accounts

app/routines/populate_trial_course_content.rb
Dantemss: Hah was that part of the requirements? Maybe I missed it...

app/subsystems/course_membership/models/period.rb
Dantemss: Yeah it gets saved when the period is saved

db/migrate/20161101220241_add_is_demo_to_courses.rb
Dantemss: Actually I figured out recently that change_column_null accepts a default value and will change existing nulls to that value (in this case false) without setting the column's actual default value.

app/routines/populate_trial_course_content.rb
Dantemss: So they do exist in Accounts, just cannot be claimed because no password or email.

app/routines/populate_trial_course_content.rb
Dantemss: Currently it's using the first N chapters that have valid exercises.

PR #1314, opened 2016-11-16
Production hotfixes missing from master

Dantemss:
These are production hotfixes that have not yet been applied to master, now rebased onto master

- Added research identifiers to ExportUsersInfoToMatchWithConsentForms
- Renamed GetExcludedExercises to ExportExerciseExclusions
- Fixed the exercise exclusions report
- Added book title, uuid and book location to exclusions by course 
- Added exclusion timestamps to the exercise exclusions export
- Added "null" placeholder on exclusion spreadsheet for exercises that got removed from the book
- Added instructor-only notifications


lib/settings/notifications.rb
```diff
@@ -1,46 +1,43 @@
 module Settings
   module Notifications
 
-    KEY = 'system:notifications'
```
jpslav: Anything we need to do to clean out old notifications with this key base?

lib/settings/notifications.rb
```diff
@@ -1,46 +1,43 @@
...
+        RequestStore.store[key] ||= Settings::Redis.store.hgetall(key)
       end
 
-      def update!(json)
-        Settings::Redis.store.set(KEY, ActiveSupport::JSON.encode(json))
-      end
+      protected
 
+      def key(type)
+        "#{KEY_BASE}/#{type}"
```
jpslav: minor point - are separators in redis keys normally `:`?  Can't remember

Dantemss: It seems `:` is indeed the Redis convention... Maybe we can one day change all our `/` to `:` but for now I think we should stick to `/` since we use it everywhere else...

lib/settings/notifications.rb
Dantemss: Seems like we would need to manually delete the key since it has no expiration... space is cheap though...

lib/settings/notifications.rb
Dantemss: On a side note, I think Rails might use `/` by default...

PR #1321, opened 2016-12-05
TermYear adjustments

Dantemss:
- Changed legacy courses to end Dec 31st 2016 at 11:59:59 PM
- Added helper text to admin edit courses screen for editing the term/year and dates
- Added JS to clear dates on admin edit courses screen so the course gets default dates when the term/year are changed
- Changed demo courses to somewhat respect the course year to prevent confusion
- Changed demo rake task courses to use standard demo term dates

### DIFF
--- a/app/controllers/admin/courses_controller.rb
+++ b/app/controllers/admin/courses_controller.rb
@@ -88,9 +88,16 @@ class Admin::CoursesController < Admin::BaseController
   def update
     handle_with(Admin::CoursesUpdate,
                 params: course_params,
-                complete: ->(*) {
+                success: ->(*) {
                   flash[:notice] = 'The course has been updated.'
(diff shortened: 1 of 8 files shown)


app/controllers/admin/courses_controller.rb
```diff
@@ -88,9 +88,16 @@ def edit
...
+                success: ->(*) {
                   flash[:notice] = 'The course has been updated.'
                   redirect_to admin_courses_path
+                },
+                failure: ->(*) {
+                  flash[:error] = @handler_result.errors.full_messages
+                  @course = CourseProfile::Models::Course.find(params[:id])
+                  get_schools
+                  get_catalog_offerings
+                  render :edit
```
jpslav: Can there be errors?  If so, how do they get displayed?

PR #1328, opened 2016-12-12
Add title and order to catalog offerings

Dantemss:
- Added title and number columns to catalog offerings
- Added title and description to offering representer
- Add title and number fields to admin offerings edit form

Needs FE PR

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161129003301) do
+ActiveRecord::Schema.define(version: 20161212211910) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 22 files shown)


db/schema.rb
```diff
@@ -30,10 +30,14 @@
...
     t.string   "appearance_code"
     t.boolean  "is_available",                         null: false
+    t.string   "title",                                null: false
+    t.integer  "number",                               null: false
   end
 
   add_index "catalog_offerings", ["content_ecosystem_id"], name: "index_catalog_offerings_on_content_ecosystem_id", using: :btree
+  add_index "catalog_offerings", ["number"], name: "index_catalog_offerings_on_number", unique: true, using: :btree
   add_index "catalog_offerings", ["salesforce_book_name"], name: "index_catalog_offerings_on_salesforce_book_name", unique: true, using: :btree
+  add_index "catalog_offerings", ["title"], name: "index_catalog_offerings_on_title", using: :btree
```
nathanstitt: I don't think the index on `number` and `title` will ever be used, since they're always going to be fetched by either `all` or maybe explicitly by `id`, then sorted.  My understanding of indexes are that they're only used by sorting if you're also fetching using them.  This table is going to be tiny though so isn't a big deal at all either way.

PR #1329, opened 2016-12-13
Block access to more things after course end date and improve error messages

Dantemss:
- Improved error messages for forbidden actions after a course has ended
- Specs

Needs FE


nathanstitt:
Looks good to me @Dantemss

PR #1335, opened 2016-12-29
Support logging multiple entries at once

nathanstitt:
This pattern is more convent for the FE since it needs to wrap it's logging in
a guard so that if something really goes wrong it won't log 1,000
entries in a few milli-seconds.


jpslav:
Looks good to me.

PR #1340, opened 2017-01-06
Fixed cloned assignment forecast error

Dantemss:
This bug is causing exceptions in Production

- Fixed forecast error when a (cloned) assignment has an ecosystem not in the course's list of ecosystems

### DIFF
--- a/app/routines/get_course_ecosystems_map.rb
+++ b/app/routines/get_course_ecosystems_map.rb
@@ -15,7 +15,11 @@ class GetCourseEcosystemsMap
     to_ecosystem_strategy = ecosystem_strategy_class.new(to_content_ecosystem)
     to_ecosystem = ::Content::Ecosystem.new(strategy: to_ecosystem_strategy)
 
-    from_ecosystems = course.ecosystems.map do |from_content_ecosystem|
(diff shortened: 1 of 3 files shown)


app/routines/get_course_ecosystems_map.rb
```diff
@@ -15,7 +15,11 @@ def exec(course:, ecosystem_strategy_class: ::Content::Strategies::Direct::Ecosy
     to_ecosystem_strategy = ecosystem_strategy_class.new(to_content_ecosystem)
     to_ecosystem = ::Content::Ecosystem.new(strategy: to_ecosystem_strategy)
 
-    from_ecosystems = course.ecosystems.map do |from_content_ecosystem|
+    course_ecosystems = course.ecosystems
+    tp_ecosystem_ids = Tasks::Models::TaskPlan.where(owner: course).pluck(:content_ecosystem_id)
```
nathanstitt: I think this will give you a big array of `content_ecosystem_id` for task plan for the course.  If you add a `.uniq(:content_ecosystem_id)` to the query then it should be more efficient to return just a few ids, plus the query below will be a bit cleaner.

PR #1342, opened 2017-01-09
Update cloned assignments to newest book content

Dantemss:
- Added routine to update task_plans to newer ecosystems
- Call TP update routine from TaskPlansController on create if the TP is cloned
- Spec to ensure update routine is called properly
- Spec to ensure update routine works

### DIFF
--- a/app/routines/update_task_plan_ecosystem.rb
+++ b/app/routines/update_task_plan_ecosystem.rb
@@ -1,0 +1,69 @@
+class UpdateTaskPlanEcosystem
+
+  lev_routine express_output: :task_plan
+
+  protected
+
+  # Note: Does not save the task_plan, that's up to the caller
+  def exec(task_plan:, ecosystem:)
+    outputs.task_plan = task_plan
+
+    # Lock the plan to prevent concurrent publication
+    outputs.task_plan.lock!
+
+    return unless outputs.task_plan.valid?
+
+    old_ecosystem = outputs.task_plan.ecosystem
+
+    return if old_ecosystem == ecosystem
+
+    outputs.task_plan.ecosystem = ecosystem
+
+    return if old_ecosystem.nil? || !['reading', 'homework'].include?(outputs.task_plan.type)
+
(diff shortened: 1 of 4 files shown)


app/routines/update_task_plan_ecosystem.rb
```diff
@@ -0,0 +1,73 @@
+class UpdateTaskPlanEcosystem
+
+  lev_routine express_output: :task_plan
+
+  protected
+
+  # Note: Does not save the task_plan, that's up to the caller
```
jpslav: Why not just go ahead and save it?

app/routines/update_task_plan_ecosystem.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    return unless outputs.task_plan.valid?
+
+    old_ecosystem = outputs.task_plan.ecosystem
+
+    return if old_ecosystem == ecosystem
+
+    outputs.task_plan.ecosystem = ecosystem
+
+    return if old_ecosystem.nil? || !['reading', 'homework'].include?(outputs.task_plan.type)
```
jpslav: Can you add a comment as to why readings and homeworks are singled out here?

app/routines/update_task_plan_ecosystem.rb
```diff
@@ -0,0 +1,73 @@
...
+      page_map = map.map_pages_to_pages(pages: wrapped_pages_by_id.values)
+
+      updated_page_ids = page_ids.map do |page_id|
+        wrapped_page = wrapped_pages_by_id[page_id.to_i]
+        page_map[wrapped_page].try!(:id).try!(:to_s)
+      end.compact
+
+    end
+
+    outputs.task_plan.settings['page_ids'] = updated_page_ids
```
jpslav: `updated_page_ids` isn't going to exist if `page_ids` is empty

app/routines/update_task_plan_ecosystem.rb
Dantemss: Mostly because the controller (which is the only place that calls this routine) still needs to set other stuff like the publish time afterwards.

PR #1344, opened 2017-01-11
Course enrollment API changes

Dantemss:
- Added period and course info to prevalidate enrollment API
- Added course_ended error message to enrollment prevalidate and create APIs
- Specs


nathanstitt:
Waiting on FE to be completed before merging - code looks good to me though

app/controllers/api/v1/enrollment_changes_controller.rb
```diff
@@ -9,31 +9,36 @@ class Api::V1::EnrollmentChangesController < Api::V1::ApiController
...
     result = CourseMembership::ValidateEnrollmentParameters.call(
-      book_uuid: enrollment_params.book_uuid, enrollment_code: enrollment_params.enrollment_code
+      enrollment_code: enrollment_params.enrollment_code,
+      book_uuid: enrollment_params.book_uuid
     )
-    unless result.outputs.is_valid
-      render_api_errors(:invalid_enrollment_code)
+    if result.errors.any?
+      render_api_errors(result.errors)
     else
```
nathanstitt: something (completely unrelated) that just occurs to me.

We could change `render_api_errors` so it returns null if errors.empty?.  Then we could write this as:
```ruby
  render_api_errors(result.errors) || respond_with(result.outputs.period, ...)
```

nathanstitt: Probably would be better to rename the method to something like `conditionaly_render_api_errors`

Dantemss: Yep good idea but I'll do it in a separate PR

PR #1345, opened 2017-01-11
switch from coveralls to codecov

philschatz:
As part of https://github.com/openstax/napkin-notes/issues/68 this switches from coveralls to codecov.

I was not sure about a couple things but took some defaults:

- When should coveralls make comments on the PR and what should be included
- Is the coveralls formatter used for anything (I don't know if there is a corresponding gem for codecov)
- I used this recommendation from codecov.io: https://github.com/codecov/example-ruby but was not sure about whether I needed to change the `SimpleCov`

I didn't run the tests locally because I was hoping Travis would test for me ~but it looks like this repo is configured to only run Travis when a PR is created~ Ah, `tutor-server` takes a long time to run tests.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -2,10 +2,7 @@ require 'simplecov'
 require 'codecov'
 require 'parallel_tests'
 
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
-  SimpleCov::Formatter::HTMLFormatter,
-  SimpleCov::Formatter::Codecov
-]) if ParallelTests.last_process?
+SimpleCov.formatter = SimpleCov::Formatter::Codecov
 
 SimpleCov.at_exit do
   ParallelTests.wait_for_other_processes_to_finish if ParallelTests.last_process?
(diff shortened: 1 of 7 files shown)


spec/rails_helper.rb
```diff
@@ -1,10 +1,8 @@
 require 'simplecov'
-require 'coveralls'
 require 'parallel_tests'
 
 SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
```
Dantemss: Do you need something from codecov here? If not, you can remove the MultiFormatter and just use the HTMLFormatter directly.

Dantemss: Actually maybe just remove this `if`, I think `HtmlFormatter` is the default...?

philschatz: Thanks! It seems like I need to do something similar to what coveralls did: https://github.com/codecov/example-ruby#usage

but I don't think the HTMLFormatter is necessary unless you want to continue having a local coverage report (I _think_ that's what this code does ; )

philschatz: poor Travis is gonna kill me with all these builds 😅

PR #1348, opened 2017-01-12
Cleaned up some uses of render_api_errors

Dantemss:


jpslav:
Looks good.  Rebase and merge.

PR #1349, opened 2017-01-13
Disable coverage comments and submit coverage report only on CI

Dantemss:
@philschatz

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -2,12 +2,9 @@ require 'simplecov'
 require 'codecov'
 require 'parallel_tests'
 
-SimpleCov.formatter = SimpleCov::Formatter::Codecov
-
-SimpleCov.at_exit do
-  ParallelTests.wait_for_other_processes_to_finish if ParallelTests.last_process?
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -2,7 +2,10 @@
 require 'codecov'
 require 'parallel_tests'
 
-SimpleCov.formatter = SimpleCov::Formatter::Codecov
+SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
+  SimpleCov::Formatter::HTMLFormatter,
+  SimpleCov::Formatter::Codecov
+]) if ParallelTests.last_process? && ENV['CI'] == 'true'
```
philschatz: ooh nice! I saw that snippet in the example code (CI only) but wasn't sure if you wanted it 😄 .

This way you can get HTML coverage reports locally, right?

Dantemss: Actually I might change this... not sure if using `last_process?` is correct...
SimpleCov docs seem to imply we don't need it

PR #1351, opened 2017-01-15
Push course stats to SF when no SF records exist

jpslav:
Prior to this semester, courses were largely created in Tutor because some record indicating teacher interest already existed in Salesforce.  Tutor saw that record, made a course, and then had that SF record as a place to push stats back.  (exception is that when teachers taught a 2nd semester, we reused courses and Tutor had to make a new record in SF to store the 2nd semester stats).  Thankfully, we are no longer reusing courses, so when teachers teach again, we make a brand new course and then normally have to build up a few records in SF so we can store stats.  This PR does this.  It also turns off the old code that updates stats in the old paradigm.


app/routines/push_salesforce_course_stats.rb
```diff
@@ -0,0 +1,263 @@
...
+      DevMailer.inspect_object(
+        object: @errors,
+        subject: "#{self.class.name} errors",
+        to: Rails.application.secrets.salesforce['mail_recipients']
+      ).deliver_later
+    end
+  end
+
+  def is_real_production?
+    Rails.application.secrets.environment_name == "prodtutor"
```
Dantemss: Isn't this code already somewhere else? Maybe we should put it somewhere more standard.

.env.salesforce.example
```diff
@@ -0,0 +1,35 @@
+####### BEGIN SALESFORCE #########
+#
+# We have two sets of secrets for Salesforce, one for connecting to the real data and
+# another for the "TutorSpecs" sandbox.  When logging in to the sandbox, append ".tutorspecs"
+# to your email address.
+
+# Secrets for the real data
+SALESFORCE_REAL_CONSUMER_KEY=~
+SALESFORCE_REAL_CONSUMER_SECRET=~
+SALESFORCE_REAL_LOGIN_SITE=https://login.salesforce.com
```
jpslav: These variables are only used from within the `.env` file.  Was just easier for me to keep both sets of keys in the file clearly marked as what is real and what is sandbox, so I could switch back and forth.  I think people will know that this code can access real SF stuff (the only point to having SF code is to access real SF instances)

app/routines/push_salesforce_course_stats.rb
```diff
@@ -0,0 +1,263 @@
...
+      attached_record = courses_to_attached_records[course]
+      os_ancillary = attached_record.try(:salesforce_object)
+
+      # If no OSA, try to make one
+
+      if os_ancillary.nil?
+        # First figure out which SF Contact the OSA would hang off of
+
+        sf_contact_id = best_sf_contact_id_for_course(course)
+        error!(message: "No teacher SF contact", course: course) if sf_contact_id.nil?
```
jpslav: Are you saying they set "faculty_verified" in accounts in the dropdown?  Aren't we not supposed to do that?  If there is no SF ID then the error above is appropriate.  Maybe we need to make it possible on Accounts for admins to manually set a SF contact ID?

app/routines/push_salesforce_course_stats.rb
```diff
@@ -0,0 +1,263 @@
...
+        if candidate_individual_adoptions.size > 1
+          error!(message: "Too many IndividualAdoptions matching #{individual_adoption_criteria}",
+                 course: course)
+        end
+
+        individual_adoption =
+          candidate_individual_adoptions.first ||
+          begin
+            # quickie iso8601 format
+            start_date = course.fall? ? "#{course.year}-08-01" : "#{course.year}-01-01"
```
jpslav: We don't handle summer because SF doesn't handle summer.  I have told @denvergreene that we will need the ability to handle summer soon.

Is the `starts_at` ever before Jan 1 for spring courses?  (seem to recall our dates have extra padding in them)

app/routines/push_salesforce_course_stats.rb
```diff
@@ -0,0 +1,263 @@
...
+          .compact
+          .first
+  end
+
+  def applicable_courses
+    # Only update courses from this term and any term in the future (looking at future
+    # courses will get us going on courses made way in advance of the start date).
+    # After July of a year, don't update courses from the preceding spring.
+    @courses ||=
+      CourseProfile::Models::Course.where{year.gte my{Time.now.year}}.to_a.tap do |courses|
```
jpslav: Ah I thought offerings were required now.  I think if it is a course we'll want to report on, we should just have an error above "No offering for course 42".... but do you think there will be courses that we don't want to push to SF?  i.e. should we have an option on the admin course edit page to exclude the course from this whole SF reporting algorithm? (`exclude_from_salesforce`)

app/routines/push_salesforce_course_stats.rb
```diff
@@ -0,0 +1,263 @@
...
+        courses.reject!{|cc| !cc.fall? && !cc.spring?}
+        courses.reject!{|cc| cc.year == Time.now.year && !cc.fall?} if Time.now.month >= 7
+      end
+  end
+
+  def courses_to_attached_records
+    @courses_to_attached_records ||= begin
+      ars = Salesforce::AttachedRecord
+              .preload(:salesforce_objects)
+              .select{|ar| ar.attached_to_class_name == "CourseProfile::Models::Course"}
```
jpslav: Agreed

app/routines/push_salesforce_course_stats.rb
```diff
@@ -0,0 +1,263 @@
...
+
+  def applicable_courses
+    # Only update courses from this term and any term in the future (looking at future
+    # courses will get us going on courses made way in advance of the start date).
+    # After July of a year, don't update courses from the preceding spring.
+    @courses ||=
+      CourseProfile::Models::Course.where{year.gte my{Time.now.year}}.to_a.tap do |courses|
+        courses.reject!{|cc| !cc.fall? && !cc.spring?}
+        courses.reject!{|cc| cc.year == Time.now.year && !cc.fall?} if Time.now.month >= 7
+      end
```
jpslav: ok... so is it just `@courses ||= CourseProfile::Models::Course.not_ended`? with `not_ended` as a new scope

app/routines/push_salesforce_course_stats.rb
jpslav: You may be thinking of accounts.  This is the only use here so far, but I can still move it.

app/routines/push_salesforce_course_stats.rb
Dantemss: CS and I did use the dropdown to verify faculty in the cases I described above.

app/routines/push_salesforce_course_stats.rb
Dantemss: Start dates are Jan 1st for Spring, May 1st for Summer and July 1st for Fall. All of them midnight UTC.
End dates are June 30th, August 31st and Dec 31st, respectively. All of them 11:59:59 UTC.

app/routines/push_salesforce_course_stats.rb
Dantemss: Well it's fine to raise an error I guess... Hopefully no admin-created courses with no offering.

app/routines/push_salesforce_course_stats.rb
Dantemss: Yeah could be... but then you'd have to make sure summer courses do something that makes sense. I think Denver was setting them as Spring in SF?

.env.salesforce.example
Dantemss: Fair enough.

app/routines/push_salesforce_course_stats.rb
Dantemss: Hm on second thought... Don't we want to push these stats to the sandboxes even on dev/qa/staging?

app/routines/push_salesforce_course_stats.rb
jpslav: ok, then we need a way for CS to set a contact manually

jpslav: Made a change request.

app/routines/push_salesforce_course_stats.rb
jpslav: We do push stats for non-real production.  This `is_real_production?` method is only used to limit sending out error emails to real production only (so Denver isn't getting errors from -qa that make no sense to him)

PR #1354, opened 2017-01-27
set :go param for student signups to be "student_signup"

nathanstitt:
Originally i was going to add an entirely different param but then realized that we really do want to go to signup, just for a student.  So there's really no need for a different type

Goes along with https://github.com/openstax/accounts/pull/482

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -105,7 +105,9 @@ class AuthController < ApplicationController
       redirect_to openstax_accounts.dev_accounts_url
       session[:parent] = params[:parent]
     else
-      redirect_to openstax_accounts.login_url(go: params[:go])
+      redirect_to openstax_accounts.login_url(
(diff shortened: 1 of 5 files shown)


app/controllers/auth_controller.rb
```diff
@@ -105,7 +105,9 @@ def redirect_to_login_url
       redirect_to openstax_accounts.dev_accounts_url
       session[:parent] = params[:parent]
     else
-      redirect_to openstax_accounts.login_url(go: params[:go])
+      redirect_to openstax_accounts.login_url(
+                    go: params[:go] == 'signup' ? 'student_signup' : nil
+                  )
     end
```
nathanstitt: I've debated changing this so the client will set the param to be `:student_signup`, but decided that decision should be made server-side, and the client should just be in charge of saying they want "signup".  I could be convinced either way though.

PR #1356, opened 2017-01-31
Settings page path helper error

jpslav:
On the admin page, clicking "System Setting" then "Settings" was generating a 500 error because there was a missing `main_app.` prefix on a path helper in the application controller.

Also includes a commit to send out an exception email for this kind of exception.

### DIFF
--- a/config/initializers/rescue_from.rb
+++ b/config/initializers/rescue_from.rb
@@ -44,3 +44,5 @@ OpenStax::RescueFrom.register_exception(
   # only notify when real data involved (only time it really needs admin attention)
   notify: secrets['salesforce']['allow_use_of_real_data']
 )
+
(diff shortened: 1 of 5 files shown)


config/initializers/rescue_from.rb
```diff
@@ -44,3 +44,5 @@
   # only notify when real data involved (only time it really needs admin attention)
   notify: secrets['salesforce']['allow_use_of_real_data']
 )
+
+ExceptionNotifier.ignored_exceptions.reject!{|ee| ee == "ActionController::UrlGenerationError"}
```
Dantemss: I think you can just `ExceptionNotifier.ignored_exceptions.delete("ActionController::UrlGenerationError")

jpslav: Thanks... looked for something like that but didn't see `delete`

PR #1365, opened 2017-03-08
add errata_url to exercises configuration and to bootstrap data

pandafulmanda:


config/secrets.yml.example
```diff
@@ -65,6 +65,7 @@ development:
       secret: <%= ENV["OPENSTAX_EXERCISES_SECRET"] %>
       url: <%= ENV["OPENSTAX_EXERCISES_URL"] || 'https://exercises-dev.openstax.org' %>
       stub: <%= ENV["OPENSTAX_EXERCISES_STUB"] %>
+      errata_url: <%= ENV["OPENSTAX_EXERCISES_ERRATA_FORM"] || 'https://oscms.openstax.org/errata/form' %>
```
jpslav: Can this env var end in `_URL` to match the others?

jpslav: Also this needs to go in an osweb section I believe? (not part of Exercises).

app/representers/api/v1/bootstrap_data_representer.rb
```diff
@@ -29,6 +29,11 @@ class BootstrapDataRepresenter < ::Roar::Decorator
                )
              }
 
+    property :exercises_errata_url,
+             readable: true,
+             writeable: false,
+             getter: ->(*) { OpenStax::Exercises::V1.configuration.errata_url }
```
jpslav: See below -- this can just be

```
getter: ->(*) { Rails.application.secrets["openstax"]["osweb"]["errata_form_url"] }
```

(I think quotes are right for those keys, tho they might need to be symbols)

spec/representers/api/v1/bootstrap_data_representer_spec.rb
```diff
@@ -18,6 +18,7 @@
       "courses" => [], # not testing this since it's too expensive to generate meaningful course data
       "accounts_api_url" => OpenStax::Accounts.configuration.openstax_accounts_url + 'api',
       "accounts_profile_url" => OpenStax::Accounts.configuration.openstax_accounts_url + 'profile',
+      "exercises_errata_url" => OpenStax::Exercises::V1.configuration.errata_url,
```
jpslav: since hardcoded in the secrets.yml in the `test` section, can just hardcode URL here (will test that the URL is actually in the secrets properly)

jpslav:
Looks good.  Merge after complimentary tutor-deployment PR merged.

PR #1366, opened 2017-03-08
Admin courses screen fixes

Dantemss:
- Admin courses screen now displays ecosystem ID instead of book ID
- Ecosystem titles are now autogenerated from book title and cnx_id

I removed the Marshal stuff because it broke and I couldn't figure out why (message was something like... can't marshal hash). Just used ActiveRecord models instead, since ActiveJob handles those.


app/routines/fetch_and_import_book_and_create_ecosystem.rb
```diff
@@ -7,15 +7,13 @@ class FetchAndImportBookAndCreateEcosystem
...
 
   # Returns a Content::Ecosystem containing a book obtained from the given CNX id
-  def exec(book_cnx_id:, archive_url: nil, ecosystem_title: nil,
+  def exec(book_cnx_id:, archive_url: nil,
            reading_processing_instructions: nil, exercise_uids: nil, comments: nil)
     archive_url ||= OpenStax::Cnx::V1.archive_url_base
     OpenStax::Cnx::V1.with_archive_url(archive_url) do
       cnx_book = OpenStax::Cnx::V1.book(id: book_cnx_id)
-      ecosystem_title ||= \
-        "#{cnx_book.title} (#{cnx_book.uuid}@#{cnx_book.version}) - #{Time.now.utc}"
```
jpslav: Is the time of import recorded elsewhere?

app/routines/search_courses.rb
```diff
@@ -36,17 +36,23 @@ def exec(params = {}, options = {})
...
+            teachers.role.profile.account.first_name.like_any(sanitized_queries) |
+            teachers.role.profile.account.last_name.like_any(sanitized_queries) |
+            teachers.role.profile.account.full_name.like_any(sanitized_queries) |
+            ecosystems.books.title
+              .op('||', ' (')
+              .op('||', ecosystems.books.uuid)
+              .op('||', '@')
+              .op('||', ecosystems.books.version)
+              .op('||', ')')
+              .like_any(sanitized_queries)
```
jpslav: What is lines 50 through 54 doing?

app/routines/search_courses.rb
```diff
@@ -102,7 +108,15 @@ def exec(params = {}, options = {})
...
 
-          @items = @items.joins(:ecosystems).where{ecosystems.title.like_any sanitized_titles}
+          @items = @items.joins(ecosystems: :books).where do
+            ecosystems.books.title
+              .op('||', ' (')
+              .op('||', ecosystems.books.uuid)
+              .op('||', '@')
+              .op('||', ecosystems.books.version)
+              .op('||', ')')
+              .like_any(sanitized_titles)
```
jpslav: Is there somewhere else these op lines can exist so they are not duplicated in this file? (without even knowing what they are doing, can they be a scope in the model or something?)

app/routines/fetch_and_import_book_and_create_ecosystem.rb
Dantemss: The `created_at` field. It's also what is displayed in the admin courses screen.

app/routines/search_courses.rb
Dantemss: We used to search the ecosystem title, but that is no longer a column so to maintain the current behavior I'm concatenating the columns used to generate the title and searching that.

app/routines/search_courses.rb
Dantemss: Hm they really are only useful in searches... Hmm maybe I should leave the title column for caching/searches but not allow it to be set...

PR #1368, opened 2017-03-13
Update SF code for new production schema changes

jpslav:
The Salesforce production schema changes (contacts required on OSAs, term-based class start dates on IAs, etc).  This PR contains updates to match those changes.  

Note that to be tested on QA or staging, the TutorQA Salesforce sandbox will have to be refreshed.


jpslav:
There are build errors on travis b/c I didn't update a default in the secrets.  I'm going to merge this since it is in the RC, and then push another change.

PR #1375, opened 2017-04-10
return entire task when a step is marked complete

nathanstitt:
Return entire task when a step is marked complete

This is a bit more efficient for the FE since it won't have to re-request the task each time.


jpslav:
How much of a computational load is this for a long interactive reading?

nathanstitt:
@jpslav we're not entirely sure, but @Dantemss and I chatted about it and we believe it's the best approach. The FE could turn around after each step completion and reload the steps, but that seems like it'd be even more expensive because of the multiple HTTP requests.

Dantemss:
PR seems fine to me. Not sure about CPU/DB impact, hopefully not that bad but depends on assignment size.

nathanstitt:
@Dantemss - I think if the load does get to high, we could get smarter about the logic and only return the task if the steps have actually changed.  Not sure how to detect that right now though

PR #1379, opened 2017-04-19
More preview fixes

Dantemss:
- Allow preview courses with only 1 period
- Fix preview assignment dates
- Fix preview course duration
- More strict parameter checking during course creation
- Change preview student names
- More specs


Dantemss:
Wait for Kim and Kef to approve these...

PR #1382, opened 2017-04-25
Add is_preview flag to task plans

nathanstitt:


Dantemss:
`is_preview`?

nathanstitt:
@Dantemss - yeah, i was on the fence about wether the flag should be `is_preview` or `is_demo`

I felt like is_demo might be more appropriate because it wouldn't be confused with the preview course.  But maybe naming them the same is the right call?

Dantemss:
The reason I don't like is_demo is because it's the same as actual demo courses/assignments (that script you made a long time ago, people still use it to demo to foundations, etc)

Dantemss:
I'm fine with a different name though, maybe `is_automatic` or the opposite, `is_created_by_the_teacher`?

nathanstitt:
@Dantemss You know the more I think about it I think we should "go with the flow" and just use `is_preview` since it's generated with the for the purpose as the preview course.   Gonna make that change unless you've now flipped your opinion and disagree 😁

PR #1383, opened 2017-04-27
add winter term

nathanstitt:
Needed for FE per https://trello.com/c/BFtqSFn4/657-update-the-create-or-copy-a-course-screens

### DIFF
--- a/spec/lib/term_year_spec.rb
+++ b/spec/lib/term_year_spec.rb
@@ -47,6 +47,26 @@ RSpec.describe TermYear, type: :lib do
     end
   end
 
+  context 'winter' do
+    let(:term) { 'winter' }
+
+    TESTED_YEARS.each do |year|
+      context year.to_s do
+        let(:year) { year }
+
+        it "returns Nov 31st, #{year} 00:00:00 AM as the start date" do
+          start_date = DateTime.parse("Jan 1st, #{year} 00:00:00 AM")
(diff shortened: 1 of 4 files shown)


nathanstitt:
@jpslav looks like this PR is failing the salesforce specs, no idea why though since it doesn't add any fields or anything.

https://travis-ci.org/openstax/tutor-server/builds/226913632

jpslav:
The course factory sets the course term randomly, which is why this spec just happened to fail here. 
 Happened to me in a different PR today too, and I fixed it in https://github.com/openstax/tutor-server/pull/1384/commits/a6a24bff336cbc73ced43062be44cd2a1c4ef745 - which has been merged, so if you rebase to master you should be good here.

spec/lib/term_year_spec.rb
```diff
@@ -47,6 +47,26 @@
...
   end
 
+  context 'winter' do
+    let(:term) { 'winter' }
+
+    TESTED_YEARS.each do |year|
+      context year.to_s do
+        let(:year) { year }
+
+        it "returns Nov 31st, #{year} 00:00:00 AM as the start date" do
```
jpslav: any reason this says "Nov 31st" but the code says "Jan 1st"?  Ditto the next example

nathanstitt: nope, looks like a typo, i'll fix it

PR #1386, opened 2017-05-02
Set time zone and record # of students when course is created

nathanstitt:
Previously the create course only accepted a `TimeZone` model, this supports setting as string as well.

Also adds "estimated_student_count" field for teachers to record what they think the class size will be

### DIFF
--- a/app/controllers/api/v1/courses_controller.rb
+++ b/app/controllers/api/v1/courses_controller.rb
@@ -153,7 +153,7 @@ class Api::V1::CoursesController < Api::V1::ApiController
     Creates a copy of the course with the given ID
     All JSON attributes in the schema are optional
     They will default to the given course's attributes if ommitted
-    #{json_schema(Api::V1::CourseRepresenter, include: :writeable)}
+    #{json_schema(Api::V1::CourseCloneRepresenter, include: :writeable)}
   EOS
   def clone
     OSU::AccessPolicy.require_action_allowed!(:clone, current_api_user, @course)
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/courses_controller.rb
```diff
@@ -153,7 +153,7 @@ def roster
     Creates a copy of the course with the given ID
     All JSON attributes in the schema are optional
     They will default to the given course's attributes if ommitted
-    #{json_schema(Api::V1::CourseRepresenter, include: :writeable)}
+    #{json_schema(Api::V1::CourseCloneRepresenter, include: :writeable)}
```
jpslav: I actually just saw this unused representer earlier today -- FE is changing to reflect this?

app/routines/create_course.rb
```diff
@@ -16,8 +16,8 @@ class CreateCourse
 
   def exec(name:, term:, year:, is_preview:, is_college:, is_concept_coach: nil, num_sections: 0,
            catalog_offering: nil, appearance_code: nil, starts_at: nil, ends_at: nil,
-           school: nil, time_zone: nil, cloned_from: nil,
-           default_open_time: nil, default_due_time: nil)
+           school: nil, time_zone: nil, cloned_from: nil, default_open_time: nil,
+           default_due_time: nil, estimated_student_count: 0)
```
jpslav: make the default `nil` so we know we don't know it is zero?

app/controllers/api/v1/courses_controller.rb
nathanstitt: No, it's in use - the code in the controller uses it to consume the params, but the docs had the wrong one listed.

It does inherit from `CourseRepresenter` so the docs were mostly right though.

app/routines/create_course.rb
nathanstitt: sure, sounds good

PR #1387, opened 2017-05-03
ecommerce BE infrastructure #1

jpslav:
Rebase after BL PR merged (use its migration for the student UUID)

## TODO

* [x] Add a `prompt_student_to_pay` field on the student JSON that is `payments_site_online && course.does_cost && !course.is_preview && !(student.is_paid || student.is_comped)` so that the FE can look at this one place to know whether or not to prompt a student instead of having to compile the data/logic itself.
* [x] Specs for representer changes (including roster showing paid status)
* [x] Spec payment_due_at 1 second before and after
* [x] Spec correct setting of `payment_due_at` on enroll.  Spec that this value doesn't change if dropped and re-enrolled.
* [x] Admin ability to globally ignore that payments are due (should impact BE code check for "has paid" and should be provided to FE in bootstrap data)
* [x] Feature specs showing unpaid students have all intended actions limited
* [x] Spec needs to pay before_filter when only have `@task_step`

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -31,6 +31,7 @@ class Api::V1::ApiController < OpenStax::Api::V1::ApiController
   protected
 
   def payment_overdue?(course, student)
+    return false if !Settings::Payments.payments_enabled
     return false if student.nil?                        # only students need to pay
     return false if course.is_preview                   # preview courses should never cost
     return false if !course.does_cost
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -2,4 +2,43 @@
 # this class could be V7 and still inherit from V1 in the API gem.
 #
 class Api::V1::ApiController < OpenStax::Api::V1::ApiController
+
+  def error_if_student_and_needs_to_pay
+    return true if current_api_user.human_user.is_anonymous?
+
+    if @student.present?
+      student = @student
+      course = @student.course
+    elsif @course.present?
+      course = @course
+      student = UserIsCourseStudent.call(
+        user: current_api_user.human_user, course: @course
+      ).outputs.student
+    elsif @task.present?
+      student = @task.taskings.first.role.student
+      course = student.try!(:course)
+    elsif @task_step.present?
+      # Assumes all tasks are assigned to one student
+      student = @task_step.task.taskings.first.role.student
+      course = student.try!(:course)
+    else
+      raise "Either @student, @course, @task, or @task_step must be set"
+    end
+
+    payment_overdue?(course, student) ? render_api_errors(:payment_overdue) : true
+  end
+
+  protected
+
+  def payment_overdue?(course, student)
+    return false if student.nil?                        # only students need to pay
+    return false if course.is_preview                   # preview courses should never cost
+    return false if !course.does_cost
+    return false if student.payment_due_at.nil?
+    return false if Time.now < student.payment_due_at   # not overdue yet
+    return false if student.is_paid
+    return false if student.is_comped
+    return true
+  end
+
 end
--- a/app/controllers/purchases_controller.rb
+++ b/app/controllers/purchases_controller.rb
@@ -1,8 +1,7 @@
 class PurchasesController < ApplicationController
 
   def show
-    student = CourseMembership::Models::Student.find_by(uuid: params[:id])
-    raise ActiveRecord::RecordNotFound if student.nil?
+    student = CourseMembership::Models::Student.find_by!(uuid: params[:id])
     raise SecurityTransgression if student.role.role_user.profile != current_user.to_model
     redirect_to course_dashboard_path(student.course)
   end
--- a/app/controllers/purchases_controller.rb
+++ b/app/controllers/purchases_controller.rb
@@ -1,0 +1,10 @@
+class PurchasesController < ApplicationController
+
+  def show
+    student = CourseMembership::Models::Student.find_by(uuid: params[:id])
+    raise ActiveRecord::RecordNotFound if student.nil?
+    raise SecurityTransgression if student.role.role_user.profile != current_user.to_model
+    redirect_to course_dashboard_path(student.course)
+  end
+
+end
--- a/app/routines/create_course.rb
+++ b/app/routines/create_course.rb
@@ -19,7 +19,7 @@ class CreateCourse
            num_sections: 0, catalog_offering: nil, appearance_code: nil,
            starts_at: nil, ends_at: nil, school: nil, time_zone: nil, cloned_from: nil,
            default_open_time: nil, default_due_time: nil, estimated_student_count: nil,
-           does_cost: false)
+           does_cost: nil)
 
     # TODO eventually, making a course part of a school should be done independently
     # with separate admin controller interfaces and all work done in the SchoolDistrict SS
@@ -35,7 +35,7 @@ class CreateCourse
     end
 
     is_concept_coach = catalog_offering.try!(:is_concept_coach) if is_concept_coach.nil?
-    does_cost = catalog_offering.try!(:does_cost) if does_cost.nil?
+    does_cost = (catalog_offering.try!(:does_cost) || false) if does_cost.nil?
 
     fatal_error(
       code: :is_concept_coach_blank,
--- a/app/routines/create_course.rb
+++ b/app/routines/create_course.rb
@@ -18,7 +18,9 @@ class CreateCourse
   def exec(name:, is_preview:, is_college:, is_concept_coach: nil, term: nil, year: nil,
            num_sections: 0, catalog_offering: nil, appearance_code: nil,
            starts_at: nil, ends_at: nil, school: nil, time_zone: nil, cloned_from: nil,
-           default_open_time: nil, default_due_time: nil, estimated_student_count: nil)
+           default_open_time: nil, default_due_time: nil, estimated_student_count: nil,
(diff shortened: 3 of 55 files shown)


app/subsystems/course_membership/is_course_student.rb
```diff
@@ -24,7 +24,8 @@ def exec(course:, roles:, include_dropped: false, include_archived: false)
     end
 
     is_course_student = valid_student.present? || outputs.is_dropped || outputs.is_archived
-
     outputs.is_course_student = !!is_course_student
+    
+    outputs[:student] = valid_student
```
Dantemss: Uh we already have way too many routines that do this or something very similar... I would say we should get rid of some (eventually): `ChooseCourseRole`, `GetUserCourseRoles`, `CourseMembership::GetCourseRoles`, `CourseMembership::GetRoleCourses`

db/migrate/20170428230057_add_payment_fields.rb
```diff
@@ -0,0 +1,11 @@
+class AddPaymentFields < ActiveRecord::Migration
+  def change
+    add_column :catalog_offerings, :does_cost, :boolean, default: false, null: false
+    add_column :course_profile_courses, :does_cost, :boolean, default: false, null: false
```
Dantemss: I would argue that these columns shouldn't have a default, they should always be set by... something.

Could do this by creating them without `null: false` and then calling `change_column_null :course_profile_courses, :does_cost, false, false` etc to set current courses to false without setting a default

db/migrate/20170428230057_add_payment_fields.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    add_column :course_membership_students, :first_paid_at, :datetime
+    add_column :course_membership_students, :is_paid, :boolean, default: false, null: false
```
Dantemss: `last_paid_at` might provide us with more info than a boolean... `is_paid` could be a method (that maybe also checks `is_comped`)

db/migrate/20170428230057_add_payment_fields.rb
```diff
@@ -0,0 +1,11 @@
...
+    add_column :course_membership_students, :is_comped, :boolean, default: false, null: false
```
Dantemss: `comped_at` with `is_comped` being a method?

app/controllers/api/v1/api_controller.rb
```diff
@@ -2,4 +2,44 @@
...
+    elsif @task.present?
+      student = @task.taskings.first.role.student
+      course = student.try!(:course)
+    elsif @task_step.present?
+      # Assumes all tasks are assigned to one student
+      student = @task_step.task.taskings.first.role.student
+      course = student.try!(:course)
+    else
+      raise "Either @student, @course, @task, or @task_step must be set"
+    end
```
jpslav: I agree it is a lot.  I might leave it tho.

app/controllers/api/v1/api_controller.rb
```diff
@@ -2,4 +2,44 @@
...
+  def payment_overdue?(course, student)
+    return false if !Settings::Payments.payments_enabled
+    return false if student.nil?                        # only students need to pay
+    return false if course.is_preview                   # preview courses should never cost
+    return false if !course.does_cost
+    return false if student.payment_due_at.nil?
+    return false if Time.now < student.payment_due_at   # not overdue yet
+    return false if student.is_paid
+    return false if student.is_comped
+    return true
```
jpslav: The reason I like separate statements is for debugging.  When stepping through you can see which line triggers as opposed to manually typing out the conditions to see what is tripping the return.

app/controllers/purchases_controller.rb
```diff
@@ -0,0 +1,10 @@
+class PurchasesController < ApplicationController
+
+  def show
+    student = CourseMembership::Models::Student.find_by(uuid: params[:id])
+    raise ActiveRecord::RecordNotFound if student.nil?
+    raise SecurityTransgression if student.role.role_user.profile != current_user.to_model
```
jpslav: I read this comment and I immediately thought "yes of course".  But when I started writing it it got weird.  As you know, access policies test some action on some object for some user.  The only object we have here is `Student`.  So we could have an action like `:go_to_purchase`, but somehow that doesn't seem right.  Seems more like the access policy should control access to a `Purchase` but we don't have such an object (yet).  If we added one, we'd have to have some way to know this is a "student" purchase, have some case statement in the `PurchaseAccessPolicy` etc.  All seems like a lot of work to replace this one line.  

Any thoughts on a way to make this good and easy?  I may have had too much wine to think through it clearly :-)

app/routines/create_course.rb
```diff
@@ -33,16 +35,14 @@ def exec(name:, is_preview:, is_college:, is_concept_coach: nil, term: nil, year
     end
 
     is_concept_coach = catalog_offering.try!(:is_concept_coach) if is_concept_coach.nil?
+    does_cost = catalog_offering.try!(:does_cost) if does_cost.nil?
```
jpslav: Good point.  I kind of like making it not be required.  But it should really just default to the offering's setting if `does_cost` is not specified.  And I do want courses that cost to be intentionally set somehow (want the fallback position to be that they don't cost).

So maybe making `does_cost: nil` be the argument and then 

```
does_cost = catalog_offering.try!(:does_cost) || false
```

In any case I need a spec to lock the behavior down.

app/subsystems/course_membership/is_course_student.rb
jpslav: eventually :-)

we should pow-wow about the biggest wins we can shoot for in simplifying our setup.

db/migrate/20170428230057_add_payment_fields.rb
jpslav: I want `does_cost = true` to be very intentional.  Nothing should cost unless we _really_ intend for it to.

db/migrate/20170428230057_add_payment_fields.rb
jpslav: Yeah, I think I started with this and switched back... may have been for a dumb reason like making spec setup easier. So I'm definitely up for reconsidering. About a `last_paid_at`... if that being non nil is how we know someone paid, what would you expect the behavior to be if a student was issued a refund?  Would we just set `last_paid_at` to `nil`?  One could argue that keeping the `last_paid_at` set to its pre-refund value would make sense to (as that is still the time the student last paid even if a refund was later issued).  

Like instead, do we need `paid_at` and `refunded_at`  and...

```
def is_paid 
  paid_at.present? && (refunded_at || 1000.years.ago < paid_at)
end
```

db/migrate/20170428230057_add_payment_fields.rb
jpslav: This one is clearer to me than `last_paid_at` since it'll just be admins comping or un-comping.

jpslav:
Now wondering if we need a more standalone AR for purchases instead of double-booking `Student` with some extra fields (instead of `first_paid_at`, `last_paid_at`, `refunded_at`, etc etc, do we just need a local copy of transactions related to a `Student`, kind of like `Enrollment` to track purchase history?)...

db/migrate/20170428230057_add_payment_fields.rb
Dantemss: In that case either:

A) paid_at.present? && (refunded_at.nil? || paid_at > refunded_at)

or

B) Keep the is_paid field (but might be a good idea to still add last_paid_at)

db/migrate/20170428230057_add_payment_fields.rb
Dantemss: But similarly we don't want anything to not cost unless we really intend for it to? Which means we should always specify the field?

app/routines/create_course.rb
Dantemss: Well the line you proposed would completely ignore the does_cost argument

app/controllers/purchases_controller.rb
Dantemss: I would just name the action :purchase and keep it on Student...

Dantemss: Actually... what is this action doing anyway? It just checks a couple of things and... redirects to the dashboard?

jpslav: It is so an invoice on Payments (which just has a UUID for what it thinks of as the thing being bought, happens to be the student UUID but payments doesn't know that), can link to the thing being bought (the student's enrollment/course)

app/routines/create_course.rb
jpslav: ha!

```
does_cost ||= catalog_offering.try!(:does_cost) || false
```

app/controllers/purchases_controller.rb
jpslav: If we put it on `Student`'s access policy I think the closest action name would be `:view_purchase`, `:access_purchase`, or `:go_to_purchase`.

jpslav:
Instead of `Transaction` we could avoid polymorphism and have...

```
class Student
  has_many :student_transactions
end
```
```
StudentTransaction Table

ID  | student_id | transaction_type (enum) | happened_at          | recorded_at
------------------------------------------------------------------------------------------------------
23  | 6          | paid                    | <time from payments> | <when received>
24  | 6          | refunded                | <time from payments> | <when received>
25  | 6          | paid                    | <time from payments> | <when received>
26  | 6          | chargeback              | <time from payments> | <when received>
27  | 7          | paid                    | <time from payments> | <when received>
```

PR #1389, opened 2017-05-08
Speed up CC a bit more with a view

nathanstitt:


nathanstitt:
Should have targeted  existing pr

PR #1395, opened 2017-05-11
More Biglearn fixes

Dantemss:
See commit messages for bug info...


app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -0,0 +1,121 @@
+class OpenStax::Biglearn::Api::JobWithSequenceNumber < OpenStax::Biglearn::Api::Job
```
Dantemss: Sure

app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -0,0 +1,121 @@
...
+  queue_as :default
+
+  def perform(method:, requests:, create:,
+              sequence_number_model_key:, sequence_number_model_class:)
+    req = [requests].flatten
+    sequence_number_model_key = sequence_number_model_key.to_sym
+
+    model_ids = req.map { |request| request[sequence_number_model_key].id }.compact
+    model_id_counts = {}
+    model_ids.each { |model_id| model_id_counts[model_id] = (model_id_counts[model_id] || 0) + 1 }
```
Dantemss: Yeah my complaint about this is that it would create a hash, then an array, then another hash... probably unnecessary...

app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    cases = model_id_counts.map do |model_id, count|
+      "WHEN #{model_id} THEN #{count}"
+    end
+    increments = "CASE \"id\" #{cases.join(' ')} END"
+    sequence_number_sql = <<-SQL.strip_heredoc
+      UPDATE #{table_name}
+      SET "sequence_number" = "sequence_number" + #{increments}
+      WHERE "#{table_name}"."id" IN (#{model_ids.join(', ')})
+      #{'  AND "sequence_number" > 0' unless create}
```
Dantemss: It's not really a comment, just a value substitution in a heredoc... I guess github doesn't render heredocs properly.

app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -0,0 +1,121 @@
...
+        next if sequence_number.nil?
+
+        next_sequence_number = sequence_number + 1
+        sequence_numbers_by_model_id[model.id] = next_sequence_number
+
+        # Make sure the provided model has the new sequence_number
+        # and mark the attribute as persisted
+        model.sequence_number = next_sequence_number
+        model.previous_changes[:sequence_number] = model.changes[:sequence_number]
+        model.send :clear_attribute_changes, :sequence_number
```
Dantemss: The changes are already in the DB because of the update statement above. The change stuff is like doing a reload without actually hitting the DB. (so fixing up the model representation in memory to match the DB)

app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -0,0 +1,121 @@
...
+
+        # Call the given block with the previous sequence_number
+        request.merge(sequence_number: sequence_number)
+      end.compact
+
+      # If an array was given, call the block with an array
+      # If another type of argument was given, extract the block argument from the array
+      modified_requests = requests.is_a?(Hash) ? requests_with_sequence_numbers.first :
+                                                 requests_with_sequence_numbers
```
Dantemss: 1. Oh this used to be in a method that took a block but I guess that's gone now so I'll change the comment
2. This whole logic is to handle calling several bulk biglearn methods with either a hash or an array. I could change it to always be an array, but would require refactoring a lot of things...

app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      # nil can happen if the request was suppressed
+      return {} if modified_requests.nil?
+      return [] if modified_requests.empty?
```
Dantemss: See above. Would need to change everywhere where bulk biglearn methods are called with hashes to expect an array result...

app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -0,0 +1,121 @@
...
+
+        OpenStax::Biglearn::Api::Job.perform_later(
+          method: method.to_s, requests: modified_requests
+        )
+      else
+        super(method: method, requests: modified_requests)
+      end
+    end
+
+    return job_or_result unless can_perform_later
```
Dantemss: Yes, good catch.

app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -0,0 +1,121 @@
...
+    worker = Delayed::Worker.new
+    ready_scope = Delayed::Job.ready_to_run(worker.name, Delayed::Worker.max_run_time)
+                              .where(id: delayed_job_id)
+    # Pretend the lock is much closer to expiring so we get a shorter timeout
+    now = Delayed::Job.db_time_now - Delayed::Worker.max_run_time + BIGLEARN_API_TIMEOUT
+    delayed_job = Delayed::Job.reserve_with_scope(ready_scope, worker, now)
+
+    return job_or_result if delayed_job.nil?
+
+    worker.run(delayed_job)
```
Dantemss: It creates a background job to guarantee that it will retry requests to Biglearn... then it tries to do the request right away to reduce the time we spend waiting for Biglearn...

app/routines/concerns/create_practice_task_routine.rb
```diff
@@ -18,6 +21,17 @@ module CreatePracticeTaskRoutine
 
   protected
 
+  def add_task_steps
+  end
```
Dantemss: Yeah I think raising some error would be the correct thing to do here

app/routines/create_practice_specific_topics_task.rb
```diff
@@ -16,29 +16,38 @@ def setup(course:, role:, page_ids: nil, chapter_ids: nil)
...
+    end
+
+    after_transaction do
+      outputs.task = Tasks::PopulatePlaceholderSteps.call(task: @task).outputs.task
+
+      nonfatal_error(
+        code: :no_exercises,
+        message: "No exercises were returned from Biglearn to build the Practice Widget." +
+                 " [Course: #{@course.id} - Role: #{@role.id}" +
+                 " - Task Type: #{@task_type} - Ecosystem: #{@ecosystem.title}]"
```
Dantemss: It becomes a 422 for the user so I guess you could find the 422 in the logs?

app/routines/create_practice_worst_topics_task.rb
```diff
@@ -6,19 +6,32 @@ class CreatePracticeWorstTopicsTask
...
-  def get_biglearn_exercises
-    OpenStax::Biglearn::Api.fetch_practice_worst_areas_exercises(
+  def add_task_steps
+    exercises = OpenStax::Biglearn::Api.fetch_practice_worst_areas_exercises(
       student: @role.student,
       inline_retry_proc: ->(response) { response[:student_status] != 'student_ready' }
-    )
+    ).first(CreatePracticeTaskRoutine::NUM_BIGLEARN_EXERCISES)
+
+    fatal_error(
```
Dantemss: Oh I think the nonfatal_error might be a leftover from what I was trying earlier (was trying to raise a nonfatal_error but that ended up not working (the transaction was still being rolled back or something) so I had to move it outside the transaction... I think I can change it back to fatal_error now

app/routines/get_task_core_page_ids.rb
```diff
@@ -1,85 +1,33 @@
...
-      homework_exercise_ids = homework_exercise_id_to_task_ids_map.keys.flatten
-      Content::Models::Exercise.where(id: homework_exercise_ids)
-                               .pluck(:id, :content_page_id)
-                               .each do |exercise_id, content_page_id|
-        task_ids = homework_exercise_id_to_task_ids_map[exercise_id]
-
-        task_ids.each do |task_id|
-          task_id_to_core_page_ids_map[task_id] << content_page_id
-        end
-      end
```
Dantemss: Because I forgot in which PR we added a content_page_id field to the task steps, so that works for all task types.

app/routines/create_practice_worst_topics_task.rb
Dantemss: Ah actually just tested this and the reason that nonfatal is there is because it's running outside the transaction and lev does not catch fatal_errors_encountered outside the transaction.

PR #1399, opened 2017-05-18
Onboarding journey event tracking

jpslav:
`TrackTutorOnboardingEvent[user: current_user, event: "created_preview_course"]` etc etc

See `/app/routines/track_tutor_onboarding_event.rb` for the possible events and the associated data that must be supplied.

### TODO

* [x] Handle arrive at marketing page not via Pardot


app/routines/track_tutor_onboarding_event.rb
```diff
@@ -0,0 +1,127 @@
...
+  def exec(event:, user:, data: {})
+    begin
+
+      toa = nil
+
+      case event.to_sym
+      when :arrived_tutor_marketing_page_from_pardot
+        raise(MissingArgument, "pardot_reported_contact_id") if data[:pardot_reported_contact_id].blank?
+        raise(MissingArgument, "pardot_reported_piaid") if data[:pardot_reported_piaid].blank?
+        raise(MissingArgument, "pardot_reported_picid") if data[:pardot_reported_picid].blank?
```
Dantemss: Could maybe have used `data.fetch(:whatever)` below and added KeyError to the instant failure exceptions, but this works too.

app/routines/track_tutor_onboarding_event.rb
```diff
@@ -0,0 +1,127 @@
...
+      end
+
+      outputs.tutor_onboarding_a = toa
+
+    rescue CannotGetToa => ee
+      log(:error) { "Cannot get TOA for #{event} event because '#{ee.message}'" }
+      raise
+    rescue MissingArgument => ee
+      log(:error) { "Missing the `#{ee.message}` argument for event #{event}" }
+      raise
```
Dantemss: Could put these messages in the exception themselves (like some other exceptions take string arguments on initialize), that way we would (hopefully) get them in the exception emails and not have to look through the logs to find them. Would also get rid of the need for this rescue.

app/routines/track_tutor_onboarding_event.rb
```diff
@@ -0,0 +1,127 @@
...
+        if user.salesforce_contact_id.present?
+
+      toa ||= TOA.find_by(accounts_uuid: user.uuid) if user.uuid.present?
+
+      toa ||= TOA.find_by(pardot_reported_contact_id: pardot_reported_contact_id) \
+        if pardot_reported_contact_id.present?
+
+      toa ||= TOA.new(accounts_uuid: user_uuid!(user))
+
+      # TODO think through / spec different cases (e.g. teacher forwards pardot email to colleage at different times)
```
Dantemss: Hopefully this has been thought through? :)

app/routines/track_tutor_onboarding_event.rb
```diff
@@ -0,0 +1,127 @@
...
+
+      toa
+    end
+  end
+
+  def log(level, &block)
+    Rails.logger.send(level) { "[OnboardingTracking] #{block.call}" }
+  end
+
+  def user_uuid!(user)
```
Dantemss: Maybe add a verb to this method name? Like `require_and_return` or even just `get`... It's not clear what it's doing just from the name.

Dantemss: I also feel a simple method like this shouldn't do things like raising exceptions... Maybe the raise belongs in `find_or_initialize` instead? That would also mean you only have to check that the uuid is blank once.

app/subsystems/user/user.rb
```diff
@@ -1,6 +1,7 @@
 module User
   class User
     include Wrapper
+    include GlobalID::Identification
```
Dantemss: Even though we're getting rid of these wrappers eventually, this is pretty neat.

app/subsystems/user/user.rb
```diff
@@ -22,7 +23,10 @@ def create!(account_id:, strategy_class: ::User::Strategies::Direct::User)
       end
 
       def find(*args, strategy_class: ::User::Strategies::Direct::User)
-        verify_and_return strategy_class.find(*args), klass: self, error: StrategyError
+        # convention that anonymous user has an ID of -1, helps with globalID lookup
+        [-1] == args.map(&:to_i) ?
```
Dantemss: This is flaky because this method has many different signatures. See http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find

Maybe this would do:

```
ids = args.first.is_a?(Array) ? args.first : args
ids.map(&:to_i).include?(-1)
```

config/initializers/delayed_job.rb
```diff
@@ -53,9 +53,20 @@
     end
   }
 
+  def self.with_delay_jobs(value, &block)
```
Dantemss: Not ThreadSafe(TM) but I guess that doesn't matter right now.
Maybe just add a comment saying it's not thread-safe.

config/initializers/openstax_salesforce.rb
```diff
@@ -28,8 +28,15 @@
...
-    config.sandbox_refresh_token = secrets['tutor_specs_refresh_token']
     config.sandbox_instance_url = secrets['tutor_specs_instance_url']
+
+    # DO NOT set the refresh token, because if the oauth token has expired the
+    # specs will use the refresh token to get a new oauth token, but then recorded
+    # cassettes will contain that unfiltered token and future spec runs may end up
+    # having the "unused interactions" messages because they don't expect the refresh
+    # token interactions.
+    #
+    #config.sandbox_refresh_token = secrets['tutor_specs_refresh_token']
```
Dantemss: We could use the VCR substitution thing to fix that but if it works without the token then I guess this is fine.

app/routines/track_tutor_onboarding_event.rb
```diff
@@ -0,0 +1,127 @@
...
+        raise(MissingArgument, "decision") if data[:decision].blank?
+
+        toa = find_or_initialize_toa(user: user)
+        toa.latest_adoption_decision_at = DateTime.current
+        toa.latest_adoption_decision = data[:decision]
+      else
+        raise InstantFailStandardError, "unknown tutor onboarding event: #{event}"
+      end
+
+      toa.save_if_changed
```
nathanstitt: it looks to me like each of the conditions calls `find_or_initialize_toa(user: user)` except `made_adoption_decision` raises first if `data[:decision].blank?` 

Maybe the `find_or_initialize_toa` could be moved to before the case block since it seems like a default operation?

jpslav: also the `arrived_marketing_page_from_pardot` passes in different data and requires different data to be set.

app/routines/track_tutor_onboarding_event.rb
jpslav: haha yes I still need to ponder this a bit, but it isn't that pressing.  we could change it in july and be ok

config/initializers/openstax_salesforce.rb
jpslav: I started to go down this path of making it resilient to needing a refresh... but it gets tricky because sometimes it doesn't need a refresh so you may or may not have the refresh interactions in the cassette.  I went down a path of turning off the cassette while testing the connection and then setting the VCR substitution thing with the new oauth token, but it was more trouble than it was worth -- just decided to make us (me) clear and reset the user in the dev database and copy the fresh oauth key in for the specs.  Also reduces the chances that we'd miss something and end up with an oauth token in the cassettes.

app/routines/track_tutor_onboarding_event.rb
jpslav: uuid can be blank at one point in the middle of `find_or_initialize_toa`, so I can't just check once. 
 Other times I need it to not be blank (or freak out).  `require_and_return` is good.

app/routines/track_tutor_onboarding_event.rb
jpslav: Ok, I implemented by raising a new exception with the longer message so I didn't have to repeat the message in a few places (or override the message method)

PR #1400, opened 2017-05-19
create standby preview courses

nathanstitt:
Adds admin setting for number of courses to prebuild, and a scheduled task to check if the count is low and make more.


app/subsystems/course_profile/build_preview_courses.rb
```diff
@@ -0,0 +1,65 @@
...
+
+          run(:populate_preview_course_content, course: course)
+        end
+      end
+    end
+  end
+
+
+  def self.start_in_background
+    if offerings_that_need_previews.any? && Jobba.where(job_name: self.name).none?
```
Dantemss: I don't think Jobbas are ever deleted, so maybe this would only run once?

Dantemss: Since this is running in a cron task, it's already in the background, so you could skip this part.

app/subsystems/course_profile/build_preview_courses.rb
```diff
@@ -0,0 +1,65 @@
...
+    courses = CourseProfile::Models::Course
+                .where("is_preview = 't' and preview_claimed_at is null")
+                .group(:catalog_offering_id)
+                .select([:catalog_offering_id, 'count(*) as course_preview_count'])
+
+
+    Catalog::Models::Offering
+      .joins { courses.as('course_preview_counts')
+                 .on { id == course_preview_counts.catalog_offering_id }
+                 .outer }
```
Dantemss: I think you should be able to just do `joins { courses.outer }` here. And in Rails 5 that would become `left_outer_joins(:courses)`.

Dantemss: On second thought, there is no call to `count()` here, so how is this working?

app/subsystems/course_profile/claim_preview_course.rb
```diff
@@ -0,0 +1,44 @@
...
+  lev_routine express_output: :course
+
+  protected
+
+  def exec(catalog_offering:, name:)
+    course = CourseProfile::Models::Course
+               .where(is_preview: true,
+                      preview_claimed_at: nil,
+                      catalog_offering_id: catalog_offering.id)
+               .first
```
Dantemss: Might be a good idea to add `.lock` here (before the `.first`). Won't make it more correct or anything (because we use Repeatable Read isolation) but might prevent some useless work if 2 `ClaimPreviewCourse` happen to run at the same time.

app/subsystems/course_profile/claim_preview_course.rb
```diff
@@ -0,0 +1,44 @@
...
+    )
+
+    offset = (Time.now - course.created_at).seconds.to_i
+    update = lambda { |fields|
+      return fields.map{|f| "#{f} = #{f} + interval '#{offset} seconds'"}.join(', ')
+    }
+
+    CourseProfile::Models::Course
+      .where({id: course.id})
+      .update_all(update[%w{created_at updated_at}])
```
Dantemss: Couldn't you just set `created_at` and `updated_at` to `Time.current` (and maybe combine that with the `update_attributes` above?

app/subsystems/course_profile/claim_preview_course.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    course.taskings.update_all(update[%w{created_at updated_at}])
```
Dantemss: Same as above, minus the combining part.

app/subsystems/course_profile/claim_preview_course.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    tasks = Tasks::Models::Task
+              .joins(taskings: :period )
+              .where(taskings: { period: { course_profile_course_id: course.id } })
+    tasks.update_all(
+      update[%w{opens_at_ntz due_at_ntz feedback_at_ntz last_worked_at created_at updated_at deleted_at}]
```
Dantemss: OK I guess the offset makes sense for these :)
Except created_at/updated_at but might as well keep them together in this case... (But see above note about maybe not updating `created_at`/`updated_at`).

app/subsystems/course_profile/claim_preview_course.rb
Dantemss: On second thought, several records' `created_at` and `updated_at` have not been updated (periods, enrollments, enrollment_changes, students...).
Maybe we should just leave `created_at` and `updated_at` alone?

app/subsystems/course_profile/claim_preview_course.rb
```diff
@@ -0,0 +1,44 @@
...
+    )
+    outputs.course = course.reload
```
Dantemss: If we get rid of the `update_all`, could skip the reload:

```
course.taskings.reset
course.periods.reset (maybe)
outputs.course = course
```

config/schedule.rb
```diff
@@ -21,3 +21,7 @@
 every 30.minutes do
   runner "Tasks::CcPageStatsView.refresh"
 end
+
+every 10.minutes do
+  runner "CourseProfile::BuildPreviewCourses.start_in_background"
```
Dantemss: As I said above, this is already in the background so maybe we could just call the routine.

spec/subsystems/course_profile/build_preview_courses.spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe CourseProfile::ClaimPreviewCourse do
```
Dantemss: Wrong class here. Also might wanna use `described_class` below, makes refactoring easier later :)

app/subsystems/course_profile/build_preview_courses.rb
nathanstitt: Well sometimes task will take substantially longer than 30 minutes.  For instance they're discussing pre-building 1,000 courses in prep for fall classes.

Good point regarding Jobba's not being deleted. I'll have to update that to see if the task is running...

app/subsystems/course_profile/build_preview_courses.rb
nathanstitt: it's joining to the the `courses` query on line 44, which has the count.

app/subsystems/course_profile/claim_preview_course.rb
nathanstitt: yeah, can't hurt.  i'll add that

app/subsystems/course_profile/claim_preview_course.rb
nathanstitt: hm, yeah let's leave `created_at` and `updated_at` alone.  They're not really used anywhere and that way we'll be able to track when they were really created if we ever need to reconstruct something.

app/subsystems/course_profile/build_preview_courses.rb
Dantemss: In that case you might wanna use a pidfile or an advisory lock... I don't think Jobba will work properly.

app/subsystems/course_profile/build_preview_courses.rb
Dantemss: Oh I see it now... All good then.

PR #1402, opened 2017-05-24
Handle incoming /pardot/toa links

jpslav:
When pardot marketing campaign emails conclude, they'll send users to `https://tutor.openstax.org/pardot/toa?<params here>`.  This PR implements that route, which has the purpose of initializing a TutorOnboardingA record in SF and redirecting users to a configurable marketing page.

### DIFF
--- a/app/controllers/pardot_controller.rb
+++ b/app/controllers/pardot_controller.rb
@@ -1,0 +1,27 @@
+class PardotController < ApplicationController
+
+  skip_before_filter :authenticate_user!
+
+  def toa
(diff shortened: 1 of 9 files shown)


app/controllers/pardot_controller.rb
```diff
@@ -0,0 +1,27 @@
+class PardotController < ApplicationController
+
+  skip_before_filter :authenticate_user!
+
+  def toa
```
jpslav: Pardot email campaigns terminate at this action

nathanstitt:
Looks good to me @jpslav

PR #1404, opened 2017-05-25
track events for pardot

nathanstitt:
Adds a `/log/onboarding/<code>` endpoint and tracks when a new course is created


nathanstitt:
@jpslav - once I added the on boarding call in `AddUserAsCourseTeacher` I got a ton of spec failures when various other routines call that and attempts to log on boarding without SF setup.  The backtrace isn't very clear but I believe it's actually originating from:

https://github.com/openstax/openstax_salesforce/blob/master/lib/openstax/salesforce/client.rb#L5

I think we need to stub that out in development, but not sure if you'd prefer to do so in the `TrackTutorOnboardingEvent` routine or somewhere in the SF gem?

jpslav:
@nathanstitt one thing that will help (slightly) is to upgrade `openstax_salesforce` to `0.15.0`.  This will probably not fix the errors but it will make the error messages more clear as I changed `User.first` to `OpenStax::Salesforce::User.first` (https://github.com/openstax/openstax_salesforce/compare/3e4ef6ba61c82ccb7e88221a23e98154847dff42...64fb6a6e266efe3e76e8083012f337e1220e6408).

This change should make the code start raising `OpenStax::Salesforce::UserMissing`.  Looks like a ton of these spec failures are repeated.  Is this because the background job is being retried a zillion times?

Probably the fix for this is just to disable SF-stuffs in these specs which you should be able to do with:

```ruby
before(:each) { disable_sfdc_client }
```

app/routines/create_or_claim_course.rb
```diff
@@ -18,6 +21,7 @@ def exec(attributes)
     else
       run(:create_course, attributes)
     end
+    run(:add_user_as_teacher, course: outputs.course, user: attributes[:teacher]) if errors.none?
```
jpslav: Not exactly part of this PR, but it is a little strange to me that this routine isn't two routines, or rather, why we don't just call this `if` block from the /api/courses POST?  If like I suggest elsewhere the TOA tracking call gets added to the end of this routine, it'll have to be in an another `if preview? ... else ...` block effectively. And at that point, it seems like this routine is really just two different routines.  

You know really the problem here is that we never figured out how to use handlers in API endpoints.  Because that's what this routine is -- a handler.  i.e. an action-unique routine that parses input, calls some other routines based on that input, and does a controller action specific thing (adds a teacher to the newly created course).

Eh... don't worry about this comment too much.  Just me rambling :-)

app/controllers/api/v1/log_controller.rb
```diff
@@ -49,4 +49,14 @@ def entry
...
+
+  api :POST, '/log/event/<event code>', 'Log that a user event has occured'
+  def event
+    if OSU::AccessPolicy.action_allowed?(params[:code], current_human_user, TrackTutorOnboardingEvent)
+      TrackTutorOnboardingEvent.perform_later(event: params[:code], user: current_human_user)
+      head :created
+    else
+      head :forbidden
+    end
+  end
```
nathanstitt: I split the difference 😁 Rather than complicate the code for something that may never happen, I just updated the route to be `/log/onboarding/<event code>` and renamed the action to `onboarding_event`  

That way if/when we need to support other categories we can add them without breaking urls

app/routines/create_or_claim_course.rb
nathanstitt: Yes, you're entirely correct - this is a routine masquerading as a handler. 

If you happen to noticed, I actually moved the `add_user_as_teacher` from the controller into the new `CreateOrClaim` routine.  I did that because it seemed like assigning a teacher would be the last step in creating or claiming a course, and so it'd be inside the lev transaction just in case something really weird happened and we wouldn't end up with a claimed course without a teacher.

app/routines/add_user_as_course_teacher.rb
```diff
@@ -15,6 +15,10 @@ def exec(user:, course:)
     else
       run(Role::CreateUserRole, user, :teacher)
       run(CourseMembership::AddTeacher, course: course, role: outputs.role)
+      TrackTutorOnboardingEvent.perform_later(
+        event: (course.is_preview? ? 'created_preview_course' : 'created_real_course'),
+        user: user
+      ) if errors.none?
```
nathanstitt: Yes, great point. that helps a lot with cutting down on spots where it's called form in specs

nathanstitt:
@jpslav this is getting VCR errors for `/spec/features/salesforce/push_salesforce_course_stats_spec.rb`, can you advise if I should re-record it (and how to do so), or is something else wrong?

PR #1405, opened 2017-05-25
Improve enrollment query performance

Dantemss:
Should fix QA performance issue


nathanstitt:
I think it looks good @Dantemss - don't spot any obvious bugs at least, but we'll have to test on qa to be sure 😁

PR #1406, opened 2017-05-28
admin ability to edit payment fields on students

jpslav:
Changes up the tab names on the edit course page:

image

"Student Roster" just had upload capability, no listing functionality, so it got renamed to "Upload Students".

Added a new "Roster" tab that lists the students and has options to change payment info, edit and see info on the user, removed unused "username" column, etc:

image

The "Paid" column is only updated by the Payments server, but admins can "Comp" a student enrollment by clicking the "Yes"/"No" under that column.  Admins can also change the payment date for a student.  Note that whichever date is picked Tutor will set the due time to be 11:59:59PM on that day.  This can't be changed by admins.

This PR also changes the feature specs to use `poltergeist`.

### DIFF
--- a/app/controllers/manager/course_details.rb
+++ b/app/controllers/manager/course_details.rb
@@ -5,6 +5,7 @@ module Manager::CourseDetails
     @course = CourseProfile::Models::Course.find(params[:id])
     @periods = @course.periods
     @teachers = @course.teachers.includes(role: { profile: :account })
+    @students = @course.students.includes(role: { profile: :account }).sort_by{|ss| ss.last_name}
     @ecosystems = Content::ListEcosystems[]
 
     @course_ecosystem = nil
--- a/app/views/admin/courses/_students.html.erb
+++ b/app/views/admin/courses/_students.html.erb
@@ -1,17 +1,55 @@
-<% method ||= :post %>
+<% @page_header = "Students for course #{@course.name}" %>
 
-<%= form_for(:course, url: url, method: method, html: { multipart: true }) do |f| %>
-  <div class='form-group'>
-    <%= f.label :period %>
-    <%= select_tag 'course[period]', options_from_collection_for_select(@periods, :id, :name),
-                                     class: 'form-control' %>
-  </div>
+<table class="table table-striped" id="students">
+  <thead>
+    <tr style="background-color: white">
+      <th>Name</th>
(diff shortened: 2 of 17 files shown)


jpslav:
@Dantemss any thoughts on the travis failure?  Something about transaction isolation conflict avoidance, and IIRC you were doing something with that recently.

app/controllers/manager/course_details.rb
```diff
@@ -5,6 +5,7 @@ def get_course_details
     @course = CourseProfile::Models::Course.find(params[:id])
     @periods = @course.periods
     @teachers = @course.teachers.includes(role: { profile: :account })
+    @students = @course.students.includes(role: { profile: :account }).sort_by{|ss| ss.last_name}
     @ecosystems = Content::ListEcosystems[]
```
nathanstitt: You could sort in sql, but it would require using the table name:
```ruby
@course.students.includes(role: { profile: :account }).order('openstax_accounts_accounts.last_name')
```

app/views/admin/courses/_students.html.erb
```diff
@@ -1,17 +1,55 @@
...
+</table>
 
-  <%= f.submit 'Upload', class: 'btn btn-primary' %>
+
+<% content_for :javascript do %>
+  <script type="text/javascript">
+    $(document).ready(function() {
+      $('table#students').stickyTableHeaders();
+      jQuery(".best_in_place").best_in_place();
+    });
```
nathanstitt: no big thing, but `$(` and `jQuery` are equivalent, could use one or the other

spec/features/admin/edit_course_spec.rb
```diff
@@ -23,7 +24,7 @@
     visit admin_courses_path
     click_link 'Edit'
 
-    expect(page).to have_content('Edit Course')
+    expect(page).to have_content('Edit course')
     fill_in 'Name', with: 'Changed777888'
     click_button 'edit-save'
```
nathanstitt: Since we're testing the same thing over & over, seems like a ripe area for abstracting  into a `expect_edit_page` helper or something.

nathanstitt:
Looks good @jpslav but the specs are failing on `DistributeTasks` with:

       expected result to have changed by 3, but was changed by 0

PR #1408, opened 2017-05-30
send email on preview pool exhaustion

nathanstitt:
This was missed on my earlier pr :(

### DIFF
--- a/app/subsystems/course_profile/claim_preview_course.rb
+++ b/app/subsystems/course_profile/claim_preview_course.rb
@@ -11,6 +11,11 @@ class CourseProfile::ClaimPreviewCourse
                .lock
                .first
     if course.nil?
+      DevMailer.inspect_object(
+        object: catalog_offering,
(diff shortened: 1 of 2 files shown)


app/subsystems/course_profile/claim_preview_course.rb
```diff
@@ -11,6 +11,11 @@ def exec(catalog_offering:, name:)
                .lock
                .first
     if course.nil?
+      DevMailer.inspect_object(
+        object: catalog_offering,
+        subject: "Failed to find preview course to claim for offering id #{catalog_offering.id}",
+        to: Rails.application.secrets.salesforce['mail_recipients']
```
jpslav: Don't want salesforce recipients here.  Use exception recipient probably

nathanstitt: Thanks, I'll just switch to using `WarningMailer.log_and_deliver` since it also will log and backtrace

PR #1411, opened 2017-06-01
Relax ui-settings "previous_settings" requirement

nathanstitt:
This continues to cause 422 errors when QA and others are using multiple copies of Tutor concurrently, either though multiple tabs or qa test script.

Also fixes a bug from the pardot pr #1404 where i'd failed to track the `data`  That's needed for FE https://github.com/openstax/tutor-js/pull/1684/commits


jpslav:
Looks good.

PR #1415, opened 2017-06-05
Remove unused preload

Dantemss:


Dantemss:
Removing this preload seems to affect the performance forecast specs, so leaving it in for now...

PR #1419, opened 2017-06-08
is_test fixes

Dantemss:
- Fix is_test not being set during course creation
- Test courses are no longer considered real courses for Biglearn


app/handlers/admin/courses_create.rb
```diff
@@ -38,6 +39,7 @@ def handle
                         num_sections: course_params.num_sections,
                         starts_at: course_params.starts_at,
                         ends_at: course_params.ends_at,
+                        is_test: course_params.is_test,
```
jpslav: guessing you need `course_params.is_test || false` here.

lib/openstax/biglearn/api/real_client.rb
```diff
@@ -159,7 +159,7 @@ def create_course(request)
     biglearn_request = {
       course_uuid: course.uuid,
       ecosystem_uuid: request.fetch(:ecosystem).tutor_uuid,
-      is_real_course: !course.is_preview,
+      is_real_course: !course.is_preview && !course.is_test,
```
jpslav: nice

jpslav:
Looks good; travis failure

PR #1420, opened 2017-06-09
Legacy course roster fix

Dantemss:
- Hide payment columns from roster when course does not cost
- Change student payment_due_at to be NOT NULL

### DIFF
--- a/spec/features/admin/edit_course_spec.rb
+++ b/spec/features/admin/edit_course_spec.rb
@@ -17,6 +17,7 @@ RSpec.feature 'Admin editing a course' do
     click_on 'Save'
 
     @course = CourseProfile::Models::Course.order(:id).last
+    @course.update_attribute :does_cost, true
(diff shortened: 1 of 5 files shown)


spec/features/admin/edit_course_spec.rb
```diff
@@ -83,12 +84,19 @@
...
+    expect(CourseProfile::Models::Course.first.does_cost).to eq false
+
     check 'course_does_cost'
     click_button 'edit-save'
 
     expect(current_path).to eq(edit_admin_course_path(@course))
     expect(page).to have_css('.flash_notice', text: 'The course has been updated.')
-    expect(CourseProfile::Models::Course.first.does_cost).to be_truthy
+    expect(CourseProfile::Models::Course.first.does_cost).to eq true
   end
```
nathanstitt: just curious, but why remove the `be_truthy` in favor of `eq true` ?  I always thought they were the same?

PR #1425, opened 2017-06-10
Update SF stats pushing to new approach

jpslav:
In the new world order...

1. There can be multiple `OsAncillary` (OSA) records under an `IndividualAdoption` (IA) for the same term.
2. We are reporting an IA's `Source`
3. We have added stats to the OSA: # students paid, # students comped, # students refunded, teacher's estimated number of students (collected when course created)
4. We are keying more off a course's UUID.
5. We are no longer setting start dates on the IA, but are setting them on the OSA, and the IA will roll them up.


app/routines/push_salesforce_course_stats.rb
```diff
@@ -7,6 +7,10 @@ class PushSalesforceCourseStats
   # the current course via `throw`/`catch` calls (to avoid always remembering to
   # `return` after `error!` and to avoid using exceptions which we catch all of)
 
+  # shortcuts
+  OSA = OpenStax::Salesforce::Remote::OsAncillary
+  IA = OpenStax::Salesforce::Remote::IndividualAdoption
```
jpslav: This looks like a lot of changes, but a lot of it is just moving code around so that the main method calls private methods instead of just being a humongous method.  There are some new things tho, so you may just want to review by looking at the current version.

app/routines/push_salesforce_course_stats.rb
```diff
@@ -232,9 +234,15 @@ def teachers(course)
...
   def best_sf_contact_id_for_course(course)
-    teachers(course).map{|tt| tt.role.role_user.profile.account.salesforce_contact_id}
-                    .compact
-                    .first
+    @cache ||= {}
+    (
+      @cache[course.uuid] ||=
+        teachers(course).map{|tt| tt.role.role_user.profile.account.salesforce_contact_id}
+                        .compact
+                        .first
```
Dantemss: If this is rerun later (next time it's called by cron or something else), this might return a different result if there is more than 1 teacher in the course. Not sure if this matters. Could avoid this by sorting the teachers by created_at or something.

spec/features/salesforce/push_salesforce_course_stats_spec.rb
```diff
@@ -25,13 +30,20 @@
                        year: 2017,
                        starts_at: Chronic.parse("January 1, 2017"),
                        offering: chemistry_offering,
-                       is_concept_coach: false
+                       is_concept_coach: false,
+                       estimated_student_count: 42,
+                       uuid: @uuids.shift
```
Dantemss: Honestly I would probably just create the courses in a before(:all) and then call the VCR stuff using the generated UUID.

spec/routines/push_salesforce_course_stats_spec.rb
```diff
@@ -10,13 +10,13 @@
 
     subject { instance.best_sf_contact_id_for_course(course) }
 
-    it 'returns nil if there are no teachers' do
-      expect(subject).to be_nil
+    it 'errors if there are no teachers' do
+      expect{ subject }.to throw_symbol(:go_to_next_course)
```
Dantemss: Interesting... Didn't know about throw_symbol

app/routines/push_salesforce_course_stats.rb
```diff
@@ -54,123 +58,105 @@ def call_for_course(course)
...
 
-        candidate_os_ancillaries = OpenStax::Salesforce::Remote::OsAncillary.where(os_ancillary_criteria).to_a
+    if candidate_individual_adoptions.size > 1
+      error!(message: "Too many IndividualAdoptions matching #{individual_adoption_criteria}",
+             course: course)
+    end
 
-        if candidate_os_ancillaries.size > 1
-          error!(message: "Too many OsAncillaries matching #{os_ancillary_criteria}", course: course)
+    candidate_individual_adoptions.first || begin
```
jpslav: yep much better, this was left over from before it was in a method like this

app/routines/push_salesforce_course_stats.rb
```diff
@@ -54,123 +58,105 @@ def call_for_course(course)
...
+      sf_contact = OpenStax::Salesforce::Remote::Contact.where(id: sf_contact_id).first
+
+      individual_adoption_options = {
+        contact_id: sf_contact_id,
+        book_id: book_names_to_sf_ids[book_name],
+        school_id: sf_contact.school_id,
+        adoption_level: "Confirmed Adoption Won",
+        source: "Tutor Signup",
+        description: Time.now.in_time_zone('Central Time (US & Canada)').iso8601.gsub(/T.*/,'') + ", " +
```
jpslav: nice on `to_date`

format of this data was requested from Denver... actually all of this data is elsewhere in the object  so one day he could potentially make this autopopulate if blank

app/routines/push_salesforce_course_stats.rb
```diff
@@ -198,15 +184,31 @@ def push_stats(course, os_ancillary)
...
       os_ancillary.num_teachers = course.teachers.length
-      os_ancillary.num_students = periods.flat_map(&:latest_enrollments_with_deleted).length
       os_ancillary.num_sections = periods.length
 
+      os_ancillary.estimated_enrollment = course.estimated_student_count
+
       os_ancillary.status = OpenStax::Salesforce::Remote::OsAncillary::STATUS_APPROVED
       os_ancillary.product = course.is_concept_coach ? "Concept Coach" : "Tutor"
+
+      os_ancillary.course_start_date = course.term_year.starts_at.iso8601.gsub(/T.*/,'')
```
jpslav: You win the time zone prize.  I think this would have been a problem :-)

jpslav: We are changing SF to not really use this date other than for reporting, so time zone and how the day is split won't really matter any longer.  Thanks for pointing this out, pushed us to make some things better in how Tutor and SF talk to each other.

app/routines/push_salesforce_course_stats.rb
jpslav: Probably shouldn't matter, because once this info is used it won't be called again for the same course (unless there was some error on the first call).  But... I put in an order clause just in case.

jpslav:
addressed your review comments @Dantemss

PR #1426, opened 2017-06-12
related_content/label fixes (Comes from)

Dantemss:
- Removed related_content as a step field (load it from the page instead)
- Removed add_labels method (assign directly instead)
- Cleaned up TaskExercise a bit

### DIFF
--- a/app/routines/task_exercise.rb
+++ b/app/routines/task_exercise.rb
@@ -4,7 +4,7 @@ class TaskExercise
 
   protected
 
-  def exec(exercise:, title: nil, task: nil, task_step: nil)
+  def exec(exercise:, task: nil, task_step: nil, title: nil)
     # This routine will make one step per exercise part.
     # If provided, the incoming `task_step` will be used as the first step.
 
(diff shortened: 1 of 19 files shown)


app/routines/task_exercise.rb
```diff
@@ -61,13 +53,13 @@ def exec(exercise:, title: nil, task: nil, task_step: nil)
...
       # Both of these only save the steps if the task or the step are already persisted
-      if new_step
+      if is_new_step
         task.task_steps << current_step
       elsif current_step.persisted?
         current_step.save!
       end
 
-      new_step = true
+      is_new_step = true
```
jpslav: Why does this matter?

Dantemss: It's used above to determine if we should use `task.task_steps << current_step` or just save the current_step. I had to add it because I was getting duplicated steps in the task due to calling `<<` on steps that were already in `task.task_steps`. It's set to true here because it's always true after the first iteration of the loop.

PR #1427, opened 2017-06-13
Fixed reading step content_page_ids

Dantemss:
`20170320194017_add_content_page_id_to_tasks_task_steps.rb` contains a mistake for reading steps (wrong page number due to intro pages being numbered 0)
This change fixes the mistake before we remove the `related_content` column and expose it (since we now load the `related_content` from the referenced `page`)

The migration has not yet been ran on production, so this fix will be effective there.


Dantemss:
@jpslav changed this PR to only affect the migration that contains the error
Will have to manually run it on production.

PR #1428, opened 2017-06-13
Fix preview course dates; Commit BuildPreviewCourses result more often

Dantemss:
Fixes the preview course start/end dates and commits the transaction after every course is built so we don't lose multiple courses if the process crashes.


app/subsystems/course_profile/build_preview_courses.rb
```diff
@@ -1,66 +1,55 @@
...
+      .joins do
+        courses.as('course_preview_counts')
+               .on { id == course_preview_counts.catalog_offering_id }
+               .outer
+      end
+      .where(is_tutor: true, is_available: true)
+      .where(['coalesce(course_preview_count, 0) < ?', desired_count])
+      .select('coalesce(course_preview_count, 0) as course_preview_count, catalog_offerings.*')
+      .reorder(1, :number) # Work on offerings with lower course_preview_count first
+      .lock('FOR UPDATE OF catalog_offerings SKIP LOCKED') # Skip offerings being worked on
```
nathanstitt: great idea to work the ones that need previews first

spec/subsystems/course_profile/claim_preview_course_spec.rb
```diff
@@ -3,42 +3,39 @@
...
         is_college: true,
         catalog_offering: offering,
         estimated_student_count: 42
-      ).outputs.course
-    }
+      ]
+    end
 
-    it 'finds the course and updates it’s attributes' do
+    it 'finds the course and updates its attributes' do
```
nathanstitt: heh, no love for UTF apostrophe?

nathanstitt:
Looks good @Dantemss - i like that you left the comment about not using the routines at the top so we're not tempted to put them back 👍

PR #1434, opened 2017-06-15
Add num students dropped, num with work, and course name to OSA stats

jpslav:


app/routines/push_salesforce_course_stats.rb
```diff
@@ -187,18 +187,28 @@ def push_stats(course, os_ancillary)
...
         os_ancillary.num_students += 1
         os_ancillary.num_students_paid += 1 if student.is_paid
         os_ancillary.num_students_comped += 1 if student.is_comped
         os_ancillary.num_students_refunded += 1 if student.first_paid_at.present? && !student.is_paid
+        os_ancillary.num_students_dropped += 1 if student.deleted?
+
+        num_steps_completed = student.role.taskings.map{ |tasking|
+          tasking.task.completed_steps_count
+        }.sum
+
```
nathanstitt: fwiw sum accepts a block, could use:

```ruby
 num_steps_completed = student.role.taskings.sum{ |tasking| tasking.task.completed_steps_count }
```

app/routines/push_salesforce_course_stats.rb
```diff
@@ -187,18 +187,28 @@ def push_stats(course, os_ancillary)
...
+        os_ancillary.num_students_with_work += 1 if num_steps_completed >= 10
       end
```
nathanstitt: what's the significance of completing at least more than 10 steps ?  Is that just an internal threshold or something?  If it's going to be changed maybe we should make it a constant or something

app/routines/push_salesforce_course_stats.rb
jpslav: really thought I tried that and it kaboomed on me

app/routines/push_salesforce_course_stats.rb
jpslav: grimaldi requested this as a threshold to figure out how many students are "real".  first time he asks me to change it I'll make it a rails-settings-ui config option :-)

app/routines/push_salesforce_course_stats.rb
nathanstitt: To test I just did:  
```ruby
[{a: 1},{a: 2},{a: 3}].sum{|x| x[:a] }
```

nathanstitt: although I suspect that Ruby probably does map/sum under the hood anyways so that's fine

PR #1438, opened 2017-06-15
Fix performance forecast performance by removing unnecessary preload

Dantemss:
And also fix all the specs that depended on the course being reloaded due to the preload.


jpslav:
You da man @Dantemss

PR #1442, opened 2017-06-16
stub payment check API endpoint

jpslav:
This stub lets the payment team test against deployed Tutors

### DIFF
--- a/app/controllers/api/v1/purchases_controller.rb
+++ b/app/controllers/api/v1/purchases_controller.rb
@@ -1,0 +1,32 @@
+class Api::V1::PurchasesController < Api::V1::ApiController
+  resource_description do
+    api_versions "v1"
+    short_description 'Interface for purchases'
+    description <<-EOS
+    EOS
+  end
+
+  api :PUT, ':id/check', 'Instructs Tutor to check on a purchase\'s payment status'
+  description <<-EOS
+    Instructs Tutor to check on a purchase\'s payment status.  The ID is the UUID
+    of the purchase.  This endpoint is throttled.
+
+    Responses:
+    * 202 Accepted if all good
+    * 404 if the UUID does not exist for a purchase
+    * 429 if throttled
+    * 5xx if things go boom
+
+    Caller should retry later if response is not 2xx or 404.
+  EOS
+  def check
+    # Stubbed for now
+    if CourseMembership::Models::Student.find_by(uuid: params[:id])
+      # TODO Call background job to do the check callback to payments
+      head :accepted
+    else
+      head :not_found
+    end
+  end
+
+end
--- a/lib/rack-attack-settings.rb
+++ b/lib/rack-attack-settings.rb
@@ -10,6 +10,11 @@ class Rack::Attack
     req.ip if req.path.starts_with?(ROUTES.api_log_entry_path) && req.post?
   end
 
+  throttle(ROUTES.check_api_purchase_path(':id'), limit: 10, period: 1.hour) do |req|
+    # Throttled based on the ID field for requests to this endpoint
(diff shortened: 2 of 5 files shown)


lib/rack-attack-settings.rb
```diff
@@ -10,6 +10,11 @@ class Rack::Attack
     req.ip if req.path.starts_with?(ROUTES.api_log_entry_path) && req.post?
   end
 
+  throttle(ROUTES.check_api_purchase_path(':id'), limit: 10, period: 1.hour) do |req|
```
Dantemss: 10 per hour? Might be easily exceeded...

lib/rack-attack-settings.rb
```diff
@@ -10,6 +10,11 @@ class Rack::Attack
...
+    # Throttled based on the ID field for requests to this endpoint
+    req.path.match(ROUTES.check_api_purchase_path('(.*)')).try(:[],1)
```
Dantemss: Uh no idea what this matcher is doing, I'll just trust the specs.

spec/requests/throttling_spec.rb
```diff
@@ -31,4 +31,29 @@
...
+      # First to pass the limit
+      expect_any_instance_of(Rack::Attack::Request).to receive(:log_throttled!).once
+      api_put '/api/purchases/first_ID/check', nil
+      expect(response).to have_http_status(429)
+
+      # Second to pass the limit
+      api_put '/api/purchases/first_ID/check', nil
+      expect(response).to have_http_status(429)
+
+      # Different IP OK
```
Dantemss: ID?

lib/rack-attack-settings.rb
Dantemss: Oh I see... 10 per hour per ID...

app/controllers/api/v1/purchases_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    api_versions "v1"
+    short_description 'Interface for purchases'
+    description <<-EOS
+    EOS
+  end
+
+  api :PUT, ':id/check', 'Instructs Tutor to check on a purchase\'s payment status'
+  description <<-EOS
+    Instructs Tutor to check on a purchase\'s payment status.  The ID is the UUID
+    of the purchase.  This endpoint is throttled.
```
Dantemss: I'm guessing this is coming from some FE (either in payments or in Tutor). Would it be possible to check that the user owns the student id given?

lib/rack-attack-settings.rb
jpslav: Yeah.  Wanted to put something in since the API is open to anyone.

lib/rack-attack-settings.rb
jpslav: The rack request doesn't get `params` populated from the URL.  This line matches the incoming URL path against `/api/purchases/(.*)/check` and pulls out the matched ID from the `(.*)` or returns `nil` if no match (in which case don't throttle).

spec/requests/throttling_spec.rb
jpslav: Oops copy paste error

app/controllers/api/v1/purchases_controller.rb
jpslav: It is coming from the Payments BE.  This is what is called when the payment finally finishes going through all the way on the payments server so that we know to go get updated payment info.  Also happens when refunds or chargebacks are processed on payments.

Dantemss: Oh in that case... why not send us the payment info in the request (rather than telling us to do another request to get it?)

PR #1460, opened 2017-06-23
iframe stub stuff for payments glue

nathanstitt:
Decided to make this a pr against the pr since this'll probably continue to evolve

It's at the point where it'll display but the buttons aren't hooked up yet.


philschatz:
If `payment-glue` is merged first, you can always `[Edit]` this PR to point to `master` later. I use that to help make PR's more concise

PR #1470, opened 2017-07-05
Biglearn fallbacks

Dantemss:
- Send dev emails when Biglearn doesn't respond
- Assign random exercises from the same pages as the task when Biglearn doesn't respond

### DIFF
--- a/lib/openstax/biglearn/api.rb
+++ b/lib/openstax/biglearn/api.rb
@@ -552,7 +552,7 @@ module OpenStax::Biglearn::Api
                                 accepted_response_status: accepted_response_status
       else
         accepted = false
-        for ii in 1..inline_max_attempts do
+        inline_max_attempts.times do
           response = job_class.perform method: method,
                                        requests: request_with_uuid,
                                        create: create,
@@ -569,7 +569,7 @@ module OpenStax::Biglearn::Api
         Rails.logger.warn do
           "Maximum number of attempts exceeded when calling Biglearn API inline" +
           " - API: #{method} - Request: #{request}" +
-          " - Attempts: #{ii} - Sleep Interval: #{inline_sleep_interval} second(s)"
+          " - Attempts: #{inline_max_attempts} - Sleep Interval: #{inline_sleep_interval} second(s)"
         end unless accepted
 
         verify_result(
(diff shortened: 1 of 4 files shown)


lib/openstax/biglearn/api.rb
```diff
@@ -625,63 +632,107 @@ def bulk_api_request(method:, requests:, keys:, optional_keys: [],
...
+          accepted_responses_uuids = responses.map do |response|
+            next if response_status_key.present? &&
+                    !accepted_response_status.include?(response[response_status_key])
 
-            responses_map[original_request] = verify_result(
-              result: block_given? ? yield(original_request, response) : response,
-              result_class: result_class
-            )
+            response[uuid_key]
+          end.compact
```
nathanstitt: How performance critical is this code, and how large will the `responses` array be?  I would be a bit more efficient to use an `each` loop vs a `map` then `compact` since the latter will loop loop twice.
```ruby
accepted_responses_uuids = []
responses.each do |response|
   accepted_responses_uuids << response[uuid_key] unless response_status_key.nil? ||
                    accepted_response_status.include?(response[response_status_key])
end
```

lib/openstax/biglearn/api.rb
```diff
@@ -542,32 +551,30 @@ def single_api_request(method:, request:, keys:, optional_keys: [],
                                 response_status_key: response_status_key.try!(:to_s),
                                 accepted_response_status: accepted_response_status
       else
-        should_retry = true
+        accepted = false
         for ii in 1..inline_max_attempts do
           response = job_class.perform method: method,
```
nathanstitt: loop could be:
```ruby

inline_max_attempts.times do |ii|
...
end

```

lib/openstax/biglearn/api.rb
```diff
@@ -542,32 +551,30 @@ def single_api_request(method:, request:, keys:, optional_keys: [],
...
-          else
-            ![accepted_response_status].flatten.include?(response[response_status_key])
-          end
-          break unless should_retry
+          accepted = response_status_key.nil? ||
+                     [accepted_response_status].flatten.include?(response[response_status_key])
+          break if accepted
 
           sleep(inline_sleep_interval)
         end
```
nathanstitt: Should sleep amount be exponential?  `sleep(inline_sleep_interval * ii)` might be good

PR #1479, opened 2017-07-15
UUIDs needed for payments

nathanstitt:
Most were in the DB but not included in representers; however all my courses were null.  I was afraid production will be the same, so added the migration to make sure they're set.

### DIFF
--- a/db/migrate/20170715185155_set_default_uuids.rb
+++ b/db/migrate/20170715185155_set_default_uuids.rb
@@ -1,0 +1,6 @@
+class SetDefaultUuids < ActiveRecord::Migration
+  def up
+    execute 'update course_profile_courses set uuid = gen_random_uuid() where uuid is null'
+    change_column_null :course_profile_courses, :uuid, false
+  end
+end
--- a/spec/controllers/api/v1/courses_controller_spec.rb
+++ b/spec/controllers/api/v1/courses_controller_spec.rb
@@ -1319,9 +1319,11 @@ RSpec.describe Api::V1::CoursesController, type: :controller, api: true,
 
       it 'clones the course for the user' do
         api_post :clone, user_1_token, parameters: valid_params, raw_post_data: valid_body
-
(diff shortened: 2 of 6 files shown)


spec/controllers/api/v1/courses_controller_spec.rb
```diff
@@ -1319,9 +1319,11 @@ def add_book_to_course(course:)
 
       it 'clones the course for the user' do
         api_post :clone, user_1_token, parameters: valid_params, raw_post_data: valid_body
-
         expect(response).to have_http_status(:success)
-        expect(response.body_as_hash).to match expected_response
+        body = response.body_as_hash
+        expect(body[:uuid]).not_to eq(expected_response[:body])
```
jpslav: Is `expected_response[:body]` a thing?

jpslav: did you mean `expected_response[:uuid]` by chance?  instead of these changes can you just put `body: a_kind_of(String)` or `body: a_string_matching(/[0-9a-f-]{36}/`?

db/migrate/20170715185155_set_default_uuids.rb
```diff
@@ -0,0 +1,6 @@
+class SetDefaultUuids < ActiveRecord::Migration
+  def up
+    execute 'update course_profile_courses set uuid = gen_random_uuid() where uuid is null'
+    change_column_null :course_profile_courses, :uuid, false
```
Dantemss: These uuids are used by Biglearn and they are set in prod... But I do agree with setting the column to `not null`.

spec/controllers/api/v1/courses_controller_spec.rb
nathanstitt: yeah it should have been `:uuid`, what I'm testing is that the new course gets a new uuid.  I've reworked to be more clear

PR #1480, opened 2017-07-18
automatically refund dropped students

jpslav:


spec/requests/api/v1/purchases_spec.rb
```diff
@@ -26,6 +26,7 @@
 
   after(:all) do
     OpenStax::Payments::Api.client = @original_client
+    OpenStax::Payments::Api.save_static_client!
```
jpslav: without this the client wasn't really getting reset, was messing with later specs.

app/subsystems/course_membership/models/student.rb
```diff
@@ -20,7 +22,7 @@ class CourseMembership::Models::Student < Tutor::SubSystems::BaseModel
   delegate :period, :course_membership_period_id, to: :latest_enrollment, allow_nil: true
 
   def is_refund_allowed
-    is_paid ? first_paid_at + 14.days > Time.now : false # TODO make this configurable
+    is_paid ? first_paid_at + REFUND_PERIOD > Time.now : false
   end
```
nathanstitt: should we be using `Timecop` ?  On the FE we're using server time which is driven by that so that eventually QA will be able to move the time around to test easier.

PR #1483, opened 2017-07-19
let admins refund a student

jpslav:
Adds a "Refund" link in the course roster for admins.  This refund link will request a refund from Payments.  The page will be refreshed and the link will change to "Refund pending".  A later refresh should show that the refund went through.

image


nathanstitt:
Looks good @jpslav

PR #1484, opened 2017-07-19
TaskStepsController performance enhancements

Dantemss:
- Wrap TaskStepsController actions in a single transaction
- Preload task_steps, pages and chapters when rendering the task JSON

### DIFF
--- a/app/controllers/api/v1/task_steps_controller.rb
+++ b/app/controllers/api/v1/task_steps_controller.rb
@@ -1,8 +1,8 @@
 class Api::V1::TaskStepsController < Api::V1::ApiController
 
-  before_filter :get_task_step_and_tasked
-  before_filter :error_if_student_and_needs_to_pay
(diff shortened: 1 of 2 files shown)


Dantemss:
If nothing else, this will clear some confusion in the new relic traces.

app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -1,8 +1,8 @@
 class Api::V1::TaskStepsController < Api::V1::ApiController
 
-  before_filter :get_task_step_and_tasked
-  before_filter :error_if_student_and_needs_to_pay
-  before_filter :populate_placeholders_if_needed, only: :show
+  around_action :with_task_step_and_tasked
```
jpslav: why an around_action?  nothing after `yield`

Dantemss: Just wrapping the action it in the transaction

PR #1489, opened 2017-07-24
rename notifications to 'updates'; include payment status

nathanstitt:


spec/representers/api/v1/bootstrap_data_representer_spec.rb
```diff
@@ -21,7 +21,7 @@
       "errata_form_url" => 'https://oscms.openstax.org/errata/form',
       "tutor_api_url" => 'https://example.com/api',
       "payments" => a_hash_including(
-        "is_enabled" => be_falsy,
+        "is_enabled" => Settings::Payments.payments_enabled,
```
jpslav: this is what I thought I had to change because of that weird spec error -- maybe it'll pass this way tho :-)

nathanstitt: yes, it passed locally with `be_falsy` but failed on Travis.  I think we must have an unrelated spec that flips that which causes order-depended failures.  🤞 this works

jpslav: @dante also getting these kinds of fails now on https://github.com/openstax/tutor-server/pull/1488

jpslav: This is probably it @nathanstitt https://github.com/openstax/tutor-server/blob/master/spec/lib/settings/payments_spec.rb#L15 -- need to reset value in a `finally` block probably

PR #1498, opened 2017-08-04
LMS Integration

jpslav:


app/subsystems/lms/launch.rb
```diff
@@ -0,0 +1,247 @@
...
+
+    raise LmsDisabled if !course.is_lms_enabled
+
+    # In theory, we could allow multiple LTI context IDs to point to the same Tutor course (e.g. if
+    # one teacher has 3 sections and uses 3 LMS courses to point to one Tutor course).  For this reason
+    # we don't currently prohibit this with a database constraint.  But for the time being, we do
+    # restrict this here in code by requiring, in the auto create call, that a course only be used
+    # in one Context.
+
+    raise CourseKeysAlreadyUsed if Lms::Models::Context.where(course: course).exists?
```
Dantemss: This basically makes this a find_or_create... Index on course should be unique.

db/migrate/20170826122030_create_lms_tables_2.rb
```diff
@@ -0,0 +1,37 @@
...
+      t.string :description
+      t.string :url
+      t.string :contact_email
+    end
+
+    create_table :lms_contexts do |t|
+      t.string :lti_id, null: false, index: true
+      t.references :lms_tool_consumer, null: false, index: true,
+                                       foreign_key: { on_update: :cascade, on_delete: :cascade }
+      t.references :course_profile_course, null: false, index: true,
```
Dantemss: Index should be unique to prevent race condition on autocreate method

app/subsystems/lms/launch.rb
```diff
@@ -0,0 +1,247 @@
...
+
+    if trusted
+      with_warnings(warning_verbosity) do
+        @message = IMS::LTI::Models::Messages::Message.generate(request_parameters)
+        @message.launch_url = request_url
+      end
+    else
+      begin
+        Lms::Models::Nonce.create!({ lms_app_id: app.id, value: request_parameters[:oauth_nonce] })
+      rescue ActiveRecord::RecordNotUnique => ee
```
Dantemss: Make sure you are not running this inside a transaction, or else the transaction becomes invalid... Might want to add a comment about this at least.

app/subsystems/lms/models/app.rb
```diff
@@ -0,0 +1,20 @@
+class Lms::Models::App < Tutor::SubSystems::BaseModel
+
+  # Currently, apps are only owned by individual courses; later they will also
+  # be owned by schools or school systems
+
+  belongs_to :owner, polymorphic: true
+
+  before_validation :initialize_tokens
+
+  validates :owner, presence: true
```
Dantemss: Might need uniqueness validation to go with the (new) unique index? Same applies to the other models.

app/subsystems/lms/models/nonce.rb
```diff
@@ -0,0 +1,3 @@
+class Lms::Models::Nonce < Tutor::SubSystems::BaseModel
+  # TODO AR uniqueness constraint
```
Dantemss: Hehe :)

app/subsystems/user/models/profile.rb
```diff
@@ -41,3 +41,12 @@ def self.anonymous
     end
   end
 end
+
+# Leave this monkey patch of Account here.  If moved to a location that is
+# not reloaded by Rails, things break in the development environment.
```
Dantemss: You can move this to an initializer or something if you do this:

```
Rails.application.config.to_prepare do
  codehere
end
```

app/views/manager/courses/_index.html.erb
```diff
@@ -102,12 +122,42 @@
                          options_from_collection_for_select(
                            @ecosystems, :id, :unique_title, params[:ecosystem_id]
                          ),
-                         include_blank: '-- Select an ecosystem --',
+                         include_blank: '-- Select an ecosystem (does not work across pages yet) --',
```
Dantemss: Ugh ok I guess you at least mention it doesn't work

config/initializers/openstax_accounts.rb
```diff
@@ -34,11 +34,6 @@
 end
 
 OpenStax::Accounts::Account.class_exec do
```
Dantemss: Move the code back and wrap it in `to_prepare` IMO :)

config/routes.rb
```diff
@@ -192,9 +192,10 @@
       end
     end
 
-    resources :enrollment_changes, only: [:create] do
+    resources :enrollment, only: [:create] do
```
Dantemss: If you pluralize the controller, remember to change the route too :)

config/routes.rb
```diff
@@ -365,6 +375,14 @@
     end
   end
 
+  get 'lms/configuration', to: 'lms#configuration'
+  get 'lms/ci_configuration', to: 'lms#ci_configuration'
+  post 'lms/launch', to: 'lms#launch'
+  get 'lms/launch_authenticate', to: 'lms#launch_authenticate'
+  post 'lms/ci_launch', to: 'lms#ci_launch'
+  get 'lms/launch_failed', to: 'lms#launch_failed'
+  get 'lms/complete_launch', to: 'lms#complete_launch'
```
Dantemss: Could probably group those like we do for the static_pages:

`scope controller: :lms do`

config/schedule.rb
```diff
@@ -18,6 +18,10 @@
   runner "OpenStax::RescueFrom.this{ PushSalesforceCourseStats.call(allow_error_email: true) }"
 end
 
+every 1.day, at: '10:30 AM' do
+  runner "OpenStax::RescueFrom.this { Lms::Models::TrustedLaunchData.where{created_at.lt 30.weeks.ago}.destroy_all }"
```
Dantemss: Oh nice... we do clean them up.

config/secrets.yml.example
```diff
@@ -121,8 +121,8 @@ test:
     exports_folder: TestUploads
   openstax:
     accounts:
-      client_id: ~
-      secret: ~
+      client_id: 'a_client_id'
+      secret: 'secret'
```
Dantemss: Make it a bit more specific please :) like 'openstax_accounts_secret'. Same for the client_id
My issue is that the VCR code was replacing all instances of client_id and secret in the cassettes with the <salesforce> stuff.

db/migrate/20170905204817_change_fields_on_course_grade_callback.rb
```diff
@@ -0,0 +1,17 @@
...
+    add_reference :lms_course_grade_callbacks, :user_profile, null: false,
+                  index: { name: :course_grade_callbacks_on_user },
+                  foreign_key: { on_update: :cascade, on_delete: :cascade }
+
+    add_reference :lms_course_grade_callbacks, :course_profile_course, null: false,
+                  foreign_key: { on_update: :cascade, on_delete: :cascade }
+
+    add_index :lms_course_grade_callbacks,
+              [:course_profile_course_id, :user_profile_id, :result_sourcedid, :outcome_url],
+              name: :course_grade_callbacks_on_course_user_result_outcome
```
Dantemss: course_grade_callbacks_on_course_user_result_outcome needs to be unique due to find_or_create_by

lib/openstax/exercises/v1/real_client.rb
```diff
@@ -41,6 +37,22 @@ def exercises(params = {}, options = {})
     JSON.parse(request(:get, uri, with_accept_header(options)).body)
   end
 
+  protected
+
+  def oauth_worker
```
Dantemss: Please add not threadsafe comment :)

app/controllers/admin/courses_controller.rb
```diff
@@ -152,6 +155,27 @@ def bulk_set_ecosystem
     redirect_to admin_courses_path
   end
 
+  def bulk_set_flag
+    if params[:flag_name].blank?
+      flash[:error] = 'Select a flag to modify'
+    else
+      # if select all on all pages, rerun query and get all course IDs
+
+      if params[:courses_select_all_on_all_pages] == 'on'
```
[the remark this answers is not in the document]
Dantemss: I saw later that you at least added a message on the UI... guess that's fine.

app/subsystems/lms/launch.rb
```diff
@@ -0,0 +1,247 @@
...
+
+  def self.from_request(request)
+    new(request_parameters: request.request_parameters, request_url: request.url)
+  end
+
+  def persist!
+    Lms::Models::TrustedLaunchData.create!(
+      request_params: request_parameters,
+      request_url: request_url
+    ).id
```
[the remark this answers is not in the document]
Dantemss: I saw later that we do clean those up, so ignore this comment.

.travis.yml
```diff
@@ -28,6 +28,8 @@ branches:
     - master
     - alpha
     - beta
+    - lms
+    - /^int-.*$/
```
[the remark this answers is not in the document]
jpslav: Let's travis run on all future integration branches if we start them with `int-`, e.g. could have named this branch `int-lms` and wouldn't have had to change `.travis.yml` to get it to run specs.

Gemfile
```diff
@@ -63,17 +63,22 @@ gem 'whenever'
...
-gem 'openstax_api', '~> 8.1.0'
+gem 'openstax_api', '~> 8.2.0'
 
 gem 'apipie-rails'
 gem 'maruku'
 
+# Fixes unreleased CotentItemSelectionRequest validation error
+# https://github.com/instructure/ims-lti/issues/128
+# This code appears to be in 2.2.1 now
+gem 'ims-lti', git: 'https://github.com/instructure/ims-lti.git', ref: '5134532d4043ee9a3e110dc5b9b41300f6c13a07'
```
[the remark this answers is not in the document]
jpslav: Yep, will do.  Might also just do a diff of 2.2.1 vs the tag here.

Gemfile
```diff
@@ -111,7 +116,7 @@ gem 'newrelic_rpm'
 gem 'validates_timeliness'
 
 # JSON schema validation
-gem 'json-schema'
+gem 'json-schema', '~> 2.8.0'
```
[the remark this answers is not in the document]
jpslav: ok I think we are only using it for apipie?

app/controllers/lms_controller.rb
```diff
@@ -0,0 +1,220 @@
...
+      redirect_to token_enroll_url(course.uuid)
+    else
+      redirect_to course_dashboard_url(course)
+    end
+  end
+
+  protected
+
+  def allow_embedding_in_iframe
+    response.headers["X-FRAME-OPTIONS"] = 'ALLOWALL'
```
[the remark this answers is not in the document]
jpslav: hmmm.... the launch and ci_launch may be loaded in iframes in any LMS.  I admittedly put this in way early to get around loading in an iframe, so maybe there's a more restrictive way to do it?  The launch and CI launch will fail unless the launch parameters are correctly signed.  @nathanstitt what do you think?

config/secrets.yml.example
jpslav: You mean make the dummy value more specific? (think that's what you're saying but want to make sure)

db/migrate/20170826122030_create_lms_tables_2.rb
jpslav: I think this one is already protected via the bigger index? https://github.com/openstax/tutor-server/pull/1498/files#diff-1acd2e7e27a227829d5d14a91c863bb6R461

app/subsystems/lms/launch.rb
jpslav: See other comment - I think `Context`'s `create!` call is protected with a unique index on all args passed to `create!`

app/subsystems/lms/models/app.rb
jpslav: Added.  Some models already had some uniqueness checks, so if you check the next commit, you might not see everywhere you expect.

app/subsystems/lms/models/nonce.rb
jpslav: Fixed.

config/initializers/openstax_accounts.rb
jpslav: Done.

config/routes.rb
jpslav: I previously tried a bunch to group these somehow, but it never came out like I wanted.  One part was always off.  But your comment inspired me to give another go.  The following is equivalent to what I have and I'll switch to it:

```ruby
  scope '/lms', controller: :lms, as: :lms do
    get :configuration
    get :ci_configuration
    post :launch
    get :launch_authenticate
    post :ci_launch
    get :launch_failed
    get :complete_launch
  end
```

config/secrets.yml.example
Dantemss: Yep that's right

db/migrate/20170905204817_change_fields_on_course_grade_callback.rb
jpslav: there's another index that is unique on `[:result_sourcedid, :outcome_url]` which I think will cover this case too (will also trigger if these 4 fields are not unique)

Dantemss: OK that works

app/subsystems/lms/launch.rb
Dantemss: That index will allow multiple Contexts for the same course, as long as the context_id and/or tool_consumer are different. Is that what is intended? Seems like the where above would say we only want 1 Context per course.

jpslav: you are right as always

PR #1515, opened 2017-08-17
More specific error page for non-enrolled students who click short link

jpslav:
Instead of a generic "forbidden" error page, students in this situation see:

image

### DIFF
--- a/app/controllers/short_codes_controller.rb
+++ b/app/controllers/short_codes_controller.rb
@@ -26,8 +26,9 @@ class ShortCodesController < ApplicationController
             raise ShortCodeNotFound
           when :authentication_required
             authenticate_user!
-          when :invalid_user
-            raise SecurityTransgression
+          when :user_not_in_course_with_required_role
+            body = "To enroll in this course, please ask your instructor for the course enrollment link."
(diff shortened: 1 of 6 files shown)


app/controllers/short_codes_controller.rb
```diff
@@ -26,8 +26,13 @@ def redirect
...
-          when :invalid_user
-            raise SecurityTransgression
+          when :user_not_in_course_with_required_role
+            error_page('You are not enrolled in this course.',
+                       <<-BODY
+                         To enroll in this course, please ask your instructor for the
+                         course enrollment link.
+                       BODY
+                       , :forbidden)
           else
```
nathanstitt: Nice!  I didn't know we had the `error_page` thing

nathanstitt:
Looks good to me @jpslav

app/controllers/short_codes_controller.rb
nathanstitt: @jpslav travis didn't like this - said it was a syntax error

jpslav: yep, fixed it

PR #1516, opened 2017-08-17
No teacher-students

Dantemss:
Block teachers from registering as students and block students from registering as teachers

### DIFF
--- a/app/subsystems/course_membership/create_enrollment_change.rb
+++ b/app/subsystems/course_membership/create_enrollment_change.rb
@@ -10,18 +10,26 @@ class CourseMembership::CreateEnrollmentChange
     fatal_error(code: :invalid_enrollment_code,
                 message: 'The given enrollment code is invalid') if period.nil?
 
+    course = period.course
     fatal_error(code: :preview_course,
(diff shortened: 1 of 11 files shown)


app/subsystems/course_membership/create_enrollment_change.rb
```diff
@@ -31,7 +39,7 @@ def exec(user:, enrollment_code:, book_uuid: nil, requires_enrollee_approval: tr
                 message: 'The given enrollment code does not match the current book') \
       if book_uuid.present? && book_uuid != course_book_uuid
 
-    course_roles, other_roles = student_roles.partition{ |role| role.student.course == course }
+    course_roles, other_roles = student_roles.partition { |role| role.student.course == course }
```
nathanstitt: FYI rubocop says to add a space before the ` {`.  I'm  not sure why though, but if you see me do it that's why 😁

Dantemss: Oh... didn't even know that... I've been adding spaces because it lines up with do:

```
method {}
method do
end
```

PR #1520, opened 2017-08-23
auto-create account using info from LMS

nathanstitt:


app/controllers/lms_controller.rb
```diff
@@ -86,6 +92,30 @@ def launch_failed; end
...
 
+
+  def forward_user_to_accounts
+    url = openstax_accounts.login_url
+    # params must be normalized so they're deterministic for the signature
+    qp = OAuth::Helper.normalize(lti_account_params)
+    # loosely from SO:
+    # http://stackoverflow.com/questions/4084979/ruby-way-to-generate-a-hmac-sha1-signature-for-oauth
+    secret_key = Rails.application.secrets.openstax['accounts']['secret']
+    signature = Base64.encode64(OpenSSL::HMAC.digest('sha1',secret_key, qp)).gsub(/\W+$/, '')
```
Dantemss: The gsub might be problematic? Maybe drop the `Base64.encode64` and use `OpenSSL::HMAC.hexdigest` like the last answer in that SO. Otherwise you might have to `CGI.escape` all this (like the first answer) to make it url-safe, but it will probably end up with some url escape characters.

https://ruby-doc.org/stdlib-2.1.0/libdoc/openssl/rdoc/OpenSSL/HMAC.html#method-c-hexdigest

hexdigest should be url-safe by itself

app/controllers/lms_controller.rb
```diff
@@ -86,6 +92,30 @@ def launch_failed; end
 
   protected
 
+
+  def forward_user_to_accounts
```
Dantemss: I'd recommend maybe extracting the params check from the launch method into its own method as well and adding a comment here saying you MUST perform the params check before calling this method.

nathanstitt: Yeah the code's probably going to be moved into a routine or something.  Main thing I wanted to check at this point is that i was using HMAC.digest in a way that made sense

app/controllers/lms_controller.rb
nathanstitt: Yep, `OpenSSL::HMAC.hexdigest` is the way to go - thanks!  the `gsub` was to strip the ending `=\n` off the end that `Base64.encode64` added.

PR #1521, opened 2017-08-23
Get a new token if Payments gives us a 403

jpslav:


lib/openstax/payments/api/real_client.rb
```diff
@@ -64,12 +69,38 @@ def api_request(method:, url:, body: {})
...
+        if num_tries == 1
+          Rails.logger.info("OX Payments: client token expired") if token_expired?
+          Rails.logger.info("OX Payments: getting a new token to try to resolve a 403")
+          initialize_oauth_variables! # resets token
+          retry
+        else
+          Rails.logger.info("OX Payments: getting a new token didn't resolve the 403")
+        end
+      end
+
```
nathanstitt: Will this ever be called concurrently?  I'm wondering if the check should be `if num_tries >= 1`

jpslav: _Probably_ not? But... no harm in making that change just in case :-)

PR #1522, opened 2017-08-24
WIP start formalizing lms app and launching

jpslav:


app/handlers/lms_complete_launch.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    # Add the user as a teacher or student
+
+    if launch.is_student?
+      # Make sure the user is in the course as a student
+
+      # TODO how do we know which period to add them to?  Possible that we could just add
+      # the student to the course and then on the FE recognize that they are without period
+      # and have them self-select their period?
```
jpslav: This is a biggie we need to discuss with UX

app/subsystems/lms/launch.rb
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    if (include_everything)
+      data += "\nEverything:\n" + message.ai(plain:true, ruby19_syntax: true, indent: 2)
+    end
+
+    data
+  end
+
+end
```
nathanstitt: are we still only using this in the view?  If so feels like we should move it there and let the view handle the formatting and displaying.

Fine to leave for future though, this is still working towards "real" code

db/migrate/20170824032434_reorganize_lms_tables.rb
```diff
@@ -0,0 +1,29 @@
+class ReorganizeLmsTables < ActiveRecord::Migration
+
+  # Make sure this class always exist when this migration is run, in case it
+  # goes away one day.
+  class Lms::Models::Nonce < Tutor::SubSystems::BaseModel; end
+
```
nathanstitt: Great idea - no more legacy migrations that no longer run but aren't discovered until a new team member attempts to start from scratch!

db/migrate/20170826122030_create_lms_tables_2.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    create_table :lms_contexts do |t|
+      t.string :lti_id, null: false, index: true
+      t.references :lms_tool_consumer, null: false, index: true,
+                                       foreign_key: { on_update: :cascade, on_delete: :cascade }
+      t.references :course_profile_course, null: false, index: true,
+                                           foreign_key: { on_update: :cascade, on_delete: :cascade }
+    end
+
```
nathanstitt: Pretty sure indexes on `lms_tool_consumers.guid` and `lms_contexts.lti_id` should be unique

app/controllers/lms_controller.rb
```diff
@@ -1,141 +1,85 @@
...
+      launch = Lms::Launch.from_request(request)
 
+      Rails.logger.debug { launch.formatted_data(include_everything: true) }
 
-    @launch_message = authenticator.message
-    # Check that we haven't seen this nonce yet
+      # Bail early if we can't handle the launch role
+      if !(launch.is_student? || launch.is_instructor?)
+        render(:unsupported_role) and return
+      end
```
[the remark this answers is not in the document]
jpslav: good point

app/subsystems/lms/launch.rb
```diff
@@ -0,0 +1,163 @@
...
+  def self.from_id(id)
+    launch_data = Lms::Models::TrustedLaunchData.find(id)
+    raise CouldNotLoadLaunch if launch_data.nil?
+    launch = new(request_parameters: ActiveSupport::HashWithIndifferentAccess.new(launch_data.request_params),
+                 request_url: launch_data.request_url,
+                 trusted: true)
+    launch_data.destroy
+    launch
+  end
+
```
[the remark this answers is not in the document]
jpslav: true - we could leave as a TODO

app/subsystems/lms/launch.rb
```diff
@@ -0,0 +1,163 @@
...
+      lms_roles = request_parameters[:roles].split(',')
+      if lms_roles.any?{|lms_role| lms_role.match(/Instructor/)}
+        :instructor
+      elsif lms_roles.any?{|lms_role| lms_role.match(/Student|Learner/)}
+        :student
+      else
+        :other
+      end
+    end
+  end
```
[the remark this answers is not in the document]
jpslav: Yep I think we will want to do those, at least to cover all of the roles in Accounts.

app/subsystems/lms/launch.rb
jpslav: I log it to the console only now -- the view is now the real launch

db/migrate/20170826122030_create_lms_tables_2.rb
jpslav: combo of guid and lti_id should be unique - that what you're saying?  lti_id _should_ be unique but I don't think it is guaranteed across LMS instances

PR #1524, opened 2017-09-05
"Complete" launch flow

jpslav:
Finishes the rough-in of the launch flow.  Still need more specs of course.

### DIFF
--- a/app/handlers/lms_complete_launch.rb
+++ b/app/handlers/lms_complete_launch.rb
@@ -9,7 +9,7 @@ class LmsCompleteLaunch
   end
 
   def handle
-    raise "`launch` cannot be nil" if launch.nil?
+    fatal_error(code: :launch_cannot_be_nil) if launch.nil?
 
     # Get the tool consumer, and update its metadata as needed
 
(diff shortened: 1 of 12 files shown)


db/migrate/20170905204817_change_fields_on_course_grade_callback.rb
```diff
@@ -0,0 +1,16 @@
...
+    add_reference :lms_course_grade_callbacks, :user_profile, null: false,
+                  index: { name: :course_grade_callbacks_on_user },
+                  foreign_key: { on_update: :cascade, on_delete: :cascade }
+
+    add_reference :lms_course_grade_callbacks, :course_profile_course, null: false,
+                  foreign_key: { on_update: :cascade, on_delete: :cascade }
+
+    add_index :lms_course_grade_callbacks, [:course_profile_course_id, :user_profile_id],
+              name: :course_grade_callbacks_on_course_user
+  end
```
nathanstitt: Should the index include all the columns that we use for the `find_or_create_by` ?  Looks like it's missing the `outcome_url`

app/handlers/lms_complete_launch.rb
```diff
@@ -46,25 +46,27 @@ def handle
...
+        # a Student later records).  It is possible that a teacher could add the Tutor assignment
+        # more than once, so we could have multiple callback infos for ever course/user combination.
+
+        Lms::Models::CourseGradeCallback.find_or_create_by(
+          result_sourcedid: launch.result_sourcedid,
+          outcome_url: launch.outcome_url,
+          course: course,
+          profile: caller.to_model
+        )
       end
```
nathanstitt: Good idea, by using the `user` instead of student that neatly sidesteps the entire "student record doesn't exist yet" issue.

PR #1525, opened 2017-09-06
Add /enrollment/<course_uuid>/choices

nathanstitt:
Renames `enrollment_changes` to just `enrollment`


app/representers/api/v1/course_enrollments_representer.rb
```diff
@@ -0,0 +1,34 @@
...
+             type: String,
+             writeable: true,
+             readable: true,
+             schema_info: {
+               description: "The period's name"
+             }
+
+    property :enrollment_code,
+             type: String,
+             writeable: true,
```
jpslav: Make all `writeable`s in this representer `false`

app/controllers/api/v1/enrollment_controller.rb
```diff
@@ -9,6 +10,25 @@ class Api::V1::EnrollmentChangesController < Api::V1::ApiController
...
 
+  api :GET, '/:course_uuid/choices', 'Returns limited information for a given course uuid'
+  description <<-EOS
+    Returns course and period information for a course identified by it's UUID.
+    Can be called by any logged in user, but only returns the course name and enrollment codes
+    #{json_schema(Api::V1::PeriodRepresenter, include: :readable)}
+  EOS
+  def choices
+    course = CourseProfile::Models::Course.find_by(uuid: params[:id])
+    OSU::AccessPolicy.require_action_allowed!(:enrollment_info, current_human_user, course)
```
jpslav: Access policies are more for saying who is authorized to do or not do something.  Here it is being used as a check that the course of the given UUID is actually found.

Instead, suggest using `find_by!` which will raise `ActiveRecord::RecordNotFound` if not present, which in `openstax_rescue_from` will give a JSON response with a 404. Then the `require_action_allowed!` can be nixed along with the changes to the access policy.

app/controllers/api/v1/enrollment_controller.rb
```diff
@@ -1,4 +1,5 @@
-class Api::V1::EnrollmentChangesController < Api::V1::ApiController
+class Api::V1::EnrollmentController < Api::V1::ApiController
```
jpslav: yay for fewer characters!

app/representers/api/v1/course_enrollments_representer.rb
```diff
@@ -0,0 +1,34 @@
...
+             readable: true,
+             schema_info: {
+               description: "The period's enrollment code"
+             }
+  end
+
+    property :name,
```
jpslav: something weird with indentation?

app/controllers/api/v1/enrollment_controller.rb
nathanstitt: yeah, this use was intentional on my part - I wanted to prevent guessing the uuid, and all wrong answers returned access denied it wouldn't offer clue to someone brute forcing it.

However that's kinda silly since a uuid really isn't guessable and even it it was the user must be logged in.  i'll go ahead and do as you suggest

jpslav: ah ok, I see

PR #1526, opened 2017-09-11
Adding hypothesis

m1yag1:


app/controllers/webview_controller.rb
```diff
@@ -4,7 +4,15 @@ class WebviewController < ApplicationController
 
   layout :resolve_layout
 
-  skip_before_filter :authenticate_user!, only: [:home, :enroll]
+  # Requested by an OPTIONS request type
+  def cors_preflight_check # the other CORS headers are set by the before_filter
+    headers['Access-Control-Max-Age'] = '1728000'
+    headers['Access-Control-Allow-Origin'] = Rails.application.secrets[:hypothesis]['client_url']
+    render text: '', content_type: 'text/plain'
+  end
```
jpslav: @m1yag1 -- is this code operational?  (don't see it connected anywhere); if you need it, talk to @nathanstitt about reusing similar code in API methods

m1yag1: What do you mean about connected anywhere? I was not able to get CORS requests to work without the code I placed here.

m1yag1: I don't know Rails that well so maybe it should go somewhere else? 

Additionally, yes the code is operation. I am currently able to use one login with a hack that I still need to correct once tutor-js is able to load the sidebar.

app/views/layouts/webview.html.erb
```diff
@@ -7,11 +7,35 @@
...
+    window.hypothesisConfig = function() {
+        return {
+            "services": [
+            {
+                "authority": "<%= Rails.application.secrets[:hypothesis]['authority'] %>",
+                "grantToken": "<%= generate_hypothesis_token %>",
+                "icon": "https://openclipart.org/download/272629/sihouette-animaux-10.svg",
+                "apiUrl": "<%= Rails.application.secrets[:hypothesis]['api_url'] %>"
+            }
+        ]
```
nathanstitt: Can this be included in the bootstrap data, or is it needed by Hypothesis?

m1yag1: That's the hack i keep talking about. that will be changed and become an api call from tutor-js to an endpoint on the hypothesis side for the grant token.

m1yag1: OR --- it could be passed in via the bootstrap data. i don't see why not. tutor-server should know about the user.

m1yag1: ☝️ That would actually be great (passing grant_token via bootstrapdata) as it would greatly simplify things. The tutor-server would need to take the uuid of the user, the client_id, and client secret and generate that during the bootstrap process.

app/views/layouts/webview.html.erb
```diff
@@ -7,11 +7,35 @@
   <script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-MML-AM_HTMLorMML-full&amp;delayStartupUntil=configured" async></script>
   <script type='text/javascript' src='https://c.la2-c2-dfw.salesforceliveagent.com/content/g/js/39.0/deployment.js'></script>
   <script type='text/javascript' src='<%= Rails.application.secrets[:js_url] %>' async></script>
+  <script type="text/javascript">
+    window.hypothesisConfig = function() {
```
m1yag1: This entire function is a hack and i would need to load this during the tutor-js bootstrapping process.

app/views/layouts/webview.html.erb
```diff
@@ -7,11 +7,35 @@
...
   <script id="tutor-boostrap-data" type="application/json"><%=raw json_escape bootstrap_data %></script>
+  <!-- ADD HYPOTHESIS embed script here and config -->
+  <% if current_user.present? %>
+      <script type="text/javascript" src="<%= Rails.application.secrets[:hypothesis]['embed_url']%>" ></script>
+      <!-- This needs to be removed. Only here for debugging purposes.
+      <script>"<%=raw json_escape bootstrap_data %>"</script> -->
+       <!-- This needs to be removed. Only here for debugging purposes.  -->
+
+  
+  <% end %>
```
m1yag1: This is also a hack and will be removed when everything is done.

app/controllers/webview_controller.rb
jpslav: What I mean is there is no route in `config/routes.rb` that points to this method, and this method is not a `before_filter` that I can see, so I don't see how this code is being reached.  If you comment it out, do things still work?

m1yag1: I thought whatever route called the webview loaded that code. I'll comment it out and see what happens.

m1yag1: @jpslav, you were correct. I commented out the code and it worked. I wonder why I was having CORS errors before? Either way, Thank you!

PR #1527, opened 2017-09-12
match text in the ticket for students using the teacher join url

gregfitch:


gregfitch:
didn't see that the FE one was still open...

PR #1528, opened 2017-09-13
LMS feature and enable flags with admin course list setting

jpslav:
Adds two flags to `CourseProfile::Models::Course` -- `is_lms_enabling_allowed` (an admin-controlled feature flag), and `is_lms_enabled`, a teacher-controlled flag for opting in to LMS integration.  These are reported in course JSON, and `is_lms_enabled` can be set in the patch course API call, if `is_lms_enabling_allowed` is `true`.  

The feature flag can be set via a new bulk flag setting option at the bottom of the course search admin screen, or on the course edit page.  The bulk flag setting option also respects a new "Select all on all pages" toggle.

More search terms were added to the admin course search, including the ability to search by the values of these new LMS booleans (also `costs`, `term`, `year`).  Added a toggleable search help div.

image

## TODO

* [ ] ~~Use the `is_lms_enabled` field to allow/block related actions~~ (will do this in another PR)
* [x] Be able to toggle `is_lms_enabling_allowed` and `is_lms_enabled` from the course edit screen (currently can only set via the course search screen)
* [x] specs all around

### DIFF
--- a/app/controllers/admin/courses_controller.rb
+++ b/app/controllers/admin/courses_controller.rb
@@ -6,6 +6,7 @@ class Admin::CoursesController < Admin::BaseController
 
   def index
     @query = params[:query]
+    @order_by = params[:order_by]
     result = SearchCourses.call(query: params[:query], order_by: params[:order_by] || 'id')
     per_page = params[:per_page] || 25
     per_page = result.outputs.total_count if params[:per_page] == 'all'
@@ -119,6 +120,8 @@ class Admin::CoursesController < Admin::BaseController
     case params[:commit]
     when 'Set Ecosystem'
       bulk_set_ecosystem
+    when 'Set Flag'
+      bulk_set_flag
     end
   end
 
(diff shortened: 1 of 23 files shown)


app/controllers/admin/courses_controller.rb
```diff
@@ -152,6 +155,36 @@ def bulk_set_ecosystem
...
+        end
+        flash[:notice] = 'Flag values were updated'
+      rescue ActiveRecord::RecordInvalid => invalid
+        flash[:error] = "Could not update flag value for at least one course, rolled back all changes: #{invalid.message}"
+      end
+    end
+
+    redirect_to admin_courses_path(query: params[:query], order_by: params[:order_by])
+  end
+
```
nathanstitt: We had problems with updating the `does_cost` flag inline like this, in production the request timed out before it'd complete.  Not sure if that'll happen here though since the list of `course_ids` will presumably be small.

An `update_all` would be much more efficient

app/subsystems/course_profile/create_course.rb
```diff
@@ -3,6 +3,9 @@ class CourseProfile::CreateCourse
 
   protected
   def exec(attrs = {})
+    attrs[:is_lms_enabling_allowed] = Settings::Db.store.default_is_lms_enabling_allowed
+    attrs[:is_lms_enabled] = nil
+
     outputs.course = CourseProfile::Models::Course.create(attrs)
```
nathanstitt: A thought - if a teacher clones a `lms_enabled` course, should the setting also carry over to the clone?

app/controllers/admin/courses_controller.rb
jpslav: I agree.  I started with `update_all` and then got worried about the lack of callback checks.  Tho there are no such restrictions on changes to this field now.... wish there was some way to get the best of both worlds here.  I'll ponder and if nothing comes to me shortly I'll just go with `update_all` :-)

app/subsystems/course_profile/create_course.rb
jpslav: That is an interesting question.  I suspect that we still want to teachers to opt-in to LMS each time, so they make an explicit choice, know they are getting new keys, etc.... but we should ask @kjdav

jpslav:
@nathanstitt I made the `update_all` change and set a reminder for Monday to talk about course cloning behavior - based on what kim says, we can make a new card that can be done later (ok since no courses have LMS on that can be cloned now)

PR #1531, opened 2017-09-14
launch immediately into new tab before login

jpslav:
Immediately show the "open in new tab" link if in an iframe, before going to authentication


db/migrate/20170915024702_add_created_at_to_lms_trusted_launch_data.rb
```diff
@@ -0,0 +1,6 @@
+class AddCreatedAtToLmsTrustedLaunchData < ActiveRecord::Migration
+  def change
+    # add a default since some dev environment instances may already exist
+    add_column :lms_trusted_launch_data, :created_at, :datetime, default: '2017-09-14 00:00:00'
+  end
```
nathanstitt: should this also be `null: false` ?

nathanstitt:
Looks good @jpslav

PR #1535, opened 2017-09-20
reorganize launch flow to support errors as early as possible

jpslav:
With some error checking moving into the early part of the launch, I've put all of the launch code directly in the LMS controller (no longer have handlers).  Handlers are good for taking dissimilar action logic out of one big file, but here all of the controller actions are related (they are steps in a flow), so putting the logic inline made the most sense.


app/controllers/lms_controller.rb
```diff
@@ -87,4 +162,31 @@ def allow_embedding_in_iframe
...
+  def fail_for_course_keys_already_used(launch)
+    Rails.logger.info { "Launch #{session[:launch_id]} failing because course keys already used for another context"}
+    render :fail_course_keys_already_used, locals: { launch: launch }, status: :unprocessable_entity
+  end
+
+  def fail_with_catchall_message(exception)
+    Rails.logger.info { "Launch failed: #{exception.try(:message)}" }
+    render(:fail_catchall, status: :unprocessable_entity)
+  end
+
```
nathanstitt: do we really need different views for each error?  I thought we'd be able to use one view with a message lookup function.  I guess it's pretty much the same though and this way we can style them independantly

app/subsystems/lms/launch.rb
```diff
@@ -94,6 +106,64 @@ def app
...
+      @context = query.first
+    end
+    @context
+  end
+
+  def missing_required_fields
+    @missing_required_fields = REQUIRED_FIELDS.select do |required_field|
+      send(required_field).blank?
+    end
+  end
```
nathanstitt: should this use `||=` so it's cached? i.e. 

`@missing_required_fields ||= REQUIRED_FIELDS.select`

app/subsystems/lms/launch.rb
```diff
@@ -94,6 +106,64 @@ def app
...
+    raise LmsDisabled if !course.is_lms_enabled
+
+    # In theory, we could allow multiple LTI context IDs to point to the same Tutor course (e.g. if
+    # one teacher has 3 sections and uses 3 LMS courses to point to one Tutor course).  For this reason
+    # we don't currently prohibit this with a database constraint.  But for the time being, we do
+    # restrict this here in code by requiring, in the auto create call, that a course only be used
+    # in one Context.
+
+    raise CourseKeysAlreadyUsed if Lms::Models::Context.where(course: course).any?
+
```
nathanstitt: `.any?` will run the query and build the AR objects, and then call the array method `.any?` to see if the array is non-empty.

However if you use `.exist?` it'll be much more efficient since it'll change the query to be the `select 1 AS one ... limit 1` and see if anything's returned.

https://andreigridnev.com/empty-blank-any-exists-methods-of-ruby-on-rails-activerecord/

spec/subsystems/lms/launch_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 RSpec.describe Lms::Launch do
 
+  # TODO write a bunch of specs here
+
   xit 'verifies signature' do
```
nathanstitt: lol

app/views/lms/fail_catchall.html.erb
```diff
@@ -0,0 +1 @@
+<p>Something went wrong when trying to open OpenStax Tutor.  Please go back and try to open the Tutor link again; that may resolve the issue.  If not, please <%= link_to "contact Support", TUTOR_HELPDESK_URL %>.</p>
```
nathanstitt: Pretty sure Marcom's gonna want a `beta` appended to OpenStax Tutor 😁 

I think messages are still placeholders though so probably doesn't matter

app/controllers/lms_controller.rb
jpslav: Implementation wise we can use partials / layouts to share common stuff; somewhat in favor of leaving, so if we need to add more / make changes they are isolated.

app/subsystems/lms/launch.rb
jpslav: oops! good catch

app/subsystems/lms/launch.rb
jpslav: good tip didn't know that!

app/views/lms/fail_catchall.html.erb
jpslav: yep, just wanted something in there that was more real in case this gets deployed and QA/PDM/UX start to play with it.

PR #1540, opened 2017-09-25
Should fix ecosystem updates (needs testing)

Dantemss:
Fix `(NoMethodError) "undefined method 'page_id_to_page_id_map' for nil:NilClass"` when trying to update many course ecosystems at the same time

### DIFF
--- a/app/subsystems/content/models/map.rb
+++ b/app/subsystems/content/models/map.rb
@@ -118,8 +118,8 @@ class Content::Models::Map < Tutor::SubSystems::BaseModel
   end
 
   def before_save_callbacks
-    create_exercise_id_to_page_id_map,
-    create_page_id_to_page_id_map,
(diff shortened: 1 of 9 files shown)


app/subsystems/content/models/map.rb
```diff
@@ -118,6 +117,13 @@ def validate_maps
     pages_map_to_exercises
   end
 
+  def before_save_callbacks
+    create_exercise_id_to_page_id_map
+    create_page_id_to_page_id_map
+    create_page_id_to_pool_type_exercise_ids_map
+    validate_maps
```
jpslav: if one of these fails, the whole thing won't fail - that ok?

Dantemss: Don't think they prevented saving the record before. It's good to cache failed mappings too.

PR #1542, opened 2017-09-26
Add is_access_switchable flag to course profile

nathanstitt:


app/subsystems/course_membership/add_student.rb
```diff
@@ -31,5 +34,7 @@ def exec(period:, role:, student_identifier: nil,
       reassign_published_period_task_plans: reassign_published_period_task_plans,
       send_to_biglearn: send_to_biglearn
     )
+
+    run(:mark_course_enrolled, course: course)
```
jpslav: good point that it only has to be done here!

PR #1543, opened 2017-09-27
push course scores to LMS

jpslav:
Background job for pushing course scores to an LMS and an API endpoint to trigger it.

* `GET /api/lms/:course_id` changed to  `GET /api/lms/courses/:id`
* Replaces "grade" with "score", as that is our way ("grade" has too much baggage)

Need to make an LMS simulator to support testing.


app/controllers/api/v1/lms/courses_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    The number of errors + `num_missing_scores` should equal `num_callbacks` unless some other
+    unhandled error occurred.
+
+    The job progress will be updated during the push.
+  EOS
+  def push_scores
+    OSU::AccessPolicy.require_action_allowed!(:lms_sync_scores, current_api_user, @course)
+    status_id = ::Lms::SendCourseScores.perform_later(course: @course)
+    render_job_id_json(status_id)
+  end
```
nathanstitt: Something to consider for the future is that the FE doesn't use the url that's returned from jobs.  Instead [we split it on `/` to extract the uuid and then use that](https://github.com/openstax/tutor-js/blob/master/tutor/src/helpers/job.coffee#L17-L21).  Would be cleaner just to return the UUID here.

app/subsystems/lms/send_course_scores.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    if "failure" == outcome_response.code_major
+      error!({
+        score: score_data[:average_score],
+        student_name: score_data[:name],
+        student_identifier: score_data[:student_identifier],
+        lms_description: outcome_response.description
+      })
+    end
+  end
```
nathanstitt: Does `send_one_score` need to have a `rescue` ?  What happens if send times out?  I see that auth does use `begin ... ensure` but I think @response will be nil if that happens.

https://github.com/oauth-xx/oauth-ruby/blob/master/lib/oauth/tokens/access_token.rb#L10-L17

PR #1552, opened 2017-10-03
Misc LMS changes (jobs, student count in period json, more LMS enabled change validations)

jpslav:
* Timestamp fields on job JSON
* Last LMS scores push job ID now stored on course
* count of enrolled students in period JSON
* disallow changes to `is_lms_enabled` when `is_access_switchable` is false
* killed a spec supplanted by a richer request spec


app/representers/api/v1/period_representer.rb
```diff
@@ -37,6 +37,15 @@ class PeriodRepresenter < Roar::Decorator
                description: "The period's enrollment URL"
              }
 
+    property :num_enrolled_students,
+             type: Integer,
+             readable: true,
+             writeable: false,
+             getter: ->(*) { to_model.students.where(deleted_at: nil).count },
+             schema_info: {
```
nathanstitt: This is a good example of stuff that is harder because of the wrappers.  Your bypassing it via `to_model` so it seems like they're not really useful.

PR #1554, opened 2017-10-04
LMS error pages (with styles and copy)

jpslav:


app/controllers/lms_controller.rb
```diff
@@ -14,6 +14,11 @@ class LmsController < ApplicationController
   before_filter :allow_embedding_in_iframe, only: [:launch, :ci_launch]
 
   layout false
+  layout 'minimal_error', only: [:fail_unsupported_role,
+                                 :fail_lms_disabled,
+                                 :fail_missing_required_fields,
+                                 :fail_course_keys_already_used,
+                                 :fail_catchall]
```
jpslav: I need to remove these, now handled with a `layout: 'minimal_error'` arg in the `render` call -- don't think this top `layout` command even worked because wasn't redirecting to these actions.

spec/rails_helper.rb
```diff
@@ -215,3 +217,8 @@ def create_contract!(name)
     contract.content = name + ' content'
   end
 end
+
+def with_test_routes(&block)
+  test_routes = ActionDispatch::Routing::RouteSet.new.send(:eval_block, block)
+  allow(Rails.application).to receive(:routes) { test_routes }
+end
```
jpslav: Need to remove this too.  Failed attempt at test routes.

app/views/layouts/minimal_error.html.erb
```diff
@@ -0,0 +1,57 @@
...
+        }
+      }
+
+      if (!inIframe()) {
+        document.getElementById('iframe').style.display = 'none';
+        document.getElementById('full-page').style.display = 'block';
+      } else {
+        document.getElementById('minimal_error').style.background = 'white';
+      }
+    })();
```
nathanstitt: An alternative approach would be to do something like:
```javascript
if (inFrame) {
  document.body.classList.add('in-frame');
}
```

Then later in the css you can style conditionally by doing stuff like:
```css
.error_card {
  ... styles ...
 &.iframe {
   display: none;
  }
}
.in-frame .error-card {
  &.iframe {
   display: block;
  }
}
```

PR #1566, opened 2017-10-09
Remove CC courses from demo script

nathanstitt:


Dantemss:
Woohoo... though what is the smallest book now? I guess soc? Or maybe the sample apush book.

PR #1573, opened 2017-10-12
redirect launching dropped students to course, not enrollment

jpslav:
For students who have been dropped inside Tutor but not from the LMS, prevents them from relaunching from the LMS and getting re-enrolled.


Dantemss:
Need to test what happens to dropped students when they hit this, because I don't know if they are even allowed to see the course dashboard...

nathanstitt:
@Dantemss yes we also wrote a FE handler for that so they see a “you’ve been dropped message”

PR #1580, opened 2017-10-25
Keep only one score callback per link / user combo; email devs on error

jpslav:
Some LMS's change the sourcedid per launch.  We don't want to accumulate a bunch of callbacks for one user and one LMS link when they are all reporting the same score.  Furthermore, the LTI 1.1 implementation guide even says that we should only keep the latest.

Changes the code...
* to make callback records include the `resource_link_id` which is a required parameter in the launch and which identifies the link that was clicked.  
* to keep only one callback per `resource_link_id` / user combo
* to somewhat unrelatedly email devs when there's an error syncing scores.

### DIFF
--- a/app/subsystems/lms/launch.rb
+++ b/app/subsystems/lms/launch.rb
@@ -52,6 +52,10 @@ class Lms::Launch
     request_parameters[:lis_result_sourcedid]
   end
 
+  def resource_link_id
+    request_parameters[:resource_link_id]
+  end
+
   def outcome_url
     request_parameters[:lis_outcome_service_url] ||
     request_parameters[:ext_ims_lis_basic_outcome_url]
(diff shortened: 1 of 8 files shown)


app/subsystems/lms/launch.rb
```diff
@@ -168,22 +172,33 @@ def update_tool_consumer_metadata!
...
+      ).destroy_all
+
+      Lms::Models::CourseScoreCallback.create!(
+        resource_link_id: resource_link_id,
+        result_sourcedid: result_sourcedid,
+        outcome_url: outcome_url,
+        course: context.course,
+        profile: user.to_model
+      )
+    end
```
Dantemss: I suggest you do this little magic:

```
callback = Lms::Models::CourseScoreCallback.new(
  resource_link_id: resource_link_id,
  result_sourcedid: result_sourcedid,
  outcome_url: outcome_url,
  course: context.course,
  profile: user.to_model
)

Lms::Models::CourseScoreCallback.import [ callback ], on_duplicate_key_update: {
  conflict_target: [ :course, :profile, resource_link_id ],
  columns: [ :result_sourcedid, :outcome_url ]
}
```

jpslav: whoa.... why?  b/c faster?

Dantemss: It's definitely faster (1 query instead of 2). Might also be safer if multiple things attempt to change the same callback.

Dantemss: Assuming you have the validations in the DB as well (NOT NULL columns and unique indices), then you can do this which skips the Rails validations:

```
Lms::Models::CourseScoreCallback.import [ callback ], validate: false, on_duplicate_key_update: {
  conflict_target: [ :course, :profile, resource_link_id ],
  columns: [ :result_sourcedid, :outcome_url ]
}
```

PR #1582, opened 2017-10-27
Cache Task stats

Dantemss:
- Added routine for caching task stats per page
- Made caching routine be scheduled to run whenever a task is updated
- Use the cache in Performance Forecast
- Use the cache in Quick Look
- Use the cache when computing the trouble flags in the dashboard


app/routines/get_student_guide.rb
```diff
@@ -4,13 +4,8 @@ class GetStudentGuide
...
 
-  def get_course_guide(role)
-    period = role.student.period
-    course = period.course
-    history = get_history_for_roles(role)
-    ecosystems_map = get_course_ecosystems_map(course)
-
-    get_period_guide(period, role, history, ecosystems_map, :student)
+  def exec(role:)
+    outputs.course_guide = get_course_guide(students: role.student, type: :student).first
```
jpslav: Ummm... I can't find `get_course_guide` except for the two places where this PR removes it?

app/subsystems/course_content/add_ecosystem_to_course.rb
```diff
@@ -31,9 +31,21 @@ def exec(course:, ecosystem:,
...
+      # Saving is necessary so the course can be sent to Biglearn
+      # because we cannot serialize unsaved AR objects
+      course.save
+    else
+      # Recalculate all TaskPageCaches since we need to map them to the new ecosystem
+      tasks = Tasks::Models::Task
+        .select(:id)
+        .joins(taskings: { role: :student })
+        .where(taskings: { role: { student: { course_profile_course_id: course.id } } })
+        .to_a
```
jpslav: should we put some of this code into `course.tasks`?

app/routines/calculate_task_stats.rb
```diff
@@ -2,179 +2,215 @@ class CalculateTaskStats
...
-          answer_stats = compute_answer_stats(question_tasked_exercises)
-          all_answer_ids = question_tasked_exercises.first.answer_ids
-          completed_question_tasked_exercises = question_tasked_exercises.select(&:completed?)
+      periods.first
+    end
 
-          {
-            question_id: question_id,
+    # Get unmapped cached Task stats (for the Task's original ecosystem)
+    tc = Tasks::Models::TaskCache.arel_table
```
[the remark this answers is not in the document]
Dantemss: Will remove.

app/routines/get_student_guide.rb
Dantemss: It's in the CourseGuideRoutine class.

app/subsystems/course_content/add_ecosystem_to_course.rb
Dantemss: I'd rather not do that because:

1. There are lots of hops in between course and tasks so it's unclear how many records are loaded when you call course.tasks
2. There are multiple ways to go between task and periods and because of all the polymorphic records in the path all of them make assumptions. For example, you could go task -> task_plan -> owner (assuming owner is a course). Or task -> taskings -> role -> student -> course (assuming all roles are students and in the same course). Or task -> taskings -> period -> course (which might be the most reliable, but this period doesn't change even if the student moved periods, which could be a problem). Mostly relevant in tests, though, where it might be hard to figure out which record you need to update to get the course to change somewhere else.

Dantemss: If we decide to fix all this confusion (in a separate PR) and make it so there's only 1 way to go from a task to a course and vice-versa then I'll gladly do it.

PR #1583, opened 2017-10-27
Add launch url to LMS json

nathanstitt:
Maybe this should be a representer?  If so we'll have to figure out how to calc the urls in it though


jpslav:
https://github.com/openstax/tutor-server/blob/master/app/representers/api/v1/period_representer.rb#L35 URLs in representers

PR #1584, opened 2017-10-31
Added routines to archive/unarchive periods that also send updates to Biglearn

Dantemss:
The archive routine also checks if any students should be refunded.


jpslav:
Nice PR!

PR #1588, opened 2017-11-06
Add `host: localhost` to database.yml

jpslav:
Otherwise I got some PG error


nathanstitt:
should be fine I think as long as Travis doesn't care.  Some configurations may be using unix sockets

PR #1601, opened 2017-11-13
allow restoring a student even if the id is in use

nathanstitt:
Also the code had a bug where it was checking all students, including the current one so it'd always fail this check


nathanstitt:
[Bug: Cannot re-add a student back to a course after dropping that student](https://trello.com/c/6mcTvfus/1079-bug-cannot-re-add-a-student-back-to-a-course-after-dropping-that-student)

PR #1607, opened 2017-11-27
Added script to remove bad exercises from all assignments

Dantemss:
`rake exercises:remove[1@1,2@2...etc]`

### DIFF
--- a/lib/tasks/exercises.rake
+++ b/lib/tasks/exercises.rake
@@ -1,0 +1,39 @@
+require 'values_table'
+
+namespace :exercises do
+  desc 'Removes defective exercises from student assignments'
+  task remove: :environment do |task, args|
+    uids = args.extras
+    fail 'You must provide one or more exercise number@version to remove' if uids.empty?
+
+    split_values = uids.map { |value| value.split('@') }
+                       .reject{ |number, version| version.nil? }
+                       .map { |number, version| [ number.to_i, version.to_i ] }
+
+    exercises_join_query = <<-JOIN_SQL
+      INNER JOIN (#{ValuesTable.new(split_values)})
+        AS "ex" ("number", "version")
+        ON "content_exercises"."number" = "ex"."number"
+          AND "content_exercises"."version" = "ex"."version"
+    JOIN_SQL
+
+    Tasks::Models::Task.transaction do
+      tasked_exercises = Tasks::Models::TaskedExercise
+        .select(:id)
+        .joins(:exercise)
+        .joins(exercises_join_query)
+      tasked_exercise_ids = tasked_exercises.map(&:id)
+
+      task_steps = Tasks::Models::TaskStep
+        .select([ :id, :tasks_task_id ])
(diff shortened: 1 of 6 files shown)


lib/values_table.rb
```diff
@@ -0,0 +1,35 @@
+class ValuesTable
```
jpslav: Think we need a spec for this class?

jpslav: If no spec, would probably be good to add some comments about what this class does

lib/tasks/exercises.rake
```diff
@@ -0,0 +1,39 @@
+require 'values_table'
+
+namespace :exercises do
+  desc 'Removes defective exercises from student assignments'
+  task remove: :environment do |task, args|
+    uids = args.extras
+    fail 'You must provide one or more exercise number@version to remove' if uids.empty?
+
+    split_values = uids.map { |value| value.split('@') }
+                       .reject{ |number, version| version.nil? }
```
jpslav: Might the split result in `version` being whitespace? If so, need `blank?`

spec/lib/tasks/exercises_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      @tasks.each(&:reload)
+    end.to change { Tasks::Models::TaskedExercise.count }.by(-num_tasks)
+       .and change { Tasks::Models::TaskStep.count }.by(-num_tasks)
+       .and not_change { Tasks::Models::Task.count }
+       .and change { @tasks.map(&:steps_count).reduce(0, :+) }.by(-num_tasks)
+       .and change { @tasks.map(&:completed_steps_count).reduce(0, :+) }.by(-num_tasks)
+       .and change { @tasks.map(&:exercise_steps_count).reduce(0, :+) }.by(-num_tasks)
+       .and change { @tasks.map(&:completed_exercise_steps_count).reduce(0, :+) }.by(-num_tasks)
+       .and change { @tasks.map(&:correct_exercise_steps_count).reduce(0, :+) }.by(-num_tasks)
```
jpslav: Are there specs somewhere that deleting an entry from a `sortable_has_many` is good?  I looked in `openstax/sortability` but didn't see much.

lib/tasks/exercises.rake
```diff
@@ -0,0 +1,39 @@
+require 'values_table'
+
+namespace :exercises do
+  desc 'Removes defective exercises from student assignments'
```
jpslav: Can you give an example call up here?

lib/values_table.rb
Dantemss: Added spec

spec/lib/tasks/exercises_spec.rb
Dantemss: The guy who wrote sortability put this in the readme:

```
The sort_position is not guaranteed to contain consecutive numbers. When listing the records, you should compute their positions in application code as you iterate through the list. If you need the position for a single record, call compact_sort_position_peers! first to close any gaps in its peers, then read sort_position directly.
```

We already deal with this though (for example, when placeholder steps are deleted because the teacher excluded all possible exercises... etc)

PR #1614, opened 2017-12-05
Add another level of caching for PF/Trouble Flag

Dantemss:


app/subsystems/tasks/update_period_caches.rb
```diff
@@ -0,0 +1,202 @@
...
+    locked_periods = CourseMembership::Models::Period
+      .select(:id)
+      .where(id: periods.map(&:id))
+      .lock('FOR NO KEY UPDATE SKIP LOCKED')
+    period_ids = locked_periods.map(&:id)
+
+    # Retry periods that we couldn't lock later
+    skipped_periods = periods - locked_periods
+    self.class.perform_later(periods: skipped_periods) unless skipped_periods.empty?
+
```
nathanstitt: Should we look into task queues for background jobs.  Sometimes the queue gets long and I wonder if it might potentially be quite a while before this actually runs.  We could add a "fast" queue for jobs like this that will be fast

app/subsystems/tasks/update_task_caches.rb
```diff
@@ -178,8 +183,12 @@ def exec(tasks:)
...
     Tasks::Models::TaskCache.import task_caches, validate: false, on_duplicate_key_update: {
       conflict_target: [ :tasks_task_id, :content_ecosystem_id ],
-      columns: [ :opens_at, :due_at, :feedback_at, :student_ids, :as_toc ]
+      columns: [ :opens_at, :due_at, :feedback_at, :student_ids, :student_names, :as_toc ]
     }
+
+    # Update the PeriodCaches
+    periods = CourseMembership::Models::Period.select(:id).where(id: period_ids.uniq)
+    Tasks::UpdatePeriodCaches.perform_later(periods: periods.to_a)
   end
```
nathanstitt: do we want `periods` to be instances of the `Periods` model, or would id's be fine?   Using `pluck` instead of `select` might be better since you're calling `to_a` on it

app/subsystems/tasks/update_task_caches.rb
```diff
@@ -298,6 +307,7 @@ def build_task_cache(
       title: ecosystem.title,
       has_exercises: books_array.any? { |bk| bk[:has_exercises] },
       num_assigned_steps: num_assigned_steps,
+      num_known_location_steps: books_array.map { |bk| bk[:num_assigned_steps] }.reduce(0, :+),
       num_completed_steps: books_array.map { |bk| bk[:num_completed_steps] }
```
nathanstitt: Would `books_array.sum { |bk| bk[:num_assigned_steps] }`  or even `books_array.sum(&:num_assigned_steps)` work here?  That'd be more efficient but will fail if one of the values is null

app/routines/get_student_guide.rb
```diff
@@ -1,11 +1,118 @@
...
+      .select([ :tasks_task_id, :content_ecosystem_id, :task_type, :as_toc ])
+      .where(content_ecosystem_id: ecosystem_ids)
+      .where("\"tasks_task_caches\".\"student_ids\" @> ARRAY[#{student.id}]")
+      .where(tc[:opens_at].eq(nil).or tc[:opens_at].lteq(current_time))
+      .sort_by { |task_cache| index_by_ecosystem_id[task_cache.content_ecosystem_id] }
+      .uniq { |task_cache| task_cache.tasks_task_id }
+
+    # Get cached Task stats by chapter
+    chs = task_caches.flat_map do |task_cache|
+      practice = task_cache.practice?
```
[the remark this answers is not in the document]
Dantemss: `flat_map` is needed to flatten the result of the block (nested maps in the block)

`find_each` probably wouldn't help too much because we are accumulating the `chs` so memory is consumed regardless but yeah it could definitely be a problem if there are some massive classes with tons of students (1 TaskCache per student and this routine processes 1 period at a time)

app/routines/get_student_guide.rb
```diff
@@ -1,11 +1,118 @@
...
+    # Get the Student CLUes from Biglearn
+    biglearn_responses = OpenStax::Biglearn::Api.fetch_student_clues(biglearn_requests)
+    biglearn_clue_by_book_container_uuid = biglearn_responses.map do |request, response|
+      [ request[:book_container].tutor_uuid, response ]
+    end.to_h
+
+    # Merge the TaskCache ToCs and create the Student Performance Forecast
+    chapter_guides = chs.group_by { |ch| ch[:book_location] }.sort.map do |book_location, chs|
+      pgs = chs.flat_map { |ch| ch[:pages] }
+
```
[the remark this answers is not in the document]
Dantemss: The pages are not unique, but they are merged based on `book_location` below.

app/subsystems/tasks/update_task_caches.rb
Dantemss: Well the routine expects periods but I guess you are right it could take period_ids instead... Maybe a performance optimization for a future PR though?

app/subsystems/tasks/update_task_caches.rb
Dantemss: Using &: doesn't work because these are plain hashes (would throw NoMethodErrors).
How did nobody ever tell me about `sum` though?

app/routines/get_teacher_guide.rb
```diff
@@ -1,21 +1,147 @@
...
+        :student_ids,
+        :as_toc
+      ])
+      .where(content_ecosystem_id: ecosystem_ids, course_membership_period_id: period_ids)
+      .where(pc[:opens_at].eq(nil).or pc[:opens_at].lteq(current_time))
+      .sort_by { |period_cache| index_by_ecosystem_id[period_cache.content_ecosystem_id] }
+      .uniq do |period_cache|
+        [ period_cache.course_membership_period_id, period_cache.tasks_task_plan_id ]
+      end
+
```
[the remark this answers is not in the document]
Dantemss: Oh I totally missed this one when you did the actual review but no, the `index_by_ecosystem_id` is built above and it uses the same ordering that courses use to determine their newest ecosystem (order by date ecosystem was assigned to course I believe).

PR #1618, opened 2017-12-08
 Fix dashboard trouble flags and period cache creation memory consumption

Dantemss:
Also fixed a flaky spec that was failing.


app/routines/get_tp_dashboard.rb
```diff
@@ -41,27 +41,25 @@ def load_plans(course, start_at_ntz, end_at_ntz, current_time = Time.current)
 
     outputs.plans = result.plans.map do |task_plan|
       period_caches = period_caches_by_task_plan_id[task_plan.id] || []
+      total_students = period_caches.flat_map { |pc| pc.student_ids }.uniq.size
       pgs = period_caches.flat_map do |period_cache|
```
nathanstitt: I suspect that using `Set` to find the unique id count might be faster/use less memory.
```ruby
period_caches.reduce(Set.new) { |s, pc| s.merge(pc) }.size
```
I tested in IRB with: 

`[[1,2,4],[1,2]].reduce(Set.new){ |s, v| s.merge(v) }` and it returns: `#<Set: {1, 2, 4}>`

app/subsystems/tasks/update_period_caches.rb
```diff
@@ -151,7 +146,8 @@ def build_period_cache(period:, ecosystem:, task_plan:, tasking_plan:, task_cach
           num_assigned_exercises: chs.map { |ch| ch[:num_assigned_exercises] }.reduce(0, :+),
           num_completed_exercises: chs.map { |ch| ch[:num_completed_exercises] }.reduce(0, :+),
           num_correct_exercises: chs.map { |ch| ch[:num_correct_exercises] }.reduce(0, :+),
-          num_assigned_placeholders: chs.map { |ch| ch[:num_assigned_placeholders] }.reduce(0, :+),
+          num_assigned_placeholders: chs.map { |ch| ch[:num_assigned_placeholders] }.reduce(0, :+),          student_ids: chs.flat_map { |ch| ch[:student_ids] }.uniq,
+          student_names: chs.flat_map { |ch| ch[:student_names] }.uniq,
           pages: period_pgs
```
nathanstitt: if `ch[:num_assigned_placeholders]` will be completely numbers (without nil), then you could call `sum()` on it, 
and the reduce with Set from comment above would probably work in place of the `flat_map` w/`uniq`

If this works, then I think there's a few spots below where it could also be applied

spec/subsystems/tasks/update_period_caches_spec.rb
```diff
@@ -71,22 +77,8 @@
...
-                      free_response: nil,
-                      selected_answer_id: nil,
-                      completed: false,
-                      correct: false,
-                      student_ids: kind_of(Array),
-                      student_names: kind_of(Array)
-                    }
-                  end * 10
+                  student_ids: [ kind_of(Integer) ] * 10,
+                  student_names: [ kind_of(String) ] * 10
```
nathanstitt: Nifty use of `*` for a kind_of matcher 😁

Dantemss: :)

app/subsystems/tasks/update_period_caches.rb
Dantemss: I definitely plan to replace all `map { whatever }.reduce(0,  :+)` with `sum { whatever }` but felt like I should leave those for another PR.

PR #1619, opened 2017-12-11
Fix PushSalesforceCourseStats error on soft-deleted SF records

Dantemss:


app/subsystems/salesforce/attach_record.rb
```diff
@@ -3,13 +3,12 @@ class AttachRecord
...
 
     def exec(record:, to:)
-      model = Models::AttachedRecord.create(
+      outputs.attached_record = Models::AttachedRecord.create(
         tutor_gid: to.to_global_id.to_s,
         salesforce_class_name: record.class.name,
         salesforce_id: record.id
       )
-      transfer_errors_from(model, {type: :verbatim}, true)
-      outputs.attached_record = AttachedRecord.new(strategy: model)
```
jpslav: It probably doesn't matter, but from a risk-aversion perspective, getting rid of the `model` variable means that if there is an error in saving the `AttachedRecord`, `outputs.attached_record` will now be set, whereas before it wouldn't be.

PR #1623, opened 2017-12-12
Prepend "r" to the research identifiers

Dantemss:


lib/token_generator/secure_random_token_generator.rb
```diff
@@ -7,7 +7,9 @@ def initialize(mode, options)
   end
 
   def run
-    send("run_#{mode}")
+    result = send("run_#{mode}")
+
+    result.is_a?(String) ? "#{options[:prefix] || ''}#{result}#{options[:suffix] || ''}" : result
```
jpslav: Don't think you need `|| ''`

```
[5] pry(main)> "#{nil}hi"
=> "hi"
```

Dantemss: I think you are right

PR #1624, opened 2017-12-14
Updated accounts-rails to add support_identifier and is_test to user accounts

Dantemss:
Also switched the Tutor user search to use the user search routines in accounts-rails.
Courses created or cloned by test accounts marked as test courses.

### DIFF
--- a/app/subsystems/lms/launch.rb
+++ b/app/subsystems/lms/launch.rb
@@ -229,7 +229,7 @@ class Lms::Launch
           request_parameters,
           app.secret
         )
-
+debugger if !authenticator.valid_signature?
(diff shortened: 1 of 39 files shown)


app/subsystems/lms/launch.rb
```diff
@@ -229,7 +229,7 @@ def initialize(request_parameters:, request_url:, trusted: false)
           request_parameters,
           app.secret
         )
-
+debugger if !authenticator.valid_signature?
```
jpslav: Hey are you trying to break LMS? ;-)

jpslav: Probably indicates a missing spec on my part :-)

PR #1643, opened 2018-01-05
More research exports

Dantemss:
- Added CNX (fragment) export
- Added exercise export


app/controllers/api/v1/enrollment_controller.rb
```diff
@@ -61,8 +59,7 @@ def prevalidate
     end
   end
 
-  api :POST, '/enrollment_changes',
-             'Creates a new EnrollmentChange request or updates the current one'
+  api :POST, '/enrollment', 'Creates a new EnrollmentChange request or updates the current one'
```
jpslav: Is FE clued in to this route change? and the one below

Dantemss: Just a documentation change. The route change already happened apparently.

PR #1644, opened 2018-01-08
assign missing surveys routine

jpslav:
Handles assigning surveys when (1) survey plans are published, (2) courses are added to a study with previously-published survey plans, (3) a student is added to a course in a study with previously-published survey plans.

### DIFF
--- a/app/controllers/research/survey_plans_controller.rb
+++ b/app/controllers/research/survey_plans_controller.rb
@@ -7,7 +7,7 @@ class Research::SurveyPlansController < Research::BaseController
   end
 
   def create
-    @survey_plan = Research::Models::SurveyPlan.new(cleared_params)# params[:research_models_survey_plan])
+    @survey_plan = Research::Models::SurveyPlan.new(cleared_params)
 
     respond_to do |format|
       if @survey_plan.save
@@ -39,7 +39,7 @@ class Research::SurveyPlansController < Research::BaseController
   def publish
     respond_to do |format|
       begin
-        Research::PublishSurveyPlan[survey_plan: Research::Models::SurveyPlan.first]
+        Research::PublishSurveyPlan[survey_plan: @survey_plan]
         format.html { redirect_to research_survey_plans_path, notice: "Published survey plan #{@survey_plan.id}"}
       rescue
         format.html { redirect_to research_survey_plans_path, alert: "Could not publish survey plan #{@survey_plan.id}"}
--- a/app/subsystems/research/assign_missing_surveys.rb
+++ b/app/subsystems/research/assign_missing_surveys.rb
@@ -1,0 +1,101 @@
+class Research::AssignMissingSurveys
+
+  lev_routine
+
+  def exec(survey_plan: nil, course: nil, student: nil)
+    # Only one input should be present, if multiple are set only one
+    # will be used.
+
+    missing_surveys = missing_surveys_for_student(student) ||
+                      missing_surveys_for_course(course) ||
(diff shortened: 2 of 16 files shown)


app/subsystems/research/assign_missing_surveys.rb
```diff
@@ -0,0 +1,101 @@
...
+            NOT EXISTS (
+              SELECT *
+              FROM "research_surveys"
+              WHERE "research_surveys"."course_membership_student_id" = #{student.id}
+              AND "research_surveys"."research_survey_plan_id" = "research_survey_plans"."id"
+            )
+          WHERE_SQL
+        )
+        .pluck(:id)
+
```
nathanstitt: If the "raw" sql bothers you, you could fallback on `arel` to write the query.  https://stackoverflow.com/questions/6207139/how-to-do-where-exists-in-arel

I find the sql easier to read than arel though so I think this is good.   The only time I'd use arel is if there were concerns with escaping the sql, such as if `student.id` was a string.

app/controllers/research/survey_plans_controller.rb
```diff
@@ -7,7 +7,7 @@ def new
   end
 
   def create
-    @survey_plan = Research::Models::SurveyPlan.new(cleared_params)# params[:research_models_survey_plan])
+    @survey_plan = Research::Models::SurveyPlan.new(cleared_params)
```
Dantemss: It's not in this diff, but I think on line 51, cleared_params has to say params.require(:blah).permit(:blah, :blah, :blah) or else you're going to get a nil error if the main key is not provided.

app/subsystems/research/assign_missing_surveys.rb
jpslav: it all bothers me lol.  I agree this SQL is pretty easy to read, thanks to @Dantemss for writing it out :-)

spec/factories/research/survey.rb
```diff
@@ -0,0 +1,14 @@
...
+  factory :research_survey, class: '::Research::Models::ToolConsumer' do
+    association :survey_plan, factory: :research_study
+    association :student, factory: :course_membership_student
+
+    trait :completed do
+      completed_at { Time.now }
+    end
+
+    trait :hidden do
+      permanently_hidden_at { Time.now }
```
Dantemss: Nothing is permanent :) jk

spec/subsystems/research/assign_missing_surveys_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  # The specs for this routine are inside other routine specs (those
+  # that call AssignMissingSurveys).
+  #
+  #
+  # research/publish_survey_plan_spec  - tests AssignMissingSurveys[survey_plan: ...]
+  # research/add_course_to_study_spec  - tests AssignMissingSurveys[course: ...]
+  # course_membership/add_student_spec - test AssignMissingSurveys[student: ...]
+
+end
```
Dantemss: I was going to say to delete this file but I guess ideally we would split the other tests into 2 so I guess it's fine to leave it.

app/controllers/research/survey_plans_controller.rb
jpslav: Ah cool.  Things I don't know because I never really bothered to learn :-)

spec/factories/research/survey.rb
jpslav: True :-) I wrote the "unhide" routine when they ask us to undo a "permanent hide"

PR #1645, opened 2018-01-11
Do not track users without SF contact ID arriving my courses

jpslav:
Users may get to "My Courses" without a Salesforce Contact ID.  If they do, we don't want to try to track them with an onboarding event because there is nothing to track them by (and our code will explode with an exception email).


Dantemss:
Thank you for this.

PR #1646, opened 2018-01-11
Hide surveys

jpslav:
Also adds an "unhide" just in case.

### DIFF
--- a/spec/features/research/survey_plans_spec.rb
+++ b/spec/features/research/survey_plans_spec.rb
@@ -11,18 +11,12 @@ RSpec.feature 'Researcher working with survey plans' do
     @student_1 = AddUserAsPeriodStudent[period: @period, user: @student_1_user].student
     @student_2 = AddUserAsPeriodStudent[period: @period, user: @student_2_user].student
 
-    @study = FactoryBot.create :research_study
+    @study = FactoryBot.create :research_study, name: "Study 1"
     Research::AddCourseToStudy[course: @course, study: @study]
-  end
-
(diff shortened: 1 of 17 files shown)


spec/features/research/survey_plans_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    @student_1 = AddUserAsPeriodStudent[period: @period, user: @student_1_user].student
+    @student_2 = AddUserAsPeriodStudent[period: @period, user: @student_2_user].student
+
+    @study = FactoryBot.create :research_study, name: "Study 1"
+    Research::AddCourseToStudy[course: @course, study: @study]
+
+    researcher = FactoryBot.create(:user, :researcher)
+    stub_current_user(researcher)
+  end
+
```
nathanstitt: Do you prefer a before/background block with instance variables instead of using `let`?

Am curious on your reasoning, maybe good to discussion with rest of team sometime

nathanstitt:
Should we rename `permanently_hidden` to  just `hidden` or maybe `withdrawn` ?

spec/features/research/survey_plans_spec.rb
jpslav: Sometimes I do this when I need more than just one variable and the variables work together.... tho of course you can often do that with `let`s too.  Here with `stub_current_user` we need a before block

jpslav:
Probably so on `permanently_hidden`.  And I guess we could add an unhide button.

PR #1649, opened 2018-01-16
allow iframe lms by using same technique as auth controller

nathanstitt:
I think the iframe options header has different case/capitalization


nathanstitt:
[Change Request: Update Canvas iframe to allow Tutor launch](https://trello.com/c/oiw7knSu/1198-change-request-update-canvas-iframe-to-allow-tutor-launch)

PR #1650, opened 2018-01-16
Score weights

Dantemss:
- Added score weights to course profile object
- Changed score export to match mockup


app/routines/concerns/tasks/performance_report_routine.rb
```diff
@@ -13,54 +13,56 @@ def included_in_averages?(task)
...
+  end
+
+  def average(array, value_getter = nil)
+    values = array.map { |item| value_getter.nil? ? item : value_getter.call(item) }.compact
+    num_values = values.length
+    return if num_values == 0
 
-    num_values == 0 ? nil : value_sum / num_values
+    values.sum / num_values.to_f
   end
```
nathanstitt: does a nil value indicate the student hasn't worked the task and they should have a 0 for it?

I'm wondering if `num_values` should be stored before calling `.compact`.  Or maybe not call `.compact`, and instead add a check for `nil` in the `map` and return 0 if the value is nil.

PR #1661, opened 2018-02-01
add first & last name to user representer

nathanstitt:


nathanstitt:
[Bug: User initials are wrong if anything is entered in the prefix or suffix](https://trello.com/c/HLY4EAuu/1273-bug-user-initials-are-wrong-if-anything-is-entered-in-the-prefix-or-suffix)

PR #1662, opened 2018-02-01
allow numbers in tour ids

nathanstitt:
This way the FE can use the course id to name them


nathanstitt:
[Change Request: Ensure returning Tutor user sends out new, correct enrollment links for new course](https://trello.com/c/NXwmul6k/1224-change-request-ensure-returning-tutor-user-sends-out-new-correct-enrollment-links-for-new-course)

PR #1664, opened 2018-02-06
Allow exercise responses to be sent on the completed API call

Dantemss:
Should allow the FE to not have to wait for the answer to save every time you click a multiple choice answer.

### DIFF
--- a/app/controllers/api/v1/task_steps_controller.rb
+++ b/app/controllers/api/v1/task_steps_controller.rb
@@ -51,7 +51,10 @@ class Api::V1::TaskStepsController < Api::V1::ApiController
       @tasked.save
       result = MarkTaskStepCompleted.call(task_step: @task_step)
 
-      render_api_errors(result.errors) || render_api_errors(@tasked.errors) || respond_with(
+      raise(ActiveRecord::Rollback) if render_api_errors(result.errors)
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -43,14 +43,20 @@ def update
...
-    result = MarkTaskStepCompleted.call(task_step: @task_step)
-    render_api_errors(result.errors) || respond_with(
-      @task_step.task,
-      responder: ResponderWithPutPatchDeleteContent,
-      represent_with: Api::V1::TaskRepresenter
-    )
+    @tasked.with_lock do
+      consume!(@tasked, represent_with: Api::V1::TaskedRepresenterMapper.representer_for(@tasked))
+      @tasked.save
+      result = MarkTaskStepCompleted.call(task_step: @task_step)
```
jpslav: Do we want to `@tasked.save!` so the transaction bails out and the `MarkTaskStepCompleted` call isn't called if the save fails?  And if so, then we'd need the `render_api_errors` on the outside of the transaction?  Worried about one of the saves completing.  Ditto on the `call` command - maybe we need an explicit rollback if `result.errors.any?`?

PR #1667, opened 2018-02-07
Fix CNX import to work with pages that only exist in the context of a CNX book

Dantemss:
Will also need FE work (I think browse the book/QA view might have been affected by the new CNX urls)


app/subsystems/tasks/get_tp_performance_report.rb
```diff
@@ -105,9 +105,9 @@ def exec(course:)
     end
 
     def filter_and_sort_tasking_plans(taskings, course, period)
-      taskings.flat_map do |tg|
-        tg.task.task_plan.tasking_plans.select { |tp| tp.target == period  || tp.target == course }
-      end.uniq.sort { |a, b| [b.due_at_ntz, b.created_at] <=> [a.due_at_ntz, a.created_at] }
+      taskings.flat_map { |tk| tk.task.task_plan.tasking_plans }.select do |tp|
+        (tp.target == period  || tp.target == course) && tp.task_plan.is_published?
+      end.uniq.sort_by { |tp| [tp.due_at_ntz, tp.created_at] }.reverse
```
jpslav: Checking that this bit of code is supposed to be in this older PR -- seems unrelated to this but related to later scores work?

lib/openstax/cnx/v1/book.rb
```diff
@@ -12,14 +11,14 @@ def initialize(id: nil, hash: nil, title: nil,
 
     attr_reader :id
 
-    def hash
-      @hash ||= OpenStax::Cnx::V1.fetch(id)
-    end
-
     def url
       @url ||= OpenStax::Cnx::V1.archive_url_for(id)
     end
```
jpslav: Is this `url` method still always accurate given the changes this PR is accounting for?

app/subsystems/tasks/get_tp_performance_report.rb
Dantemss: Yeah could be. I'll make sure it's already in and rebase to remove it from the diff.

lib/openstax/cnx/v1/book.rb
Dantemss: Book URL's are unchanged so I believe this is still valid.

PR #1668, opened 2018-02-09
Student performance reports

Dantemss:


app/subsystems/tasks/get_tp_performance_report.rb
```diff
@@ -8,78 +8,101 @@ class GetTpPerformanceReport
...
+    def exec(course:, role:)
+      is_teacher = CourseMembership::IsCourseTeacher[course: course, roles: [role]]
+      periods = if is_teacher
+        course.periods.reject(&:archived?)
+      else
+        raise(SecurityTransgression) \
+          if role.student.nil? || role.student.course_profile_course_id != course.id
+
+        [ role.student.period ]
+      end
```
nathanstitt: won't the access check in `access_policies/course_access_policy.rb` catch this?

Dantemss: You are right that this line should not be reached during normal operation. But might be a good thing to have if we call this routine in the console.

PR #1671, opened 2018-02-13
WIP: Proof of concept - Dockerize tutor-server

m1yag1:
This is a proof of concept if anyone happens to be interested. I believe running hypothesis along with tutor-server has added some more complexity that can be helped along by utilizing docker. 

I've also included a Makefile which I hope can act as a self-documenting method of initializing the ecosystem.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+node_modules/
+.git/
+coverage/
+
+.dockerignore
+Dockerfile
+*.pem
+.gitignore
+.env
+.eslintrc
+.editorconfig
+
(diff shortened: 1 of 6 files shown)


Dantemss:
I think we could merge this (don't think it would interfere with non-Docker stuff)
Need to update the ruby version on Docker to 2.3.6 now though :)

m1yag1:
Oh, if we can merge this let me update the README.md. I'll try to get that in by later today.

m1yag1:
Closing because I don't think docker is going to be used in production any time soon.

PR #1681, opened 2018-02-22
Added missing score weight validations

Dantemss:


nathanstitt:
[Bug: able to set negative values for the score weight](https://trello.com/c/hK2bJlF0/1332-bug-able-to-set-negative-values-for-the-score-weight)

PR #1685, opened 2018-02-23
Added start and end dates for notifications

Dantemss:


lib/hypothesis.rb
```diff
@@ -5,14 +5,12 @@ def self.generate_grant_token(account_uuid)
 
       payload = {
         aud: Rails.application.secrets[:hypothesis]['host'],
-        iss: Rails.application.secrets[:hypothesis]['client_id'],
+        iss: Rails.application.secrets[:hypothesis]['client_id'] || '',
```
jpslav: why is this in here?

app/assets/javascripts/admin.js
```diff
@@ -14,9 +14,15 @@ $(document).ready(function() {
...
   var day = date.getDate();
-  var dateString = year + (month < 10 ? '-0' : '-') + month + (day < 10 ? '-0' : '-') + day;
+  var tomorrow = day + 1;
+  var timeString = ' 00:00:00';
+  var yearMonthString = year + (month < 10 ? '-0' : '-') + month;
+  var startDateTimeString = yearMonthString + (day < 10 ? '-0' : '-') + day + timeString;
+  var endDateTimeString   = yearMonthString + (tomorrow < 10 ? '-0' : '-') + tomorrow + timeString;
+
+  $('.datepicker.start').val(startDateTimeString);
+  $('.datepicker.end'  ).val(endDateTimeString  );
```
nathanstitt: this looks ok, but it'd be much more clear if you added https://momentjs.com to the bundle and used that.  You could then do: 
```
$('.datepicker.start').val(
    moment().startOf('day'). subtract(1, day).format()
```

lib/hypothesis.rb
Dantemss: Was breaking my local dev without it :)

app/assets/javascripts/admin.js
Dantemss: Will this return the format I expect? It has to match exactly for the datepicker to work. Or I need to change the datepicker format.

PR #1688, opened 2018-02-26
(Hopefully) fix codecov

Dantemss:
The main thing I'm trying here is to merge results only at the end and send them to codecov only once, rather than sending a bunch of partial merges as the test processes exit.


Dantemss:
Project coverage increased by 1% with this PR (which does not test anything)... Maybe it worked?

PR #1701, opened 2018-03-06
WIP: Include the exercise content as an object on stats

nathanstitt:
This way they match the normal exercise layout and the same FE exercise model can be re-used


nathanstitt:
[T3.21.10 Support previewing exercises as they appear in Tutor](https://trello.com/c/XQ5R81kN/1317-t32110-support-previewing-exercises-as-they-appear-in-tutor)

PR #1704, opened 2018-03-06
Release 2018-03-12

Dantemss:
- Fix LMS score sync again

### DIFF
--- a/app/subsystems/lms/send_course_scores.rb
+++ b/app/subsystems/lms/send_course_scores.rb
@@ -16,11 +16,15 @@ class Lms::SendCourseScores
 
   lev_routine transaction: :no_transaction
 
-  def exec(course:)
+  def exec(course:, role:)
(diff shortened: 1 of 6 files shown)


app/subsystems/lms/send_course_scores.rb
```diff
@@ -55,7 +59,7 @@ def save_status_data
 
   def course_score_data(user_profile_id)
     @scores_by_user_profile_id ||= begin
-      perf_report = Tasks::GetTpPerformanceReport[course: @course]
+      perf_report = Tasks::GetTpPerformanceReport[course: @course, role: @teacher_role]
```
jpslav: We actually don't care which role it is if it is a teacher.  Why not just add a `is_teacher: true` argument?  Then no need to pass in role, check that role is course teacher above (this check is already done in the access policy).

PR #1710, opened 2018-03-12
 Extend student trial if they register before the course begins

Dantemss:
Does not attempt to fix existing students, only affects new ones.

Includes https://github.com/openstax/tutor-server/pull/1709


jpslav:
Rejected in trello

PR #1712, opened 2018-03-14
Add DO_NOT_TRACK_TUTOR_ONBOARDING env var to silence SF errors

nathanstitt:
I keep hitting this locally and got tired of it.


Dantemss:
@jpslav maybe remove this completely. We get this on prod too for test users and it's meaningless.
Or at least don't raise for test users so we remember to mark them as such.

jpslav:
@nathanstitt maybe just return from routine immediately unless an ENV is set? 

```ruby
return if EnvUtilities.load_boolean(name: 'DO_NOT_TRACK_TUTOR_ONBOARDING', default: false)
```

That way you can turn it off but doesn't require a code change to work on this code in non-prod envs?

PR #1714, opened 2018-03-14
Fix course dashboard performance

Dantemss:
Fix time_zone, task_plan and task_steps N+1 queries in student dashboard

### DIFF
--- a/app/subsystems/tasks/get_tasks.rb
+++ b/app/subsystems/tasks/get_tasks.rb
@@ -8,16 +8,19 @@ class Tasks::GetTasks
   def exec(roles:, start_at_ntz: nil, end_at_ntz: nil)
     role_ids = verify_and_get_id_array(roles, Entity::Role)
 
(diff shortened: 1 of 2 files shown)


app/subsystems/tasks/get_tasks.rb
```diff
@@ -8,16 +8,19 @@ class Tasks::GetTasks
   def exec(roles:, start_at_ntz: nil, end_at_ntz: nil)
     role_ids = verify_and_get_id_array(roles, Entity::Role)
 
-    # Trying to join Tasks and Taskings here fails on deleted tasks
```
jpslav: What changed to make this comment no longer a problem?

Dantemss: I believe getting rid of the `default_scope` in the paranoia gem fixed this (task now has `acts_as_paranoid column: :hidden_at, without_default_scope: true`)

PR #1717, opened 2018-03-21
Fix external assignment average_progress

Dantemss:


app/subsystems/tasks/performance_report_routine.rb
```diff
@@ -3,13 +3,15 @@ module Tasks::PerformanceReportRoutine
   protected
 
   def included_in_progress_averages?(task:, current_time_ntz:)
-    return false if task.actual_and_placeholder_exercise_count == 0
+    return false if task.steps_count == 0
     return true if task.task_type == 'concept_coach'
-    return false unless ['reading', 'homework'].include?(task.task_type)
+
     task.due_at_ntz.present? && task.due_at_ntz <= current_time_ntz
```
nathanstitt: Do we need to exclude `event` types here?

Dantemss: Added an Event task_plan to the test to make sure it does not interfere with the scores.

PR #1718, opened 2018-03-22
Downgraded PopulatePlaceholderSteps transaction to READ COMMITTED

Dantemss:
The routine already locks the task to prevent concurrent modifications.
Should fix `[Tutor] (prodtutor) tasks#show (PG::TRSerializationFailure) "ERROR:  could not serialize access due to concurrent update...`


Dantemss:
Included in https://github.com/openstax/tutor-server/pull/1718

PR #1729, opened 2018-04-13
In the pages API, use the page ID as a tie-breaker

Dantemss:
So pages imported last display preferentially. We fixed a bug in the link rewriter (in the importer code in Tutor), but since that did not change the page version in CNX, the changes were not visible. This PR fixes that.


nathanstitt:
Looks good to me @Dantemss

PR #1734, opened 2018-04-20
Added new API to return course start/end dates for payments

Dantemss:
Also rewrote the Courses Controller spec so it runs in ~30 seconds.


app/controllers/api/v1/courses_controller.rb
```diff
@@ -177,6 +177,62 @@ def clone
...
+      unless uuid_array.all? { |uuid| uuid.is_a? String }
+
+    course_dates_map = {}
+    CourseProfile::Models::Course.where(uuid: uuid_array)
+                                 .pluck(:uuid, :starts_at, :ends_at)
+                                 .each do |uuid, starts_at, ends_at|
+      course_dates_map[uuid] = { starts_at: starts_at, ends_at: ends_at }
+    end
+
+    render json: course_dates_map
```
jpslav: Are we ok on the default rendering of the time stamps?  Elsewhere in our API we use the `DateTimeUtilities`.

PR #1750, opened 2018-06-08
Return all offerings with the is_available flag

nathanstitt:
This way the FE can tell the difference between an non-existent offering
and one that is no longer available

Goes along with https://github.com/openstax/tutor-js/pull/2260


nathanstitt:
[Update Tutor Bio to use Tutor Bio 2e - DO NOT RELEASE until we are ready to add Bio 2e content to production](https://trello.com/c/JBPYwxhE/1343-update-tutor-bio-to-use-tutor-bio-2e-do-not-release-until-we-are-ready-to-add-bio-2e-content-to-production)

PR #1756, opened 2018-07-09
add configuration for local development with docker environment

TomWoodward:


Dockerfile
```diff
@@ -0,0 +1,7 @@
+from ruby:2.3.6
+
+copy . /code
+
+workdir /code
+
+run bundle install
```
nathanstitt: hm, on some cases this will take awhile since the `log` directory can grow large unless you remember to clear it periodically.  For instance mine's currently `1.4G` 

Might be a good idea to not copy `log` , `tmp`, and maybe `public/assets`

docker-quickstart.md
```diff
@@ -0,0 +1,31 @@
...
+the app will become available at `http://localhost:3001`
+
+## Notes
+
+The first time you start the environment you'll want to:
+
+```bash
+docker-compose run tutor rake db:setup 
+docker-compose run tutor rake demo:courses
+```
```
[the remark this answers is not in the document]
Dantemss: `demo` will import the 3 books and generally take forever

`demo[soc]` (I think that's what it was) might be better for FE devs

BE devs might want `demo:staff` to get just the admin account

Dockerfile
TomWoodward: actually the bundle thing probably isn't working as is, i should change it to only copy the Gemfile during the docker build and mount the rest of the app at runtime without blowing away the vendor folder

TomWoodward:
@nathanstitt i changed it so only the Gemfile and Gemfile.lock are added to the image, so those large directories shouldn't be an issue anymore.

.dockerignore
```diff
@@ -0,0 +1,3 @@
+.git
+.gitignore
+tmp
```
nathanstitt: I think we'll also need to ignore `log` here, that's the one that gets really large unless you remember to periodically clear it out.

TomWoodward:
@nathanstitt i addressed your comment

.dockerignore
nathanstitt: @TomWoodward ^^^

TomWoodward: @nathanstitt 👇

docker-compose.dev.yml
```diff
@@ -0,0 +1,17 @@
...
+      - "3001:3001"
+    networks:
+      - openstax
+    environment:
+      - HYPOTHESIS_HOST=hypothesis.sandbox.openstax.org
+      - HYPOTHESIS_APP_URL=https://hypothesis.sandbox.openstax.org
+      - HYPOTHESIS_API_URL=https://hypothesis.sandbox.openstax.org/api/
+      - REDIS_URL=redis://redis:6379/0
+      - OXT_DB_HOST=postgres95
+networks:
```
nathanstitt: I'm guessing that `hypothesis.sandbox.openstax.org` has a fixed key that's set in `secrets.yml`? If that is indeed the case then anyone could add/remove annotations from it.  maybe that's ok though since the official server does the same thing essentially.

I see mention of `openstax/hypothesis-server` below though,  should this use that image instead?

PR #1758, opened 2018-07-20
update small screen styles for main nav on homepage

TomWoodward:
I removed the collapsing behavior from the nav so the links on the right show inline until there is literally no more room for them and then they bump below the logo. at that point it looks a little funny but I don't see how anybody could reasonably get their screen that small anyway.

before:
image

after:
image

### DIFF
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -142,6 +142,10 @@ header {
       background: #fff;
     }
 
+    .navbar-header {
+      float: left;
+    }
+
     .navbar-brand {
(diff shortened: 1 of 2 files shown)


TomWoodward:
[Bug: user menu alignment for small screens](https://trello.com/c/xmBWozWx/1580-bug-user-menu-alignment-for-small-screens)

TomWoodward:
[Bug: user menu alignment for small screens](https://trello.com/c/xmBWozWx/1580-bug-user-menu-alignment-for-small-screens)

PR #1761, opened 2018-07-30
WIP Fetch exercises by nickname

Dantemss:
Needs tests


Dantemss:
I don't think codecov is being accurate because it approved 980a4b3 and then the next commit added a bunch of tests and failed.

app/subsystems/tasks/assistants/generic_assistant.rb
```diff
@@ -64,12 +64,19 @@ def add_exercise_step!(
...
-  def get_all_page_exercises_with_tags(page:, tags:)
-    sorted_tags = [tags].flatten.uniq.sort
-
-    @tag_exercise_cache[page.id][sorted_tags] ||= ecosystem.exercises_with_tags(
-      sorted_tags, pages: page
-    )
+  def get_all_page_exercises_with_queries(page:, queries:)
+    queries.flat_map do |field, values|
+      sorted_values = [values].flatten.uniq.sort
+
```
nathanstitt: is the purpose of putting values into an array and then calling `flatten` to make sure it's an array?  If so there's `Array.wrap` https://api.rubyonrails.org/v4.2/classes/Array.html#method-c-wrap

PR #1762, opened 2018-07-31
transform note content for styling

TomWoodward:
adds a content div around everything in the note that isn't the title


Dantemss:
Can you add comments on the code saying why this transform is needed?
Also I think you might have to mess with the whitespace in the expected string to get the tests to pass.

TomWoodward:
@Dantemss added that comment you asked for

Dantemss:
@TomWoodward I don't see the comment in the diff. Maybe it wasn't pushed? I'm thinking just a comment right above `def map_note_format`

TomWoodward:
@Dantemss yea it looks like i must have rebased it out by mistake, added it back

PR #1764, opened 2018-08-06
Added new TLO and KC tag types

Dantemss:
Thematic Learning Objectives and Key Concepts for APUSH


Dantemss:
Apparently we decided these tags won't be visible for teachers and students so we don't need this.

PR #1768, opened 2018-08-17
assets in dev mode need dist prefix

nathanstitt:


nathanstitt:
Not needed since the `dist` prefix should be in the secrets.yml

PR #1770, opened 2018-08-22
Fail faster if Biglearn is not responding and tell the FE

Dantemss:
FE can use the new boolean in the dashboard to tell when to stop polling (because tasks might appear out of order now).

### DIFF
--- a/app/routines/dashboard_routine_methods.rb
+++ b/app/routines/dashboard_routine_methods.rb
@@ -47,8 +47,15 @@ module DashboardRoutineMethods
     end
 
     outputs.tasks = had_pes + got_pes
-    # TODO: Maybe make this boolean check background jobs (ReassignPublishedPeriodTaskPlans)
-    #       so we can return true when there are no tasks
-    outputs.all_tasks_are_ready = !outputs.tasks.empty? && still_need_pes.empty?
+    outputs.all_tasks_are_ready = still_need_pes.empty? && role_type != :student ||
+                                  Tasks::Models::TaskPlan
+                                    .joins(:tasking_plans)
+                                    .preload(:tasking_plans)
+                                    .where(tasking_plans: {
+                                      target_id: role.student.course_membership_period_id,
+                                      target_type: 'CourseMembership::Models::Period'
+                                    })
+                                    .where { first_published_at != nil }
+                                    .count <= outputs.tasks.size
   end
 end
--- a/app/routines/dashboard_routine_methods.rb
+++ b/app/routines/dashboard_routine_methods.rb
@@ -43,7 +43,8 @@ module DashboardRoutineMethods
 
     got_pes, still_need_pes = need_pes.partition(&:pes_are_assigned)
     still_need_pes.each do |task|
-      Tasks::PopulatePlaceholderSteps.perform_later task: task, background: true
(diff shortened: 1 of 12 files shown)


Dantemss:
@nathanstitt would be nice to have a matching FE PR to control the polling. Also, will the FE reload if the user happens to reach a personalized exercise that was not populated for some reason?

Dantemss:
I'll fix the tests... eventually

app/routines/dashboard_routine_methods.rb
```diff
@@ -38,6 +38,17 @@ def load_tasks(role, role_type, start_at_ntz = nil, end_at_ntz = nil, current_ti
       task.past_open? current_time: current_time
     end if role_type != :teacher
 
-    outputs.tasks = tasks
+    had_pes, need_pes = tasks.partition(&:pes_are_assigned)
+    need_pes.each { |task| Tasks::PopulatePlaceholderSteps[task: task] }
+
+    got_pes, still_need_pes = need_pes.partition(&:pes_are_assigned)
+    still_need_pes.each do |task|
+      Tasks::PopulatePlaceholderSteps.perform_later task: task, background: true
```
jpslav: will this retry if failed?

also, if someone keeps reloading the dashboard, will this call keep queuing up jobs over and over?

app/subsystems/tasks/populate_placeholder_steps.rb
```diff
@@ -215,11 +232,12 @@ def populate_placeholder_steps(task:, group_type:, biglearn_api_method:, biglear
       end
     end
 
-    task.update_attribute(:spy, task.spy.merge(spy_info.except('exercises')))
+    task.spy = task.spy.merge(spy_info.except('exercises'))
+    task.send "#{boolean_attribute}=", true
```
jpslav: Is this no longer saving the updates like before?

app/routines/dashboard_routine_methods.rb
nathanstitt: I think it might, this is included in `GetTpDashboard` which is called from https://github.com/openstax/tutor-server/blob/master/app/controllers/api/v1/courses_controller.rb#L119-L121

It probably finishes fairly quickly if it's up-to-date but maybe we should figure out a way to only call it once every X minutes or something

app/subsystems/tasks/populate_placeholder_steps.rb
Dantemss: They are saved at the end of the other method.

app/routines/dashboard_routine_methods.rb
Dantemss: It will lock the task so it will wait for the other one to finish. Then it will see PEs are already assigned, do the skim-filling check and exit. Should be pretty fast even if you queue up a few dozen.

nathanstitt:
[Change Request: CritSit resolution](https://trello.com/c/FkQTAwrM/1633-change-request-critsit-resolution)

app/subsystems/tasks/populate_placeholder_steps.rb
Dantemss: I'll move that here, might be why the specs are failing.

PR #1771, opened 2018-08-23
relax hypothesis token so it doesn't expire

nathanstitt:
And remove the "not before time" check so server clocks doesn't have to match.

This will hopefully fix the strange hypothesis errors we've been getting where the server returns forbidden while users attempt to annotate.  

Since most sessions will last more than 5 minutes they never get a new token

### DIFF
--- a/lib/hypothesis.rb
+++ b/lib/hypothesis.rb
@@ -1,6 +1,5 @@
 module Hypothesis
   def self.generate_grant_token(account)
-      now = Time.now.to_i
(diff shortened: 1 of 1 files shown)


nathanstitt:
yeah, that'd make sense but I figured it was acceptable to not expire since we also don't expire users sessions in general

TomWoodward:
[User getting an error while highlighting an incomplete sentence](https://trello.com/c/Ypg26v84/1619-user-getting-an-error-while-highlighting-an-incomplete-sentence)

PR #1775, opened 2018-09-04
Development docker

TomWoodward:
includes some changes to the way asset rev-manifest.json is loaded but doesn't yet fix the problem where the server needs to be re-deployed whenever the js is built


lib/tutor/assets.rb
```diff
@@ -13,7 +13,7 @@ def self.[](asset, ext)
     # called by assets initializer as it boots
     def self.read_manifest
       begin
-        @manifest = JSON.parse Rails.root.join('public', 'assets', 'rev-manifest.json').read
+        @manifest = JSON.parse open("#{Rails.application.secrets.assets_url}/rev-manifest.json").read
         @manifest.default_proc = proc do |_, asset|
```
nathanstitt: I don't think this'll work if tutor-fe is running in dev mode.    The rev-manifest.json is only generated by webpack when it builds the assets, and isn't present when `webpack-dev-server` is running.

Can the rev-manifest.json file be shared similar to how hypothesis secrets are, by being dropped in a comment directory and read from there?

TomWoodward: in development it works fine because it doesn't even use the manifest file. in non-development we can't really wrangle the volumes the same way we do with the hypothesis secrets

based on discussion in tutor-devs it sounds like this is definitely not the way we want to do it, but i'm not sure what a better way would be...

PR #1777, opened 2018-09-10
Fix the dashboard all_tasks_are_ready flag

Dantemss:
Fix the dashboard all_tasks_are_ready flag so it can return true when there are unopened or hidden assignments.


nathanstitt:
[Change Request: CritSit HOTFIX](https://trello.com/c/fQMA59sB/1640-change-request-critsit-hotfix)

PR #1778, opened 2018-09-10
Fix scores > 100% on spreadsheet for late work

Dantemss:
Better reproduce the bug somewhere before merging so you can test (should be easy).


nathanstitt:
[Bug: Unexpected behavior from downloaded scores](https://trello.com/c/2PQGUYHV/1661-bug-unexpected-behavior-from-downloaded-scores)

PR #1779, opened 2018-09-11
Fixed broken all_tasks_are_ready flag logic (missing parens)

Dantemss:
Can confirm that `false && false || true` is indeed true.


nathanstitt:
[Change Request: CritSit HOTFIX](https://trello.com/c/fQMA59sB/1640-change-request-critsit-hotfix)

PR #1781, opened 2018-09-14
Set time zones properly for the performance report export

Dantemss:
To fix the issue where last_worked_at in the excel spreadsheet does not match due_at's timezone.


nathanstitt:
[Bug: Scores export timing is off](https://trello.com/c/9iyByMWH/1650-bug-scores-export-timing-is-off)

PR #1786, opened 2018-09-21
Study brains for research

nathanstitt:
This PR:
* Alters lib/ox/exercise.rb so that the content_hash can be easily modified
* Adds a "StudyBrain" model which a cohort has many of.  It uses STI https://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html
  * 2 models inherit from it
  * The subclasses add methods which pass variables into the evaluated code
* Task gets additional relations so it can be mapped to cohorts/study brains
* The `ManipulateStudentTask` routine calls the study brains for each task
* Task get/update controller actions call the routine before the actions take place


app/controllers/research/brains_controller.rb
```diff
@@ -0,0 +1,48 @@
+class Research::BrainsController < Research::BaseController
```
Dantemss: I think this needs access control?

app/subsystems/research/manipulate_student_task.rb
```diff
@@ -0,0 +1,13 @@
+class Research::ManipulateStudentTask
+
+  lev_routine
+
+  def exec(hook:, task:, task_step: nil)
+    task.research_study_brains.each do |brain|
+      next unless brain.hook.blank? || brain.hook.to_sym == hook.to_sym
+      error = brain.evaluate(binding())
```
Dantemss: Not a huge fan of modifying the variables passed into the method (I would be more in favor of returning the new values and assigning them in the controller) but I guess it works...

app/subsystems/research/models/study_brain.rb
```diff
@@ -0,0 +1,31 @@
...
+  validates :name, :domain, :code, presence: true
+  validate :ensure_valid_hook_for_domain
+
+  VALID_HOOKS = {
+    student_task: %i{display update}
+  }
+
+  def evaluate(binding)
+    begin
+      eval(code, binding)
```
Dantemss: I think `binding.eval(code)` also works but meh... same effect.

app/subsystems/research/models/study_brain.rb
```diff
@@ -0,0 +1,31 @@
...
+      nil
+    rescue Exception => e
+      # "rescue exception is evil" is true but this is using one evil to cancel another
+      # The eval code should never crash the server
+      return e
```
Dantemss: Send to sentry?

app/subsystems/tasks/models/task.rb
```diff
@@ -50,6 +50,12 @@ def <<(*records)
   end
   has_many :tasked_exercises, through: :task_steps, source: :tasked,
                                                     source_type: 'Tasks::Models::TaskedExercise'
+  has_one :tasking
+  has_one :role, through: :tasking, subsystem: :entity
+  has_one :student, through: :role, subsystem: :course_membership
+  has_many :research_cohorts, class_name: 'Research::Models::Cohort', through: :student
+  has_many :research_study_brains, -> { student_task },
+           class_name: 'Research::Models::StudyBrain', through: :research_cohorts, source: :study_brains
```
Dantemss: Couldn't we just have used the existing `has_many`s for these? Alternatively add a comment saying this will stop working if we ever implement group tasks.

db/migrate/20180921183625_study_brains.rb
```diff
@@ -0,0 +1,11 @@
+class StudyBrains < ActiveRecord::Migration
+  def change
+    create_table :research_study_brains do |t|
+      t.references :research_cohort, null: false, index: true,
+                   foreign_key: { on_update: :cascade, on_delete: :cascade }
+      t.text :name, :code, null: false
+      t.integer :domain, limit: 2 # use smallint for enum
+      t.text :hook
```
Dantemss: null hook is valid?

db/schema.rb
```diff
@@ -338,7 +338,7 @@
     t.integer  "cloned_from_id"
     t.boolean  "is_preview",                                                                                         null: false
     t.boolean  "is_excluded_from_salesforce",                                          default: false,               null: false
-    t.uuid     "uuid",                                                                 default: "gen_random_uuid()", null: false
+    t.uuid     "uuid",                                                                 default: "gen_random_uuid()"
```
Dantemss: See if you can undo this change (maybe just set the column not null on your local db).

db/schema.rb
```diff
@@ -585,7 +585,7 @@
     t.datetime "updated_at",                                          null: false
     t.integer  "faculty_status",        default: 0,                   null: false
     t.string   "salesforce_contact_id"
-    t.uuid     "uuid",                  default: "gen_random_uuid()", null: false
+    t.uuid     "uuid",                  default: "gen_random_uuid()"
```
Dantemss: Same as above.

spec/subsystems/tasks/models/tasked_exercise_spec.rb
```diff
@@ -30,8 +30,8 @@
 
     tasked_exercise.free_response = nil
     expect(tasked_exercise).not_to be_valid
+    tasked_exercise.parser.questions_for_students.each{|q| q['formats'] = ['multiple-choice']}
```
Dantemss: Wouldn't it be simpler and more useful for different studies in the future to modify `tasked_exercise.content`? (You could even save the modified content so it's visible to the teacher or something, but make sure to skip validation). That should get rid of the need to add new methods like formats_for_students every time.

app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -102,9 +102,9 @@ def set_correct_answer_id
   protected
 
   def free_response_required
-    return true unless formats.include?('free-response') && free_response.blank?
-    errors.add(:free_response, 'is required')
-    false
+    if parser.question_formats_for_students.include?('free-response') && free_response.blank?
```
Dantemss: I would say either keep `formats` here and modify the method to return this or get rid of `formats` completely, if nothing else is using it.

app/controllers/research/brains_controller.rb
nathanstitt: nah it inherits from `Research::BaseController` which has a `before_filter :authenticate_researcher!`. I'll add a comment to note that tho

app/subsystems/research/manipulate_student_task.rb
nathanstitt: well if anything's modified it'd be inside the `brain.evaluate` which could do anything really.  It needs to be flexible enough for us to do whatever research comes up with to the task steps which makes it hard to anticipate what should be returned.

We also can't perform destructive updates on the tasks/task_steps since the study may be turned off mid-assignment

app/subsystems/research/models/study_brain.rb
nathanstitt: oh neat, I didn't know about that usage

app/subsystems/research/models/study_brain.rb
nathanstitt: Great idea!

app/subsystems/tasks/models/task.rb
nathanstitt: hm, I don't follow how these would break?  Do you mean if `tasking` was a has many?

db/migrate/20180921183625_study_brains.rb
nathanstitt: yeah, I think we may want to write some brains that will apply to all hooks

spec/subsystems/tasks/models/tasked_exercise_spec.rb
nathanstitt: I thought about doing that but we need to be able to turn off studies mid-assignment, and students may move between cohorts, which would have different study brain logic

app/subsystems/tasks/models/tasked_exercise.rb
nathanstitt: yeah, I think nothing else is using `formats` - I'll remove that

app/subsystems/tasks/models/task.rb
Dantemss: `taskings` is already a `has_many` :)

app/subsystems/research/manipulate_student_task.rb
Dantemss: OK

app/controllers/research/brains_controller.rb
Dantemss: OK

app/subsystems/research/models/study_brain.rb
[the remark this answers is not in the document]
Dantemss: The answer is that yes, it would probably be more efficient to load once. But would have to load ALL study brains that are in use.

spec/subsystems/tasks/models/tasked_exercise_spec.rb
Dantemss: What if... we do the destructive modifications but at the same time change the save/update_attribute methods to throw exceptions (just to make sure we don't save the modifications). So we would modify every time these are loaded but wouldn't need to define those new methods.

app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -28,6 +29,7 @@ def show
   api :PUT, '/steps/:step_id', 'Updates the specified TaskStep'
   def update
     ScoutHelper.ignore!(0.8)
+    Research::ManipulateStudentTask[hook: :update, task_step: @task_step, task: @task_step.task]
```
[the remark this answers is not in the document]
nathanstitt: Sure, the code for display what I added to the specs, but I'll add it here:

for display when it needs to alter all task_steps:
```ruby
task.task_steps.each{ |ts|
	ts.tasked.parser.questions_for_students.each{|q|
  	q['formats'] -= ['free-response']
  } if ts.exercise?
}
```

and when a single task is saved:
```ruby
task_step.tasked.parser.questions_for_students.each{|q|
  q['formats'] -= ['free-response']
} if task_step && task_step.exercise?
```

spec/subsystems/tasks/models/tasked_exercise_spec.rb
nathanstitt: Well, the added methods aren't really needed, the brains code could operate directly on the `content_hash_for_students`.  I added them so the brains could could be simpler and have less chance of errors occurring.  

I think I'd prefer this over brains code firing off destructive methods and relying on the model refusing to actually save them.

If you think it'd be better if we operate only on `content_hash_for_students` we can do that though

db/migrate/20180921183625_study_brains.rb
```diff
@@ -0,0 +1,11 @@
+class StudyBrains < ActiveRecord::Migration
+  def change
+    create_table :research_study_brains do |t|
+      t.references :research_cohort, null: false, index: true,
+                   foreign_key: { on_update: :cascade, on_delete: :cascade }
```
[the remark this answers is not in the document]
nathanstitt: Yes, that does mean that 3 cohorts could have 3 brains.  However I talked with research and they say that studies will almost always have 2 cohorts, one with the effect and a control without it.  In that case they'll only be 1 brain, since the control will not be affected and should use the default Tutor functionality.

In some cases they may do multivariate testing but in those cases each brain will have different code anyway.

They also said that they can't envision a test that would re-use an existing brain.  Some may build on earlier brains but they'll be something subtly different in the logic.

app/controllers/research/brains_controller.rb
```diff
@@ -2,6 +2,10 @@ class Research::BrainsController < Research::BaseController
 
   # Research::BaseController makes sure users are researchers
 
+  # use all the subclasses here so that they'll be loaded and visible
+  Research::Models::DisplayStudentTask
+  Research::Models::UpdateStudentTasked
```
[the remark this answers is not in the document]
nathanstitt: because the table uses inheritance the models aren't loaded until rails needs to autoload the class.  The 'type' drop-down on the edit form uses `Research::Models::StudyBrain.subclasses` to find all the possible types, but they weren't loaded and so the selection would be empty.

app/subsystems/tasks/models/task.rb
```diff
@@ -53,10 +53,11 @@ def <<(*records)
...
   has_one :role, through: :tasking, subsystem: :entity
   has_one :student, through: :role, subsystem: :course_membership
-  has_many :research_cohorts, class_name: 'Research::Models::Cohort', through: :student
+  has_many :research_cohorts, through: :student,
+           subsystem: :research, class_name: 'Research::Models::Cohort'
   has_many :research_study_brains, -> { student_task },
-           class_name: 'Research::Models::StudyBrain',
-           through: :research_cohorts, source: :study_brains
+           through: :research_cohorts, source: :study_brains,
+           subsystem: :research, class_name: 'Research::Models::StudyBrain'
```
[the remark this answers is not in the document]
nathanstitt: Yes, it wouldn't be needed if I used the `cohorts` name but I disliked that because it seemed to easily confused. I was afraid that people would think that "cohorts" meant either student's fellow course classmates, or the other tasks that the student might have. Adding the "research_" prefix seemed to make the name more clearer.

app/controllers/research/brains_controller.rb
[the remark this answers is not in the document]
Dantemss: What if we list them manually there? Or just keep the list somewhere else. I fear they may get lost in here.

spec/subsystems/tasks/models/tasked_exercise_spec.rb
Dantemss: Maybe consider this again for the next study.

app/controllers/research/brains_controller.rb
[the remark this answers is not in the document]
nathanstitt: yeah, great idea. a list is a list, so we might as well have it as close to the place where it's needed.

app/controllers/api/v1/task_steps_controller.rb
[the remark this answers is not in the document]
jpslav: Let's have a chat about calling brains from the controller vs somewhere else -- I think there's a use case that exists that I'm missing.

PR #1789, opened 2018-09-28
Update the doorkeeper gem

Dantemss:
Codecov didn't like me removing the auth controller, since it had specs.


Dantemss:
Codecov didn't like me removing the auth controller, since it had specs.

PR #1796, opened 2018-10-12
record research manipulations

nathanstitt:
* Record when a research study brain performs a manipulation
* Spec for study hiding free responses


db/schema.rb
```diff
@@ -338,7 +338,7 @@
     t.integer  "cloned_from_id"
     t.boolean  "is_preview",                                                                                         null: false
     t.boolean  "is_excluded_from_salesforce",                                          default: false,               null: false
-    t.uuid     "uuid",                                                                 default: "gen_random_uuid()", null: false
+    t.uuid     "uuid",                                                                 default: "gen_random_uuid()"
```
Dantemss: Missing `null: false` in your local db

db/schema.rb
```diff
@@ -585,7 +585,7 @@
     t.datetime "updated_at",                                          null: false
     t.integer  "faculty_status",        default: 0,                   null: false
     t.string   "salesforce_contact_id"
-    t.uuid     "uuid",                  default: "gen_random_uuid()", null: false
+    t.uuid     "uuid",                  default: "gen_random_uuid()"
```
Dantemss: Same here

db/schema.rb
```diff
@@ -679,6 +679,20 @@
 
   add_index "research_cohorts", ["research_study_id"], name: "index_research_cohorts_on_research_study_id", using: :btree
 
+  create_table "research_manipulations", force: :cascade do |t|
+    t.integer  "research_study_id",                    null: false
+    t.integer  "research_cohort_id"
+    t.integer  "research_study_brain_id"
```
Dantemss: When are null cohorts/study brains valid?

app/subsystems/research/export_data.rb
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  def exec
+    cmd = "pg_dump --clean #{Pgdb.cmd_line_flags.join(' ')}"
+    EXCLUDED_TABLES.each{|table|
+      cmd << ' --exclude-table '
+      cmd << table
+    }
+    cmd << " #{Pgdb.name}"
+    cmd << " | psql --quiet --output /dev/null #{Rails.application.secrets.research_db_connection}"
```
Dantemss: This makes it look like you are feeding the dump to psql, which will drop and then restore the database?

lib/tasks/dump_restore.rake
```diff
@@ -26,20 +26,19 @@ namespace :db do
...
+          "--format=c #{Pgdb.name} > #{dump_file}"
 
-      "mkdir -p #{dumps_dir}; " \
-      "pg_dump --host #{host} --username #{user} " \
-               "--clean --no-owner --no-acl " \
-               "--format=c #{db} > #{dump_file}"
-    end
     puts cmd
-    exec cmd
+    exec Pgdb.env, "#{cmd} > /dev/null 2>&1"
```
Dantemss: Is the output being redirected twice? Seems like it's already redirected above.

app/subsystems/research/export_data.rb
```diff
@@ -0,0 +1,39 @@
+require 'pgdb'
```
Dantemss: Since this is going to be loaded all the time anyway, maybe just add pgdb to `config/environments.rb`

app/subsystems/research/models/manipulation.rb
```diff
@@ -0,0 +1,24 @@
...
+  belongs_to :study_brain, inverse_of: :manipulations
+  belongs_to :cohort
+  belongs_to :study
+
+  belongs_to :target, polymorphic: true
+
+  validate :ensure_should_record
+
+  def record!
+    @should_record = true
```
Dantemss: If this is supposed to be like a log, seems like you want to always record it.

db/schema.rb
nathanstitt: because the cohort may be deleted.  when that happens we don't want to lose any manipulations that may have occurred.  Of course cohort's being deleted would be rare, but it may happen while testing and such

app/subsystems/research/export_data.rb
```diff
@@ -0,0 +1,39 @@
...
+   user_profiles openstax_accounts*
+   lms_course_score_callbacks
+   user_administrators
+   user_tour_views
+   role_role_users
+   user_researchers
+   course_membership_enrollment_changes
+   user_content_analysts
+   user_customer_services
+  }
```
jpslav: Can we switch to whitelisting instead of blacklisting?  Reason being that it will be easy to forget to update this exporter when new tables with PII are added; also easy to forget other tables on this first pass, e.g. `lms_trusted_launch_data` is not in this exclusion list but probably should be.  

```
  INCLUDED_TABLES=%w{
    whitelisted_tables
    go_here
  }

  excluded_tables = ActiveRecord::Base.connection.tables - INCLUDED_TABLES
```

app/subsystems/research/export_data.rb
jpslav: I agree this is worrisome :-)

app/subsystems/research/export_data.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def exec
+    cmd = "pg_dump --clean #{Pgdb.cmd_line_flags.join(' ')}"
```
jpslav: Are `pg_dump` and `psql` installed on our AWS web servers?  Maybe not since we use RDS instances?

lib/tasks/dump_restore.rake
jpslav: Why is this file being modified?  Is its use case separate from the `ExportData` routine above?

spec/research-studies/remove_free_response_format_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  before(:each) {
+    @study.deactivate!
+  }
+
+  it "can hide free-response formats" do
+    brain = FactoryBot.create :research_modified_task_for_display,
+                      study: @study,
+                      code: <<~EOC
```
jpslav: squiggly heredoc removes leading whitespace so you could indent these blocks of code a little?

app/subsystems/research/export_data.rb
nathanstitt: kinda. It pipes the dump directly to psql which has a connection to the research db.  It won't drop and restore the DB though, The `--clean` option will drop and re-create each individual table.

app/subsystems/research/models/manipulation.rb
nathanstitt: yeah, I thought that at first, but in many cases the manipulation will not do anything so it shouldn't be recorded.  Think non-exercise steps for the 'free-response' study.

lib/tasks/dump_restore.rake
nathanstitt: Agreed, it shouldn't redirect.  It's being modified because I extracted "reading Postgres config" into "pgdb" so it could be shared with the dump_restore routine.

nathanstitt: Also the original dump-restore didn't support supplying a password.  it now does via the ENV vars supp[lied by `Pgdb.env`

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: yeah, but then they are indented when you copy-n-paste into code-mirror. I was afraid we'd be tempted to fix the indentation there and then muck up the code.

app/subsystems/research/export_data.rb
nathanstitt: they are on dev and qa. I'll make a note on the devops card for them to check

app/subsystems/research/export_data.rb
nathanstitt: whitelisting is hard because the tables need to be in the correct order so earlier ones don't refer to latter ones.

And when one is included it ripples out because of all the foreign keys.  For instance in the blacklist above all I really wanted to remove was the openstax_accounts* tables and the others had keys that referred to it so they had to be removed as well.

Good point about lms_trusted_launch_data - I'll add that.

app/subsystems/research/models/manipulation.rb
Dantemss: Maybe make it a `should_record?` method that by default raises  some error then. To ensure we don't forget.

PR #1799, opened 2018-11-02
Update free response study brain to filter by sections

nathanstitt:
And also add a course spy mode info that has the research studies

### DIFF
--- a/spec/research-studies/remove_free_response_format_spec.rb
+++ b/spec/research-studies/remove_free_response_format_spec.rb
@@ -98,7 +98,6 @@ manipulated_sections = %w{
   1.1 2.1 1.2
 }.select.with_index{|_, i| 'A' == cohort.name ? i.even? : i.odd? }
 
-debugger
 if (manipulated_sections & chapter_sections).any?
   tasked.parser.questions_for_students.each{|q|
     q['formats'] -= ['free-response']
--- a/spec/research-studies/remove_free_response_format_spec.rb
+++ b/spec/research-studies/remove_free_response_format_spec.rb
@@ -34,8 +34,9 @@ RSpec.describe 'Task steps without free response field', type: :request, api: tr
     @course.ecosystem.books.last.pages.pluck(:book_location).map{|bl|bl.join('.')}
 
     brain = FactoryBot.create :research_modified_task_for_display,
-                      study: @study,
-                      code: <<~EOC
+                              study: @study,
+                              name: 'chapter 1-3 display no free-response',
+                              code: <<~EOC
 
 unless task.reading? || task.homework?
  manipulation.ignore!
@@ -47,7 +48,7 @@ chapter_sections = Content::Models::Page.where(
 ).pluck(:book_location).map{|cs|cs.join('.')}.uniq
 
 manipulated_sections = %w{
-  1.1 2.1 1.2
+  1.1 1.2 1.3 1.4 2.1 2.2 2.3 2.4 2.5 2.6 2.7 2.8 3.1 3.2 3.3 3.4 3.5
 }.select.with_index{|_, i| 'A' == cohort.name ? i.even? : i.odd? }
 
 task_step_ids = []
@@ -81,6 +82,7 @@ EOC
   it "can override requiring free-response when marking completed" do
     FactoryBot.create :research_modified_tasked_for_update,
                       study: @study,
+                      name: 'chapter 1-3 save without free-response',
                       code: <<~EOC
 
 task = tasked.task_step.task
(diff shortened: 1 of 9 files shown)


spec/research-studies/remove_free_response_format_spec.rb
```diff
@@ -3,51 +3,77 @@
...
-  it "can hide free-response formats" do
+  it "can hide free-response formats when displaying a task" do
+    @course.ecosystem.books.last.pages.pluck(:book_location).map{|bl|bl.join('.')}
+
     brain = FactoryBot.create :research_modified_task_for_display,
                       study: @study,
                       code: <<~EOC
-manipulation.ignore! and return unless cohort.name == 'control'
+
+unless task.content_ecosystem_id == #{@change_me_ecosystem_id} && (task.reading? || task.homework?)
```
Dantemss: Hm why does the ecosystem matter?

spec/research-studies/remove_free_response_format_spec.rb
```diff
@@ -3,51 +3,77 @@
...
+ manipulation.ignore!
+ return {}
+end
+
+chapter_sections = Content::Models::Page.where(
+  id: task.task_plan.settings['page_ids']
+).pluck(:book_location).map{|cs|cs.join('.')}.uniq
+
+manipulated_sections = %w{
```
Dantemss: Can you write comments saying which cohort gets which sections?

spec/research-studies/remove_free_response_format_spec.rb
```diff
@@ -3,51 +3,77 @@
...
+  1.1 2.1 1.2
+}.select.with_index{|_, i| 'A' == cohort.name ? i.even? : i.odd? }
+
 task_step_ids = []
-task.task_steps.each do |ts|
-  if ts.exercise?
-    task_step_ids << ts.id
-    ts.tasked.parser.questions_for_students.each{|q| q['formats'] -= ['free-response'] }
+
+if (manipulated_sections & chapter_sections).any?
```
Dantemss: If the task has any of the manipulated sections, then all steps (even those not in the manipulated sections) will be affected. I think you might have to check per step instead?

spec/research-studies/remove_free_response_format_spec.rb
```diff
@@ -57,16 +83,37 @@
     FactoryBot.create :research_modified_tasked_for_update,
                       study: @study,
                       code: <<~EOC
```
Dantemss: This code is very similar to the above... why is it needed?

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: because they'll want to target say physics but not bio

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: sure, it's 2 cohorts, one should get every other section and the other should get the rest

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: hm this is a good point. I'll have to get clarification on what should happen then

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: this brain is used to remove the free response when a task step is saved.  without it the save fails because the free response is blank.  We need the same logic as above because we don't want to white-list steps as not requiring free-response unless they're truly part of the study.

spec/research-studies/remove_free_response_format_spec.rb
Dantemss: But you have to specifically add courses to the study right? The ecosystem should be irrelevant in that case, just don't add bio courses?

spec/research-studies/remove_free_response_format_spec.rb
Dantemss: Well does that mean the code here is not what we are going to use in production? This code will only manipulate 2 sections for one cohort and 1 section for the other cohort.

spec/research-studies/remove_free_response_format_spec.rb
Dantemss: Can't we just save the modified tasks? Otherwise it will be hard to tell what happened here when we look back.

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: ah, ye that's very true. I'll remove it

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: we're not sure what sections will be actually used in production, just that cohorts should receive every other one.  We will have to make sure the number of sections is even otherwise it will be unbalanced

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: turns out the sections should be per-task, not per-assignment. I've updated code to do that

spec/research-studies/remove_free_response_format_spec.rb
nathanstitt: no because what happens if a study is disabled or student moves between cohorts

Dantemss: Wouldn't it be better if we had a rollback code section for these cases? Like a migration.

PR #1805, opened 2018-11-09
Update the last_completed_at date when updating task steps

Dantemss:
Make sure we record the date they last worked on something accurately.

### DIFF
--- a/app/controllers/api/v1/task_steps_controller.rb
+++ b/app/controllers/api/v1/task_steps_controller.rb
@@ -29,8 +29,21 @@ class Api::V1::TaskStepsController < Api::V1::ApiController
   api :PUT, '/steps/:step_id', 'Updates the specified TaskStep'
   def update
     ScoutHelper.ignore!(0.8)
+
     @tasked = Research::ModifiedTaskedForUpdate[tasked: @tasked]
-    standard_update(@tasked, Api::V1::TaskedRepresenterMapper.representer_for(@tasked))
+
+    if @task_step.completed?
+      # Update last_completed_at if the task_step is already completed
+      update_and_complete_task_step
+
+      respond_with(
+        @tasked,
+        responder: ResponderWithPutPatchDeleteContent,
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -116,4 +120,18 @@ def populate_placeholders_if_needed
...
+    consume!(@tasked, represent_with: Api::V1::TaskedRepresenterMapper.representer_for(@tasked))
+    @tasked.save
+
+    # Task already locked in around_action
+    result = MarkTaskStepCompleted.call(task_step: @task_step, lock_task: false)
+
+    raise(ActiveRecord::Rollback) if render_api_errors(result.errors)
+    raise(ActiveRecord::Rollback) if render_api_errors(@tasked.errors)
+  end
+
```
nathanstitt: seems like you could also move the `respond_with` reply here as well since both places this is called from are doing the same thing

app/routines/mark_task_step_completed.rb
```diff
@@ -15,7 +15,7 @@ def exec(task_step:, completed_at: Time.current, lock_task: true)
 
     course = task.taskings.first.try!(:role).try!(:student).try!(:course)
     OpenStax::Biglearn::Api.record_responses(course: course, tasked_exercise: task_step.tasked) \
-      if task_step.exercise? && !task.taskings.first.try!(:role).try!(:student).try!(:course).nil?
+      if task_step.exercise? && !course.nil?
```
nathanstitt: nice, my added associations are being useful elsewhere 😁

app/controllers/api/v1/task_steps_controller.rb
Dantemss: Well apparently completed responds with the entire task, but update does not (only the task_step)... should we change this?

Dantemss: Changed the API in the latest commit

PR #1808, opened 2018-11-12
Defer migration to the background

Dantemss:
Defer migration to add core_and_personalized_placeholder_exercise_steps_count to tasks to the background


db/background_migrate/20181112190157_set_core_and_personalized_placeholder_exercise_steps_counts.rb
```diff
@@ -0,0 +1,39 @@
+class SetCoreAndPersonalizedPlaceholderExerciseStepsCounts < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def up
+    Tasks::Models::Task.reset_column_information
+
+    loop do
+      tasks = Tasks::Models::Task.transaction do
+        tasks = Tasks::Models::Task.where(
+          <<-WHERE_SQL
```
nathanstitt: you're assigning to `tasks` twice .  Seems like the outer might not be correct

Dantemss: Changed the migration to be clearer

PR #1821, opened 2018-12-10
Update README.md

brittany-johnson:
- According to `/.ruby-version` we're using ruby version 2.3.7
-  `sudo service postgresql restart` did not work for me on OS X. I'm making the assumption that the command provided is for Linux, but please correct me if I am wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ cd tutor-server
 Then install the proper version of ruby using rbenv:
 
 ```
-rbenv install 2.3.6
+rbenv install 2.3.7
 ```
 
 Make sure that `which ruby` and `which gem` point to your `.rbenv` folder,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,7 +97,8 @@ Open that file with your favorite text editor:
 Change `peer` to `md5` or create a new `md5` entry for `localhost` (127.0.0.1).
 Then restart the PostgreSQL daemons with:
 
-`sudo service postgresql restart`
+* Linux: `sudo service postgresql restart`
+* OS X: `sudo brew services restart postgresql`
```
nathanstitt: great catch!

nathanstitt:
thanks @brittany-johnson !

PR #1823, opened 2018-12-11
Add user roster to cohorts

nathanstitt:
Roster is needed so that testers can tell which student is in which cohort

also fix bug with study brain so it only manipulates exercise steps for study
Otherwise it has an error with `parser` not being defined.

### DIFF
--- a/app/views/research/cohorts/members.html.erb
+++ b/app/views/research/cohorts/members.html.erb
@@ -1,0 +1,23 @@
+  <div>
+    Study: <%= link_to @cohort.study.name, research_study_path(@cohort.study) %>
+  </div>
+  <div>
+    <h3>Cohort <%= @cohort.name %> Members</h3>
+  </div>
+<hr>
+<table class="table">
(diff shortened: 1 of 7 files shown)


app/views/research/cohorts/members.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    Study: <%= link_to @cohort.study.name, research_study_path(@cohort.study) %>
+  </div>
+  <div>
+    <h3>Cohort <%= @cohort.name %> Members</h3>
+  </div>
+<hr>
+<table class="table">
+  <thead>
+    <tr>
+      <td>Login</td>
```
jpslav: @nathanstitt check with research team about whether they are ok with seeing user logins (they may not want to)

nathanstitt: Good idea, I checked with research and they're fine with seeing student-id.  They currently can see it in the research db as well so no additional info is exposed

PR #1828, opened 2019-01-03
bump ruby version

nathanstitt:
support for 2.3.x will end on 2019-03-31 so we should upgrade


Gemfile
```diff
@@ -172,7 +169,7 @@ gem 'active_force', git: 'https://github.com/openstax/active_force', ref: '9efe1
 
 # Global settings
 gem 'rails-settings-cached', '~> 0.4.0'
-gem 'rails-settings-ui'
+gem 'rails-settings-ui', '~> 0.5.0'
```
jpslav: Can you spot check that settings still work?   This gem has been problematic in the past.

config/initializers/rails_settings_ui.rb
```diff
@@ -20,13 +20,15 @@ def main_app_route_method?(method, include_private_methods = false)
...
-# RailsSettingsUi::ApplicationController.layout 'admin'
 
 RailsSettingsUi.settings_class = "Settings::Db::Store"
 
 Rails.application.config.to_prepare do
   # If you use a *custom layout*, make route helpers available to RailsSettingsUi:
   RailsSettingsUi.inline_main_app_routes!
+  RailsSettingsUi::ApplicationController.module_eval do
+    layout 'admin'
+  end
```
jpslav: Is this related to the Ruby upgrade?

spec/features/admin/queued_jobs_spec.rb
```diff
@@ -58,14 +59,15 @@
   end
 
   scenario 'Getting more details about a job' do
-    error = Lev::Error.new(code: 'bad', message: 'awful')
+    error = Lev::Error.new(code: 'bad', message: 'awful').as_json
```
jpslav: Why was this needed?

Gemfile
nathanstitt: yes, I did check, plus there's specs that render the view, which failed below until I added `as_json`

config/initializers/rails_settings_ui.rb
nathanstitt: yeah, after it upgraded it no longer used our layout until I added this

spec/features/admin/queued_jobs_spec.rb
nathanstitt: This is a tricky one.  

Something changed with how `to_json` works on `Object` and now this line: https://github.com/openstax/jobba/blob/master/lib/jobba/status.rb#L166 will return a string  debug of the object which is: `"Lev::Error.new(code: 'bad', message: 'awful')"`  Calling `as_json` will work fine though.

We could update `jobba` to use `as_json` but then I'm afraid it would probably fail in other cases.

I also attempted to set various options on `oj` to change the behavior but it didn't seem to have any affect.

jpslav: Worries me that jobba is going to break with 2.5 outside of this spec

spec/features/admin/queued_jobs_spec.rb
```diff
@@ -58,14 +59,15 @@
...
     status.add_error(error)
     status.save(something_spectacular: 'For all the good children')
 
     visit admin_jobs_path
     click_link status.id
 
     expect(current_path).to eq(admin_job_path(status.id))
+    save_and_open_page
```
jpslav: This needs to come out

nathanstitt: ug.. so it does.. thanks!

PR #1829, opened 2019-01-08
Stop waiting on BL for readings after 30 minutes

nathanstitt:
Before it'd wait forever and readings would never show up if BL was down or had other difficulties

### DIFF
--- a/app/routines/dashboard_routine_methods.rb
+++ b/app/routines/dashboard_routine_methods.rb
@@ -1,4 +1,6 @@
 module DashboardRoutineMethods
+  BL_TIMEOUT = 30.minutes
+
   def self.included(base)
     base.lev_routine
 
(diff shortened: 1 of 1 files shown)


app/routines/dashboard_routine_methods.rb
```diff
@@ -55,5 +61,10 @@ def load_tasks(role, role_type, start_at_ntz = nil, end_at_ntz = nil, current_ti
                                  .where { first_published_at != nil }
                                  .count <= all_tasks.size
     )
+    if ready_tasks != outputs.tasks
+      err = "TASK READINESS TIMED OUT! student role: #{role.id}"
```
kevinburleigh75: Should we be recording more useful information here to aid debugging?  Like which tasks, the timeout, etc.?  Also, it seems like there might be other reasons these two lists might not be equal - can we pinpoint the timeout message to only the scenarios where we _know_ a timeout occurred?

nathanstitt:
Turns out this fails badly when testing with a "real" BL.  There's multiple other spots that expect BL to be functioning and they throw 500 errors if it's not.

May pick this back up at some point but will close for now.

PR #1833, opened 2019-01-23
remove old score callbacks that have result_sourcedid

nathanstitt:


app/subsystems/lms/launch.rb
```diff
@@ -186,12 +186,33 @@ def store_score_callback(user)
...
+      # remove any duplicated callbacks
+      Lms::Models::CourseScoreCallback.where(
+        result_sourcedid: result_sourcedid,
+        outcome_url: outcome_url
+      ).each do |csc|
+        # if the profile for the duplicate CourseScoreCallback was
+        # abandoned then it's safe to destroy it
+        # But something is broken if the profile has joined a course
+        if UserIsCourseStudent[course: csc.course, user: csc.profile]
+          raise AlreadyUsed
```
jpslav: This exception maps to a "Nonce already used" message in the initial launch call.  Can we use a new exception like  `CourseScoreCallbackAlreadyUsed` so we don't confuse our future selves?  Can be added at the top of this file.

If this exception happens I think the users will get a generic 500 right?  Are you ok with that?

app/subsystems/lms/launch.rb
```diff
@@ -186,12 +186,33 @@ def store_score_callback(user)
...
+        else
+          csc.destroy!
+        end
+      end
+#        .destroy_all
```
jpslav: delete this old code

spec/subsystems/lms/launch_spec.rb
```diff
@@ -11,6 +11,28 @@
...
+      expect { dupe.reload }.to raise_error ActiveRecord::RecordNotFound
+    end
+
+    it 'fails if dupe is in use' do
+      expect(UserIsCourseStudent).to receive(:[]).and_return(true)
+      expect {
+        launch.store_score_callback(dupe.profile)
+      }.to raise_error Lms::Launch::AlreadyUsed
+    end
+  end
```
jpslav: nice spec!

app/subsystems/lms/launch.rb
nathanstitt: no, they see the generic lms "contact your instructor" messages.  However that's not ideal since in this case they need to contact support.  I've now added a custom message to that effect

PR #1835, opened 2019-01-24
add a worked late/accepted late fields to student dash

nathanstitt:
Needed for 
https://github.com/openstax/tutor-js/pull/2499


nathanstitt:
yeah probably should since it didn't fail there's probably coverage lacking somewhere

PR #1836, opened 2019-01-25
Remove course score callbacks and archive students from admin

nathanstitt:
Needed so an LMS account can be un-associated with a tutor student and re-launched to join a different account

goes along with https://github.com/openstax/accounts/pull/644

### DIFF
--- a/app/controllers/admin/courses_controller.rb
+++ b/app/controllers/admin/courses_controller.rb
@@ -38,6 +38,18 @@ class Admin::CoursesController < Admin::BaseController
     redirect_to admin_courses_path
   end
 
+  def drop_student
+    student = CourseMembership::Models::Student.find(params[:id])
+    CourseMembership::InactivateStudent[student: student]
+    redirect_to edit_admin_course_path(student.course) + '#roster'
+  end
+
+  def restore_student
+    student = CourseMembership::Models::Student.find(params[:id])
+    CourseMembership::ActivateStudent[student: student]
+    redirect_to edit_admin_course_path(student.course) + '#roster'
+  end
+
   def create
     handle_with(Admin::CoursesCreate,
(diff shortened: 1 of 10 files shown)


app/controllers/admin/courses_controller.rb
```diff
@@ -38,6 +38,18 @@ def unpair_lms
     redirect_to admin_courses_path
   end
 
+  def drop_student
+    student = CourseMembership::Models::Student.find(params[:id])
```
jpslav: isn't this `:id` the ID of the course?

app/views/layouts/_bootstrap_tab_activate_on_load.html.erb
```diff
@@ -0,0 +1,11 @@
+<script>
+  $(document).ready(function() {
+      var hash = window.location.hash;
+      hash && $('.nav-tabs a[role="tab"][href="' + hash + '"]').tab('show');
+      $('ul.nav.nav-pills a').click(function (e) {
+        $(this).tab('show');
+        var scrollmem = $('body').scrollTop();
+        window.location.hash = this.hash;
+      });
```
jpslav: it is like you know javascript or something

app/controllers/admin/courses_controller.rb
jpslav: per the routes this is the ID of the course, but I see you are explicitly passing in the student ID.  we probably need to make the routes consistent?

app/views/layouts/_bootstrap_tab_activate_on_load.html.erb
nathanstitt: heh, yeah I got tired of loading the page, then switching tabs

PR #1842, opened 2019-02-15
Use Willo Labs key/secret when sending scores

nathanstitt:
Code was previously using the configured LMS App record but those key/secrets aren't valid for willo, instead it needs to use their configuration.

### DIFF
--- a/spec/factories/lms/context.rb
+++ b/spec/factories/lms/context.rb
@@ -1,0 +1,9 @@
+FactoryBot.define do
+  factory :lms_context, class: 'Lms::Models::Context' do
+    association :course, factory: :course_profile_course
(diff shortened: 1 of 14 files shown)


spec/factories/lms/context.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :lms_context, class: 'Lms::Models::Context' do
+    association :course, factory: :course_profile_course
+    association :tool_consumer, factory: :lms_tool_consumer
+
+    lti_id { SecureRandom.uuid }
+    app_type 'Lms::Models::App'
```
jpslav: Make this `Lms::Models::App.class.name` to protect against this getting out of sync if model name is changed later? Super minor

PR #1844, opened 2019-02-15
don't error if nonce already exists

nathanstitt:
We've had Moodle and blackboard send non-unique nonce.   
https://www.google.com/search?q=lms+nonce+duplicate suggests that this is fairly common 😭


nathanstitt:
Am going to close this (but not remove branch) in favor of #1844 - if that change reveals it's happening often we may end up doing this, but we shouldn't do it until we have better data.

PR #1846, opened 2019-02-19
Finish notes PR

nathanstitt:


app/controllers/api/v1/notes_controller.rb
```diff
@@ -47,6 +47,19 @@ def destroy
...
+    #pages = Content::Models::Note.where(role: @role).group(:content_page_id).select(:content_page_id)
+    
+    page_ids = Content::Models::Note.where(role: @role).group(:content_page_id).pluck(:content_page_id)
+    pages = Content::Models::Page.where(id: page_ids)
+    #payload = pages.map{|p| [p.page.book_location, p.page.title] }
+    #respond_with json: { locations: payload }
+
+    #abort HighlightRepresenter.new(pages).inspect
+    respond_with HighlightRepresenter.new(pages)
+
```
nathanstitt: The `resond_with` call needs to be passed an object to render and then a "represent_with" argument.

With the `HighlightRepresenter` the way it is, it's expecting a object with a `pages` method.  You could do something like this to get it working:
```ruby
respond_with(OpenStruct.new({ pages: pages }), represent_with: Api::V1::HighlightRepresenter)
```

Or you could change this to use the `HighlightsRepresenter ` (plural, with `s`) representer to be a collection  and then you could do:

```ruby
respond_with pages, represent_with: Api::V1::HighlightsRepresenter
```

app/representers/api/v1/highlight_representer.rb
```diff
@@ -0,0 +1,17 @@
...
+               as: :children,
+               writeable: false,
+               readable: true,
+               extend: PageTocRepresenter,
+               schema_info: {
+                 required: false,
+                 description: "The pages of the book"
+               }
+  end
+end
```
nathanstitt: I'd suggest adding the properties we want onto this representer from the `BookTocRepresenter` one, maybe something like:

```ruby
module Api::V1

    class HighlightRepresenter < Roar::Decorator

           property :id,
             type: String,
             writeable: false,
             readable: true,

    property :chapter_section,
             type: Array,
             writeable: false,
             readable: true,
             getter: ->(*) { [] },
             schema_info: {
               required: true,
               description: 'Always [] for books'
             }

    end
end
```

app/representers/api/v1/highlights_representer.rb
```diff
@@ -0,0 +1,8 @@
+module Api::V1
+
+    class HighlightsRepresenter < Roar::Decorator
+        include Representable::JSON::Collection
+        items extend: HighlightRepresenter
+    end
+
+end
```
nathanstitt: This is the representer you want to render from the controller 😁

app/representers/api/v1/highlight_representer.rb
```diff
@@ -2,16 +2,17 @@ module Api::V1
   class HighlightRepresenter < Roar::Decorator
 
     include Roar::JSON
-    include Representable::Coercion
+    include Representable::Coercion         
 
     collection :pages,
                as: :children,
```
nathanstitt: The `as: children` renames the property, causing the JSON to be like this:

image

app/controllers/api/v1/notes_controller.rb
```diff
@@ -52,11 +52,9 @@ def highlighted_sections
     
     page_ids = Content::Models::Note.where(role: @role).group(:content_page_id).pluck(:content_page_id)
     pages = Content::Models::Page.where(id: page_ids)
-    #payload = pages.map{|p| [p.page.book_location, p.page.title] }
-    #respond_with json: { locations: payload }
+    
+    respond_with(OpenStruct.new({ pages: pages }), represent_with: Api::V1::HighlightRepresenter)
 
-    #abort HighlightRepresenter.new(pages).inspect
-    respond_with HighlightRepresenter.new(pages)
```
nathanstitt: I played with this some more and the OpenStruct is not needed if the representer uses a getter and I'll outline below.

The representers names also seem to be switched.  The plural `HighlightsRepresenter` should render the collection, and then the singular `HighlightRepresenter` would render each item.  They could also be in the same file, which I'll sketch out below.

app/representers/api/v1/highlight_representer.rb
```diff
@@ -2,16 +2,17 @@ module Api::V1
...
                readable: true,
-               extend: PageTocRepresenter,
+               extend: PageHighlightRepresenter,
                schema_info: {
                  required: false,
                  description: "The pages of the book"
                }
+           
   end
 end
```
nathanstitt: This could combine both representers and if it uses a getter the `OpenStuct` wouldn't be needed.  Something like:

Combining the representers is fine since the `PageHighlightRepresenter` isn't used anywhere outside the main representer. 

```ruby
module Api::V1
  class HighlightRepresenter < Roar::Decorator

    class PageHighlightRepresenter < Roar::Decorator

      include Roar::JSON

      property :id,
               type: String,
               writeable: false,
               readable: true,

      property :chapter_section,
               type: Array,
               writeable: false,
               readable: true,
               getter: ->(*) { book_location },
               schema_info: {
                 required: true,
                 description: 'The chapter and section in the book, e.g. [5, 2]'
               }

      property :title,
               type: String,
               readable: true,
               writeable: false,
               schema_info: {
                 required: true,
                 description: 'The title of the page'
               }

    end

    include Roar::JSON
    include Representable::Coercion

    collection :pages,
               getter: ->(*) { self },
               writeable: false,
               readable: true,
               extend: PageHighlightRepresenter,
               schema_info: {
                 required: false,
                 description: "The pages of the book"
               }
  end
end
```

app/controllers/api/v1/notes_controller.rb
```diff
@@ -47,6 +47,14 @@ def destroy
     render_api_errors(@note.errors) || head(:ok)
   end
 
+  def highlighted_sections
+    OSU::AccessPolicy.require_action_allowed!(:index, current_human_user, @role)
+
```
nathanstitt: we should probably add a comment here noting that we're deliberately passing @role into the `AccessPolicy`.  Maybe just: 
```ruby
# using @role as a substitute for a note in the `AccessPolicy`
```
```

File: app/controllers/api/v1/notes_controller.rb [hunk 8]
```diff
@@ -47,6 +47,14 @@ def destroy
...
 
+  def highlighted_sections
+    OSU::AccessPolicy.require_action_allowed!(:index, current_human_user, @role)
+    
+    page_ids = Content::Models::Note.where(role: @role).group(:content_page_id).pluck(:content_page_id)
+    pages = Content::Models::Page.where(id: page_ids)
+    respond_with pages, represent_with: Api::V1::HighlightRepresenter  
+  end
+
   protected
```
nathanstitt: We need to add documenation to the controller.  take a look at all the comments on `tasks_controller` as an example.  The `resource_description` blocks are turned into documentation for the api.

nathanstitt:
Thanks @KarinaMendez !

PR #1848, opened 2019-02-22
Re-word LMS error when student launches and the ID is already taken, …

KarinaMendez2:
…added my name to authors.


nathanstitt:
Looks like the specs failed. I think you'll have to update the expected text match some of the new wording here https://github.com/openstax/tutor-server/blob/master/spec/requests/lms_launch_spec.rb#L81

PR #1849, opened 2019-03-06
Add a new metatasks controller and json serialization

steveburkett:
Metatasks are lightweight versions of tutoring tasks.  Used to load first more quickly, then when clicked, the full task would load thru the regular /api/tasks API.

https://app.zenhub.com/workspaces/openstax-tutor-560aef617a763234615d6a0f/issues/openstax/tutor-server/1849

### DIFF
--- a/app/controllers/api/v1/metatasks_controller.rb
+++ b/app/controllers/api/v1/metatasks_controller.rb
@@ -1,0 +1,55 @@
+class Api::V1::MetatasksController < Api::V1::ApiController
+
+  before_action :get_task
+  before_filter :error_if_student_and_needs_to_pay, only: [:show, :destroy]
+  before_action :populate_placeholders, only: :show
+
+  resource_description do
+    api_versions "v1"
+    short_description 'Represents a metatask, a light weight task'
+    description <<-EOS
+      Metatasks are a light weight representation of something that a user
+      needs to do in the system (something that has been assigned by
+      another part of the system).  They contain dates like due_at as
+      well as information about whether the task is shared
+    EOS
+  end
+
+  ###############################################################
+  # show
+  ###############################################################
+
+  api :GET, '/metatasks/:id', 'Gets the specified metatask'
+  description <<-EOS
+    #{json_schema(Api::V1::MetataskRepresenter, include: :readable)}
+  EOS
+  def show
+    ScoutHelper.ignore!(0.8)
+    standard_read(Research::ModifiedTaskForDisplay[task: @task], Api::V1::MetataskRepresenter)
+  end
+
+  api :DELETE, '/tasks/:id', 'Hide the task from the student\'s dashboard'
+  description <<-EOS
+    Hides the Task from the student's dashboard
+  EOS
+  def destroy
+    OSU::AccessPolicy.require_action_allowed!(:hide, current_api_user, @task)
+    @task.hide.save!
+    respond_with @task, represent_with: Api::V1::MetataskRepresenter,
+                 responder: ResponderWithPutPatchDeleteContent
+  end
+
+  protected
+
+  def get_task
+    @task = ::Tasks::Models::Task.preload(
+      :research_study_brains,
+      task_steps: [:tasked, :page]
+    ).find(params[:id])
+  end
+
+  def populate_placeholders
+    Tasks::PopulatePlaceholderSteps[task: @task]
+  end
+
+end
--- a/app/representers/api/v1/metatask_representer.rb
+++ b/app/representers/api/v1/metatask_representer.rb
@@ -32,8 +32,7 @@ module Api::V1
                description: "When the metatask was last worked (nil means not yet worked)"
              }
 
-    collection :metatask_steps,
-               as: :metatask_steps,
+    collection :steps,
                writeable: false,
                readable: true,
                extend: MetataskStepRepresenter,
--- a/app/representers/api/v1/metatask_representer.rb
+++ b/app/representers/api/v1/metatask_representer.rb
@@ -1,0 +1,59 @@
+module Api::V1
+  class MetataskRepresenter < Roar::Decorator
+
+    include Roar::JSON
+    include Representable::Coercion
+
+    property :id,
+             type: String,
+             writeable: false,
+             readable: true,
+             schema_info: { required: true }
+
+    property :title,
+             type: String,
+             writeable: false,
+             readable: true,
+             schema_info: { required: true }
+
+    property :due_at,
+             type: String,
+             writeable: false,
+             readable: true,
+             getter: ->(*) { DateTimeUtilities.to_api_s(due_at) },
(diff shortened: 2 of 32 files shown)


app/controllers/api/v1/metatasks_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+    ScoutHelper.ignore!(0.8)
+    standard_read(Research::ModifiedTaskForDisplay[task: @task], Api::V1::MetataskRepresenter)
+  end
+
+  api :PUT, '/tasks/:id/accept_late_work', 'Accept late work in the task score'
+  description <<-EOS
+    Changes the Task so that it is marked as late work being accepted, and changes
+    the `score` that it computes.  Does not change the underlying counts in any way.
+  EOS
+  def accept_late_work
```
steveburkett: @nathanstitt e.g., what to do here for these member actions

app/representers/api/v1/metatasks/task_step_representer.rb
```diff
@@ -0,0 +1,67 @@
...
+               required: true,
+               type: 'boolean',
+               description: "Whether or not this step is complete"
+             },
+             if: NOT_FEEDBACK_ONLY
+
+    alias_method :to_hash_without_cache, :to_hash
+
+    def to_hash(*args)
+      Rails.cache.fetch(represented.cache_key, expires_in: NEVER_EXPIRES) { super }
```
steveburkett: @nathanstitt and here...i think we need to improve this cache_key to differ between meta tasks and regular tasks.

app/representers/api/v1/metatasks/tasked_exercise_representer.rb
```diff
@@ -0,0 +1,80 @@
...
+               description: "The title of this Exercise"
+             },
+             if: NOT_FEEDBACK_ONLY
+
+    property :content,
+             type: String,
+             writeable: false,
+             readable: true,
+             as: :content_preview,
+             getter: ->(*) { self.content.truncate(35, separator: ' ', omission: '... (preview)') },
```
steveburkett: @nathanstitt and here...what to set metatask/step content to.

app/controllers/api/v1/metatasks_controller.rb
nathanstitt: those are used by the teacher to accept (or reject in the corresponding method), work the student has performed after the assignment is due.

app/representers/api/v1/metatasks/task_step_representer.rb
nathanstitt: Actually we want to remove the cache altogether for the new actions.  The cache was just a bandaid on the problem of too much data being generate.

It's caused issues with the Research::ModifiedTaskForDisplay calls because they sometimes need to modify deep properties of the tasks, which the cache doesn't catch.

app/representers/api/v1/metatasks/tasked_exercise_representer.rb
nathanstitt: yes, this is going to be a tricky part of the code.  We need to come up with per-step type rules for generating the previews, and then use a lookup table or some such to run that code for each type.

These previews are used to populate what we call "milestones".  that's a slide down pane that shows each step and its status when you do a reading assignment.  It's triggered by clicking the grid icon on the navbar when you're in a reading :
image


this is roughly what I think we'll need for each step type:
 * exercise: the `stem_html` of the first question
 * external: the title and description of the task itself
 * event: does not have steps so no need for preview
 * reading: currently we're parsing the html of the step and extracting the step title using the logic here: https://github.com/openstax/tutor-js/blob/master/tutor/src/flux/step-title.js#L117-L134
   * I find that pretty hard to follow and it's always fun when we need to tweak it, but I think basically what it does is look for https://github.com/openstax/tutor-js/blob/master/tutor/src/flux/step-title.js#L36-L40 and then read it's text.
  * hopefully something like this is clearer to express using nokogiri or Oga (pick your poison) I'd thought we used nokogiri already but don't see it in the Gemfile

steveburkett:
https://github.com/openstax/tutor-server/pull/1849

app/representers/api/v1/metatasks/task_step_representer.rb
steveburkett: @nathanstitt good call. i like the idea of cleaning up the "hacks" out of the representers and making them just pure clean json DSL.

app/representers/api/v1/metatasks/tasked_exercise_representer.rb
steveburkett: @nathanstitt how about the "interactive" step type?

app/representers/api/v1/metatask_representer.rb
```diff
@@ -0,0 +1,59 @@
...
+             writeable: false,
+             readable: true,
+             getter: ->(*) { DateTimeUtilities.to_api_s(last_worked_at) },
+             schema_info: {
+               description: "When the metatask was last worked (nil means not yet worked)"
+             }
+
+    collection :metatask_steps,
+               as: :metatask_steps,
+               writeable: false,
```
nathanstitt: the `:as` isn't needed if the name is the same as the collection.  

Can we rename it to be just `steps`?  The front-end will map this to an association name and it'll be a bit clunky to have it be `metatask_steps`

app/representers/api/v1/metatasks/task_step_representer.rb
```diff
@@ -0,0 +1,61 @@
...
+             writeable: false,
+             readable: true,
+             getter: ->(*) { task_step.completed? },
+             schema_info: {
+               required: true,
+               type: 'boolean',
+               description: "Whether or not this step is complete"
+             },
+             if: NOT_FEEDBACK_ONLY
+  end
```
nathanstitt: We can take out the lambda `if: NOT_FEEDBACK_ONLY` lines on all the blocks , they'll always be true here

app/representers/api/v1/metatasks/task_step_representer.rb
```diff
@@ -0,0 +1,61 @@
...
+             type: String,
+             writeable: false,
+             readable: true,
+             getter: ->(*) { task_step.tasks_task_id },
+             schema_info: {
+                 required: true,
+                 description: "The id of the Task"
+             },
+             if: NOT_FEEDBACK_ONLY
+
```
nathanstitt: we don't need the task_id since the steps will be nested under the task

app/representers/api/v1/metatasks/tasked_exercise_representer.rb
```diff
@@ -0,0 +1,34 @@
+module Api::V1::Metatasks
```
nathanstitt: we don't need the url or title on any of the step types below.

nathanstitt:
Thanks Steve!

PR #1853, opened 2019-03-19
update to rails 5

nathanstitt:
Is still very incomplete, I started to get a sense for how much work will be required


Dantemss:
I removed `protect_from_forgery with: :exception` because Rails 5.2 (with 5.2 framework defaults) adds that to `ActionController::Base`.

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,25 @@
...
+#   # policy.report_uri "/csp-violation-report-endpoint"
+# end
+
+# If you are using UJS then enable automatic nonce generation
+# Rails.application.config.content_security_policy_nonce_generator = -> request { SecureRandom.base64(16) }
+
+# Report CSP violations to a specified URI
+# For further information see the following documentation:
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
+# Rails.application.config.content_security_policy_report_only = true
```
nathanstitt: commented code ;)

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
+
+# ActiveSupport::Reloader.to_prepare do
+#   ApplicationController.renderer.defaults.merge!(
+#     http_host: 'example.org',
+#     https: false
+#   )
+# end
```
nathanstitt: commented

config/initializers/content_security_policy.rb
Dantemss: These are part of the default Rails initializers: https://github.com/rails/rails/tree/master/railties/lib/rails/generators/rails/app/templates/config/initializers

I think they were added by the upgrade command.
We can discuss and decide what to do with them.

config/initializers/application_controller_renderer.rb
Dantemss: Same as above.

app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -32,7 +32,6 @@ def show
   api :PUT, '/steps/:step_id', 'Updates the specified TaskStep'
   def update
     OSU::AccessPolicy.require_action_allowed!(:update, current_api_user, @tasked)
-    OSU::AccessPolicy.require_action_allowed!(:mark_completed, current_api_user, @tasked)
```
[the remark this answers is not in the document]
Dantemss: This is the only action that calls require_action_allowed! with :mark_completed and it also calls it with :update and they had the same logic in tasked_access_policy so I just combined them...

Gemfile
```diff
@@ -63,13 +63,16 @@ gem 'whenever'
 gem 'omniauth-oauth2', '~> 1.3.1'
 
 # OpenStax Accounts integration
-gem 'openstax_accounts', '~> 7.13.1'
+#gem 'openstax_accounts', '~> 7.13.1'
```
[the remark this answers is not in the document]
Dantemss: Yes I'll do that

PR #1854, opened 2019-03-28
Redo tasks so they load each step individually

nathanstitt:
Continues work from #1849 

This slims down the JSON to only include the bits the FE needs.  it removes quite a few fields, and changes others so they're only returned when the step is loaded, and not as part of the main task payload.


app/controllers/api/v1/task_steps_controller.rb
```diff
@@ -1,6 +1,7 @@
 class Api::V1::TaskStepsController < Api::V1::ApiController
```
nathanstitt: This and the representer files represent the real changes.  We're removing the "completed" action and only using the standard "update".   We can do that since task steps no longer need to be explicitly completed like they did in the past, but the FE wasn't updated to support that until now

reedstrm:
I've connected this to tutor#105, seems like the related work.

app/subsystems/content/routines/import_exercises.rb
```diff
@@ -15,8 +15,10 @@ class Content::Routines::ImportExercises
   # and returns a Content::Models::Page for that exercise
   def exec(ecosystem:, page:, query_hash:, collaborators: [ 'openstax' ])
     # Query the exercises to get a list of OpenStax::Exercises::V1::Exercise
-    query_hash = query_hash.merge collaborator: collaborators.join(',')
+    # query_hash = query_hash.merge collaborator: collaborators.join(',')
+
```
reedstrm: Hmm, looks like your import testing slipped into the PR?

PR #1855, opened 2019-03-30
add feature flags for response validation

nathanstitt:
Adds flag to disable completely and one to disable only the UI

### DIFF
--- a/app/representers/api/v1/bootstrap_data_representer.rb
+++ b/app/representers/api/v1/bootstrap_data_representer.rb
@@ -47,7 +47,13 @@ module Api::V1
              type: Object,
              readable: true,
              writeable: false,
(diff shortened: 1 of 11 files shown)


reedstrm:
Is there an issue for this, so  can see why we're disabling validation? (Normally not a thing I'm in favor of …)

reedstrm:
Ahhh, I see - new, friendly name "response validation" rather than "gibberish/garbage detection" Got it.

nathanstitt:
yep, that's just to unify the naming across the UI and the data model

reedstrm:
A bit confusing on the backend, when 'validation' has so many other meanings.

reedstrm:
Don't know why codecoverage is so wacko on this. Approved.

PR #1857, opened 2019-04-03
allow updating steps to record the free-response info

nathanstitt:
Currently any step updates after the initial one fails because of this validation.


nathanstitt:
@reedstrm the business logic for changing answers is in the tasked_exercise model, not the task_step.  it has a check that the answer_id and free_response cannot be updated after the feedback is available, plus specs to ensure that 😉

PR #1858, opened 2019-04-11
Remove one lms context per course restriction

nathanstitt:
In response to a situation where we encountered multiple tool consumers using
the same context we decided to remove the artificial restriction on a LMS context being linked
to a single course.


db/migrate/20190411181631_lti_course_multi_context.rb
```diff
@@ -0,0 +1,7 @@
+class LtiCourseMultiContext < ActiveRecord::Migration
+  def change
+    # undo the uniqueness added in 20171006150216_add_lms_index_uniqueness_2.rb
+    remove_index :lms_contexts, name: "index_lms_contexts_on_course_profile_course_id"
+    add_index :lms_contexts, :course_profile_course_id, name: "index_lms_contexts_on_course_profile_course_id", unique: false
+  end
+end
```
nathanstitt: our little message in lms_launch about not enforcing uniqueness with a DB constraint was a lie

app/subsystems/lms/unpair_course.rb
```diff
@@ -6,8 +6,8 @@ def exec(course:)
     unless course.is_access_switchable?
       fatal_error(code: :course_access_is_locked, message: "Course access is locked")
     end
-    if course.lms_context.present?
-      course.lms_context.destroy
+    if course.lms_contexts.any?
+      course.lms_contexts.clear
```
jpslav: why not `destroy_all`?

PR #1859, opened 2019-04-12
use page title as fallback if preview can't be parsed

nathanstitt:


app/subsystems/tasks/models/tasked_reading.rb
```diff
@@ -16,8 +16,7 @@ def has_learning_objectives?
   end
 
   def content_preview
-    text = document_title.presence || data_title.presence || class_title.presence
-    text || "Reading Step ##{id}"
+    document_title.presence || data_title.presence || class_title.presence || task_step.page.title
```
reedstrm: So there is _always_ a page title? (no desperate last fallback of a fixed string)

nathanstitt: yeah, the "page" here corresponds to a CNX module, which should always be titled.  the DB column is also non-null

PR #1862, opened 2019-05-02
Fix task step population

nathanstitt:


nathanstitt:
looks great @Dantemss !

PR #1864, opened 2019-05-06
Sparfa admin settings

Dantemss:
- Change the Biglearn algorithm selection in the admin settings page to a dropdown select
- Make all courses use the chosen setting unless manually overridden

### DIFF
--- a/db/migrate/20170208182122_add_uuid_and_group_uuid_to_content_exercises.rb
+++ b/db/migrate/20170208182122_add_uuid_and_group_uuid_to_content_exercises.rb
@@ -2,9 +2,10 @@ class AddUuidAndGroupUuidToContentExercises < ActiveRecord::Migration
   def change
     enable_extension 'pgcrypto'
 
-    add_column :content_exercises, :uuid, :uuid, null: false, index: true
+    add_column :content_exercises, :uuid, :uuid, null: false
     add_column :content_exercises, :group_uuid, :uuid, null: false
 
+    add_index :content_exercises, :uuid
     add_index :content_exercises, [:group_uuid, :version]
   end
 end
--- a/spec/factories/course_profile/courses.rb
+++ b/spec/factories/course_profile/courses.rb
@@ -18,12 +18,7 @@ FactoryBot.define do
 
     uuid                  { SecureRandom.uuid }
 
-    sequence_number                              { rand(1000) + 1 }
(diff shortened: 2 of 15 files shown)


db/migrate/20170208182122_add_uuid_and_group_uuid_to_content_exercises.rb
```diff
@@ -2,9 +2,10 @@ class AddUuidAndGroupUuidToContentExercises < ActiveRecord::Migration
   def change
     enable_extension 'pgcrypto'
 
-    add_column :content_exercises, :uuid, :uuid, null: false, index: true
+    add_column :content_exercises, :uuid, :uuid, null: false
     add_column :content_exercises, :group_uuid, :uuid, null: false
 
+    add_index :content_exercises, :uuid
     add_index :content_exercises, [:group_uuid, :version]
```
nathanstitt: interesting.  what's the difference between `index: true` and `add_index` ?

spec/factories/course_profile/courses.rb
```diff
@@ -18,12 +18,7 @@
 
     uuid                  { SecureRandom.uuid }
 
-    sequence_number                              { rand(1000) + 1 }
-    biglearn_student_clues_algorithm_name        { Faker::Hacker.abbreviation }
-    biglearn_teacher_clues_algorithm_name        { Faker::Hacker.abbreviation }
-    biglearn_assignment_spes_algorithm_name      { Faker::Hacker.abbreviation }
-    biglearn_assignment_pes_algorithm_name       { Faker::Hacker.abbreviation }
-    biglearn_practice_worst_areas_algorithm_name { Faker::Hacker.abbreviation }
+    sequence_number       { rand(1000) + 1 }
```
nathanstitt: FYI, I recently learned `rand` accepts a range argument.  you can do: `rand(1..100)`

db/migrate/20170208182122_add_uuid_and_group_uuid_to_content_exercises.rb
Dantemss: `index: true` actually does not work with `add_column` (weird right?)

spec/factories/course_profile/courses.rb
Dantemss: Oh nice :) but I think I'll leave this one for now

PR #1865, opened 2019-05-10
Fix drop RoleUsers migration so it doesn't delete everything

Dantemss:
Needed so we don't lose data.

### DIFF
--- a/db/migrate/20190426205807_drop_role_role_users.rb
+++ b/db/migrate/20190426205807_drop_role_role_users.rb
@@ -7,7 +7,7 @@ class DropRoleRoleUsers < ActiveRecord::Migration
         Entity::Role.reset_column_information
 
         Entity::Role.find_each do |role|
-          user_profile_id = ActiveRecord::Base.connection.execute(
+          row = ActiveRecord::Base.connection.execute(
             <<-SQL.strip_heredoc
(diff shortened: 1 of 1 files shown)


db/migrate/20190426205807_drop_role_role_users.rb
```diff
@@ -7,15 +7,17 @@ def change
...
               SELECT "role_role_users"."user_profile_id"
               FROM "role_role_users"
               WHERE "role_role_users"."entity_role_id" = #{role.id}
             SQL
           ).to_a.first
 
-          role.update_attribute(:user_profile_id, user_profile_id) unless user_profile_id.nil?
+          next if row.nil?
+
+          role.update_attribute(:user_profile_id, row['user_profile_id'])
```
nathanstitt: how about using a bang so it'll die if it fails:  
```ruby
role.update_attributes!(user_profile_id: row['user_profile_id'])
```

Dantemss: I gave it a try in the console but even this version wants to set the user_profile_id to null if you pass it a hash instead of failing.

nathanstitt: ah, ok then

PR #1866, opened 2019-05-15
Restored the optional role_id in the student guide route

Dantemss:
Turns out this was actually used by something.


nathanstitt:
this fixes https://github.com/openstax/tutor/issues/423

PR #1867, opened 2019-05-17
Allow teacher_students to use the practice widget

Dantemss:
Fixes the 500 error when the ghost student tries to use the practice widget.


nathanstitt:
Can we return both the student role and the student record from the call to `PUT period/<id>/teacher_student`. Right now [I'm creating a fake student record](https://github.com/openstax/tutor-js/pull/2616/files#diff-9bb1effba6722ac2d015ee996aed33a2R64) but I might as well use the real one if it's easy to return

PR #1868, opened 2019-05-23
remove openstax collaborator requirement when importing exercises

nathanstitt:
this is temporary until we can update exercises so it adds a copyright holder


reedstrm:
verified that the calling location does not set `collaborators`, so this is sufficent.

PR #1871, opened 2019-06-04
Don't send null exercise_uuids to biglearn-api

Dantemss:
If some exercise has been deleted from the database. Don't ask how they have been deleted...


Dantemss:
Closing. Will fix this issue in biglearn-sparfa-server instead.

Dantemss:
Kevin prefers to fix the invalid events, so this patch should be merged to ensure they do not reoccur.

PR #1877, opened 2019-06-13
Include period_id with roles

nathanstitt:
This will allow the FE teacher as student code to find the role and re-use it

### DIFF
--- a/app/representers/api/v1/role_representer.rb
+++ b/app/representers/api/v1/role_representer.rb
@@ -10,6 +10,13 @@ module Api::V1
              writeable: false,
              schema_info: { required: true }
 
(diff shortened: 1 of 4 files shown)


app/representers/api/v1/role_representer.rb
```diff
@@ -10,6 +10,13 @@ class RoleRepresenter < Roar::Decorator
              writeable: false,
              schema_info: { required: true }
 
+    property :period_id,
+             getter: ->(*) { teacher? ? nil : course_member.course_membership_period_id },
```
Dantemss: This could be instead `if: ->(*) { !teacher? }` but it's very minor.

PR #1882, opened 2019-07-15
Fix Tutor not sending (some) data to Salesforce

Dantemss:
This PR updates accounts-rails to the newest version, which fixes a bug where the `salesforce_contact_id` would not be set on Account creation, causing any actions the user takes within 1 minute of Account creation to fail to be sent to Salesforce.

It also fixes roles never syncing properly and adds the new `homeschool` role type.

And finally it replaces the DevMailer (which was broken in production) with sending the errors to Sentry.


app/routines/push_salesforce_course_stats.rb
```diff
@@ -212,24 +207,35 @@ def skip!(message: nil, course: nil, period: nil)
...
-    DevMailer.inspect_object(
-      object: @errors,
-      subject: "#{self.class.name} errors",
-      to: Rails.application.secrets.salesforce[:mail_recipients]
-    ).deliver_later if allow_error_email && is_real_production?
+      if exception.nil?
+        Raven.capture_message error[:message], error.except(:message)
+      else
+        Raven.capture_exception exception, error.except(:exception)
+      end
```
nathanstitt: 💯 will be so nice to kill those emails 😁

PR #1884, opened 2019-07-16
Fix loading scores report as teacher_student

nathanstitt:
The query lacked support for joining to the teacher_student role

### DIFF
--- a/app/subsystems/tasks/get_performance_report.rb
+++ b/app/subsystems/tasks/get_performance_report.rb
@@ -154,11 +154,12 @@ module Tasks
     # reorder(nil) is required for distinct to work
     # distinct is required for preloading to work
     def get_course_tasks(course, role, is_teacher, current_time_ntz)
+      is_teacher_student = !is_teacher && role.present? && role.teacher_student?
       task_types = Tasks::Models::Task.task_types.values_at(:reading, :homework, :external)
       tt = Tasks::Models::Task.arel_table
       er = Entity::Role.arel_table
       st = CourseMembership::Models.const_get(
-        role.teacher_student? ? 'TeacherStudent' : 'Student'
+        is_teacher_student ? 'TeacherStudent' : 'Student'
       ).arel_table
       up = User::Models::Profile.arel_table
       ac = OpenStax::Accounts::Account.arel_table
@@ -167,11 +168,11 @@ module Tasks
           [
             tt[ Arel.star ],
             er[:id].as('"role_id"'),
-            role.teacher_student? ?
+            is_teacher_student ?
               Arel::Nodes::SqlLiteral.new("'' as student_identifier") : st[:student_identifier],
             st[:course_membership_period_id],
-            role.teacher_student? ?
-              Arel::Nodes::SqlLiteral.new('null as dropped_at') :  st[:dropped_at],
+            is_teacher_student ?
+              Arel::Nodes::SqlLiteral.new('null as dropped_at') : st[:dropped_at],
             up[:id].as('"user_id"'),
             ac[:username],
             ac[:first_name],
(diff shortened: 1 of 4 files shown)


app/subsystems/tasks/get_performance_report.rb
```diff
@@ -151,20 +154,25 @@ def exec(course:, role: nil, is_teacher: nil)
...
     # distinct is required for preloading to work
     def get_course_tasks(course, role, is_teacher, current_time_ntz)
+      is_teacher_student = !is_teacher && role.present? && role.teacher_student?
       task_types = Tasks::Models::Task.task_types.values_at(:reading, :homework, :external)
       tt = Tasks::Models::Task.arel_table
       er = Entity::Role.arel_table
-      st = CourseMembership::Models::Student.arel_table
+      st = CourseMembership::Models.const_get(
+        is_teacher_student ? 'TeacherStudent' : 'Student'
+      ).arel_table
```
Dantemss: Could do it without `const_get` I think... both classes should be loaded.

app/subsystems/tasks/get_performance_report.rb
```diff
@@ -151,20 +154,25 @@ def exec(course:, role: nil, is_teacher: nil)
     # reorder(nil) is required for distinct to work
     # distinct is required for preloading to work
     def get_course_tasks(course, role, is_teacher, current_time_ntz)
+      is_teacher_student = !is_teacher && role.present? && role.teacher_student?
```
Dantemss: Theoretically no need to check `is_teacher` because it would always be false for a `teacher_student` role.

app/subsystems/tasks/get_performance_report.rb
```diff
@@ -151,20 +154,25 @@ def exec(course:, role: nil, is_teacher: nil)
...
           [
             tt[ Arel.star ],
             er[:id].as('"role_id"'),
-            st[:student_identifier],
+            is_teacher_student ?
+              Arel::Nodes::SqlLiteral.new("'' as student_identifier") : st[:student_identifier],
             st[:course_membership_period_id],
-            st[:dropped_at],
+            is_teacher_student ?
+              Arel::Nodes::SqlLiteral.new('null as dropped_at') : st[:dropped_at],
```
Dantemss: Might be a good idea to do `st[:deleted_at].as('dropped_at')` here (or do that in raw sql)

app/subsystems/tasks/get_performance_report.rb
```diff
@@ -173,18 +181,32 @@ def get_course_tasks(course, role, is_teacher, current_time_ntz)
...
         )
         .where(tt[:opens_at_ntz].eq(nil).or tt[:opens_at_ntz].lteq(current_time_ntz))
         .preload(task_plan: :tasking_plans)
         .reorder(nil).distinct
 
-      rel = rel.joins(:taskings).where(taskings: { entity_role_id: role.id }) unless is_teacher
+      if is_teacher
+        rel = rel.joins(
+          taskings: { role: [ :student, profile: :account ] }
+        )
```
Dantemss: This one needs the where that was removed: `where(taskings: { role: { student: { course_profile_course_id: course.id } } })`
Otherwise the server will blow up :)

app/subsystems/tasks/get_performance_report.rb
```diff
@@ -173,18 +181,32 @@ def get_course_tasks(course, role, is_teacher, current_time_ntz)
...
+      elsif is_teacher_student
+        rel = rel.joins(
+          taskings: { role: [ :teacher_student, profile: :account ] }
+          ).where(
+            taskings: { role: role }
+          )
+      elsif role.student?
```
Dantemss: Might as well just make this into an `else` just to make sure we avoid loading all tasks and exploding the server if all the if checks fail.

app/subsystems/tasks/get_performance_report.rb
nathanstitt: yeah, I was attempting to be clever..

Dantemss:
Oh I just realized the teacher_student will not show up when the teacher is viewing the scores (as a teacher). But I think that is what we want.

nathanstitt:
yeah, I think that's what we want.  Although the old UI was still showing all the teacher controls for export, LMS etc.  I updated the UI to remove them

PR #1885, opened 2019-07-16
include student role id with tasks

nathanstitt:
This allows the FE to detect when a teacher is viewing a student or
working a ghost student assignment


nathanstitt:
nah the FE knows the task and step ids, but it doesn't know what role they belong to.  Once it nows that it can decide if the role is "teacher-student" or just student

PR #1886, opened 2019-07-18
Change study brain to alternate between all sections

nathanstitt:
Removes the hard-coded sections list and just checks for `even?` or `odd?` on the page


nathanstitt:
I'll leave this open until we actually test the brain on QA.  may need to tweak it based on that

PR #1888, opened 2019-08-02
Fix exception rescues failing due to nil backtrace

Dantemss:
Greg was getting some failed jobs on -dev and we couldn't figure out what they were because the exception reporting was failing due to a nil backtrace.


Dantemss:
I think getting codecov to cover these lines is outside the scope of this PR.

PR #1890, opened 2019-08-05
Fix the bootstrap data test

Dantemss:
To fix tests on master


Dantemss:
I'm going to merge this to fix tests on master. It's only a test file anyway.

PR #1895, opened 2019-08-20
send non-open tasks to teacher-students

nathanstitt:
And allow them to be worked

### DIFF
--- a/app/access_policies/tasked_access_policy.rb
+++ b/app/access_policies/tasked_access_policy.rb
@@ -4,16 +4,14 @@ class TaskedAccessPolicy
     task = tasked.task_step.task
     case action
     when :read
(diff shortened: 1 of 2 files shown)


app/access_policies/tasked_access_policy.rb
```diff
@@ -1,18 +1,19 @@
...
   def self.action_allowed?(action, requestor, tasked)
+    return false unless requestor.is_human?
+    task = tasked.task_step.task
     case action
     when :read
-      return false unless requestor.is_human?
       return true if DoesTaskingExist[task_component: tasked, user: requestor] &&
-                     tasked.task_step.task.past_open?
-      period = tasked.task_step.task.taskings.first.try(:period)
+        tasked.task_step.task.past_open?
```
Dantemss: Could shorten this since you defined `task` above, but meh it works.

PR #1897, opened 2019-08-21
More demo routine fixes

Dantemss:
- Re-enable calling the rake task with a book name (e.g. `rake demo[soc]`)
- Switch to read committed transactions for demo routines so we can successfully retry parts of transactions
- Allow demo:all to be called with only some of its arguments
- Allow only some of the users to be created with `rake demo:users`


app/routines/demo/all.rb
```diff
@@ -1,6 +1,6 @@
 # Performs all demo routines (except Users and Show) in succession
 class Demo::All < Demo::Base
-  lev_routine use_jobba: true
+  lev_routine transaction: :read_committed, use_jobba: true
```
reedstrm: Ah this is where the discussion about read_repeatable came from. Odd that you have to set this on all of these, individually. I guess even if we find where the default is being set, this is belt-and-suspenders, in case it changes again.

Dantemss: The default is set somewhere in github.com/lml/lev

PR #1906, opened 2019-09-05
Keep all preview course background jobs in the lower priority queues

Dantemss:
That way they should not interfere with higher priority background jobs.


app/jobs/openstax/biglearn/api/job_with_sequence_number.rb
```diff
@@ -15,7 +15,7 @@ def self.perform_later(*_)
   end
 
   def perform(
-    method:, requests:, create:, sequence_number_model_key:, sequence_number_model_class:,
+    method:, requests:, create:, sequence_number_model_key:, sequence_number_model_class:, queue:,
     response_status_key: nil, accepted_response_status: []
```
nathanstitt: travis has found a few other spots that call this and don't provide the `queue` argument

reedstrm:
looks like we need some specs for submitting inline jobs, and for jobs without a sequence number

PR #1907, opened 2019-09-06
Fixed assignments loading forever when an assignment was not open yet

Dantemss:
And added specs

### DIFF
--- a/spec/factories/tasks/task_plans.rb
+++ b/spec/factories/tasks/task_plans.rb
@@ -25,10 +25,8 @@ FactoryBot.define do
       task_plan.assistant ||= Tasks::Models::Assistant.find_by(code_class_name_hash) ||
                               build(:tasks_assistant, code_class_name_hash)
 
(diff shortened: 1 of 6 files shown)


spec/factories/tasks/task_plans.rb
```diff
@@ -29,12 +28,18 @@
...
       task_plan.ecosystem ||= build(:content_ecosystem)
 
-      task_plan.tasking_plans = evaluator.num_tasking_plans.times.map do
-        build(:tasks_tasking_plan,
-              task_plan: task_plan,
-              opens_at: evaluator.opens_at,
-              due_at: evaluator.due_at,
-              time_zone: task_plan.owner.try(:time_zone))
+      task_plan.owner ||= evaluator.target.try(:course) || build(:course_profile_course)
+
```
nathanstitt: seems strange that owner wouldn't be set, but I'm guessing this is needed 😁

Dantemss: I removed the `association :owner` so we could build it dynamically based on the target in `after(:build)`

PR #1910, opened 2019-09-18
Rename all the things

Dantemss:
- Renamed task_steps core_group to fixed_group
- Added task_steps is_core boolean
- Updated specs
- Switched from chromedriver-helper to webdrivers, since the former is no longer maintained

Need to ensure FE can handle the slight API changes (changing the group_name and adding is_core).


db/background_migrate/20190918184315_set_tasks_task_steps_is_core.rb
```diff
@@ -0,0 +1,32 @@
...
+      tasked_type: Tasks::Models::TaskedExternalUrl.name
+    ).limit(BATCH_SIZE).update_all(group_type: :fixed_group, is_core: true) >= BATCH_SIZE do
+    end
+
+    while Tasks::Models::TaskStep.where(
+      is_core: nil
+    ).limit(BATCH_SIZE).update_all(is_core: false) >= BATCH_SIZE do
+    end
+
+    change_column_null :tasks_task_steps, :is_core, false
```
nathanstitt: This seems strange to me.  I've never encountered `update_all` with a limit.  I also can't find any reference to it returning the # of updated records but it seems like that's what you're using with the `>= BATCH_SIZE` at the end and the `do end` loop.

https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-update_all

I guess what I'm asking is how well have you tested this with a large sample size to make sure the loop does indeed update everything?

db/migrate/20190918184534_add_is_core_to_tasks_task_steps.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    add_column :tasks_task_steps, :is_core, :boolean
+    # We still do migrations inline, with the server down, so we can still use rename_column...
+    rename_column :tasks_tasks, :core_and_personalized_placeholder_exercise_steps_count,
+                                :core_placeholder_exercise_steps_count
+
+    reversible do |dir|
+      dir.up   { BackgroundMigrate.perform_later 'up',   20190918184315 }
+      dir.down { BackgroundMigrate.perform_later 'down', 20190918184315 }
+    end
```
nathanstitt: Performing in background will mean that until the migration runs the task steps will have incorrect names.  Is that something we can deal with?

Maybe we could add a `order_by` clause so the most recent get updated first since that's what students will be working.

app/subsystems/tasks/models/task_step.rb
```diff
@@ -23,18 +23,29 @@ class Tasks::Models::TaskStep < ApplicationRecord
...
-    task.try! :lock!, *args
+    task&.lock! *args
 
     super
   end
 
+  def is_core?
+    is_core
+  end
+
```
nathanstitt: Boolean columns automatically get a method with `?` at the end so this method probably isn't needed

app/routines/get_task_core_page_ids.rb
```diff
@@ -12,16 +8,13 @@ class GetTaskCorePageIds
...
+    unloaded_task_steps = Tasks::Models::TaskStep.where(
+      tasks_task_id: unloaded_tasks.map(&:id), is_core: true
+    ).pluck(:tasks_task_id, :content_page_id).group_by(&:first)
 
     task_id_to_core_page_ids_map = {}
     loaded_tasks.each do |task|
-      task_steps = task.task_steps.to_a.select do |task_step|
-        task_step.core_group? || task_step.personalized_group?
-      end
+      task_steps = task.task_steps.to_a.select(&:is_core?)
```
nathanstitt: instead of calling `to_a.select` you can use `filter` which works on a AR result set.  the result set includes Enumerable so you can call any of them except `select` because that selects

db/background_migrate/20190918184315_set_tasks_task_steps_is_core.rb
Dantemss: From my own testing and https://stackoverflow.com/a/52189304 it seems `update_all` returns the number of records updated. That said, it might be simpler to use this: https://github.com/rails/rails/blob/98a57aa5f610bc66af31af409c72173cdeeb3c9e/activerecord/lib/active_record/relation/batches.rb#L158 which is actually documented.

So I'll change it to use `in_batches`.

db/migrate/20190918184534_add_is_core_to_tasks_task_steps.rb
Dantemss: They will get the change from `core` to `fixed` instantly because that is just an integer in the DB and I'm changing the enum on the ruby side. `is_core` will be incorrect until the background migration runs. Are you using the `group_name` for anything right now?

app/routines/get_task_core_page_ids.rb
Dantemss: You mean `Array#filter` right? Yeah I guess it makes sense to use that rather than `.to_a.select`

Dantemss: Just `select` should also work according to the docs (the second sentence in https://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/select) but `filter` is clearer.

nathanstitt: Ah, interesting. I wasn't aware that calling `select` with a block will cause it to use Enumerable select instead of sql select.  Neat, thanks!

Dantemss:
@nathanstitt updated... let me know if you think the FE will have problems with the change from `core` to `fixed`. I couldn't find anything in tutor-js that seemed to care about that name directly.

db/migrate/20190918184534_add_is_core_to_tasks_task_steps.rb
nathanstitt: Nope, in fact I don't see that we've even added `group_name` to the model so it doesn't even get stored currently.

PR #1911, opened 2019-09-20
Performance forecast fixes

Dantemss:
- Don't use withdrawn tasks in the student guide and period caches.
  Period caches are used to display teacher stuff and teachers don't care about withdrawn assignments.
- Added task_plan-related fields to task_caches.
- Add first_worked_at/last_worked_at to task_caches. Only exercise steps set those.
- Send first_worked_at/last_worked_at to the frontend.

Needed for: https://github.com/openstax/tutor-js/pull/2741

### DIFF
--- a/db/migrate/20190923160455_add_task_plan_and_withdrawn_at_to_task_caches.rb
+++ b/db/migrate/20190923160455_add_task_plan_and_withdrawn_at_to_task_caches.rb
@@ -1,0 +1,11 @@
+class AddTaskPlanAndWithdrawnAtToTaskCaches < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :tasks_task_caches, :tasks_task_plan,
+                  foreign_key: { on_update: :cascade, on_delete: :cascade }
+    add_column :tasks_task_caches, :withdrawn_at, :datetime
+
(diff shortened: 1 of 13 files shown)


db/migrate/20190923160455_add_task_plan_and_withdrawn_at_to_task_caches.rb
```diff
@@ -0,0 +1,11 @@
+class AddTaskPlanAndWithdrawnAtToTaskCaches < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :tasks_task_caches, :tasks_task_plan,
+                  foreign_key: { on_update: :cascade, on_delete: :cascade }
+    add_column :tasks_task_caches, :withdrawn_at, :datetime
+
```
nathanstitt: how likely is it that the `add_reference` will fail either because there's null or other keys that aren't present?

If there are those, we could add a delete statement before this to remove them

nathanstitt: yep, I think I'd just mis-read it and thought there was a not_null in there as well

Dantemss:
Per Carly we are changing several things in Tutor to be exercise-based so the last_worked_at etc dates here will be based on exercise steps only.

PR #1915, opened 2019-10-02
Save and let Biglearn know which calculation and ecosystem matrix uuids we used

Dantemss:
Research data etc

- Receive calculation and ecosystem matrix uuids from biglearn-api
- Save used calculation and ecosystem matrix uuids in tasks
- Send used calculation and ecosystem matrix uuids back to biglearn-api


app/representers/api/v1/book_toc_representer.rb
```diff
@@ -90,10 +90,6 @@ class BookTocRepresenter < Roar::Decorator
              }
 
     property :chapter_section,
-             getter: ->(*) {
-               baked_chapter_section.blank? ?
-                 chapter_section : baked_chapter_section
-             },
```
reedstrm: Seems unrelated to the rest of the PR?

lib/openstax/biglearn/api/real_client.rb
```diff
@@ -396,6 +396,19 @@ def create_update_assignments(requests)
...
+      spe_ecosystem_matrix_uuid = task.spe_ecosystem_matrix_uuid
+
+      pes = {}
+      pes[:calculation_uuid] = pe_calculation_uuid if pe_calculation_uuid.present?
+      pes[:ecosystem_matrix_uuid] = pe_ecosystem_matrix_uuid if pe_ecosystem_matrix_uuid.present?
+
+      spes = {}
+      spes[:calculation_uuid] = spe_calculation_uuid if spe_calculation_uuid.present?
+      spes[:ecosystem_matrix_uuid] = spe_ecosystem_matrix_uuid if spe_ecosystem_matrix_uuid.present?
+
```
reedstrm: I assume these have to be guarded because old task-records won't have them?

app/representers/api/v1/book_toc_representer.rb
nathanstitt: yeah, it's duplicated with another `getter` below that will override this one, causing it to never be called

lib/openstax/biglearn/api/real_client.rb
Dantemss: Both tasks that never had PEs and/or SPEs assigned and old records will have `nil`s here, so I decided to send a {} instead of sending the nulls...

PR #1918, opened 2019-10-09
return baked chapter section if book is collated

nathanstitt:
Otherwise non-number sections like "Connection for AP Courses" are numbered 1.1 and duplicate the baked numbers

@Dantemss I'm not sure if this is the best way to accomplish this, perhaps it'd be better to no check the book on the page_toc representer but that was a n+1 query so it'd take some careful preloading


Dantemss:
Let's ask Carly about this in scrum.

Dantemss:
Oh right if you go page -> book it will probably be N+1. But if you go page -> chapter -> book it will probably work. One of the more annoying things about preloading.

nathanstitt:
Yep, following page->chapter->book.is_collated worked to avoid the n+1.  Thanks!

PR #1927, opened 2019-10-29
Tell Biglearn when task_plans have been withdrawn or restored

Dantemss:
Probably want to merge/release openstax_api and update the Gemfile before merging this one.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,7 +71,7 @@ gem 'action_interceptor'
 gem 'chronic'
 
 # API versioning and documentation
-gem 'openstax_api'
+gem 'openstax_api', github: 'openstax/openstax_api', ref: 'ee2f081141211352460c6a8a71d8040cd3f78766'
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -71,7 +71,7 @@ gem 'action_interceptor'
 gem 'chronic'
 
 # API versioning and documentation
-gem 'openstax_api'
+gem 'openstax_api', github: 'openstax/openstax_api', ref: 'ee2f081141211352460c6a8a71d8040cd3f78766'
```
nathanstitt: Hm, is a ref intended?  Usually we do either tag or version, right?

Dantemss: Ah yes need to release the gem

Dantemss:
Travis failed with a completely unrelated failure. I pushed a fix for that. Specs pass locally. Merging...

PR #1928, opened 2019-10-30
Fix tasks not being assigned when a new section is added

Dantemss:
- Combined DistributeTasks and PropagateTaskPlanUpdates

The combined routine does everything necessary to publish and/or update a task_plan's tasks, so this should prevent us from missing some updates because we picked the wrong routine.


spec/subsystems/tasks/assistants/homework_assistant_spec.rb
```diff
@@ -21,7 +21,7 @@
       dynamic_exercises = dynamic_pools.flat_map(&:exercises).sort_by(&:uid)
 
       @teacher_selected_exercises = core_exercises[1..5]
-      teacher_selected_exercise_ids = @teacher_selected_exercises.map{|e| e.id.to_s}
+      teacher_selected_exercise_ids = @teacher_selected_exercises.map { |e| e.id.to_s }
```
nathanstitt: why do the ids need to be strings again?  If it's for the FE we no longer care about that, numeric is fine.

Dantemss: I think it's a leftover from the days FE cared :D

PR #1933, opened 2019-11-08
 Changed Salesforce login to use the password flow

Dantemss:
To try to prevent non-prod environments from disconnecting from the SF sandbox.

Just need to release openstax_salesforce v4.0.0 and update the Gemfile.


app/routines/track_tutor_onboarding_event.rb
```diff
@@ -84,9 +84,6 @@ def exec(event:, user:, data: {})
       raise
     rescue MissingArgument => ee
       raise(MissingArgument, "Missing the `#{ee.message}` argument for event #{event}")
-    rescue OpenStax::Salesforce::UserMissing => ee
-      log(:error) { "Cannot track onboarding event because Salesforce user not set" }
-      raise if Settings::Db.raise_if_salesforce_user_missing
     end
```
nathanstitt: Hooray!  this will be great to get rid of

PR #1936, opened 2019-11-11
break apart reading/homework task plans, add nudge stats

nathanstitt:
This way the assignment will use same format as task completion

### DIFF
--- a/app/subsystems/stats/calculations/exercises.rb
+++ b/app/subsystems/stats/calculations/exercises.rb
@@ -10,6 +10,17 @@ class Stats::Calculations::Exercises
                tasked_type: "Tasks::Models::Tasked#{task_type.classify}")
         .count
     end
-  end
 
+
+    nudge_stats = Tasks::Models::TaskedExercise
+      .where('response_validation is not null')
+      .where(:updated_at => interval.range)
+      .select(
+      "coalesce(sum((case when jsonb_array_length(response_validation->'attempts') > 0 then 1 else 0 end)), 0) as nudge_calculated,
+      coalesce(sum((case when (response_validation->'attempts'->0->'valid') = 'true' then 0 else 1 end)), 0) as nudge_initially_invalid,
+      coalesce(sum((case when (response_validation->'attempts'->1->'valid') = 'true' then 1 else 0 end)), 0) as nudge_retry_correct"
+    ).all[0].as_json(except: :id)
+
+    interval.stats.merge!(nudge_stats)
+  end
 end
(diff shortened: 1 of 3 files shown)


app/subsystems/stats/calculations/exercises.rb
```diff
@@ -10,6 +10,17 @@ def exec(interval:)
                tasked_type: "Tasks::Models::Tasked#{task_type.classify}")
         .count
     end
-  end
 
+
+    nudge_stats = Tasks::Models::TaskedExercise
+      .where('response_validation is not null')
```
Dantemss: Might wanna exclude `'{}'` too? Though I'm not 100% sure it happened in the data.

app/subsystems/stats/calculations/exercises.rb
```diff
@@ -10,6 +10,17 @@ def exec(interval:)
...
+      .where(:updated_at => interval.range)
+      .select(
+      "coalesce(sum((case when jsonb_array_length(response_validation->'attempts') > 0 then 1 else 0 end)), 0) as nudge_calculated,
+      coalesce(sum((case when (response_validation->'attempts'->0->'valid') = 'true' then 0 else 1 end)), 0) as nudge_initially_invalid,
+      coalesce(sum((case when (response_validation->'attempts'->1->'valid') = 'true' then 1 else 0 end)), 0) as nudge_retry_correct"
```
Dantemss: Oh I guess the coalesce is for `'{}'`?

app/subsystems/stats/calculations/exercises.rb
```diff
@@ -10,6 +10,17 @@ def exec(interval:)
...
+    ).all[0].as_json(except: :id)
```
Dantemss: Maybe it's easier to return the whole json and process in the FE? Considering the FE is the one that makes it...

app/subsystems/stats/calculations/exercises.rb
nathanstitt: Yeah, I'm not sure either.  I'll update to filter those also

app/subsystems/stats/calculations/exercises.rb
nathanstitt: nah, that's for the case where there were no rows found, in which case `sum` will return `null`

app/subsystems/stats/calculations/exercises.rb
nathanstitt: probably not the best idea since were currently at 63,000 or so results 😁

PR #1945, opened 2019-11-22
Add stats for practice and limit to current courses

nathanstitt:
mostly just a few minor tweaks to how data is gathered

### DIFF
--- a/app/subsystems/stats/calculations/exercises.rb
+++ b/app/subsystems/stats/calculations/exercises.rb
@@ -26,6 +26,7 @@ class Stats::Calculations::Exercises
 
     nudge_stats = Tasks::Models::TaskedExercise
       .joins(task_step: :task)
+      .where(:updated_at => interval.range)
       .where("coalesce(jsonb_array_length(response_validation->'attempts'), 0) > 0")
       .select(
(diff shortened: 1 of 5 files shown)


app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -5,7 +5,7 @@ class Tasks::Models::TaskedExercise < IndestructibleRecord
   auto_uuid
 
   belongs_to :exercise, subsystem: :content, inverse_of: :tasked_exercises
-
+  has_one :task_step, as: :tasked
```
Dantemss: `acts_as_tasked` already does this

app/subsystems/stats/calculations/exercises.rb
```diff
@@ -6,14 +6,28 @@ class Stats::Calculations::Exercises
...
         .count
     end
 
+    interval.stats["practice_steps"] = Tasks::Models::TaskStep
+      .joins(:task)
+      .where(
+        :first_completed_at => interval.range,
+        task: { task_type: %w[practice_worst_topics page_practice] }
+      )
+      .count
```
Dantemss: Well this will count all courses including preview/test... is that ok?

nathanstitt: nope, we don't want that. I'll update. thanks!

PR #1946, opened 2019-11-23
Fix reading assignment links

Dantemss:
The fragments were being cached before the content was modified.


app/subsystems/content/models/book.rb
```diff
@@ -50,6 +50,8 @@ def set_ecosystem_title
   end
 
   def reference_view_url
+    raise('Unpersisted Ecosystem') if ecosystem.id.nil?
+
```
nathanstitt: I think you can also do `ecosystem.persisted?` https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-persisted-3F

app/subsystems/content/models/page.rb
```diff
@@ -131,6 +145,8 @@ def context_for_feature_ids(feature_ids)
   end
 
   def reference_view_url(book = chapter.book)
+    raise('Unpersisted Page') if id.nil?
+
```
nathanstitt: same, could use `persisted?`

app/subsystems/content/models/book.rb
Dantemss: Yeah same deal but I thought using the id made sense since it's the id we care about

nathanstitt: yes, only diff would be if the page was deleted, in which case it'd have an id, but `persisted?` would be false.  I don't think we delete pages tho

PR #1947, opened 2019-11-25
Removed the CSV roster import

Dantemss:
No longer in use


Dantemss:
Failure is just the coverage decreasing due to code being deleted. I'm not going to bother...

PR #1952, opened 2019-12-06
Prevent page links from being converted twice

Dantemss:
Should fix the issue with inter-modular links.


Dantemss:
Could be the cause but not 100% sure.

PR #1960, opened 2019-12-20
include html in preview text

nathanstitt:
This way content can include bold/italic tags

### DIFF
--- a/app/subsystems/tasks/models/tasked_reading.rb
+++ b/app/subsystems/tasks/models/tasked_reading.rb
@@ -22,16 +22,15 @@ class Tasks::Models::TaskedReading < IndestructibleRecord
   private
 
   def class_title
-    text = content_dom.xpath("//*[contains(@class, 'os-title')]").first.try(:text).try(:strip)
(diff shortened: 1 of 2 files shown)


app/subsystems/tasks/models/tasked_reading.rb
```diff
@@ -22,16 +22,15 @@ def content_preview
...
   end
 
   def document_title
-    content_dom.xpath("//*[@data-type='document-title']").first.try(:text).try(:strip)
+    content_dom.xpath("//*[@data-type='document-title']").first&.inner_html&.strip
   end
 
   def data_title
-    content_dom.xpath("//*[@data-type='title']").first.try(:text).try(:strip)
+    content_dom.xpath("//*[@data-type='title']").first&.inner_html&.strip
```
reedstrm: I'm going to assume that removing the `try()`s here is handled by how these are actually called, in a fallback sort of context.

nathanstitt: yeah, `&.` is the new `try` for ruby.  It's called the "safe navigation operator"

PR #1962, opened 2020-01-15
params.slice is not a method, use permit and to_h

nathanstitt:
https://sentry.cnx.org/openstax/tutor/issues/47101/


nathanstitt:
I also grepped through `app/` and didn't see any other uses of `params.permit`

PR #1963, opened 2020-01-16
Fix NaN CLUes due to division by 0 when merging

Dantemss:
It was breaking the performance forecast

### DIFF
--- a/app/routines/concerns/clue_merger.rb
+++ b/app/routines/concerns/clue_merger.rb
@@ -15,9 +15,11 @@ module ClueMerger
     end
     weights_by_clues = weights_by_clues.select { |clue, weight| clue[:is_real] } if is_real
     total_weight = weights_by_clues.values.sum
-    most_likely = total_weight == 0 ? 0.5 : weights_by_clues.map do |clue, weight|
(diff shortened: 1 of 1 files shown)


app/routines/concerns/clue_merger.rb
```diff
@@ -14,9 +14,10 @@ def merge_clues(book_containers, clue_by_book_container_uuid)
       weights_by_clues[clue] = weight
     end
     weights_by_clues = weights_by_clues.select { |clue, weight| clue[:is_real] } if is_real
-    most_likely = weights_by_clues.map do |clue, weight|
+    total_weight = weights_by_clues.values.sum
+    most_likely = total_weight == 0 ? 0.5 : weights_by_clues.map do |clue, weight|
       clue[:most_likely] * weight
-    end.sum/weights_by_clues.values.sum
+    end.sum/total_weight
```
nathanstitt: it looks like `total_weight` will be 0 in some cases based on line 18 checking for 0.  In that case won't you still get a divide by 0 error on line 20?

nathanstitt: that's much clearer, thanks!

PR #1965, opened 2020-01-17
display paired count and only remove last LMS pairing

nathanstitt:
This way admins won't need to remove all the LMS pairs, only the last which is what they almost always want

### DIFF
--- a/app/controllers/admin/courses_controller.rb
+++ b/app/controllers/admin/courses_controller.rb
@@ -34,7 +34,7 @@ class Admin::CoursesController < Admin::BaseController
   end
 
   def unpair_lms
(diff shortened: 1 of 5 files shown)


nathanstitt:
The main reason to unlink is because a LMS user takes an existing course's keys and uses it on a new LMS course instead of getting new keys. That happens sometimes when the LMS clones a course.

In that case what we really want to do is remove the latest pairing since the first pairing is linked to the original LMS course as it should be.  

Clearing everything like we were doing is ok since it's fairly easy to re-pair, but it wasn't quite right

reedstrm:
Ah, yes, I remember the case now. So the latest is the broken pairing. Got it.

PR #1970, opened 2020-01-29
Populate practice task placeholder steps in the background

Dantemss:
Research stuff was causing errors in dev, so I added some ::

The real changes are:
- Removing the 30 retries for practice tasks
- skip_unready for practice tasks so their placeholders don't get immediately deleted

Goes with https://github.com/openstax/tutor-js/pull/2890


nathanstitt:
@Dantemss code looks good but you've got a few failures in `./spec/controllers/api/v1/practices_controller_spec`

PR #1971, opened 2020-01-31
Generate stats for new courses/students/instructors

nathanstitt:
Also fixes bug where last week wouldn't be included due to date being less than today vs tomorrow


nathanstitt:
yep, I forgot to add the new counters to the spec.  Thanks!

PR #1974, opened 2020-02-04
Fix practice tasks

Dantemss:
- Don't send practice tasks to Biglearn multiple times
- Ensure practice tasks sent to Biglearn always request 5 exercises

The practice tasks were being sent twice and apparently the second time Tutor would request only 1 exercise (due to the task having only 1 placeholder step).

### DIFF
--- a/app/subsystems/tasks/populate_placeholder_steps.rb
+++ b/app/subsystems/tasks/populate_placeholder_steps.rb
@@ -13,7 +13,6 @@ class Tasks::PopulatePlaceholderSteps
   def exec(task:, force: false, lock_task: true, background: false,
            skip_unready: false, populate_spes: true)
     outputs.task = task
-    outputs.accepted = true
 
     return if already_populated?(task, populate_spes)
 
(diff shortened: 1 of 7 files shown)


app/subsystems/tasks/populate_placeholder_steps.rb
```diff
@@ -30,37 +29,41 @@ def exec(task:, force: false, lock_task: true, background: false,
...
         background: background,
         skip_unready: skip_unready
       )
-      outputs.accepted = false unless accepted
+      spes_populated = task.spes_are_assigned
     end
 
+    return unless pes_populated || spes_populated
+
     # Save pes_are_assigned/spes_are_assigned and step counts
```
nathanstitt: it looks like you're no longer assigning to `outputs.accepted`,  are you sure nothings relying on that being set?

nathanstitt:
I did a search through the code base and don't see any places that `accepted` from `Tasks::PopulatePlaceholderSteps` is being used, will go ahead and merge

PR #1976, opened 2020-02-04
When deleting a teacher, also delete their teacher_students

Dantemss:
For example, teacher01 in prod was removed for a course but still sees the course in their dashboard due to the teacher_student roles, even though they get a 403 error if they click the course.

### DIFF
--- a/db/migrate/20200204192400_delete_teacher_students_for_deleted_teachers.rb
+++ b/db/migrate/20200204192400_delete_teacher_students_for_deleted_teachers.rb
@@ -1,0 +1,14 @@
+class DeleteTeacherStudentsForDeletedTeachers < ActiveRecord::Migration[5.2]
+  def up
+    CourseMembership::Models::Teacher.where.not(deleted_at: nil).preload(
+      role: { profile: { roles: :teacher_student } }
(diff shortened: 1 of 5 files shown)


Dantemss:
Codecov is just mad because I deleted some covered lines in a file that has uncovered lines (so coverage goes down...)

db/migrate/20200204192400_delete_teacher_students_for_deleted_teachers.rb
```diff
@@ -0,0 +1,14 @@
+class DeleteTeacherStudentsForDeletedTeachers < ActiveRecord::Migration[5.2]
+  def up
+    CourseMembership::Models::Teacher.where.not(deleted_at: nil).preload(
+      role: { profile: { roles: :teacher_student } }
+    ).find_each do |teacher|
```
nathanstitt: `CourseMembership::Models::Teacher` uses acts_as_paranoid, so I think you could use the `only_deleted` scope here.

Dantemss: Ah yes

PR #1977, opened 2020-02-11
Fix Preview::AnswerExercise for blank strings

Dantemss:
The free response gets set to a blank string in WorkTask for math-only asnwers and then Preview::AnswerExercise fails because it does not replace blank strings.


Dantemss:
We tested this fix manually in load-ae so I'm going to merge.

PR #1985, opened 2020-02-26
Speed up preview and demo course data creation

Dantemss:
Preview courses should be created 30% faster and Demo::Work should be a lot faster I think.

### DIFF
--- a/app/routines/preview/answer_exercise.rb
+++ b/app/routines/preview/answer_exercise.rb
@@ -1,35 +1,49 @@
-module Preview
-  class AnswerExercise
-    lev_routine
+class Preview::AnswerExercise
+  lev_routine
 
-    uses_routine MarkTaskStepCompleted, as: :mark_completed
+  uses_routine MarkTaskStepCompleted, as: :mark_completed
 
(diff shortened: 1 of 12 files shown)


app/routines/preview/answer_exercise.rb
```diff
@@ -1,35 +1,49 @@
...
+      # puts "tasked:    #{tasked.inspect}"
+      raise "Cannot answer a #{tasked.class.name} - Group: #{task_step.group_type}"
     end
+
+    if is_correct
+      free_response = 'A sentence explaining all the things!' if free_response.blank?
+      answer_id = tasked.correct_answer_id
+    else
+      free_response = 'A sentence explaining all the wrong things...' if free_response.blank?
+      wrong_answer_ids = tasked.answer_ids.reject { |id| id == tasked.correct_answer_id }
```
nathanstitt: really?  We're artificially putting a dummy free response?  This isn't new of course so I guess it's fine

Dantemss: Yeah...

PR #1986, opened 2020-02-26
remove ActiveRecord connect/disconnect in before/after fork

nathanstitt:
For unknown reasons on OSX 10.15.3 this causes segfaults


nathanstitt:
closing since we're moving to puma in https://github.com/openstax/tutor-server/pull/1989

PR #1991, opened 2020-03-04
Implement fetching not-complete practice exercise tasks, rather than …

reedstrm:
…generating new one

Current code triggers generating a new  practice task (set of 5 exercises), then repeatedly fetching until timeout. Attempting to practice the same topic will start all over. This code will instead serve any incomplete practice tasks for that topic (page numbers) that may exist. This would include ones that took some time to generate, or ones that the student did not complete.


app/routines/concerns/find_or_create_practice_task_routine.rb
```diff
@@ -45,23 +48,34 @@ def exec(course:, role:, **args)
...
+      role: @role,
       task_type: @task_type,
-      time_zone: time_zone,
-      title: 'Practice',
-      ecosystem: @ecosystem
-    ).outputs.task
+      page_ids: args[:page_ids]]
+
+    if @task.nil? 
+         # Create the new practice widget task
```
Dantemss: Indented too much

app/routines/concerns/find_or_create_practice_task_routine.rb
```diff
@@ -45,23 +48,34 @@ def exec(course:, role:, **args)
...
-      :build_task,
+    # Find a not completed practice task
+    @task = Tasks::GetPracticeTask[
+      role: @role,
       task_type: @task_type,
-      time_zone: time_zone,
-      title: 'Practice',
-      ecosystem: @ecosystem
-    ).outputs.task
+      page_ids: args[:page_ids]]
```
Dantemss: Maybe it's better to set `@page_ids` in the setup method? It's a bit weird to me to use args directly here.
Also you might want to sort it there.
And we should decide if we want to sort or shuffle the `@pages` too since it seems to me they currently use whatever order postgres returns.

app/subsystems/tasks/get_practice_task.rb
```diff
@@ -3,18 +3,11 @@ class Tasks::GetPracticeTask
...
+  def exec(role:, task_type:, page_ids:)
     outputs[:task] = Tasks::Models::Task
-                       .joins(:taskings)
-                       .where(taskings: { entity_role_id: role.id }, task_type: task_types)
-                       .order(:created_at)
-                       .last
+                     .joins(:taskings)
+                     .where(taskings: { entity_role_id: role.id },
+                            task_type: task_type, page_ids: page_ids)
+                     .where('completed_steps_count < steps_count').first
```
Dantemss: I like the `.order(:created_at).last`... otherwise we'll get a random one?

db/migrate/20200303163234_add_page_ids_to_tasks_tasks.rb
```diff
@@ -0,0 +1,5 @@
+class AddPageIdsToTasksTasks < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tasks_tasks, :page_ids, :integer, array: true
```
Dantemss: I think this column might be useful for other things (for example the Biglearn FakeClient could use it) so I feel tempted to migrate existing tasks to use it too. And maybe set tasks that have no pages to `[]` and make it not-null.

spec/subsystems/tasks/export_performance_report_spec.rb
```diff
@@ -71,7 +71,7 @@
   end
 
   it 'does not blow up if the course name is too long' do
-    @course.update_attribute(:name, 'Tro' + (['lo'] * 150).join)
+    @course.update_attribute(:name, 'Tro' + (['lo'] * 50).join)
```
Dantemss: Interesting... did the really long course name blow up your Postgres? :grin:

app/subsystems/tasks/get_practice_task.rb
reedstrm: Well, the idea is to enforce that there is only one. If we do order, I'd suggest .order(:created_at).first, to get them "caught up" by doing oldest first.

spec/subsystems/tasks/export_performance_report_spec.rb
reedstrm: No it's a problem with the version of encrypted disk software I use for my homedir: the encrypted files are named from the base name w/ a hash appended, then stored in a regular ext3 FS, so you hit the underlying limit on filenames sooner than you'd think. I had to do that to one other test, I think in exercises, too.

app/routines/concerns/find_or_create_practice_task_routine.rb
reedstrm: I agree sort is probably a good idea. - it really should be treated as a set. I'll address

app/routines/concerns/find_or_create_practice_task_routine.rb
reedstrm: will fix

app/routines/concerns/find_or_create_practice_task_routine.rb
reedstrm: in setup, got it. Will do. (now hat I use it in two places) I did check to be sure this would not blow up if page_ids is not in args (like practiceworst)

reedstrm: Also, first pass, I tried to use lev (run) to call this - that seemed to give me some sort of problems, so I reverted to calling it directly. I think the problems may have been elsewhere, so, do you think there's any advantage to using lev to call GetPracticeTask?

reedstrm: Hmm, I tried run()  again, and it failed miserably. I apparently do not understand lev

db/migrate/20200303163234_add_page_ids_to_tasks_tasks.rb
reedstrm: Hmm, not all tasks are linked to pages, are they? Does practice-worst end up picking pages? (i.e. it gets page_ids back from BL?) If so, I need to capture those.) If not, then NULL is correct for "doesn't matter" cases.

Should we list all pages in a Reading here? What about external and event?

reedstrm: (besides, making it not null and setting it to a value akes the migration potentially slow :-) )

spec/support/create_student_history.rb
```diff
@@ -23,17 +23,17 @@ def exec(course:, roles: setup_student_role, book_id: '93e2b09d-261c-4007-a987-0
...
 
       practice_task = create_practice_widget(
-        course: course, role: role, page_ids: ecosystem.chapters[3].pages[2].id
+        course: course, role: role, page_ids: [ecosystem.chapters[3].pages[2].id]
       )
       answer_correctly(practice_task, 5) # 5 out of 5
 
       create_practice_widget(
-        course: course, role: role, page_ids: ecosystem.chapters[3].pages.map(&:id)
+        course: course, role: role, page_ids: [ecosystem.chapters[3].pages.map(&:id)]
```
reedstrm: these are the "page_ids should be a list" fix.

db/migrate/20200303163234_add_page_ids_to_tasks_tasks.rb
Dantemss: Homeworks and readings at least are linked to pages.

I only suggested [] instead of null to keep the type the same so we don't have to check for null in methods that call page_ids.

spec/subsystems/tasks/export_performance_report_spec.rb
Dantemss: Hmmmm it's exactly what the test is testing though... well leave it like this for now and maybe later we can either limit the course name or add a switch to only run this test in CI or just remove it.

spec/support/create_student_history.rb
Dantemss: Personally I'd also put something in setup to wrap it in an array to prevent it from exploding if we do accidentally pass in a non-array.

app/subsystems/tasks/get_practice_task.rb
Dantemss: Oldest first works too. If we do migrate the tasks we'll probably want this.

app/routines/concerns/find_or_create_practice_task_routine.rb
Dantemss: run() returns the same things as call() which is a results object which has an outputs object which has a task object.
[] returns whatever output is defined as express_output so you get the task directly.

Dantemss:
We decided to migrate existing tasks and change the current code to use the new page_ids column for non-practice tasks in a separate PR. Ross is going to add some tests here and then we can merge it.

reedstrm:
@dantemss tests added, should I go ahead and merge?

Dantemss:
I think I'll go ahead and modify the migration before merging so I don't have to make another migration. Then use the page_ids in the Biglearn FakeClient PR.

reedstrm:
With this "reuse pracitce" code in place, the easiest way to test if it is working is to use the side-effect of incomplete practice tasks:

Currently, if a student launches a practice for a given section, a new practice task is created, and the browser is redirected to it., as so:

https://tutor-load-ae.openstax.org/course/12/task/44950

This is task 44950 is course 12, created by clicking on 'Practice 4.2: Newton's First law …"

Work a step, but leave it incomplete. Go back to course dashboard, and practice the same section. Will give the same url, with the same task ID. (current production will give a new task, with potentially different steps)

PR #1992, opened 2020-03-05
Bring Biglearn into Tutor + Make preview courses not cause Biglearn load

Dantemss:
Turning the Biglearn API FakeClient on means you can never go back to the RealClient.

If we do need the option to fallback to the RealClient, we might need to make the FakeClient into a hybrid that sends some methods to the RealClient (so biglearn-api still receives Tutor data). This won't work for preview though because load is caused by sending data to BL, which we want to remove for previews.

I haven't fully reasoned about why the question counts change in the specs, but it's likely due to the use of different pool and due to the FakeClient now prohibiting duplicate questions.


Dantemss:
I looked at the code again and decided it needs to use the ecosystem map to map spaced practice exercises cross-ecosystem, otherwise it won't work properly for spaced practice after an ecosystem update.

app/subsystems/course_profile/build_preview_courses.rb
```diff
@@ -69,9 +68,25 @@ def offering_that_needs_previews(desired_count)
...
+      )
+      .joins(
+        <<~JOIN_SQL
+          CROSS JOIN LATERAL (
+            SELECT "course_content_course_ecosystems"."content_ecosystem_id"
+            FROM "course_content_course_ecosystems"
+            WHERE "course_content_course_ecosystems"."course_profile_course_id" =
+              "course_profile_courses"."id"
+            ORDER BY "course_content_course_ecosystems"."created_at"
+            LIMIT 1
```
Dantemss: We could save this cross join in an "initial_ecosystem_id" column in the courses table to simplify these queries.

PR #1995, opened 2020-03-10
Rename property to "students" and include the student id

nathanstitt:
This way the FE can detect which answer belongs to a student


nathanstitt:
The spec failure is in `./spec/subsystems/tasks/models/task_spec.rb` which is also failing in #1955

Dantemss:
We decided to make a new API for scores instead. Is this PR OBE?

nathanstitt:
up, closing it

PR #1996, opened 2020-03-10
Make sure all task types set and maintain the page_ids array

Dantemss:
Also renamed update_step_counts to update_cached_attributes since it sets the page_ids now
Travis passed, it just didn't like the PR target branch name or something like that.

### DIFF
--- a/db/migrate/20200303163234_add_core_page_ids_to_tasks_tasks.rb
+++ b/db/migrate/20200303163234_add_core_page_ids_to_tasks_tasks.rb
@@ -1,17 +1,18 @@
-class AddPageIdsToTasksTasks < ActiveRecord::Migration[5.2]
+class AddCorePageIdsToTasksTasks < ActiveRecord::Migration[5.2]
   def up
-    add_column :tasks_tasks, :page_ids, :integer, array: true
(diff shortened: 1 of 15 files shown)


db/migrate/20200303163234_add_core_page_ids_to_tasks_tasks.rb
```diff
@@ -0,0 +1,30 @@
...
+          SELECT "unsorted"."content_page_id"
+          FROM (
+            SELECT DISTINCT ON ("tasks_task_steps"."content_page_id")
+              "tasks_task_steps"."content_page_id", "tasks_task_steps"."number"
+            FROM "tasks_task_steps"
+            WHERE "tasks_task_steps"."tasks_task_id" = "tasks_tasks"."id"
+              AND "tasks_task_steps"."is_core" = true
+            ORDER BY "tasks_task_steps"."content_page_id", "tasks_task_steps"."number"
+          ) "unsorted"
+          ORDER BY "unsorted"."number"
```
reedstrm: Ah, yup: the only way to guarantee an array order is to use a select from subselect. Or a plugin-language, of course.

PR #2001, opened 2020-03-13
Fix cached task attributes for practice tasks with real Biglearn

Dantemss:
Practice tasks were not getting their caches set until Biglearn responded.

### DIFF
--- a/app/routines/concerns/find_or_create_practice_task_routine.rb
+++ b/app/routines/concerns/find_or_create_practice_task_routine.rb
@@ -55,17 +55,16 @@ module FindOrCreatePracticeTaskRoutine
         task_type: @task_type,
         time_zone: time_zone,
         title: 'Practice',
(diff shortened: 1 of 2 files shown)


app/routines/concerns/find_or_create_practice_task_routine.rb
```diff
@@ -55,18 +55,17 @@ def exec(course:, role:, **args)
         task_type: @task_type,
         time_zone: time_zone,
         title: 'Practice',
-        ecosystem: @ecosystem,
-        core_page_ids: @page_ids
```
reedstrm: Hmm, what happens to the page IDs? Are they set elsewhere?

Dantemss: The call to update_cached_attributes should set them

PR #2007, opened 2020-03-26
Add student id to grading scores api

nathanstitt:
FE needs this to track which students have been graded


Dantemss:
Need to add role_id to some tests but I have modified those in my PR too so might as well fix them there to avoid conflicts.

PR #2011, opened 2020-03-30
Drop questions

Dantemss:
- Added DroppedQuestion model
- Added DroppedQuestion to TaskPlan's representer
- Made dropped questions affect scores
- Specs

### DIFF
--- a/app/representers/api/v1/task_plan/dropped_question_representer.rb
+++ b/app/representers/api/v1/task_plan/dropped_question_representer.rb
@@ -1,0 +1,16 @@
+class Api::V1::TaskPlan::DroppedQuestionRepresenter < Roar::Decorator
+  include Roar::JSON
+  include Representable::Coercion
+
+  property :question_id,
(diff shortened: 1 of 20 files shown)


app/representers/api/v1/task_plan/scores/question_heading_representer.rb
```diff
@@ -16,5 +16,4 @@ class Api::V1::TaskPlan::Scores::QuestionHeadingRepresenter < Roar::Decorator
            type: Float,
            readable: true,
            writeable: false
-
 end
```
nathanstitt, line 19: I think we'll need to add the question_id and exercise id here so that the FE will know what to send when the question is dropped.  It could probably look at the 1st student by assuming all students are the same but if it was here we'd know for sure.

nathanstitt: did this in #2015

PR #2013, opened 2020-04-02
Added tmp/cache/assets/.keep and updated bootsnap

Dantemss:
Should fix permissions during deployment


reedstrm:
Color me not convinced. Permissions where wrong on log and tmp, not just tmp/cache

PR #2018, opened 2020-04-04
Tweaked preview course probabilities and documented the LQC CLUe calculation

Dantemss:
Clarified all the terms and removed extraneous terms.


nathanstitt:
thanks for documenting this!

PR #2019, opened 2020-04-06
update scores api to return tasking_plans and work with all types

jivey:
- Return tasking_plans and grading_template in the scores API.
- Safely navigate to late_work_penalty, as events, external, and old assignments don't use grading templates

### DIFF
--- a/app/representers/api/v1/task_plan/scores/representer.rb
+++ b/app/representers/api/v1/task_plan/scores/representer.rb
@@ -22,6 +22,17 @@ class Api::V1::TaskPlan::Scores::Representer < ::Roar::Decorator
            readable: true,
            writeable: true
 
+  property :grading_template,
+           extend: Api::V1::GradingTemplateRepresenter,
+           readable: true,
+           writeable: false
+
+  collection :tasking_plans,
+             extend: Api::V1::TaskPlan::TaskingPlanRepresenter,
+             readable: true,
+             writeable: false,
+             getter: ->(*) { tasking_plans }
+
   collection :periods,
              extend: Api::V1::TaskPlan::Scores::PeriodRepresenter,
              readable: true,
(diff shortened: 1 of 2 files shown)


app/representers/api/v1/task_plan/scores/representer.rb
```diff
@@ -22,6 +22,17 @@ class Api::V1::TaskPlan::Scores::Representer < ::Roar::Decorator
...
+           readable: true,
+           writeable: false
+
+  collection :tasking_plans,
+             extend: Api::V1::TaskPlan::TaskingPlanRepresenter,
+             readable: true,
+             writeable: false,
+             getter: ->(*) { tasking_plans }
+
   collection :periods,
```
nathanstitt, line 36: @jivey can you remind me why we need the tasking_plan here again?

app/routines/calculate_task_plan_scores.rb
```diff
@@ -126,7 +126,7 @@ def exec(task_plan:)
         id: period.id,
         name: period.name,
         question_headings: question_headings_array,
-        late_work_fraction_penalty: task_plan.grading_template.late_work_penalty,
+        late_work_fraction_penalty: task_plan.grading_template&.late_work_penalty,
         num_questions_dropped: num_questions_dropped,
```
nathanstitt, line 130: should we use a `||` here so that late_work_fraction_penalty is always set?  Something like:

`late_work_fraction_penalty: (task_plan.grading_template&.late_work_penalty || 0)`

jivey:
Oops didn't see the existing review, you can ignore @Dantemss :D

app/representers/api/v1/task_plan/scores/representer.rb
jivey: Needed displaying dates on the details screen

app/routines/calculate_task_plan_scores.rb
jivey: Yeah, maybe that would be better to keep front end logic from working around null.

PR #2021, opened 2020-04-09
Enable MPQ in APUSH

Dantemss:
Add them to the dynamic exercise pools then attempt to assign the correct number of them...

### DIFF
--- a/app/routines/choose_exercises.rb
+++ b/app/routines/choose_exercises.rb
@@ -1,11 +1,14 @@
 class ChooseExercises
+  # The maximum number of question parts that will be dropped per exercise
+  # This corresponds to how many numbers above the given count we need to search
+  MAX_DROPPED_QUESTION_PARTS = 4
+
   lev_routine express_output: :exercises
 
   def exec(
     exercises:,
     count:,
     already_assigned_exercise_numbers: [],
-    allow_repeats: true,
     randomize_exercises: true,
     randomize_order: true
   )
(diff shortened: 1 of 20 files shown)


app/routines/find_or_create_practice_worst_topics_task.rb
```diff
@@ -19,24 +19,33 @@ def add_task_steps
...
                " - Task Type: #{@task_type} - Ecosystem: #{@ecosystem.title}]"
-    ) if exercises.empty?
+    ) if result[:exercises].empty?
+
+    # It's probably fine to give students slightly bigger practice tasks
+    # if that means they get whole MPQ questions
+    remaining = FindOrCreatePracticeTaskRoutine::NUM_EXERCISES
+    exercises = result[:exercises].select do |exercise|
+      (remaining > 0).tap { remaining -= exercise.number_of_questions }
+    end
```
nathanstitt, line 35: Hm, I've never seen `boolean.tap` called before? Pretty sure it'lll only be called once, since that's what `true.tap` does in `irb`.   Is it's use here doing anything?

db/migrate/20200410002543_add_number_of_questions_to_content_exercises.rb
```diff
@@ -0,0 +1,15 @@
...
+      dir.up do
+        Content::Models::Exercise.update_all(
+          '"number_of_questions" = JSONB_ARRAY_LENGTH("question_answer_ids")'
+        )
+      end
+    end
+
+    change_column_null :content_exercises, :number_of_questions, false
+  end
+end
```
nathanstitt, line 15: Will there be any exercises without question_answer_ids where this might return null?  Do we need to add a COALESCE to it?

app/routines/choose_exercises.rb
```diff
@@ -14,20 +17,130 @@ def exec(
...
       already_assigned_exercise_numbers_set.include?(ex.number)
     end
 
-    new_exercises_count = [new_exercises.size, count].min
-    repeated_exercises_count = allow_repeats ? \
-                                 [repeated_exercises.size, count - new_exercises_count].min : 0
+    outputs.exercises = choose_exercises(new_exercises, exercises, count, randomize_exercises)
+
+    outputs.exercises = outputs.exercises.shuffle if randomize_order
+  end
```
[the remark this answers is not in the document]
Dantemss: One of them shuffles the exercises that are picked, the other shuffles only the output.

app/routines/find_or_create_practice_worst_topics_task.rb
Dantemss: I want to select based on the boolean but also update remaining after that, which is what tap is doing.

db/migrate/20200410002543_add_number_of_questions_to_content_exercises.rb
Dantemss: The background migration to remove duplicate content sets question_answer_ids to NOT NULL at the end so hopefully not.

PR #2023, opened 2020-04-13
update to include openstax_accounts 9.1.0

nathanstitt:
This was missed in #2022 😭

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -401,7 +401,7 @@ GEM
     omniauth-oauth2 (1.3.1)
       oauth2 (~> 1.0)
       omniauth (~> 1.2)
-    openstax_accounts (9.0.2)
+    openstax_accounts (9.1.0)
       action_interceptor
       keyword_search
       lev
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -411,7 +411,7 @@ GEM
       openstax_api
       openstax_utilities
       pg
-      rails (< 6.0)
```
reedstrm, line 414: I don't remember the exact reason we couldn't migrate to 6, but I assume (given it's a major number) that it's a non-trivial migration - is this safe to remove this limit?

nathanstitt: that's just saying what the openstax_accounts gem works with.  I'd changed that in https://github.com/openstax/accounts-rails/pull/86 because `< 6.0` would take almost an hour to bundle due to it testing every single rails version ever.

reedstrm: Ah, gotcha.

reedstrm: Yeah, I missed the indentation.

PR #2027, opened 2020-04-15
Grader API

Dantemss:
- Grading model, migration and validations
- Grading representers
- Grading API
- Specs


app/access_policies/tasked_access_policy.rb
```diff
@@ -5,14 +5,19 @@ def self.action_allowed?(action, requestor, tasked)
...
     when :update
       DoesTaskingExist[task_component: tasked, user: requestor] &&
         !task.withdrawn? &&
         !task.past_close? &&
         (task.past_open? || task.roles.first.teacher_student?)
+    when :grade
+      return false if !task.past_due?
+      period = task.taskings.first&.period
+      return false if period.nil?
+      UserIsCourseTeacher[user: requestor, course: period.course]
```
nathanstitt, line 20: Why would the period be nil?  I thought that should always be set

app/representers/api/v1/tasks/tasked_exercise_grading_representer.rb
```diff
@@ -0,0 +1,32 @@
+module Api::V1::Tasks
+  class TaskedExerciseGradingRepresenter < Api::V1::Tasks::TaskedExerciseRepresenter
+    property :answer_id,
+             inherit: true,
+             writeable: false
+
```
nathanstitt, line 6: I think we'll need the exercise and question ids here as well so the FE can easily match them to the student's score

app/routines/calculate_task_plan_scores.rb
```diff
@@ -105,19 +125,24 @@ def exec(task_plan:)
...
         }
       end.compact.sort_by { |student| [ student[:last_name], student[:first_name] ] }
 
       {
-        id: period.id,
-        name: period.name,
+        id: tasking_plan.id,
+        period_id: tasking_plan.target_id,
+        period_name: tasking_plan.target.name,
         question_headings: question_headings_array,
```
nathanstitt, line 138: Above there's some logic for tasking_plans that target courses, instead of per-period ones.  The FE doesn't currently make course wide plans and it would be a pain to support them here. 

Maybe it's time to remove the ability to create task_plans that target a course and instead require one plan per period?

app/access_policies/tasked_access_policy.rb
Dantemss: It should, but this is just a safety check in case the tasking got deleted somehow...

app/routines/calculate_task_plan_scores.rb
Dantemss: Yeah I could remove the polymorphism from target and rename it to course_membership_period_id.

app/representers/api/v1/tasks/tasked_exercise_grading_representer.rb
Dantemss: Ah how about I add the task_step_id to the StudentQuestionRepresenter and you match that? You'll need that id to post to this route.

app/routines/calculate_task_plan_scores.rb
Dantemss: Actually I just found out that even though they are never saved, the BE uses tasking_plans that point to individual roles internally in IndividualizeTaskingPlans. So maybe let's do this one later.

PR #2032, opened 2020-04-23
default lms enabling allowed to true

nathanstitt:
This will allow teahers to "self service" lms integration instead of
forcing them to contact us


nathanstitt:
yes for pre-existing courses, but new ones will use the default, which is what we want

PR #2033, opened 2020-04-23
remove mention of $10 cost

nathanstitt:
Tutor is going to be free for the majority of courses next semester

image


nathanstitt:
@Dantemss  I think we want to leave them as costing by default and only KIP will be free once your pr is released.    I think the intent is to remove mention of cost since most courses will be freee

PR #2036, opened 2020-04-27
Block cloned assignment creation unless the content is properly updated

Dantemss:
Also adds old content to existing cloned courses to fix existing issues

Need to test this with the FE to see if cloned_from_id is being set properly for assignments.

### DIFF
--- a/app/subsystems/tasks/models/task_plan.rb
+++ b/app/subsystems/tasks/models/task_plan.rb
@@ -26,7 +26,7 @@ class Tasks::Models::TaskPlan < ApplicationRecord
 
   json_serialize :settings, Hash
 
-  before_validation :trim_text, :set_ecosystem
+  before_validation :trim_text, :set_and_return_ecosystem
 
   validates :title, presence: true
   validates :type, presence: true
@@ -76,7 +76,7 @@ class Tasks::Models::TaskPlan < ApplicationRecord
     Jobba.find(publish_job_uuid) if publish_job_uuid.present?
   end
 
-  def set_ecosystem
+  def set_and_return_ecosystem
     self.ecosystem ||= cloned_from&.ecosystem ||
                        get_ecosystems_from_settings&.first ||
                        owner.try(:ecosystems)&.first
--- a/app/subsystems/tasks/models/task_plan.rb
+++ b/app/subsystems/tasks/models/task_plan.rb
@@ -1,7 +1,6 @@
 require 'json-schema'
 
 class Tasks::Models::TaskPlan < ApplicationRecord
-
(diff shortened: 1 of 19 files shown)


app/subsystems/tasks/models/task_plan.rb
```diff
@@ -73,6 +76,12 @@ def publish_job
     Jobba.find(publish_job_uuid) if publish_job_uuid.present?
   end
 
+  def set_ecosystem
+    self.ecosystem ||= cloned_from&.ecosystem ||
+                       get_ecosystems_from_settings&.first ||
+                       owner.try(:ecosystems)&.first
+  end
+
```
nathanstitt, line 84: it seems a little dangerous to rely on `set_ecosystem` to return the ecosystem.  I can see someone changing this  in a way that causes it to no longer return the ecosystem which would break the `update_task_plan_ecosystem` routine above.  Maybe change this to be `set_and_return_ecosystem`, or just have the routine call set and then get the ecosystem?

db/migrate/20200427121628_fix_cloned_courses.rb
```diff
@@ -0,0 +1,20 @@
...
+        content_ecosystem_id: valid_ecosystem_ids
+      ).distinct.pluck(:content_ecosystem_id)
+
+      missing_ecosystem_ids.each do |ecosystem_id|
+        CourseContent::Models::CourseEcosystem.create!(
+          course: course, content_ecosystem_id: ecosystem_id
+        )
+      end
+    end
+  end
```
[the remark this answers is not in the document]
Dantemss: It should actually be pretty fast, there are only 6,000 or so total courses and about 50 broken ones.

app/subsystems/tasks/models/task_plan.rb
Dantemss: OK I'll rename it

app/subsystems/tasks/models/task_plan.rb
```diff
@@ -112,7 +115,16 @@ def valid_settings
...
 
-  def same_ecosystem
+  def valid_ecosystem
+    valid_ecosystems = owner.try(:ecosystems)
+    # Remove nil check if we remove polymorphism.
+    return if valid_ecosystems.nil? || valid_ecosystems.include?(ecosystem)
+
+    errors.add(:ecosystem, 'is not valid for this course')
+    throw :abort
+  end
```
nathanstitt, line 125: something I've been meaning to ask, is the `throw :abort` really needed?  Seems like it'd be ok to continue running the validations so they're all returned?

PR #2038, opened 2020-04-28
CLUe performance

Dantemss:
- Added StudentBookPart and PeriodBookPart models
- Cached student and period CLUes in StudentBookPart and PeriodBookPart when an immediate feedback question is answered or feedback becomes available
- Read CLUes from StudentBookPart and PeriodBookPart in the FakeClient
- Recorded whether the book_part_uuid belongs to a page or chapter and use only page CLUes in practice worst areas to avoid duplicate exercises.
- Specs


Dantemss:
These new caches should have all the fields we need in the student/teacher guides (when combined with the book structure from the ecosystem itself). So once we get rid of the trouble flags we can remove the old TaskCache and PeriodCache tables.

db/background_migrate/20200501125646_initialize_clue_caches.rb
```diff
@@ -0,0 +1,49 @@
...
+            period: role.student.period, book_part_uuid: book_part_uuid, is_page: false
+          ) if role.student?
+        end
+      end
+    end
+  end
+
+  def down
+  end
+end
```
nathanstitt, line 49: this seems like it'll take a long time to compute?

db/migrate/20200428154625_add_parent_book_part_uuid_to_content_pages.rb
```diff
@@ -0,0 +1,33 @@
...
+  def change
+    add_column :content_pages, :parent_book_part_uuid, :uuid
+
+    reversible do |dir|
+      dir.up do
+        Content::Models::Book.find_each do |book|
+          update_tree_and_pages! book.tree, book.uuid
+
+          book.save!
+        end
```
nathanstitt, line 11: what changes on the book that causes it to need saved?

lib/tutor/subsystems/association_extensions.rb
```diff
@@ -31,20 +31,24 @@ def set_subsystem_options(association_name, scope, options, is_belongs_to=false)
...
+        return if ['none', 'ignore'].include?(subsystem_name)
 
         # Find the string before the first ::, thats the subsystem module
-        my_subsystem_name = self.name.to_s[/(\w+?)::/,1].underscore
+        my_subsystem_name = self.name.to_s[/(\w+?)::/, 1]&.underscore
 
-        # Only extend subsystem's that are listed as valid.
-        return unless Tutor::SubSystems.valid_name?(my_subsystem_name)
+        # Only extend subsystems that are listed as valid.
+        return unless my_subsystem_name.blank? || Tutor::SubSystems.valid_name?(my_subsystem_name)
```
nathanstitt, line 41: this seems safe but I'm not sure why the additional `my_subsystem_name.blank?` check is needed on line 40

app/routines/mark_task_step_completed.rb
```diff
@@ -26,9 +26,37 @@ def exec(task_step:, completed_at: Time.current, lock_task: true, save: true)
...
 
+    queue = task.is_preview ? 'preview' : 'dashboard'
+    role_run_at = task.feedback_available? ? completed_at :
+                                             [ task.feedback_at, completed_at ].compact.max
+
+    page = Content::Models::Page
+      .select(:uuid, :parent_book_part_uuid)
+      .find_by(id: task_step.content_page_id)
+
+    unless page.nil?
```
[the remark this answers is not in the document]
Dantemss: Exists triggers another query though...

db/background_migrate/20200501125646_initialize_clue_caches.rb
Dantemss: Well it is a background migration... and we can leave the CLUes in the old mode until it finishes

db/migrate/20200428154625_add_parent_book_part_uuid_to_content_pages.rb
Dantemss: The units and chapters that were missing uuids had those added.

lib/tutor/subsystems/association_extensions.rb
Dantemss: I think you might be right, since the other method would probably return false.

Dantemss: Ah no now I remember... I wanted to be able to use the `subsystem: :whatever` key to refer to subsystem stuff from stuff that was outside of subsystems (I ended up making a cache subsystem in the end though).

PR #2040, opened 2020-04-29
Support free-response only exercises

nathanstitt:


app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -125,19 +128,16 @@ def content_preview
     content_preview_from_json || "Exercise step ##{id}"
   end
 
+  def has_answer_missing?
+    return answer_id.blank? && answer_ids.any?
```
Dantemss, line 132: Hmm remember that [nil].any? is true but hopefully we don't have that...

nathanstitt: yeah, the fact that `[nil].any?` returns false is weird, but I think that's actually what we'd want here.

PR #2045, opened 2020-05-06
Converted controller specs to request specs

Dantemss:
So they test the routes and views too.

Also removed some CC tests.


app/views/layouts/_authentication.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if current_user.is_anonymous? %>
-  <%= link_to 'Sign in!', openstax_accounts.login_path %>
+  <%= link_to 'Sign in!', openstax_accounts_login_path %>
```
Dantemss, line 2: So this is something weird I found while testing... We have this method in accounts-rails that sends them to this path AND adds some login params, but it was not being used here.

https://github.com/openstax/accounts-rails/blob/378350df64546947d0d74b9ad933054095bcb1f7/lib/openstax/accounts/action_controller/base.rb#L47

Maybe we need to discuss this.

Dantemss: In Tutor it's used by these methods: https://github.com/openstax/tutor-server/blob/d1bc85f1f56fd60eb647ac7b970ccad657b1f19f/app/controllers/application_controller.rb#L19-L32

Dantemss: Turns out this authentication layout partial is only used in the API docs page. I think it doesn't matter too much what we do with it.

PR #2050, opened 2020-05-12
fix namespace of tasks

jivey:
Found another one here, probably a side effect of load order from the Puma switch.


Dantemss:
It doesn't like the fact that the Tasks subsystem and some other Tasks module inside something else both exist, but it mainly affect dev mode... In prod etc eager loading of the app prevents all this.

PR #2052, opened 2020-05-14
Fixed FakeClues (preview courses)

Dantemss:
Should prevent 500 error in preview courses when loading the guide.


Dantemss:
This is no longer needed because this issue is also fixed by Glicko.

PR #2063, opened 2020-05-26
send grading templates to FE along with deleted_at

nathanstitt:
This way it can prevent dupe names

### DIFF
--- a/app/subsystems/tasks/models/tasking_plan.rb
+++ b/app/subsystems/tasks/models/tasking_plan.rb
@@ -40,7 +40,7 @@ class Tasks::Models::TaskingPlan < ApplicationRecord
   end
 
   def due_at_on_or_after_opens_at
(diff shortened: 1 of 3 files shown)


app/subsystems/tasks/models/tasking_plan.rb
```diff
@@ -40,7 +40,7 @@ def due_at_in_the_future
   end
 
   def due_at_on_or_after_opens_at
-    return if due_at.nil? || opens_at.nil? || due_at > opens_at
+    return if due_at.nil? || opens_at.nil? || due_at >= opens_at
```
Dantemss, line 43: I have to point out that `==` means the student has 0 time to do the assignment :)

nathanstitt: yeah but that's what they've been testing and getting errors with so I guess if we really want to allow teachers to do it we should 😁

PR #2075, opened 2020-05-30
Added course setting to track past-due unattempted ungraded WRQ

Dantemss:
Whether or not past-due unattempted ungraded WRQ should be zero.

### DIFF
--- a/app/representers/api/v1/course_representer.rb
+++ b/app/representers/api/v1/course_representer.rb
@@ -198,6 +198,15 @@ module Api::V1
                              "an LMS; can be `nil` which indicates no choice yet"
              }
 
+    property :past_due_unattempted_ungraded_wrq_are_zero,
+             writeable: true,
+             readable: true,
+             schema_info: {
+               required: true,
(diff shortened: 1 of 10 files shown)


app/representers/api/v1/course_representer.rb
```diff
@@ -198,6 +198,15 @@ class CourseRepresenter < Roar::Decorator
...
              }
 
+    property :past_due_unattempted_ungraded_wrq_are_zero,
+             writeable: true,
+             readable: true,
+             schema_info: {
+               required: true,
+               type: 'boolean',
+               description: 'True iff past-due unattempted ungraded WRQ automatically receive zero'
+             }
```
nathanstitt, line 208: typo, one two many `f` in iff 😁

Dantemss: If and only if :)

Dantemss: For the longest time I used to think the abbreviation was iif too...

PR #2076, opened 2020-05-30
Include score in dashboard api

nathanstitt:
Also think I found bug where score would be nil if all steps were incorrect instead of 0.0

### DIFF
--- a/app/representers/api/v1/courses/dashboard_representer.rb
+++ b/app/representers/api/v1/courses/dashboard_representer.rb
@@ -129,6 +129,11 @@ module Api::V1::Courses
                readable: true,
                writeable: false
 
+      property :score,
+               readable: true,
+               type: Float,
+               writeable: false
+
       property :provisional_score?,
                as: :is_provisional_score,
                readable: true,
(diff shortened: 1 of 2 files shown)


app/subsystems/tasks/models/task.rb
```diff
@@ -262,7 +262,7 @@ def score_without_lateness(current_time: Time.current)
       end.sum
     end
 
-    points_without_lateness/total unless total == 0.0
+    total.zero? ? 0.0 : points_without_lateness/total
```
Dantemss, line 265: If the total is zero the percentage really is undefined, not 0.
This would apply to tasks that don't have exercises.

Dantemss: Oh I see what the problem is... if the task is not yet due we dynamically adjust the total so if you haven't worked anything the total is zero.

Well we should consult with Carly and Devika what to display here.

Dantemss: From Carly's answer it seems we want to keep the nil here and display --- for that.

nathanstitt: But what about a student who's worked the assignment but got everything wrong? Ih that case total will be zero "for real"

PR #2082, opened 2020-06-02
Load extensions dynamically from the task_plan

Dantemss:
And try to preload them in places that use multiple tasks.

I think one day we should delete taskings and put the role ids in the tasks but maybe not today.


reedstrm:
You missed an opportunity: "… but it is not this day!"

PR #2084, opened 2020-06-02
New late work penalty

Dantemss:
- Apply late work penalty based on how late each step is
- Specs


jivey:
Need to merge FE at the same time: https://github.com/openstax/tutor-js/pull/3049

PR #2093, opened 2020-06-05
Added new task_plan and tasking_plan fields for the grade/regrade button

Dantemss:
Needs FE PR to use these counts instead of the scores API

To be used to display the different states of the Grade/regrade button.

if task_plan.wrq_count == 0, then hide button, else display button (this is the only one that looks at the task_plan instead of the tasking_plan)
if tasking_plan.due_at > now, then button should be disabled with some tooltip about being unable to grade before the due date
if tasking_plan.gradable_step_count == 0, then button should be disabled with some tooltip about no work being done
if tasking_plan.ungraded_step_count == 0, then button should be the Re-grade button
otherwise, display the ungraded_step_count on the orange Grade button.


Dantemss:
I'm going to re-work this so we have completed_wrq_count on both tasking_plans and task_plans instead of num_wrqs on the task_plan

app/subsystems/tasks/models/task_plan.rb
```diff
@@ -4,7 +4,12 @@ class Tasks::Models::TaskPlan < ApplicationRecord
...
 
   UPDATEABLE_ATTRIBUTES_AFTER_OPEN = [
-    'title', 'description', 'last_published_at', 'tasks_grading_template_id', 'ungraded_step_count'
+    'title',
+    'description',
+    'last_published_at',
+    'tasks_grading_template_id',
+    'completed_wrq_step_count',
+    'ungraded_wrq_step_count'
   ]
```
nathanstitt, line 13: FYI, they just had us hide al editing capability after the assignment is open.  Not sure if we need to restrict it on the BE tho

db/migrate/20200605151734_add_wrq_count_completed_wrq_count_and_ungraded_wrq_count.rb
```diff
@@ -0,0 +1,30 @@
...
+    add_column :tasks_tasks, :completed_wrq_step_count, :integer, default: 0, null: false
+    rename_column :tasks_tasks, :ungraded_step_count, :ungraded_wrq_step_count
+
+    add_column :tasks_tasking_plans, :completed_wrq_step_count, :integer, default: 0, null: false
+    add_column :tasks_tasking_plans, :ungraded_wrq_step_count, :integer, default: 0, null: false
+
+    add_column :tasks_task_plans, :wrq_count, :integer, default: 0, null: false
+    add_column :tasks_task_plans, :completed_wrq_step_count, :integer, default: 0, null: false
+    rename_column :tasks_task_plans, :ungraded_step_count, :ungraded_wrq_step_count
+
```
nathanstitt, line 12: how likely do you think it is that we have other types that are gradable?  I'm wondering if we should call these fields `wrk_` ?  Maybe `gradable_count` might be better?  That'd also go along with the `ungraded_step_count`

reedstrm: Yeah not a fan of wrq in this context, either. I thought it was a cute contraction of "work" at first, rather than "written response question". Which is ambiguous, since we do written responses with the multiple choice as well. Some flavor of "gradable"  seems reasonable to me as well.

app/subsystems/tasks/models/task_plan.rb
Dantemss: I think I'll leave those because they are OK to update from the BE's point of view.

db/migrate/20200605151734_add_wrq_count_completed_wrq_count_and_ungraded_wrq_count.rb
Dantemss: Going to leave ungraded_step_count alone and rename the new field to gradable_step_count.

camliu89:
@Dantemss @nathanstitt Are these new wrm counts per period? We want to change the state of the grade/regrade button when instructor selects a different period.

PR #2095, opened 2020-06-08
Fix exercises that require context from collated sections

Dantemss:
2 changes:

- Exercises with context now look at pages that come after the current one if a tag with the given id does not exist in the current page (because of collation).
- Exercises with context now use the content pre-reading-processing-instructions (so they still work if the reading processing instructions exclude the content containing the chosen tag).

### DIFF
--- a/app/subsystems/content/routines/transform_and_cache_page_content.rb
+++ b/app/subsystems/content/routines/transform_and_cache_page_content.rb
@@ -23,12 +23,6 @@ class Content::Routines::TransformAndCachePageContent
     # Get all page uuids and cnx_ids given
     pages_by_uuid = pages.index_by(&:uuid)
 
-    # Get all exercises that require context
-    context_exercises = Content::Models::Exercise.requires_context.where(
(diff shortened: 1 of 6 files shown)


Dantemss:
The build passed, Travis just failed to report it.

app/subsystems/content/routines/transform_and_cache_page_content.rb
```diff
@@ -78,27 +72,44 @@ def exec(book:, pages: nil, save: true)
...
 
+    lookahead_exercises.each do |context_exercise|
+      Rails.logger.warn do
+        "Exercise #{context_exercise.uid} requires context but its feature ID(s) [ #{
+          context_exercise.feature_ids.join(', ')
+        } ] could not be found on its page or any subsequent pages"
+      end
+    end
+
     outputs.pages = pages
```
nathanstitt, line 113: yeah.. while I guess this is what we have to do for this feature, it sure seems like it's going to really slow down imports.  I guess that's fine tho sine we do them so rarely.

PR #2097, opened 2020-06-09
don't allow any fields to be updated if template has task_plans

jivey:
As requested 1048, don't allow any fields to be changed if there are task_plans.

### DIFF
--- a/app/subsystems/tasks/models/grading_template.rb
+++ b/app/subsystems/tasks/models/grading_template.rb
@@ -123,7 +123,7 @@ class Tasks::Models::GradingTemplate < ApplicationRecord
     return unless has_open_task_plans?
 
     errors.add :base,
-               'cannot be changed because this template is assigned to one or more task_plans'
+               'cannot be changed because this template is assigned to one or more open task_plans'
     throw :abort
   end
 
--- a/app/subsystems/tasks/models/grading_template.rb
+++ b/app/subsystems/tasks/models/grading_template.rb
@@ -34,7 +34,7 @@ class Tasks::Models::GradingTemplate < ApplicationRecord
 
   validate :weights_add_up, :default_times_have_good_values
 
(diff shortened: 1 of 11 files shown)


Dantemss:
Hmm this is the kind of change I'd rather leave in the FE since it's not really necessary in the BE, but I'd not be opposed to merge if you added a comment saying that blocking all updates is not really 100% necessary, just our policy on grading template updates (so we remember in 6 months when they change their minds :).

nathanstitt:
Yeah seems to me that this is a low stakes validation that could be handled in the FE.  I see the grading_templates api has a `has_task_plans` property so we should be able to just disable editing when that's true.

Dantemss:
@nathanstitt we might have to modify it or add a new one that only counts published and open task plans to meet the requirement

jivey:
I like the idea of enforcing this on the backend because it's supposed to check if the assignment is open + it affects students.

nathanstitt:
yeah, that's true. I'm fine with it

app/subsystems/tasks/models/grading_template.rb
```diff
@@ -76,6 +76,10 @@ def has_task_plans?
     !task_plans.reject(&:withdrawn?).empty?
   end
 
+  def has_open_task_plans?
+    task_plans.reject(&:withdrawn?).any?(&:out_to_students?)
+  end
```
jivey, line 81: @Dantemss Does this cover the cases we want?

jivey:
(as an aside, it would also be good to check it on the FE and show an error message when clicking Edit instead of opening the form, cause the experience with this approach isn't ideal.)

app/subsystems/tasks/models/grading_template.rb
Dantemss: I believe this is correct.

PR #2103, opened 2020-06-12
Added (student-visible) points to the task representer and dashboard representer

Dantemss:
So FE can read them and display


Dantemss:
Travis is green but didn't report the status (again).

PR #2106, opened 2020-06-15
Added tasked_exercise fields to update assignment score table while working

Dantemss:
Added published_points_without_lateness, published_late_work_point_penalty to the tasked exercise representation

Also limit late work penalty to 100%.

### DIFF
--- a/app/subsystems/tasks/models/tasked_exercise.rb
+++ b/app/subsystems/tasks/models/tasked_exercise.rb
@@ -228,7 +228,9 @@ class Tasks::Models::TaskedExercise < IndestructibleRecord
     when 'immediately'
       task.late_work_penalty_per_period
     when 'daily'
-      ((task_step.last_completed_at - due_at)/1.day).ceil * task.late_work_penalty_per_period
+      [
+        ((task_step.last_completed_at - due_at)/1.day).ceil * task.late_work_penalty_per_period, 1.0
+      ].min
     when 'not_accepted'
       1.0
     else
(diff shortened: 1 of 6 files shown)


nathanstitt:
@Dantemss the spec for the controller grading is failing in a way that makes me think something might be incorrect about the PR.  the `published_points` and `published_points_without_lateness` values appear incorrect to me.

The failing spec is: `rspec ./spec/controllers/api/v1/task_steps_controller_spec.rb:329`

Dantemss:
The value looked weird in the spec because the task was late... I changed it to check the published_points_without_lateness instead.

app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -172,13 +172,13 @@ def grader_points
     task_step.task.course.past_due_unattempted_ungraded_wrq_are_zero ? 0.0 : gp
   end
 
-   # This is really published_grader_points only
-  def published_points
+  # TODO: Rename published_points column to published_grader_points
+  def published_grader_points
     case task_step.task.manual_grading_feedback_on
```
nathanstitt, line 177: you can go ahead and do this if you want.  `grader_points` is easy to search for but the other `points` was too common to find easily

Dantemss: OK I'll do it

Dantemss: We decided not to do this now, since it won't affect the API.

PR #2115, opened 2020-06-19
Added total_fraction to tasking_plans in the teacher scores API

Dantemss:
The total here will differ from the gradebook if before due date/not published but should match if after due and published...


Dantemss:
https://travis-ci.org/github/openstax/tutor-server/builds/700167062

PR #2116, opened 2020-06-19
move setting to background job

reedstrm:


db/background_migrate/20200617142820_cache_points_for_existing_tasks.rb
```diff
@@ -60,9 +62,11 @@ def up
     change_column_null :tasks_tasks, :published_points_after_due, false
     change_column_null :tasks_tasks, :is_provisional_score_before_due, false
     change_column_null :tasks_tasks, :is_provisional_score_after_due, false
+    change_column_null :tasks_tasks, :gradable_steps_count, false
```
Dantemss, line 65: gradable_step_count (and below too)

app/subsystems/tasks/models/task.rb
```diff
@@ -672,4 +672,8 @@ def closes_at_on_or_after_due_at
     errors.add(:closes_at, 'must be on or after due_at')
     throw :abort
   end
+
+  def gradable_step_count
+      super.nil? ? 0 : super
+  end
```
Dantemss, line 678: This method needs to be public (move it up in the file).

db/background_migrate/20200617142820_cache_points_for_existing_tasks.rb
reedstrm: Gaaaa!

reedstrm: at least I was consistent (in that file)

app/subsystems/tasks/models/task.rb
reedstrm: Ah found it "protected"

PR #2121, opened 2020-06-22
Fixed interactions between MCQ and publishing

Dantemss:
Also require that the step be completed to give full credit if the question was dropped with that option, as specified in the behavior doc.

### DIFF
--- a/app/subsystems/tasks/models/tasked_exercise.rb
+++ b/app/subsystems/tasks/models/tasked_exercise.rb
@@ -181,7 +181,7 @@ class Tasks::Models::TaskedExercise < IndestructibleRecord
   end
 
   def points_without_lateness
-    return available_points if full_credit?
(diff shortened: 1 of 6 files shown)


app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -181,7 +181,7 @@ def published_grader_points
   end
 
   def points_without_lateness
-    return available_points if full_credit?
+    return available_points if full_credit? && task_step.completed?
```
nathanstitt, line 185: this surprised me.  Why would the task_step not be completed ?

Dantemss: full_credit? just refers to dropping questions and giving full credit
But they want the credit to only be given if the student actually did the work.

PR #2123, opened 2020-06-23
Fixed 500 error in the student guide

Dantemss:
The loop was accidentally overwriting the task_steps variable (only if update_cached_attributes was true) with the task_steps for the last task.

We probably just didn't have a code path before with multiple tasks and update_cached_attributes true.

### DIFF
--- a/app/subsystems/tasks/update_task_caches.rb
+++ b/app/subsystems/tasks/update_task_caches.rb
@@ -65,8 +65,7 @@ class Tasks::UpdateTaskCaches
     # Update step counts for each task
     if update_cached_attributes
       tasks = tasks.map do |task|
-        task_steps = task_steps_by_task_id.fetch(task.id, [])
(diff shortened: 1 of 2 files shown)


app/subsystems/tasks/update_task_caches.rb
```diff
@@ -65,8 +65,7 @@ def exec(task_ids:, update_cached_attributes: false, background: true, queue: 'd
     # Update step counts for each task
     if update_cached_attributes
       tasks = tasks.map do |task|
-        task_steps = task_steps_by_task_id.fetch(task.id, [])
-        task.update_cached_attributes steps: task_steps
+        task.update_cached_attributes steps: task_steps_by_task_id.fetch(task.id, [])
       end
```
nathanstitt, line 69: hm yeah, this is where GitHub's review fails us by not showing quite enough context

PR #2124, opened 2020-06-24
add school_location to user representation

nathanstitt:


Dantemss:
My opinion is that the real fix is to block preview/new course for foreign school in the FE so we don't have to request the offerings.
And fix the ones incorrectly marked in SF.

nathanstitt:
Code from this was put into https://github.com/openstax/tutor-server/pull/2124

PR #2127, opened 2020-06-25
add drop question info in student task step

camliu89:
Adds a drop_question object in the student task step if the question was dropped.

### DIFF
--- a/app/representers/api/v1/tasks/tasked_exercise_representer.rb
+++ b/app/representers/api/v1/tasks/tasked_exercise_representer.rb
@@ -175,8 +175,8 @@ module Api::V1::Tasks
              writeable: false,
              if: FEEDBACK_AVAILABLE
 
-    property :dropped,
-            type: Object,
+    property :dropped_method,
+            type: String,
             readable: true,
             writeable: false
   end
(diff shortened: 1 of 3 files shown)


nathanstitt:
I think he needs the question that was dropped so that it can be identified.  I"m a little afraid this might trigger a N+1 query though.

@camliu89 if you do need to know what question was dropped you could change the task representer to include a `dropped_question_ids` array that would include the ids of each dropped question.  The FE could then look at that to detect which ones' were dropped.

camliu89:
If it is dropped, I need to know if the drop_method is `zeroed` or `full_credit`, should I just return a field called `dropped` as a string with either `zeroed` or `full_credit`? If it is not dropped, then it won't return anything. Does that work?

Dantemss:
Works for me... I think there won't be an N+1 because the way the dropped question object is being loaded right now it'll load the entire array (then if it's called again the array is already loaded) so it should be fine.

Dantemss:
OK now I would say name it `drop_method` and make the type String

camliu89:
oh yeah, forgot to change that. Thanks @Dantemss

PR #2128, opened 2020-07-07
More aggressively cache due_at/closes_at and use the cached fields

Dantemss:
10-20% performance improvement


nathanstitt:
specs did pass, travis didn't report status

PR #2129, opened 2020-07-07
Test using GH actions for CI

nathanstitt:


Dantemss:
Sweet. Documentation format is very verbose to me but I don't mind trying it out.
But it's weird that the Demo::All test fails. Seems like it's not loading the demo config properly...

Dantemss:
Codecov is still not working from GH actions but everything else works.

PR #2133, opened 2020-07-08
[Wait for release after old scores are frozen] remove unused fields from performance api

nathanstitt:
specs need cleanup still


Dantemss:
I am actually surprised is_extended and is_late are not used... don't we need these for some icons?
Oh is_late was never added to the representer huh?

Dantemss:
Ah ok... the late work column makes is_late unnecessary... but the extension icon should still be there...

nathanstitt:
is_extended can be determined by if the due_at field differs from the heading, and is_late wasn't ever used.  we may need to add that at some point but figured we can do that if needed

Dantemss:
See comment on FE-side PR about freezing old scores first.

PR #2136, opened 2020-07-10
Tell the FE whether or not a preview course should be reused

Dantemss:
We shouldn't reuse a preview course that has been updated if we can help it.

https://github.com/openstax/tutor-js/pull/3165


app/subsystems/course_profile/models/course.rb
```diff
@@ -88,6 +88,21 @@ def ecosystem
...
+    course_ecosystems.first&.content_ecosystem_id
+  end
+
+  def should_reuse_preview?
+    return unless is_preview
+
+    current_ecosystem_id == initial_ecosystem_id &&
+    current_ecosystem_id == offering&.content_ecosystem_id
+  end
+
```
nathanstitt, line 105: we should add a `order_by` to the `course_ecosystems` has_many.  Otherwise we're relying on the natural sort which **probably** does the right thing but we shouldn't rely on it

Dantemss: CourseEcosystem has a `default_scope -> { order created_at: :desc }` which makes the first one the most recent one... it's weird but it's how it works right now.

PR #2138, opened 2020-07-14
Move footnotes to bottom of page

nathanstitt:
Adds a wrapper div and "footnotes" title as well


Dantemss:
Travis is complaining that the cassette has extra stuff.

PR #2140, opened 2020-07-14
Relax the APLO regex so the new APLOs are recognized properly

Dantemss:
Just allow the stuff after the second colon to be anything.


Dantemss:
I made a tutor-js PR to allow the tags with the right format to be entered. I think exercises BE already allows anything for the tags and doesn't assign them any meaning, so hopefully no updates needed there.

PR #2141, opened 2020-07-15
strip os-part-text from titles

nathanstitt:
it's unused and causes duplication with the existing labels


Dantemss:
Merge once the hotfix is verified.

Dantemss:
We decided NOT to merge this. The fix will be done in the FE.

nathanstitt:
@Dantemss actually we now need to keep everything in the title, I updated the PR to do that

nathanstitt:
will fix conflicts after release

PR #2144, opened 2020-07-16
[Wait for release where Glicko is turned on] Task/Period Cache cleanup

Dantemss:
- Removed trouble flags
- Removed detailed stats completely
- Removed most of the normal stats API
- Remove Task/PeriodCache models
- Renamed UpdatePeriodCache to UpdateTaskPlanCache
- UpdateTaskCache only updates the cache in the Task model itself (and the TaskPlan Cache)


app/routines/get_dashboard.rb
```diff
@@ -30,59 +30,17 @@ def load_plans(course, start_at_ntz, end_at_ntz, current_time = Time.current)
...
-        pg[:num_assigned_exercises] + pg[:num_assigned_placeholders]
-      end.reduce(0, :+)
-      num_completed = pgs.map { |pg| pg[:num_completed_exercises] }.reduce(0, :+)
-      next past_due if num_completed < 0.5 * num_assigned
-
-      num_correct = pgs.map { |pg| pg[:num_correct_exercises] }.reduce(0, :+)
-      num_correct < 0.5 * num_completed
-    end
-  end
 end
```
nathanstitt, line 46: great cleanup!  good to see all that go away!

app/subsystems/tasks/update_task_plan_caches.rb
```diff
@@ -0,0 +1,16 @@
...
+  def exec(task_plan_ids:)
+    ScoutHelper.ignore!(0.995)
+
+    Tasks::Models::TaskPlan
+      .where(id: task_plan_ids)
+      .lock('FOR NO KEY UPDATE')
+      .preload(:tasks)
+      .each(&:update_gradable_step_counts!)
+  end
+end
```
nathanstitt, line 16: this seems like it'll be pretty slow due to it updating step counts for each task plan which involves reading each step, right?

Maybe background it?

Dantemss: Yeah that's the plan... It only really runs when you drop/undrop a student. But maybe I should make it update all task plans in 1 query rather than having 1 query per plan.

Dantemss: I ended up changing this to this so it's called more often but updates the plans all in 1 query.
The entire thing runs in the background anyway.

PR #2147, opened 2020-07-17
[NO MERGE] Temporarily disabled Glicko

Dantemss:
To give it time to migrate. Release this branch but do not merge.

Files were copied from v38.1.6 except I also had to fix the TeacherGuide title matchers in the spec.


Dantemss:
This has been released, but will not be merged.

PR #2155, opened 2020-07-27
Switch back to admin after impersonation

nathanstitt:
also send a flag to FE so it can show a indicator


Dantemss:
Need to add a session hash to the fake request in https://github.com/openstax/tutor-server/blob/master/spec/lib/openstax/accounts/configuration_spec.rb to fix Travis

PR #2164, opened 2020-08-04
Prevent LMS actions on courses copied from a different environment

Dantemss:
This PR prevents copied courses from running LMS sync. The routine does nothing and sends an error to Sentry. The controller returns a 403.

Works only on courses copied after the migration runs.


db/migrate/20200804171351_add_environment_name_to_courses.rb
```diff
@@ -0,0 +1,15 @@
...
+      dir.up do
+        CourseProfile::Models::Course.update_all(
+          environment_name: Rails.application.secrets.environment_name
+        )
+      end
+    end
+
+    change_column_null :course_profile_courses, :environment_name, false
+  end
+end
```
nathanstitt, line 15: instead of a string `environment_name` how about an `environments` table, and a foreign_key to it on courses?

That would prevent us from having 99% of the courses all set to the same `tutor-prod` string .  We could add a `Environment.find_or_create` method for so it'd be easy to fetch or create one for unknown ones like the AE.

Better names welcomed of course, `Environment` is maybe to overloaded of a term and `RuntimeEnviroment` or something might be better

Dantemss: sure

app/models/environment.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  validates :name, presence: true, uniqueness: true
+
+  def self.current
+    find_or_create_by! name: Rails.application.secrets.environment_name
+  end
+
+  def current?
+    name == Rails.application.secrets.environment_name
+  end
```
nathanstitt, line 12: Oh nice idea.  I like being able to `course.environment.current?`

PR #2165, opened 2020-08-04
Include students in the performance report even if they didn't receive any tasks

Dantemss:
So the gradebook always displays all enrolled students.


camliu89:
@Dantemss this is working in the FE. But it seems that some tests failed.

PR #2177, opened 2020-08-14
Admin edit styles

nathanstitt:
Fixes huge checkboxes and re-arranges admin fields a bit 
before:
image

after:
image


Dantemss:
Something is interfering with the flash alerts but not sure what...

PR #2184, opened 2020-08-18
add uuid to launch data, use it in url

nathanstitt:
Avoids the need for session cookies to be sent while inside the iframe.

Chrome had issues with this earlier (fixed in #2174), and now Safari also has issues as well

### DIFF
--- a/app/views/lms/fail_could_not_load_launch.html.erb
+++ b/app/views/lms/fail_could_not_load_launch.html.erb
@@ -3,5 +3,7 @@
 <% end %>
 
 <% content_for :body do %>
+  The launch identifier for your request was not found.
+  <br>
   Make sure your browser is set to allow cookies from OpenStax Tutor. If that does not resolve the issue, please <%= link_to "contact support", TUTOR_CONTACT_SUPPORT_URL %>.
 <% end %>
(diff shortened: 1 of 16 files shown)


db/migrate/20200818184342_uuid_launch_data.rb
```diff
@@ -0,0 +1,7 @@
+class UuidLaunchData < ActiveRecord::Migration[5.2]
+  def change
+    # TODO: next release replace id with uuid
+    # if we do that now launches in use will break
+    add_column :lms_trusted_launch_data, :uuid, :uuid, default: "gen_random_uuid()", null: false
```
Dantemss, line 5: Needs a unique index I think.

app/views/lms/fail_could_not_load_launch.html.erb
```diff
@@ -3,5 +3,7 @@
 <% end %>
 
 <% content_for :body do %>
+  The launch identifier for your request was not found.
+  <br>
   Make sure your browser is set to allow cookies from OpenStax Tutor. If that does not resolve the issue, please <%= link_to "contact support", TUTOR_CONTACT_SUPPORT_URL %>.
```
Dantemss, line 8: We can probably remove the cookie message now, but maybe leave the support link (ask Carly?)

nathanstitt: I thought about that but the message also appears on screens after this one where the uuid is in the session.

PR #2185, opened 2020-08-18
[Wait for migration tests] Fixed UI settings by completely removing json_serialize

Dantemss:
Rails has much better support nowadays for Postgres data types, so we might as well use them.
This migration might be... problematic though (but maybe it's a good test case for the migration tests? :)


Dantemss:
This PR would need to be reworked to do this change in 2 releases, we can't have the migration lock the entire server.

PR #2195, opened 2020-09-02
add metadata of viewport to get the responsiveness in mobile screen

camliu89:
it was not getting the responsive CSS queries on mobile phones.

if testing on Chrome, please make sure to use the inspector to select a mobile device.

### DIFF
--- a/app/assets/stylesheets/enroll.scss
+++ b/app/assets/stylesheets/enroll.scss
@@ -11,13 +11,6 @@
 
 $mobile-collapse-breakpoint: 600px;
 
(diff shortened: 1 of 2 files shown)


nathanstitt:
@camliu89 @dkhowala did we also want to restore the mention of $10 cost on this screen for courses that are no-free?

Alex, if we need to do that we'll need to lookup the course for the code and check it's `does_cost` property.  @jivey can probably help you with some pointers on that

nathanstitt (approved):

PR #2197, opened 2020-09-11
fetch and use manifest.json for JS assets

nathanstitt:
Goes along with https://github.com/openstax/tutor-js/pull/3241


lib/tutor/assets.rb
```diff
@@ -2,28 +2,40 @@
 
 module Tutor
   module Assets
+    # TODO: Re-read manifest (on every request?) so we can deploy FE without restarting BE
+    mattr_accessor :manifest
```
nathanstitt, line 6: yes, I puzzled over how to do that.  was thinking maybe make a admin endpoint that we could hit that'd trigger a re-fetch

Dantemss (approved):

PR #2199, opened 2020-09-16
rename manifest.json to assets.json

nathanstitt:
Rails sprockets gem will remove a manifest.json file

Goes along with https://github.com/openstax/tutor-js/pull/3293


Dantemss (approved):
Some Payments cassette was accidentally changed but it seems to not matter.

nathanstitt:
That was my bad, I did a search and replace for manifest.json   I'll put that one back, but it is strange that it didn't matter

PR #2203, opened 2020-09-28
Parse assets manifest on every request but with faraday-http-cache

Dantemss:
So we can deploy the FE without restarting the BE


nathanstitt (approved):
looks good to me

PR #2204, opened 2020-10-01
add lms score link to student roster in admin

nathanstitt:
Will allow support to partially debug why scores do not sync in the common case where a student hasn't clicked a link

image


Dantemss (approved):
Seems good to me

PR #2207, opened 2020-10-02
send date when the question was dropped

camliu89:


app/representers/api/v1/task_plan/dropped_question_representer.rb
```diff
@@ -13,4 +13,10 @@ class Api::V1::TaskPlan::DroppedQuestionRepresenter < Roar::Decorator
            readable: true,
            writeable: true,
            schema_info: { required: true }
+
+  property :updated_at,
+           type: String,
+           readable: true,
+           writeable: true,
```
Dantemss, line 20: Writeable should be false

app/representers/api/v1/task_plan/dropped_question_representer.rb
```diff
@@ -13,4 +13,10 @@ class Api::V1::TaskPlan::DroppedQuestionRepresenter < Roar::Decorator
...
+           schema_info: { required: true }
```
Dantemss, line 21: I would say leave this out

Dantemss (changes requested):

Dantemss:
Pushed change to hopefully fix graded question count when dropping questions.

nathanstitt (approved):

PR #2208, opened 2020-10-09
import ecosystems from cloudfront

nathanstitt:
Needs tests written and to discuss if we want to add a new field for the code version.

This uses urls like in this format for books:
https://openstax.org/apps/archive/20200827.155539/contents/02040312-72c8-441e-a685-20e9333f3e1d@14.4.json

The ecosystem yaml needs to be updated with the archive url including
the code version like so:

- archive_url: https://openstax.org/apps/archive/20200827.155539/


Dantemss:
Closing since @jivey made another PR that includes this commit plus the tests https://github.com/openstax/tutor-server/pull/2263

PR #2215, opened 2020-10-22
Save to practice

jivey:


app/routines/find_or_create_practice_specific_topics_task.rb
```diff
@@ -12,7 +12,7 @@ def setup(page_ids:)
     ) if ecosystems.size > 1
     @ecosystem = ecosystems.first
     @pages = pages.reject { |page| page.practice_widget_exercise_ids.empty? }
-    @page_ids = @pages.map(&:id)
+    @page_ids = @pages.map(&:id).sort
```
nathanstitt, line 16: I don't think it makes sense to sort by page id.  Probably better to sort by book_location which what the BE calls the "chapter.section" identifier

app/subsystems/tasks/complete_practice_task.rb
```diff
@@ -0,0 +1,34 @@
...
+      task_type: ::Tasks::Models::Task::PRACTICE_TASK_TYPES
+    ]
+    return unless task
+
+    task.task_steps.reject(&:completed?).map(&:delete)
+    task.task_steps.reload
+    task.due_at = Time.current
+    task.closes_at = Time.current
+    task.save!
+
```
nathanstitt, line 21: @Dantemss curious what you think about deleting the non-completed task steps here, since Josiah mentioned  that'll orphan any related records like the tasked_exercise

app/subsystems/tasks/get_practice_task.rb
```diff
@@ -3,14 +3,16 @@ class Tasks::GetPracticeTask
...
+                 .joins(:taskings)
+                 .where(taskings: { entity_role_id: role.id }, task_type: task_type)
+                 .where('completed_steps_count < steps_count')
+                 .order(:created_at)
+
+    base_query = base_query.where(core_page_ids: [page_ids].flatten) if page_ids
+    base_query = base_query.where(id: id) if id
+
+    outputs.task = base_query.first
   end
```
nathanstitt, line 17: not important at all, but the `base_` prefix seems weird to me

nathanstitt (commented):
looking good @jivey

app/routines/find_or_create_practice_saved_task.rb
```diff
@@ -0,0 +1,43 @@
+class FindOrCreatePracticeSavedTask
+  include FindOrCreatePracticeTaskRoutine
+
+  uses_routine GetCourseEcosystem, as: :get_course_ecosystem
+  uses_routine TaskExercise, as: :task_exercise
+
+  protected
+
+  def setup(**args)
+    @exercise_ids = @role.practice_questions.where(id: args[:question_ids]).select(&:available?).pluck(:content_exercise_id)
```
jivey, line 10: This might seem redundant because of using `FilterExcludedExercises` in `#add_task_steps`, but we have to check for the exercise itself being attempted. Otherwise `auto_grading_feedback_on` being `answer` will cause `task#feedback_available?` to be true (what `FilterExcludedExercises` checks), allowing the exercise to be practiced before the "real" attempt inside the assignment.

`FilterExcludedExercises` is probably still good to use for other exclusion rules.

app/routines/find_or_create_practice_specific_topics_task.rb
```diff
@@ -12,7 +12,7 @@ def setup(page_ids:)
...
```
jivey, line 15: This combined with https://github.com/openstax/tutor-server/pull/2215/files#diff-1577640ff86b21fa929caf708954818b280dd1875ad829996ac9fa3ff4bf7fd6R397 fixes a bug with the page id ordering breaking finding an existing practice task. (The inserted ordering is different than the query expects.)

Example:
```
query  1617,1618,1619,1620
insert 1618,1617,1620,1619
```

app/routines/find_or_create_practice_specific_topics_task.rb
jivey: I'll take a look. This was a small fix outlined here:
https://github.com/openstax/tutor-server/pull/2215#discussion_r511051531

app/subsystems/tasks/get_practice_task.rb
jivey: Yeah good catch, this was a remnant of having a more complicated query setup early on

app/routines/find_or_create_practice_specific_topics_task.rb
nathanstitt: ah I see.  I'd thought you were sorting for presentation 😁

jivey: I could probably also sort the Postgres array in the query 🤔

app/controllers/api/v1/practice_questions_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+             'Saves a practice question'
+  description <<-EOS
+    #{json_schema(Api::V1::PracticeQuestionRepresenter, include: :writeable)}
+  EOS
+  def create
+    step = ::Tasks::Models::TaskStep.joins(task: :taskings).find_by(
+      task: { taskings: { entity_role_id: @role.id} },
+      tasked_id: params[:tasked_exercise_id],
+      tasked_type: ::Tasks::Models::TaskedExercise.name
+    )
```
Dantemss, line 32: Could probably have loaded `tasked = Tasks::Models::TaskedExercise...` directly here and saved having to deal with tasked_type but this is very minor and the code is fine as is

app/routines/filter_excluded_exercises.rb
```diff
@@ -42,11 +42,7 @@ def exec(
 
       # Get tasks are not yet due or do not yet have feedback
       exercise_numbers_by_task_id.values_at(
-        *tasks.filter do |task|
-          past_due = task.past_due?(current_time: current_time)
-          (!task.due_at.nil? && !past_due) ||
-          !task.auto_grading_feedback_available?(past_due: past_due)
-        end.map(&:id)
+        *tasks.reject(&:feedback_available?).map(&:id)
```
Dantemss, line 45: Hmm I'm wondering here if we have to check if feedback is available per step, for the case where they haven't done an assignment that should have feedback available and then decide to practice.

app/routines/find_or_create_practice_saved_task.rb
Dantemss: As I mentioned above, I think we might want to move this into Filter...
I'm ok with merging without it since it only matters in certain corner cases, but it's something we probably want to fix soon.

app/routines/find_or_create_practice_specific_topics_task.rb
Dantemss: Hmm I am surprised the `order(:id)` in the page query wasn't enough.

app/subsystems/tasks/complete_practice_task.rb
Dantemss: Would just setting the due and close be enough? If so, leave the steps there.

db/migrate/20201006191442_create_practice_questions.rb
```diff
@@ -0,0 +1,15 @@
...
+    create_table :tasks_practice_questions do |t|
+      t.references :tasks_tasked_exercise, null: false, foreign_key: { on_update: :cascade,
+                                                                       on_delete: :cascade }
+      t.references :content_exercise, null: false,
+                                      foreign_key: { on_update: :cascade,
+                                                     on_delete: :cascade }
+      t.references :entity_role, null: false, foreign_key: { on_update: :cascade,
+                                                             on_delete: :cascade }
+      t.timestamps
+      t.index ['entity_role_id', 'content_exercise_id'], name: 'index_question_on_role_and_exercise', unique: true
```
Dantemss, line 12: Maybe overkill, but I generally tend to add indices on all the foreign keys (the multicolumn one is fine for the role since it's the first column, but I would add a separate one for content_exercise and tasked_exercise as well)

app/routines/find_or_create_practice_specific_topics_task.rb
Dantemss: I would say don't sort the postgres array in the query, I've been learning DB CPU time is precious.

Dantemss (commented):
Looks very good overall

app/subsystems/tasks/complete_practice_task.rb
jivey: @nathanstitt @Dantemss I think we have to delete them because we can't mark the step as completed without an answer, and the "is there an existing practice task?" stuff compares completed task steps (maybe we change that to check close date instead?)

db/migrate/20201006191442_create_practice_questions.rb
jivey: @Dantemss I think Rails adds the index by default when using `references` but I added a commit for clarity.

app/routines/filter_excluded_exercises.rb
nathanstitt: I think we'd decided that in that case it was fine to still prevent them from practicing it.  We wanted to encourage them to finish the assignment

app/subsystems/tasks/complete_practice_task.rb
Dantemss: Yeah checking the close date should work I think

db/migrate/20201006191442_create_practice_questions.rb
nathanstitt: yes rails adds indexes so now we have duplicate ones.  we should remove that 😁 

```
ox_tutor_dev=# \d tasks_practice_questions
                                               Table "public.tasks_practice_questions"
          Column          |            Type             | Collation | Nullable |                       Default
--------------------------+-----------------------------+-----------+----------+------------------------------------------------------
 id                       | bigint                      |           | not null | nextval('tasks_practice_questions_id_seq'::regclass)
 tasks_tasked_exercise_id | bigint                      |           | not null |
 content_exercise_id      | bigint                      |           | not null |
 entity_role_id           | bigint                      |           | not null |
 created_at               | timestamp without time zone |           | not null |
 updated_at               | timestamp without time zone |           | not null |
Indexes:
    "tasks_practice_questions_pkey" PRIMARY KEY, btree (id)
    "index_question_on_role_and_exercise" UNIQUE, btree (entity_role_id, content_exercise_id)
    "index_tasks_practice_questions_on_content_exercise_id" btree (content_exercise_id)
    "index_tasks_practice_questions_on_entity_role_id" btree (entity_role_id)
    "index_tasks_practice_questions_on_tasks_tasked_exercise_id" btree (tasks_tasked_exercise_id)
Foreign-key constraints:
    "fk_rails_0f03b8f995" FOREIGN KEY (tasks_tasked_exercise_id) REFERENCES tasks_tasked_exercises(id) ON UPDATE CASCADE ON DELETE CASCADE
    "fk_rails_ea5a42ff00" FOREIGN KEY (entity_role_id) REFERENCES entity_roles(id) ON UPDATE CASCADE ON DELETE CASCADE
    "fk_rails_f60dbdc821" FOREIGN KEY (content_exercise_id) REFERENCES content_exercises(id) ON UPDATE CASCADE ON DELETE CASCADE
```

nathanstitt (approved):

db/migrate/20201006191442_create_practice_questions.rb
Dantemss: Ah you are right... thanks

Dantemss: You could remove the role one since it's covered by the multi column index... `index: false` I think

PR #2219, opened 2020-11-02
refactor exercise filter to check each tasked

jivey:
This moves the PracticeQuestion logic into the exercises filter, which needs to check for feedback on the tasked exercise itself. MPQs should be covered here, as the number will be added for exclusion if _any_ of the parts are unavailable.

While this logic is simpler/faster, it does have an edge case. If a completed question is assigned again in a second assignment, it will be unavailable despite the student having seen answers. This also prevents allowing the Save to Practice button inside of a practice assignment (like Practice Worst.)

If we want to support that in the future, we could do something like this:

```ruby
exercise_numbers_with_status = Hash.new {|hash, key| hash[key] = [] }
taskeds.map {|t|
  exercise_numbers_with_status[t.exercise.number] << t.parts.all?(&:feedback_available?)
}
exercise_numbers_with_status.select {|k, v| v.all?(false) }.keys
```

(To address that edge case, you have to treat the feedback availability of all the MPQ steps of a tasked as a single unit, then only exclude instances where all are unavailable. Completing an MPQ (all steps) would allow availability for all steps somewhere else.)

### DIFF
--- a/app/routines/filter_excluded_exercises.rb
+++ b/app/routines/filter_excluded_exercises.rb
@@ -27,23 +27,21 @@ class FilterExcludedExercises
       outputs.already_assigned_exercise_numbers = []
       []
     else
-      tasks = role.taskings.preload(task: :course).map(&:task)
+      tasks   = role.taskings
+                  .joins(:task)
+                  .where.not(task: { task_type: Tasks::Models::Task::PRACTICE_TASK_TYPES } )
+                  .preload(task: :course).map(&:task)
+      taskeds = Tasks::Models::TaskedExercise
+                  .joins(:exercise, task_step: :task)
+                  .where(task_step: { task: { id: tasks.map(&:id) } })
+                  .preload(:exercise, task_step: { task: { task_plan: :grading_template } })
 
-      exercise_numbers_by_task_id = Hash.new { |hash, key| hash[key] = [] }
-      Content::Models::Exercise
-        .joins(tasked_exercises: :task_step)
-        .where(tasked_exercises: { task_step: { tasks_task_id: tasks.map(&:id) } })
-        .pluck(:tasks_task_id, :number)
-        .each do |task_id, number|
-        exercise_numbers_by_task_id[task_id] << number
-      end
-
-      outputs.already_assigned_exercise_numbers = exercise_numbers_by_task_id.values.flatten.uniq
+      outputs.already_assigned_exercise_numbers = taskeds.map {|t| t.exercise.number }.flatten.uniq
 
       # Get tasks are not yet due or do not yet have feedback
-      exercise_numbers_by_task_id.values_at(
-        *tasks.reject(&:feedback_available?).map(&:id)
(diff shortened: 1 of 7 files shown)


app/subsystems/tasks/models/practice_question.rb
```diff
@@ -9,13 +9,6 @@ class Tasks::Models::PracticeQuestion < ApplicationRecord
...
   def available?
-    parts.all?(&:feedback_available?)
-  end
-
-  def parts
-    return [tasked_exercise] unless tasked_exercise.is_in_multipart?
-
-    task = tasked_exercise.task_step.task
-    task.task_steps.exercises.preload(:tasked).map(&:tasked).select{|tasked| tasked.content_exercise_id == content_exercise_id }
+    tasked_exercise.parts.all?(&:feedback_available?)
```
jivey, line 12: There is an edge case here where a question that is assigned twice — one complete, one incomplete — may be available to select in the UI, but will be excluded by the filter. Double assigned questions are probably rare, but if we do want to handle this in the future, I think we can stop using `#available?` entirely and return a list of unavailable ids in the representer using the excluded list from the filter.

Dantemss: Double assigned questions happen a lot because of spaced practice, but I think it's fine to have the behavior you described, so the current code should be fine.

app/routines/filter_excluded_exercises.rb
```diff
@@ -27,23 +27,21 @@ def exec(
...
     else
-      tasks = role.taskings.preload(task: :course).map(&:task)
+      tasks   = role.taskings
+                  .joins(:task)
+                  .where.not(task: { task_type: Tasks::Models::Task::PRACTICE_TASK_TYPES } )
+                  .preload(task: :course).map(&:task)
+      taskeds = Tasks::Models::TaskedExercise
+                  .joins(:exercise, task_step: :task)
+                  .where(task_step: { task: { id: tasks.map(&:id) } })
+                  .preload(:exercise, task_step: { task: { task_plan: :grading_template } })
```
Dantemss, line 37: This is a minor thing, but we could combine these 2 to save 1 query.

app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -339,6 +339,13 @@ def feedback_available?(current_time: Time.current)
     )
   end
 
+  def parts
+    return [self] unless is_in_multipart?
+
+    task = task_step.task
+    task.task_steps.exercises.preload(:tasked).map(&:tasked).select{|tasked| tasked.content_exercise_id == content_exercise_id }
```
Dantemss, line 346: Maybe a TODO for the future: select only some columns from tasked_exercises to avoid loading the big `content` field.

Dantemss (approved):
Looks good. The test failure is probably related to the Daylight Savings Time change, not the code in the PR.

PR #2220, opened 2020-11-06
Exercise editing for teachers

nathanstitt:
main integration branch for exercise editing


app/controllers/api/v1/storage_controller.rb
```diff
@@ -0,0 +1,4 @@
+class Api::V1::StorageController < ApplicationController
+
+
+end
```
Dantemss, line 4: Is this used at all?

db/schema.rb
```diff
@@ -556,14 +583,14 @@
...
     t.string "salesforce_contact_id"
-    t.uuid "uuid", default: -> { "gen_random_uuid()" }, null: false
     t.integer "role", default: 0, null: false
     t.citext "support_identifier"
     t.boolean "is_test"
     t.integer "school_type", default: 0, null: false
     t.boolean "is_kip"
     t.integer "school_location", default: 0, null: false
     t.boolean "grant_tutor_access"
+    t.uuid "uuid", default: -> { "gen_random_uuid()" }
```
Dantemss, line 593: Where did the `null: false` go?

db/schema.rb
```diff
@@ -575,7 +602,6 @@
     t.index ["school_type"], name: "index_openstax_accounts_accounts_on_school_type"
     t.index ["support_identifier"], name: "index_openstax_accounts_accounts_on_support_identifier", unique: true
     t.index ["username"], name: "index_openstax_accounts_accounts_on_username"
-    t.index ["uuid"], name: "index_openstax_accounts_accounts_on_uuid", unique: true
```
Dantemss, line 578: Same for this index

app/routines/get_exercises.rb
```diff
@@ -29,9 +29,13 @@ def exec(ecosystem: nil, course: nil, page_ids: nil, exercise_ids: nil, pool_typ
...
 
+    profile_ids = [User::Models::OpenStaxProfile::ID]
+    profile_ids << course.related_teacher_profile_ids if course
+
     # Preload exercises, pages and teks tags
     exercises_by_id = Content::Models::Exercise
       .where(id: exercise_ids_by_pool_type.values.flatten.uniq)
+      .where(user_profile_id: profile_ids.flatten.uniq)
       .preload(:page, tags: :teks_tags)
       .index_by(&:id)
```
Dantemss, line 40: Alex noted that for teacher created exercises this routine probably has to ensure only the latest version that you have access to is returned

Dantemss (commented):
Just a few comments

app/routines/create_teacher_exercise.rb
```diff
@@ -0,0 +1,77 @@
...
+      has_interactive: wrapper.has_interactive?,
+      has_video: wrapper.has_video?,
+      derived_from: derived_from,
+      anonymize_author: anonymize || false,
+      is_copyable: copyable || true
+    )
+    exercise.images.attach(images) if images
+
+    if save
+      exercise.set_teacher_exercise_number
```
camliu89, line 37: @jivey I think we should be calling this only if `derived_from` is `null`. I believe we can set the `uid` as `#{derived_from.number}@#{version}`.

Dantemss: What did we decide was the rule for when to create new versions of the same exercise vs when to create a new exercise number?

Keeping in mind that if you create a new number you can have duplicates in assignments unless you exclude the original.

app/routines/create_teacher_exercise.rb
```diff
@@ -0,0 +1,77 @@
...
+  uses_routine Content::Routines::PopulateExercisePools, as: :populate_exercise_pools
+
+  protected
+
+  def exec(course:, page:, content:, profile:, derived_from_id: nil, images: nil, tags: [], copyable: nil, anonymize: nil, save: false)
+    tags << ['type:practice']
+
+    wrapper = OpenStax::Exercises::V1::Exercise.new(content: content.to_json)
+    derived_from = derived_from_id && find_derivable_exercise(course, derived_from_id)
+    version = (derived_from&.version || 0) + 1
```
Dantemss, line 14: I think 2 different uses of derived_from are being conflated here:
1. determining which exercise is being edited (to set number and version)
2. determining the original exercise for attribution purposes (what is stored as derived_from_id)

They are the same initially, but not always.
We'll probably need a new parameter to determine which exercise is being edited.

app/subsystems/content/models/exercise.rb
```diff
@@ -104,4 +111,28 @@ def is_free_response_only?
...
+  end
+
+  def authored_by_teacher?
+    user_profile_id.present? && user_profile_id != User::Models::OpenStaxProfile::ID
+  end
+
+  def set_teacher_exercise_number
+    return unless authored_by_teacher?
+
+    self.number = generate_next_teacher_exercise_number
```
Dantemss, line 132: When a new number if set we could reset the version to 1.

app/routines/create_teacher_exercise.rb
Dantemss: This got moved to the model and now checks that you are the author of the exercise being edited.

app/subsystems/content/models/exercise.rb
```diff
@@ -147,6 +168,21 @@ def set_teacher_exercise_number_and_version
...
+    self
+  end
+
+  def set_teacher_exercise_coauthors
+    return unless authored_by_teacher? && derived_from
+    return if derived_from.coauthor_profile_ids.empty? && derived_from_same_profile?
+
+    coauthor_ids = derived_from.coauthor_profile_ids.dup
+
+    [derived_from.author.id, author.id].each do |id|
```
Dantemss, line 194: I think this means the first author is not a coauthor on the original exercise (because derived_from is nil) but then on the first derivation the current author is also added as a coauthor which is weird.

app/subsystems/content/models/exercise.rb
```diff
@@ -147,6 +168,21 @@ def set_teacher_exercise_number_and_version
...
+      coauthor_ids << id unless !id.in?(abstract_profile_ids) && id.in?(coauthor_ids)
```
Dantemss, line 195: Since Nathan is always telling me to avoid these double negatives,

`coauthor_ids << id if id.in?(abstract_profile_ids) || !id.in?(coauthor_ids)`

Dantemss: But does this mean abstract profiles can be added multiple times? That's a bit weird.

app/subsystems/course_profile/models/course.rb
```diff
@@ -32,6 +32,9 @@ class CourseProfile::Models::Course < ApplicationRecord
   has_many :periods, subsystem: :course_membership, inverse_of: :course
 
   has_many :teachers, subsystem: :course_membership, inverse_of: :course
+  has_many :teacher_roles, through: :teachers, source: :role, class_name: 'Entity::Role'
+  has_many :teacher_profiles, through: :teacher_roles, source: :profile, class_name: 'User::Models::Profile'
```
Dantemss, line 36: I'm generally not a fan of has_many :through because it breaks preloading...

Dantemss (commented):
Looks good but see comments about coauthors.

app/routines/get_exercises.rb
Dantemss: This got implemented.

app/subsystems/course_profile/models/course.rb
nathanstitt: I'm not sure if these are used.  I'd added them but then later saw that @jivey had added a method to calculate them.  @jivey are they needed?

app/routines/create_teacher_exercise.rb
Dantemss: We decided that derived_from tracks the previous exercise in the chain and we keep an array of coauthors for attribution purposes.

app/subsystems/content/models/exercise.rb
Dantemss: This got implemented.

app/controllers/api/v1/storage_controller.rb
nathanstitt: i don't think so, i think sometimes it's used to allow custom uploads, but we're not doing that

db/schema.rb
Dantemss: Seems to have been fixed.

db/schema.rb
Dantemss: Seems to have been fixed.

app/views/webview/home.html.erb
```diff
@@ -16,53 +16,53 @@
...
-    Advanced Placement® and AP® are trademarks registered and/or owned by the College Board, which is not affiliated with, and does not endorse, this site.
+    <footer>
+      <p>
+        <%= render 'layouts/copyright_terms' %>
+        &middot;
+        <a href="http://gdpr.rice.edu/" target="_blank">GDPR</a>
+      </p>
+      <p class="disclaimer">
+        Advanced Placement® and AP® are trademarks registered and/or owned by the College Board, which is not affiliated with, and does not endorse, this site.
   </p>
```
nathanstitt, line 66: I don't think anything changed here.  I'm torn between reverting the changes or allowing them since they do fix the indentation.

lib/openstax/exercises/v1/fake_client.rb
```diff
@@ -63,7 +63,7 @@ def self.new_exercise_hash(uuid: SecureRandom.uuid, group_uuid: SecureRandom.uui
         }
       ],
       questions: num_questions.times.map do |index|
-        question_number = number + index
+        question_number = (number || -1) + index
```
Dantemss, line 66: I think generally those start at 1 but shouldn't hurt anything

spec/lib/has_timezone_spec.rb
```diff
@@ -62,19 +62,17 @@
...
 
-    test_tz_names = [ 'UTC' ] + CourseProfile::Models::Course::VALID_TIMEZONES
-    test_tz_names.each do |tz_name|
-      expect do
-        task.course.update_attribute :timezone, tz_name
-      end.to change { task.reload.closes_at.zone }
+    test_tz_names = CourseProfile::Models::Course::VALID_TIMEZONES + [ 'UTC' ]- [ 'US/Central' ]
+    expect do
+      task.course.update_attribute :timezone, test_tz_names.sample
+    end.to change { task.reload.closes_at.zone }
```
nathanstitt, line 68: hm, why are we removing central time again?  I thought it was Mountain that had the weirdness?  At any rate can we add a comment here so future us will remember 😁

app/subsystems/course_profile/models/course.rb
jivey: @nathanstitt oops yeah I added `#teacher_profiles` to map what I needed in `CollectCourseInfo`, maybe you added these because the representer gets reused somewhere `CollectCourseInfo` doesn't provide the data?

lib/openstax/exercises/v1/fake_client.rb
jivey: On my end they started at `-1` so I preserved that incase it was important, I wasn't really sure.

app/subsystems/content/models/exercise.rb
jivey: We return early with `derived_from.coauthor_profile_ids.empty? && derived_from_same_profile?` so I think the only case a teacher could be both author and in the coauthors list is if a 2nd teacher copies their exercise, and the 1st teacher copies that. I think that's probably fine. If they only copy their own exercises, it _should_ continue to keep coauthors empty.

app/subsystems/content/models/exercise.rb
jivey: Good note, I didn't like the double negative either. I wish I could just call `uniq` here, but we decided to continue adding anonymous authors rather than collapse them.

app/subsystems/content/models/exercise.rb
Dantemss: 2 copies 1, then 3 copies 2... wouldn't 3 immediately be a co-author? (why is the current exercise's author.id added to the coauthors array?)

app/subsystems/content/models/exercise.rb
Dantemss: Maybe it makes sense for anonymous, but it would still be weird to have multiple OpenStax.

app/subsystems/content/models/exercise.rb
jivey: I think we'd decided (outside of the single author case) that the list should be anyone that has contributed to the question. I can't remember if that was a legal or semantic reason... either way is fine to me. Should I change it to not include the current author?

app/subsystems/content/models/exercise.rb
jivey: I can't see why that case would ever happen (`derived_from.author` would only be OS in an initial copy,) but I can update it to skip

app/subsystems/course_profile/models/course.rb
jivey: @nathanstitt I removed them and it didn't seem to break anything, so I'm hoping the specs pass here: https://github.com/openstax/tutor-server/pull/2233

PR #2222, opened 2020-11-20
add teacher profiles to course representer

jivey:
Need these on the front end to list selectable authors on an exercise.


camliu89:
@jivey this is working in the FE.

camliu89 (approved):

PR #2225, opened 2020-12-02
bump version number by looking at max version and generate uuids

jivey:
The version generation needs to consider that the given derivable exercise might not be the original, and bump the latest version it can find globally rather than from the derivable.


jivey:
@nathanstitt @Dantemss Updated PR with the logic we discussed, basically:

1) If no derivable sent (fresh exercise) then generate number + version 1
2) Derivable sent: is it owned by current user? Then reuse number + bump version. Otherwise do the 1st option above (derived_from is always set when sent regardless.)

jivey:
@nathanstitt @Dantemss One last update to this PR, since the `uuid` changes feel at home with the version stuff already here.

Based on our conversations, a `uuid` will be generated every time, as every teacher exercise is always a new version. If there is a valid derivable source (the teacher's own exercise,) then the `group_uuid` will use the derivable's (just like `number`). Otherwise a new `group_uuid` will be generated (so will a new `number`.)

PR #2229, opened 2020-12-04
Delete teacher exercise API

Dantemss:
I added API specs for this API but not specs for the exclusion of deleted exercises from other APIs.


app/routines/filter_excluded_exercises.rb
```diff
@@ -25,9 +25,11 @@ def exec(
...
     # Exclude exercises that aren't created by OS or course teacher(s)
     profile_ids << User::Models::OpenStaxProfile::ID
-    no_ownership_excluded_numbers = exercises.map do |exercise|
-      exercise.number unless exercise.user_profile_id.in?(profile_ids.flatten.uniq)
-    end
+    no_ownership_excluded_numbers = exercises.select do |exercise|
+      exercise.user_profile_id.in?(profile_ids.flatten.uniq)
+    end.map(&:number)
+
+    deleted_excluded_numbers = exercises.select(&:deleted?).map(&:number)
```
nathanstitt, line 32: @Dantemss I think this check is inverted and should be not in

When I load on this branch I don't see any exercises at all returned and think I've tracked the cause down to this check

PR #2233, opened 2020-12-07
Remove teacher relations

jivey:
I don't think we need these since we get what we need in `CollectCourseInfo`.


nathanstitt (approved):
You beat me to this by about 10 seconds!

jivey:
Oops thought I merged, thanks!

PR #2234, opened 2020-12-08
don't include current author in the coauthors list

jivey:
We decided it was better to keep these separate, for a definitive list of contributors you'll need author + coauthors


nathanstitt (approved):
perfect (I hope).  Love the comment explaining the logic

PR #2236, opened 2020-12-09
remove url from presence validation

jivey:
URL is no longer required because of Teacher-created exercises.


camliu89 (approved):
:+1: it can save answers now.

PR #2239, opened 2020-12-11
Update tasks with new teacher exercise version

jivey:
#### Overview 

When a new version of a teacher exercise is created (edited by the same profile that created the original version), find any tasked exercises that have an old version id, and if the task plan isn't open yet, update them to the new version id. This also updates the `settings` exercises and page ids (page ids _could_ change between versions) on the task plan.

#### A 🤔 filter thing
If you look at the diff here: https://github.com/openstax/tutor-server/commit/b1b2bd48bd3b1d2c9c2d81d2910aa371fbb78ce1#diff-dc0c01a1d84a4b8c0f82d31dfbe079880d0c619929f98e24811c501ef12d569cL7-L13, I was initially filtering by the task ids the profile has access to, but it doesn't feel like the extra query is necessary, since numbers are already associated so closely with the profile. The only long term issue I can think of is if a "community questions" area is added, and assignments can be created with someone else's number — you probably don't want their assignments getting updated if the original author updates their exercise. Thoughts?

### DIFF
--- a/app/routines/update_assigned_exercise_version.rb
+++ b/app/routines/update_assigned_exercise_version.rb
@@ -24,7 +24,7 @@ class UpdateAssignedExerciseVersion
       next if plan.out_to_students?
 
       plan.settings['exercises'].each {|ex| ex['id'] = new_version.id.to_s if ex['id'].in?(old_ids) }
-      plan.settings['page_ids'] << new_version.page.id.to_s
+      (plan.settings['page_ids'] ||= []) << new_version.page.id.to_s
       plan.settings['page_ids'].uniq!
       plan.save
 
--- a/app/routines/update_assigned_exercise_version.rb
+++ b/app/routines/update_assigned_exercise_version.rb
@@ -4,13 +4,14 @@ class UpdateAssignedExerciseVersion
   protected
 
   def exec(number:, profile:)
-    task_ids     = profile.roles.select {|r| r.role_type == 'teacher' }.map(&:course).map(&:tasks).flatten.map(&:id)
+    updated_task_plan_ids = []
+    updated_tasked_exercise_ids = []
+
     all_versions = Content::Models::Exercise.where(number: number)
     new_version  = all_versions.max_by(&:version)
     old_ids      = all_versions.without(new_version).map {|v| v.id.to_s }
     update_ids   = Content::Models::Exercise
                      .joins(tasked_exercises: { task_step: { task: :task_plan } })
-                     .where(tasked_exercises: { task_step: { tasks_task_id: task_ids } })
                      .where(number: number)
                      .where.not(version: new_version.version)
                      .pluck('tasks_task_plans.id', 'tasks_tasked_exercises.id')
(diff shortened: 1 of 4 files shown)


Dantemss (changes requested):
Looks good overall. See comments.

app/routines/update_assigned_exercise_version.rb
```diff
@@ -0,0 +1,42 @@
...
+    update_ids.each {|set| taskeds_by_plan_id[set[0]] << set[1] }
+
+    taskeds_by_plan_id.each do |plan_id, tasked_ids|
+      plan = Tasks::Models::TaskPlan.find(plan_id)
+      next if plan.out_to_students?
+
+      plan.settings['exercises'].each {|ex| ex['id'] = new_version.id.to_s if ex['id'].in?(old_ids) }
+      (plan.settings['page_ids'] ||= []) << new_version.page.id.to_s
+      plan.settings['page_ids'].uniq!
+      plan.save
```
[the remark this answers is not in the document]
jivey: Yeah I think you're right, I ran into this while writing tests when I forgot to put the new version into the same ecosystem :D Lots to think about when we do the ecosystem fixes.

app/routines/update_assigned_exercise_version.rb
```diff
@@ -0,0 +1,46 @@
...
+                     .pluck('tasks_task_plans.id', 'tasks_tasked_exercises.id')
+
+    taskeds_by_plan_id = Hash.new {|hash, key| hash[key] = [] }
+    update_ids.each {|set| taskeds_by_plan_id[set[0]] << set[1] }
+
+    taskeds_by_plan_id.each do |plan_id, tasked_ids|
+      plan = Tasks::Models::TaskPlan.find(plan_id)
+      next if plan.out_to_students?
+
+      (plan.settings['exercises'] ||= []).each do |ex|
```
Dantemss, line 27: I think you actually can't set plan.settings['exercises'] for readings or they'll fail validation
So if plan.settings['exercises'] is nil, just skip updating them

app/routines/update_assigned_exercise_version.rb
```diff
@@ -0,0 +1,49 @@
...
+    all_versions = Content::Models::Exercise.where(number: number)
+    new_version  = all_versions.max_by(&:version)
+    old_ids      = all_versions.without(new_version).map {|v| v.id.to_s }
+    update_ids   = Content::Models::Exercise
+                     .joins(tasked_exercises: { task_step: { task: :task_plan } })
+                     .where(number: number)
+                     .where.not(user_profile_id: User::Models::OpenStaxProfile::ID)
+                     .where.not(version: new_version.version)
+                     .pluck('tasks_task_plans.id', 'tasks_tasked_exercises.id')
+
```
nathanstitt, line 19: should we limit this to non-openstax exercises?  I thought we were going to switch to also looking for possible ecosystem updated exercises here?

Dantemss: I think it's fine to limit it to non-openstax. If this is called when the teacher saves an exercise, the openstax exercises shouldn't have been updated.

Dantemss (approved):
Looks good to me... send it to testing

nathanstitt (approved):

PR #2241, opened 2020-12-12
generate question and answer ids for teacher exercises

jivey:
The FE uses question and answer ids for filtering student responses in some UI areas, so start giving these a UUID.


jivey:
Sorry about the big spec diff, I was making it more rspec-y with context blocks.

jivey:
We could also use an exercise number prefix instead of a UUID, I have no clear thoughts if one would be better besides length and readability.

nathanstitt:
I think uuid is good, nothing should rely on these other than being unique

nathanstitt (approved):

PR #2245, opened 2020-12-18
remap teacher exercises and practice questions

jivey:
- Remap teacher exercises to new ecosystem page ids during book import
- Remap practice question exercise ids to the new id when updating course ecosystem
- Always run UpdateTaskPlanEcosystem when creating a task in a cloned course, which solves 1) always getting the latest teacher exercise version and 2) setting new page ids so the checked page/section is always correct when editing the details

### DIFF
--- a/app/routines/update_task_plan_ecosystem.rb
+++ b/app/routines/update_task_plan_ecosystem.rb
@@ -19,8 +19,6 @@ class UpdateTaskPlanEcosystem
 
     old_ecosystem = outputs.task_plan.set_and_return_ecosystem
 
-    return if old_ecosystem == ecosystem
-
     outputs.task_plan.ecosystem = ecosystem
 
     return if old_ecosystem.nil? || !TPS_THAT_NEED_UPDATES.include?(outputs.task_plan.type)
@@ -63,7 +61,7 @@ class UpdateTaskPlanEcosystem
 
     return unless exercises.present?
 
-    # Update exercises to the new ecosystem by exercise number
+    # Update exercises to the new ecosystem or new version by exercise number
     exercise_number_by_id = Content::Models::Exercise.where(
       id: exercises.map { |ex| ex['id'] }
     ).pluck(:id, :number).to_h
@@ -71,9 +69,15 @@ class UpdateTaskPlanEcosystem
     new_exercise_id_by_number = ecosystem.exercises.where(
       number: exercise_numbers
     ).pluck(:number, :id).to_h
+    page_id_by_exercise_id = ecosystem.exercises.where(
+      number: exercise_numbers
+    ).pluck(:id, :content_page_id).to_h
     outputs.task_plan.settings['exercises'] = exercises.each_with_index.map do |hash, index|
       id = new_exercise_id_by_number[exercise_numbers[index]]
       next if id.nil?
+
+      (outputs.task_plan.settings['page_ids'] ||= []) << page_id_by_exercise_id[id].to_s
+      outputs.task_plan.settings['page_ids'].uniq!
 
       { 'id' => id.to_s, 'points' => hash['points'] }
     end.compact
--- a/app/subsystems/content/import_book.rb
+++ b/app/subsystems/content/import_book.rb
@@ -5,6 +5,7 @@ class Content::ImportBook
   uses_routine Content::Routines::ImportExercises, as: :import_exercises
   uses_routine Content::Routines::TransformAndCachePageContent, as: :transform_and_cache_content
   uses_routine Content::Routines::PopulateExercisePools, as: :populate_exercise_pools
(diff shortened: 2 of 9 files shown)


jivey:
@nathanstitt @Dantemss I still need to write a spec for that UpdateTaskPlanEcosystem change, but I'd appreciate some 👀 on this in case changes are needed.

app/routines/update_task_plan_ecosystem.rb
```diff
@@ -63,18 +61,24 @@ def update_task_plan(task_plan:, ecosystem:)
...
     ).pluck(:number, :id).to_h
+    page_id_by_exercise_id = ecosystem.exercises.where(
+      number: exercise_numbers
+    ).pluck(:id, :content_page_id).to_h
     outputs.task_plan.settings['exercises'] = exercises.each_with_index.map do |hash, index|
       id = new_exercise_id_by_number[exercise_numbers[index]]
       next if id.nil?
 
+      (outputs.task_plan.settings['page_ids'] ||= []) << page_id_by_exercise_id[id].to_s
+      outputs.task_plan.settings['page_ids'].uniq!
```
jivey, line 80: Something I'm wondering about this addition, did new ecosystem page ids not get picked up before?

app/routines/update_task_plan_ecosystem.rb
```diff
@@ -19,8 +19,6 @@ def update_task_plan(task_plan:, ecosystem:)
 
     old_ecosystem = outputs.task_plan.set_and_return_ecosystem
 
-    return if old_ecosystem == ecosystem
```
nathanstitt, line 22: I'm curious why this was removed.  it seems like if the ecosystems match there's nothing to do and we can quit early

app/routines/update_task_plan_ecosystem.rb
nathanstitt: there's probably pages that have no exercises?

nathanstitt (approved):

app/routines/update_task_plan_ecosystem.rb
Dantemss: The page_ids get set on line 52.

app/routines/update_task_plan_ecosystem.rb
jivey: The docs wanted a cloned task plan to always grab the latest exercise even if the ecosystem is the same.

nathanstitt: ah that makes sense

app/subsystems/content/routines/remap_teacher_exercises.rb
```diff
@@ -0,0 +1,50 @@
...
+      from_ecosystems: ecosystems, to_ecosystem: ecosystem
+    )
+    mapped = map.map_page_ids(page_ids: page_ids).compact.reject do |k, v|
+      k.to_s == v.to_s
+    end
+
+    mapped.each do |from, to|
+      exercises = Content::Models::Exercise.created_by_teacher.where(content_page_id: from)
+      updated_exercise_ids_by_page_id[from.to_s] = exercises.map(&:id).sort
+      exercises.update_all(content_page_id: to) if save
```
Dantemss, line 31: We can't really update the existing ones or existing assignments will be inconsistent. We'd have to make copies.

app/routines/update_task_plan_ecosystem.rb
Dantemss: I think because of version updates.

app/subsystems/content/import_book.rb
```diff
@@ -132,6 +133,8 @@ def exec(cnx_book:, ecosystem:, reading_processing_instructions: nil, exercise_u
       ecosystem: ecosystem, page: page_block, query_hash: query_hash, all_tags: all_tags
     )
 
+    run(:remap_teacher_exercises, ecosystem: ecosystem, save: true)
```
Dantemss, line 136: This runs before we assign the new book to the courses, another reason why we can't modify existing exercises here.

app/subsystems/content/routines/remap_practice_question_exercises.rb
```diff
@@ -0,0 +1,33 @@
...
+    outputs.mapped_ids = mapped_ids
+
+    return unless save
+
+    mapped_ids.each do |old_id, new_id|
+      Tasks::Models::PracticeQuestion
+        .where(entity_role_id: role_ids)
+        .where(content_exercise_id: old_id)
+        .update_all(content_exercise_id: new_id)
+    end
```
Dantemss, line 31: This can be up to 10000 queries which is pretty slow. Fine for now but let's fix for Fall 2021.

Dantemss (changes requested):
Main thing is we can't mutate existing Content::Exercises

Dantemss (approved):
Looks good

PR #2246, opened 2020-12-21
Added exercise editing terms placeholder for local testing and new installations

Dantemss:
`rake db:seed`


nathanstitt (approved):
thanks @Dantemss !

PR #2262, opened 2021-01-27
add a course code field in course_profiles table

camliu89:
Add a code field in the `course_profiles` table


db/schema.rb
```diff
@@ -1155,7 +1157,7 @@
     t.integer "completed_on_time_exercise_steps_count", default: 0, null: false
     t.integer "completed_on_time_steps_count", default: 0, null: false
     t.datetime "hidden_at"
-    t.uuid "uuid", default: -> { "gen_random_uuid()" }, null: false
+    t.uuid "uuid", default: -> { "public.gen_random_uuid()" }, null: false
     t.integer "content_ecosystem_id", null: false
```
nathanstitt, line 1161: @Dantemss do we want to record the changes that add the `public` to the functions?  maybe that's the new default for PG

Dantemss: I get the column order changes when I rerun migration but I don't get the public changes. Shouldn't matter either way, since public is implicit.

Maybe what we should do is have GH actions regenerate the schema or something? Wonder how hard that is to setup.

nathanstitt (commented):
@camliu89 we'll also to add the subject field to the admin form to allow admins to set it. the form's here: https://github.com/openstax/tutor-server/blob/main/app/views/admin/catalog_offerings/_edit_row.html.erb

the routine that updates the model will probably also need changed https://github.com/openstax/tutor-server/blob/main/app/subsystems/catalog/update_offering.rb

Dantemss (approved):

PR #2263, opened 2021-01-27
Fix cf import tests

jivey:
This updates the existing cassettes with the new url format — pages prefixed with book id, no redirects, etc. Also fixes a small bug with `archive_url_for` where if a full path like `https://openstax.org/apps/archive/20201222.172624/contents` was passed, it would add an extra `contents`: `https://openstax.org/apps/archive/20201222.172624/contents/contents`, and also made it friendly to top-level `/resources/` paths.

This branch includes the latest `main` (some unrelated specs were fixed there) + `cf-imports`, but I'm targeting `main` for now so the diff will be useful.


Dantemss (commented):
See my comment about the archive_url but otherwise looks good.

Dantemss:
Needs revisiting now that we have the mini book for specs

PR #2268, opened 2021-02-23
add offerings to bootstrap data

jivey:
We decided to include this in the bootstrap data as it's needed most of the time anyway.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -39,7 +39,7 @@ class Api::V1::UsersController < Api::V1::ApiController
                      tutor_api_url: api_root_url,
                      flash: flash.to_hash,
                      is_impersonating: session[:admin_user_id].present?,
-                     is_teacher: current_human_user.roles.any?(&:teacher?)
+                     is_teacher: current_human_user.can_create_courses?
                    }
     end
   end
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -38,7 +38,8 @@ class Api::V1::UsersController < Api::V1::ApiController
                    user_options: {
                      tutor_api_url: api_root_url,
                      flash: flash.to_hash,
-                     is_impersonating: session[:admin_user_id].present?
+                     is_impersonating: session[:admin_user_id].present?,
+                     is_teacher: current_human_user.roles.any?(&:teacher?)
                    }
     end
   end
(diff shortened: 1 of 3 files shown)


app/representers/api/v1/bootstrap_data_representer.rb
```diff
@@ -104,5 +104,11 @@ class BootstrapDataRepresenter < ::Roar::Decorator
...
                writeable: false,
                getter: ->(user_options:, **) { CollectCourseInfo[user: self] }
+
+    collection :offerings,
+               extend: Api::V1::OfferingRepresenter,
+               readable: true,
+               writeable: false,
+               getter: ->(*) { Catalog::Models::Offering.without_deleted }
   end
 end
```
nathanstitt, line 114: while they're not really sensitive, we should only send the offerings to teachers. No read to send to students.  You can use `user_options` to the representer to indicate if they should be included.  https://github.com/openstax/tutor-server/blob/main/app/representers/api/v1/exercise_representer.rb#L37-L40 has an example

nathanstitt (approved):

nathanstitt:
looks good, but specs failed though :(

app/controllers/api/v1/users_controller.rb
```diff
@@ -38,7 +38,8 @@ def bootstrap
                    user_options: {
                      tutor_api_url: api_root_url,
                      flash: flash.to_hash,
-                     is_impersonating: session[:admin_user_id].present?
+                     is_impersonating: session[:admin_user_id].present?,
+                     is_teacher: current_human_user.roles.any?(&:teacher?)
```
Dantemss, line 42: This will be false before they create their first course, making the offerings not be returned. Will that work?

nathanstitt: good catch @Dantemss - yes we'll need the offerings to allow teachers to create their first course.

Josiah, I think you can use the same logic that the existing offering access policy uses:
https://github.com/openstax/tutor-server/blob/main/app/access_policies/offering_access_policy.rb#L5

nathanstitt (changes requested):

app/controllers/api/v1/users_controller.rb
jivey: Yes good call, fixed

nathanstitt (approved):
The homework_assistant spec's still flaky. we can merge and make a followup for that though

PR #2269, opened 2021-02-25
redirect to new url '/courses' instead of '/dashboard'

camliu89:


app/controllers/webview_controller.rb
```diff
@@ -10,7 +10,7 @@ def home
     if params[:cc] == "1"
       redirect_to 'http://cc.openstax.org'
     elsif current_user.is_signed_in?
-      redirect_to dashboard_path
```
camliu89, line 13: @nathanstitt @Dantemss where is `dashboard_path` defined? I can't find so hardcoded `/courses` here.

jivey: It's in `routes.rb` line 23, I haven't tested it but something like this might do what you want:
```
get :courses, as: :dashboard
```

camliu89: thanks @jivey!

nathanstitt:
@camliu89 can you update the homework_assitant spec to use the values that it expects here: https://github.com/openstax/tutor-server/pull/2269/checks?check_run_id=1983333200#step:6:710

lets see if that fixes the flakiness or if the values continue to change

PR #2276, opened 2021-03-09
render book list as a select tag with titles and cnx_ids

jivey:
Use a select tag with titles and cnx_ids to make it easier to associate offerings with books in admin.

image

### DIFF
--- a/app/views/admin/catalog_offerings/_edit_row.html.erb
+++ b/app/views/admin/catalog_offerings/_edit_row.html.erb
@@ -33,8 +33,8 @@
     <%= form.text_field :subject, class: 'form-control' %>
   </div>
   <div class="form-group">
-    <%= form.label :os_book_id %>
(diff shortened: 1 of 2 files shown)


app/views/admin/catalog_offerings/_edit_row.html.erb
```diff
@@ -34,7 +34,7 @@
   </div>
   <div class="form-group">
     <%= form.label :os_book_id %>
-    <%= form.text_field :os_book_id, class: 'form-control' %>
+    <%= form.select :os_book_id, options_for_select(@books, selected: form.object.os_book_id), { include_blank: true }, class: 'form-control' %>
   </div>
```
nathanstitt, line 38: hm, I'm afraid I don't remember what this field even is for, what is the book used for again?

nathanstitt:
@jivey ok, now that I've refreshed my memory on what this is, I think we want to leave it as a free-form string, but probably need to update the label to read `OpenStax Book Identifier` or something like that so we can hopefully remember it's purpose in the future 🤞    why don't you update this pr to do so

nathanstitt (approved):

PR #2280, opened 2021-03-15
sort the offerings in the bootstrap data 

camliu89:
if an offering has no real course AND no preview courses, put them at the end of the array.

### DIFF
--- a/app/representers/api/v1/bootstrap_data_representer.rb
+++ b/app/representers/api/v1/bootstrap_data_representer.rb
@@ -109,6 +109,7 @@ module Api::V1
                extend: Api::V1::OfferingRepresenter,
                readable: true,
                writeable: false,
-               getter: ->(*) { Catalog::Models::Offering.without_deleted }
+               #  if offering's real course and preview courses are not available, put it at the bottom of the array
+               getter: ->(*) { Catalog::Models::Offering.without_deleted.sort_by { |o| o.is_available && o.is_preview_available ? 0 : 1 } }
   end
 end
--- a/app/representers/api/v1/bootstrap_data_representer.rb
+++ b/app/representers/api/v1/bootstrap_data_representer.rb
@@ -109,7 +109,6 @@ module Api::V1
                extend: Api::V1::OfferingRepresenter,
                readable: true,
                writeable: false,
-               #  if offering's real course and preview courses are not available, put it at the bottom of the array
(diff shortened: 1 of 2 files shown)


app/representers/api/v1/bootstrap_data_representer.rb
```diff
@@ -109,6 +109,7 @@ class BootstrapDataRepresenter < ::Roar::Decorator
                extend: Api::V1::OfferingRepresenter,
                readable: true,
                writeable: false,
-               getter: ->(*) { Catalog::Models::Offering.without_deleted }
+               #  if offering's real course and preview courses are not available, put it at the bottom of the array
+               getter: ->(*) { Catalog::Models::Offering.without_deleted.sort_by { |o| o.is_available && o.is_preview_available ? 0 : 1 } }
```
camliu89, line 113: @nathanstitt @Dantemss is it ok to do the sorting here? or somewhere else?

nathanstitt: Yes I think it'd be ok to sort here, but we typically do presentation things like sorting in the front-end.  I'd rather put this as a hook or something like that the offering store.  That way if we change it later it'd be a front-end only release.

nathanstitt: One thing that occurs to me is that we may need to send the `number` of the offering to the FE for sorting as well. with that we could use lodash sortby and return an array with first the available 0/1 and the second the number so it sorts the way admin sets it

camliu89: ok. so this PR will just be sending the `number` to the FE and the FE will deal with the sorting?

app/representers/api/v1/offering_representer.rb
```diff
@@ -86,6 +86,12 @@ class Api::V1::OfferingRepresenter < Roar::Decorator
            readable: true,
            writeable: false
 
+  
+  property :number,
+           type: String,
+           readable: true,
+           writeable: false
+
```
nathanstitt, line 94: "number" needs to be a number by setting type to 'Integer"

nathanstitt (approved):

app/representers/api/v1/offering_representer.rb
camliu89: yes, i forgot to push my changes.

PR #2281, opened 2021-03-15
Remove teacher exercises from pools

Dantemss:
- Remove teacher exercises from pools
- Manually add them to personalized exercises, spaced practice, reading assistant code that determines if pages get personalized exercises and practice tasks, all using the related_teacher_profile_ids

### DIFF
--- a/db/migrate/20210315204108_remove_teacher_exercises_from_page_pools.rb
+++ b/db/migrate/20210315204108_remove_teacher_exercises_from_page_pools.rb
@@ -1,0 +1,24 @@
+class RemoveTeacherExercisesFromPagePools < ActiveRecord::Migration[5.2]
+  def up
+    teacher_exercise_ids = Content::Models::Exercise.created_by_teacher.pluck(:id)
+    page_ids = Content::Models::Exercise.created_by_teacher.distinct.pluck(:content_page_id)
+
+    Content::Models::Page.select(
+      :id, :content_book_id,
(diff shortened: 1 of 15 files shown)


db/migrate/20210315204108_remove_teacher_exercises_from_page_pools.rb
```diff
@@ -0,0 +1,24 @@
...
+      page.all_exercise_ids -= teacher_exercise_ids
+      page.homework_core_exercise_ids -= teacher_exercise_ids
+      page.reading_dynamic_exercise_ids -= teacher_exercise_ids
+      page.reading_context_exercise_ids -= teacher_exercise_ids
+      page.homework_dynamic_exercise_ids -= teacher_exercise_ids
+      page.practice_widget_exercise_ids -= teacher_exercise_ids
+      page.save!
+    end
+  end
+
```
nathanstitt, line 20: this might take a long time, but I think it's ok since if a hw happens to be created while it's running it'll do the right thing and not include the bad exercises

nathanstitt (approved):

db/migrate/20210315204108_remove_teacher_exercises_from_page_pools.rb
Dantemss: There are not many teacher exercises, so hopefully it's quick

PR #2285, opened 2021-03-23
✨ New Homepage

jivey:


.github/workflows/tests.yml
```diff
@@ -52,4 +52,5 @@ jobs:
           gem install bundler
           bundle config path vendor/bundle
           bundle install --jobs 4 --retry 3
+          bin/rails assets:precompile
```
Dantemss, line 55: Is this necessary for the tests to run now?

Dantemss (approved):
Looks good, just 1 question

.github/workflows/tests.yml
jivey: Seems so, I think this was the error that is thrown otherwise:

https://github.com/openstax/tutor-server/runs/2177076355?check_suite_focus=true#step:6:418

Dantemss (approved):

PR #2286, opened 2021-03-25
Updated simple_oauth to a fork with HMAC-SHA256 support

Dantemss:
Fix for 500 error encountered when some LMS tried to sign an LTI request with HMAC-SHA256


jivey (approved):
Failure should work after getting the updated Gemfile.lock

PR #2287, opened 2021-03-25
Updated mimemagic gem

Dantemss:
Necessary to fix builds


Dantemss:
The migration flow will keep failing because it tries to install the gem that no longer exists. There are no migrations in this PR anyway.

nathanstitt (approved):

PR #2289, opened 2021-03-29
Removed unused cron jobs and consolidated remaining cron into fewer processes

Dantemss:
Should reduce memory usage in the tutor-cron instances

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -4,7 +4,7 @@
 #   * https://github.com/javan/whenever/issues/481
 #   * https://github.com/javan/whenever/pull/239
 
-every 1.minute { rake 'cron:minute' }
+every(1.minute) { rake 'cron:minute' }
 
 every 10.minutes do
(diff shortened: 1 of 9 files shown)


config/schedule.rb
```diff
@@ -4,38 +4,14 @@
...
-  runner "OpenStax::RescueFrom.this { GetSalesforceBookNames.call(true) }"
-  runner "OpenStax::RescueFrom.this { PushSalesforceCourseStats.call }"
-end
-
-every 1.day, at: '10:30 AM' do
-  runner "OpenStax::RescueFrom.this { Lms::Models::TrustedLaunchData.cleanup }"
-end
+# These hours are chosen to allow plenty of space between them
+# and to place them in low traffic hours for US timezones
+every(1.day, at: '8 AM') { rake 'cron:day' }  # Midnight-1AM Pacific/2-3AM Central/3-4AM Eastern
```
nathanstitt, line 15: hah, thanks for the comment, I was like "no we don't want 8am"

nathanstitt (approved):

PR #2299, opened 2021-04-08
Fix division by 0 on PushSalesforceCourseStats

Dantemss:
Since we now send courses with all archived sections, we have to be careful not to divide the course-wide enrollment estimate by 0 when calculating the section estimate.

This PR also sets the estimate for archived sections to 0.

### DIFF
--- a/app/routines/push_salesforce_course_stats.rb
+++ b/app/routines/push_salesforce_course_stats.rb
@@ -99,8 +99,6 @@ class PushSalesforceCourseStats
 
   def process_course(course, students_by_period_id, sf_tutor_course_periods_by_period_uuid)
     begin
-      num_teachers = course.teachers.reject(&:deleted?).length
-      num_periods = course.periods.reject(&:archived?).length
       sf_teacher = best_sf_teacher(course)
 
       course_wide_stats = {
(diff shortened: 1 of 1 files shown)


app/routines/push_salesforce_course_stats.rb
```diff
@@ -166,6 +164,18 @@ def process_period(
...
+        # 0 if the section is archived
+        # null if the course doesn't have an estimate
+        # Otherwise, the course-wide estimate divided by the number of non-archived sections
+        sf_tutor_course_period.estimated_enrollment = if period.archived?
+          0
+        elsif course.estimated_student_count.nil?
+          nil
+        else
+          course.estimated_student_count/sf_tutor_course_period.num_periods
+        end
```
jivey, line 177: Do we validate that `estimated_student_count` can't be 0 in the backend?

jivey (approved):

app/routines/push_salesforce_course_stats.rb
Dantemss: No, but 0 in the numerator should be fine, it only blows up if 0 is in the denominator.

nathanstitt: no, but I think that's ok, right?

nathanstitt (approved):

PR #2301, opened 2021-04-09
remove home cache

jivey:
This was causing the browser to reuse the cached response post-login instead of hitting the route and letting us check for a signed in status to redirect to the dashboard.


nathanstitt:
@jivey can you base this branch off of the `v52.0.3` tag?  that way we can release it as a hot fix

jivey:
@nathanstitt Done!

Dantemss (approved):

PR #2302, opened 2021-04-12
add validation for is_preview_available if offering is_available

jivey:
- Validate `is_preview_available` to be true if `is_available` is true.
- Add JS to auto-check `is_preview_available` when checking `is_available` and prevent it from being unchecked when it shouldn't be.

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -87,8 +87,18 @@ $(document).ready(function() {
     window.location.search = $.param(paramsObject);
   });
 
+  $('#offering_is_available').change(function(e) {
+    if ($(this).is(':checked')) {
+      $('#offering_is_preview_available').prop('checked', true);
+      onIsPreviewChange();
+    }
+  });
+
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/admin.js
```diff
@@ -87,8 +87,18 @@ $(document).ready(function() {
...
+    if ($(this).is(':checked')) {
+      $('#offering_is_preview_available').prop('checked', true);
+      onIsPreviewChange();
+    }
+  });
+
   //========== Hides and shows the preview message box when is_preview selected ==========//
   function onIsPreviewChange() {
+    if ($('#offering_is_available').is(':checked')) {
+      $(this).prop('checked', true)
```
Dantemss, line 100: Hmm maybe we want `$('#offering_is_preview_available')` instead of `$(this)` here?
Since `$('#offering_is_available').change()` also calls `onIsPreviewChange()`
Not sure it matters though.

Dantemss (approved):

PR #2305, opened 2021-04-21
Fixed practice worst areas when teacher exercises are present

Dantemss:
Also added a test for practice worst areas

### DIFF
--- a/app/subsystems/tasks/fetch_practice_worst_areas_exercises.rb
+++ b/app/subsystems/tasks/fetch_practice_worst_areas_exercises.rb
@@ -16,25 +16,23 @@ class Tasks::FetchPracticeWorstAreasExercises
       role_book_part.clue['is_real'] ? role_book_part.clue['most_likely'] : 1.5
     end.first(max_num_exercises).map(&:book_part_uuid)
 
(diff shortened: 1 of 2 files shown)


app/subsystems/tasks/fetch_practice_worst_areas_exercises.rb
```diff
@@ -16,25 +16,23 @@ def exec(student:, max_num_exercises: FindOrCreatePracticeTaskRoutine::NUM_EXERC
...
-    page_ids = page_id_by_uuid.values_at(*page_uuids)
-
-    exercise_ids_by_page_id = ecosystem
+    exercise_ids_by_page_uuid = ecosystem
       .pages
-      .where(id: page_ids)
-      .pluck(:id, :practice_widget_exercise_ids)
+      .where(uuid: page_uuids)
+      .pluck(:uuid, :practice_widget_exercise_ids)
       .to_h
```
nathanstitt, line 23: this will pull across ecosystems, right? I think that's what we want, but is a change

nathanstitt (approved):

app/subsystems/tasks/fetch_practice_worst_areas_exercises.rb
Dantemss: Fixed to respect ecosystems

PR #2308, opened 2021-04-25
add additional urls to allowed CSP in dev mode

nathanstitt:
The latest webpack has changed the ports it uses, causing browser warnings.  @jivey this is the cause of all those console messages


Dantemss (approved):
Looks good to me but I didn't test

PR #2330, opened 2021-06-21
[No merge] Temporarily revert fix for scores after dropping questions to fix scores not loading for big courses

Dantemss:
cached? is returning false all the time right now due to the order things update


nathanstitt (approved):
is good for a hot fix, we can correct further in next release

Dantemss:
Probably close this instead of merging and we can fix it properly instead

Dantemss:
Superseded by https://github.com/openstax/tutor-server/pull/2331

PR #2331, opened 2021-06-21
Fixed logic for determining when the task cache should be valid and added specs

Dantemss:
Follow up to today's hotfix

We wanted something like publish_last_requested_at, but that column is not updated in some cases so I added a new column that is always updated.


Dantemss:
We may also want to fix the N+1 that happens when the cache is invalid so the performance is better if you drop a question and immediately go to the scores page

nathanstitt (approved):

PR #2333, opened 2021-06-25
generate payment codes for students

jivey:
- Generate and store codes to be used for bookstore payments, and download a CSV of generated codes
- Downloadable report showing the redemption date and relevant student/course info


app/models/payment_code.rb
```diff
@@ -0,0 +1,31 @@
...
+    post = Array.new(10) { base.sample }.join
+    self.code = "#{prefix.parameterize.upcase}-#{post}"
+  end
+
+  def regenerate_code
+    if errors.types[:code]&.include?(:taken)
+      generate_code
+      save
+    end
+  end
```
nathanstitt, line 30: this method makes me nervous.  A code should never change, because once we hand them out we MUST honor them.  It'd suck if a student somehow paid for a code that we'd overwritten and no longer had a record of.  

I see the code column has a unique index, and since collisions should be very unlikely, how about adding a try/catch block in the generate codes routine and just loop through again if it does happen

nathanstitt (changes requested):

app/models/payment_code.rb
nathanstitt: I also think @Dantemss has an `immutablerecord`  class somewhere, we should use that so that codes cannot be deleted

jivey: @nathanstitt Thoughts if https://github.com/openstax/tutor-server/pull/2333/commits/e4a4a1570e074db74e397b89150ba9caad2c7e22 addresses those concerns? If not I'm totally good with moving this behavior into the routine 👍

app/routines/generate_payment_code_report.rb
```diff
@@ -0,0 +1,28 @@
+class GeneratePaymentCodeReport
+  lev_routine express_output: :export_path
+
+  def exec
+    path = File.join('tmp', 'exports', "payment-code-report.csv")
```
Dantemss, line 5: This has to end up in the exports S3 bucket somehow, you can look at the code on https://github.com/openstax/tutor-server/blob/c42e7bb9d3078763cfd7a530ca6e5e9905b50884/app/subsystems/tasks/models/performance_report_export.rb or roll your own using ActiveStorage because I'm not a fan of Carrierwave either.

The bucket name comes from the secrets.

app/routines/generate_payment_code_report.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    CSV.open(path, 'w') do |file|
+      file << [
+        'Code',
+        'Redeemed At',
+        'Course UUID',
+        'Student Tutor ID',
+        'Student Identifier'
+      ]
+      PaymentCode.in_batches.each_record do |pc|
```
Dantemss, line 15: Might wanna have a parameter in this routine to limit this by dates or something, since we probably won't care about really old codes in a few years.

app/models/payment_code.rb
```diff
@@ -0,0 +1,42 @@
+class PaymentCode < IndestructibleRecord
+  CONFUSED_CHARS = %w(B 8 G 6 I 1 O 0 S 5 Z 2)
+
+  validates_presence_of :code
+  validates_uniqueness_of :code
+
+  before_validation :generate_code, on: :create
+  after_rollback :regenerate_code
```
Dantemss, line 8: Not entirely sure this will help with the mass generation, since if any code fails then GeneratePaymentCodes will fail. Maybe instead add something to GeneratePaymentCodes to handle failures and retry.

app/routines/generate_payment_codes.rb
```diff
@@ -0,0 +1,28 @@
+class GeneratePaymentCodes
+  lev_routine express_output: :codes, transaction: :read_committed
+
+  protected
+
+  def exec(prefix:, amount: 1, export_to_csv: false)
+    codes = []
+
+    amount.times do
+      codes << PaymentCode.create!(prefix: prefix).code
```
Dantemss, line 10: I believe create! will fail the entire routine (and report to Sentry etc) if it fails so to handle failures you'll need either:
- Use `PaymentCode.create(...)`, assign it to a variable and regenerate or just make a new one if it fails OR
- Keep using `PaymentCode.create!(...)` but rescue the error you get (probably `ActiveRecord::RecordInvalid`)

You probably want to test the failure mode, for this I think you test can do something like this:

First split `generate_code` into `generate_code` and `set_code` (which calls `generate_code`) so the return value of `generate_code` is actually used to set the record's code (make sure the before_validation calls `set_code` instead), then add this to your test:
```
allow_any_instance_of(PaymentCode).to receive(:generate_code).and_return('1', '1', '2')
```

So the first 2 will be a duplicate code and the last one will be different.

Dantemss (changes requested):
Looks good other than my comments below

app/routines/generate_payment_codes.rb
jivey: Yeah I meant to swap to just `create` here after I made the model capable of handling the collision by itself. Good notes on testing a failure state, I'll take a look.

app/models/payment_code.rb
nathanstitt: think that's good but how about having it throw an exception if it's persisted?  Right now it'll silently fail to set_code so whatever's calling it wouldn't know that it shouldn't be doing so

nathanstitt (approved):

app/routines/generate_payment_code_report.rb
jivey: Switched it to default to `1.year.ago`, we can add a datepicker to the admin view at some point.

PR #2336, opened 2021-07-05
Fail migration GH action if any file changed (Gemfile.lock, db/schema.db)

Dantemss:
To ensure we remember to commit those files


jivey (approved):
Looks like it's already catching an issue with the sequences? :D

Dantemss:
> Looks like it's already catching an issue with the sequences? :D

It's definitely puzzling me because it keeps happening in the GH action here but not locally and also not in Exercises.

PR #2344, opened 2021-07-09
Don't sync dropped students with LMS

Dantemss:
Moodle integration needs this to work properly.


Dantemss:
Codecov is reporting 0% of diff hit but that can't be right, the test fails without the change in the PR.

nathanstitt (approved):

PR #2348, opened 2021-07-14
Added warnings about modifying the book toc representers

Dantemss:
Since the API that uses these calls `stale?(ecosystem)`, some users had the old response cached, causing issues.


nathanstitt (approved):
great idea!

PR #2356, opened 2021-07-20
update homepage to new subjects

jivey:
Switch the homepage to the "new subjects" design, and fix up a couple of issues with the carousel width. Also updated desktop-large breakpoint, and button sizes on desktop.

image


nathanstitt:
@jivey is there a list of books that will appear on the page somewhere?  They'll probably want to review that to make sure it's correct

Dantemss (approved):

PR #2357, opened 2021-07-23
Fixed the exercise copy script to properly copy exercise tags

Dantemss:
It was failing when trying to copy exercise tags


nathanstitt (approved):

app/subsystems/content/routines/tag_resource.rb
```diff
@@ -10,11 +10,11 @@ def exec(ecosystem:, resource:, tags:, tagging_class: nil, save_tags: true, all_
...
       when Hash
-        tag.merge tag_type: tag[:type]
+        tag.with_indifferent_access.merge tag_type: tag[:type]
       when Content::Models::Tag
-        tag.attributes
+        tag.attributes.with_indifferent_access
       else
-        { value: tag }
+        { value: tag }.with_indifferent_access
       end
```
nathanstitt, line 18: it's not 100% clear to me why `with_indifferent_access` is needed? Is it just to duplicate the hash into one that can use both string & symbol keys?

Dantemss: It was more necessary for the `attributes` one which had string keys but I decided it was probably a good idea to add it to all of them

PR #2360, opened 2021-08-03
Multiple attempts

Dantemss:
- Added PreviousAttempt model
- Added attempt_number column to tasks_tasked_exercises
- When a tasked exercise is completed again after the first time, increment the attempt_number and create a PreviousAttempt record
- Allow the FE to echo back the attempt_number to verify that they are seeing the most up to date information before saving
- Allow the answer to be changed if there are multiple attempts remaining
- Block the free response from being changed after a multiple choice answer is selected
- Return feedback only for the selected answer if an incorrect answer is selected and attempts remain


jivey (approved):

app/subsystems/tasks/models/grading_template.rb
```diff
@@ -84,6 +86,12 @@ def has_open_task_plans?
 
   protected
 
+  def disable_multiple_attempts_unless_immediate_feedback
+    return if auto_grading_feedback_on_answer?
+
+    self.allow_auto_graded_multiple_attempts = false
```
jivey, line 92: Thanks for adding this!

PR #2361, opened 2021-08-10
Fix instructor exercises not showing up after an ecosystem update

Dantemss:
I think this is the first time we mix exercises from different ecosystems in the same assignment so hopefully it's not a problem... this will need a lot of testing


jivey (approved):

Dantemss:
We decided not to do this. This will be properly fixed once we do https://github.com/openstax/tutor/issues/1628. In the meantime, we need to update courses immediately after we import the new ecosystems, otherwise instructor exercises will disappear.

PR #2362, opened 2021-08-10
change migration make sure existing tasked exercises get set to 0 attempts

jivey:
Small change to make sure existing records also have 0 in the column instead of NULL, otherwise existing assignments will complain that 0 is a new attempt. 

I was surprised that the Rails _changed? behavior doesn't use the default value in the comparison even though it uses it when calling the getter:

```
irb(main):040:0> old = Tasks::Models::TaskedExercise.find(5001)
irb(main):045:0> old.attempt_number
=> 0
irb(main):041:0> old.attribute_in_database('attempt_number')
=> nil
irb(main):042:0> old.attempt_number = 0
=> 0
irb(main):043:0> old.attempt_number_change
=> [nil, 0]
```

### DIFF
--- a/db/migrate/20210803214129_add_attempt_number_to_tasks_tasked_exercises.rb
+++ b/db/migrate/20210803214129_add_attempt_number_to_tasks_tasked_exercises.rb
@@ -1,6 +1,5 @@
 class AddAttemptNumberToTasksTaskedExercises < ActiveRecord::Migration[5.2]
   def change
-    add_column :tasks_tasked_exercises, :attempt_number, :integer
-    change_column_default :tasks_tasked_exercises, :attempt_number, 0
+    add_column :tasks_tasked_exercises, :attempt_number, :integer, default: 0
   end
 end


Dantemss (changes requested):
We can't really do this because this would lock the table for more than a whole day, probably (well might be less now that we have Aurora). The table is simply too big for a migration that has to rewrite the entire table.
The attempt number should also be 0 for incomplete ones and 1 for completed ones.
The way to do this would be to run a background job with something like this:

```
loop do
  num_updated = Tasks::Models::TaskedExercise.where(attempt_numbe
r: nil, first_completed_at: nil).limit(100).update_all(attempt_number: 0)
  break if num_updated == 0
end

loop do
  num_updated = Tasks::Models::TaskedExercise.where(attempt_numbe
r: nil).where.not(first_completed_at: nil).limit(100).update_all(attempt_number: 1)
  break if num_updated == 0
end
```

Dantemss:
And since the above would likely take a whole day, we might have to fix the code that uses the changes too.

jivey:
Good to know! Seems I should update the _changed? check to consider `nil -> 0` as unchanged, I'll explore that when I get a chance and update this PR.

PR #2366, opened 2021-08-18
Fix full demo import

Dantemss:
I brought back the APUSH prototype cassette for one of the tests since it's much smaller than the full book.


Dantemss:
... I somehow really doubt that adding 3 tests for code that was previously not covered causes project coverage to go down

jivey (approved):

PR #2368, opened 2021-08-23
Add detailed collaborator solutions

jivey:
Returns `collaborator_solutions` if `solutions` isn't set, which is also what teacher-created exercises use.


app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -34,7 +34,7 @@ class Tasks::Models::TaskedExercise < IndestructibleRecord
   delegate :questions, :question_formats, :question_answers,
            :question_answer_ids, :question_formats_for_students,
            :correct_question_answers, :correct_question_answer_ids,
-           :feedback_map, :solutions, :content_hash_for_students, to: :parser
+           :feedback_map, :solutions, :collaborator_solutions, :content_hash_for_students, to: :parser
```
Dantemss, line 37: As far as I can tell, `solutions` is not a thing anymore. There's only collaborator_solutions and community_solutions now. But we're not using community_solutions yet so no need to add that. Maybe remove `solutions`?

app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -150,7 +150,7 @@ def available_points
   # The following 2 methods assume only 1 Question; this is OK for TaskedExercise,
   # because each TE contains at most 1 part of a multipart exercise.
   def solution
-    solutions[0].try(:first)
+    solutions[0].try(:first) || collaborator_solutions[0].try(:first)
```
Dantemss, line 153: collaborator_solutions should probably be the default really

Dantemss (approved):

app/subsystems/tasks/models/tasked_exercise.rb
jivey: Aaaah if that's the case, I'll swap it to `collaborator_solutions`. I had been curious about it, because none of my local exercises had any `solutions`.

PR #2371, opened 2021-08-26
Not completed while attempts remain

Dantemss:
Only consider multiple attempt exercise steps as completed after the correct answer is selected or all attempts are exhausted. Excludes WRQs.


jivey:
Fix worked, thanks!

jivey (approved):

PR #2375, opened 2021-09-08
Fix exercise exclusions to work despite ecosystem updates

Dantemss:
Reported by an instructor on prod

### DIFF
--- a/app/subsystems/course_content/update_exercise_exclusions.rb
+++ b/app/subsystems/course_content/update_exercise_exclusions.rb
@@ -25,7 +25,7 @@ class CourseContent::UpdateExerciseExclusions
       fatal_error(code: :id_blank, message: 'Missing id for exercise exclusion') \
         unless exercise_params.has_key?('id')
       fatal_error(
-        code: :is_excluded_blank, message: "Missing is_exclusion for exercise with 'id'=#{id}"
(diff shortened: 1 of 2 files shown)


app/subsystems/course_content/update_exercise_exclusions.rb
```diff
@@ -8,36 +8,30 @@ class CourseContent::UpdateExerciseExclusions
...
     exercise_ids = updates_array.map { |update| update['id'] }
-    page_ids = Content::Models::Exercise.where(id: exercise_ids).pluck(:content_page_id)
+    exercises_by_id = Content::Models::Exercise.where(id: exercise_ids).index_by { |ex| ex.id.to_s }
 
     # The course is touched at the end, so we lock the course to make this update atomic
     course.lock!
 
     course_id = course.id
 
-    out = run(:get_exercises, course: course, page_ids: page_ids).outputs
```
TomWoodward, line 18: previously this line would have ensured the given exercises were actually part of this course right? i'm not sure that is necessarily a problem to not have anymore but its something this used to do that it doesn't seem to be doing anymore.

jivey (approved):

app/subsystems/course_content/update_exercise_exclusions.rb
Dantemss: I think for the exclusion code this is fine... it's not a huge deal if they exclude exercises not in the course.

PR #2377, opened 2021-09-14
Fix grading for exercises that existed before multiple attempts

Dantemss:
Possible hotfix
Also fix infinite loop if one of those is changed to multiple attempts (?)


jivey (approved):
I see the spec failure, but the fix itself looks good.

Dantemss:
I think I might also add a background migration to finally get rid of those nulls

Dantemss:
The spec that failed earlier seems to pass now. Other specs now fail because the column is set to NOT NULL. I'll remove those.

PR #2378, opened 2021-09-15
Don't delete LMS callbacks for dropped students

Dantemss:
I didn't even realize this was here... the code that sends scores to the LMS already skips callbacks for dropped students. Don't delete them when dropping because that is not recoverable if they get re-added.


jivey (changes requested):
Looks good, you probably already saw this but looks like the spec is still expecting the Callback count to change.

jivey (approved):

PR #2381, opened 2021-09-24
fix practice questions not showing up when saved in older ecosystem

jivey:
If you saved an exercise to practice when the course's ecosystem was 5, and then the ecosystem was updated to 6, the saved exercise id will not be found in 6, so we need to start querying for each ecosystem an exercise belongs to.

So far no issues have been found with creating a practice exercise with mixed ecosystems, but we should do a lot of testing on QA.


Dantemss (approved):
There's an N+1 query here but N in the number of ecosystems in a single course so I'm not too worried.

PR #2382, opened 2021-09-27
Fix preloading for task_step#completed? within task#completed?

Dantemss:
This was an N+1 query associated with a slow dashboard load in scout.
https://scoutapm.com/apps/13939/endpoints/Q29udHJvbGxlci9hcGkvdjEvY291cnNlcy9kYXNoYm9hcmQ=/trace/1469493446?nplusone=24081.689813&section=timing

### DIFF
--- a/app/subsystems/tasks/models/task.rb
+++ b/app/subsystems/tasks/models/task.rb
@@ -602,7 +602,11 @@ class Tasks::Models::Task < ApplicationRecord
     if use_cache && cache_valid?
       completed_steps_count == steps_count
     else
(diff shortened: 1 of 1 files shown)


jivey (approved):
Looks good, I wonder if we'd like to move the block checks into methods to make it easier to test and understand later on.

Dantemss:
It's a little tricky because we'd have to pass the method names to call to this new method (for example core_task_steps) since we don't really want to call them early to pass the actual array... we can worry about that later

PR #2386, opened 2021-10-01
Add teachers endpoint

TomWoodward:


TomWoodward:
i think my data structure changes are breaking a bunch of tests, and i should add a controller test for the new endpoint.

some feedback on if all the changes look ok before i get that deep would be nice

app/controllers/api/v1/courses_controller.rb
```diff
@@ -127,6 +127,21 @@ def dashboard
...
+  description <<-EOS
+    Returns the teachers for a given course
+    #{json_schema(Api::V1::TeacherRepresenter, include: :readable)}
+  EOS
+  def teachers
+    OSU::AccessPolicy.require_action_allowed!(:read, current_api_user, @course)
+
+    teachers = GetCourseTeachers[@course]
+
+    puts teachers.inspect
```
jivey, line 140: 👀

jivey (changes requested):
This seems good but did we decide not to change the bootstrap data because of query time or maybe it would break something on the front end? It would be nice to not need a new endpoint, but maybe it's necessary?

jivey (approved):
Approach looks good, wasn't sure if you prefer an approve now before tests pass or wait to re-review after.

TomWoodward:
@jivey yea sorry i should have had the tests done first, this is everything i think

PR #2389, opened 2021-10-07
Add dropped courses to student misc user info

jivey:
Also fix the `ol` tag. Renders like this:
image


jivey:
I had the same thought! I was confused by the amount of overlap GetUserCourses has, room to simplify some of these routines as you say.

PR #2401, opened 2021-10-29
Salesforce sync failures admin page

Dantemss:
- Removed what remained of the admin Salesforce controller and routes (it did not have links to it or template files)
- Added an admin page listing Salesforce sync failures
- Tests

Screen Shot 2021-10-29 at 4 45 09 PM

### DIFF
--- a/app/controllers/admin/salesforce_controller.rb
+++ b/app/controllers/admin/salesforce_controller.rb
@@ -1,9 +1,7 @@
 module Admin
   class SalesforceController < BaseController
     def failures
-      terms = CourseProfile::Models::Course.terms.values_at(
-        :spring, :summer, :fall, :winter, :preview
-      )
+      terms = CourseProfile::Models::Course.terms.values_at :spring, :summer, :fall, :winter
 
       @courses = CourseProfile::Models::Course
         .not_ended
@@ -38,7 +36,7 @@ module Admin
                     "openstax_accounts_accounts"."salesforce_contact_id" IS NOT NULL
             )
           WHERE_SQL
-        ).preload(:students, teachers: { role: { profile: :account } })
+        ).order(:id).preload(:students, teachers: { role: { profile: :account } })
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/admin/salesforce_controller.rb
```diff
@@ -1,17 +1,42 @@
...
+                      "course_membership_teachers"."deleted_at" IS NULL
+              ) OR EXISTS (
+                SELECT *
+                FROM "course_membership_students"
+                WHERE "course_membership_students"."course_profile_course_id" =
+                        "course_profile_courses"."id" AND
+                      "course_membership_students"."dropped_at" IS NULL
+              )
+            ) AND NOT EXISTS (
+              SELECT *
```
TomWoodward, line 26: can we do this through the orm with like a has many through or something? why is the inline sql necessary?

app/controllers/admin/salesforce_controller.rb
```diff
@@ -1,17 +1,42 @@
...
+      terms = CourseProfile::Models::Course.terms.values_at :spring, :summer, :fall, :winter
 
-      redirect_to admin_salesforce_path
+      @courses = CourseProfile::Models::Course
+        .not_ended
+        .where(is_test: false, is_preview: false, is_excluded_from_salesforce: false, term: terms)
+        .where(
+          <<~WHERE_SQL
+            (
+              EXISTS (
```
TomWoodward, line 12: is it really necessary to filter out courses without active students/teachers? you already have the `not_ended` filter. are there a lot of courses that are created past the preview stage and then abandoned?

if we need this we probably want to document that its happening, the help text on the page doesn't indicate this filter exists.

TomWoodward (changes requested):

app/controllers/admin/salesforce_controller.rb
Dantemss: I figured that we do not care about courses if the teacher never enrolled anyone and then left it, but yeah I guess we can still display them to document it.

jivey (approved):

PR #2402, opened 2021-11-01
Cleanup roles in course APIs

Dantemss:
Goes with https://github.com/openstax/tutor-js/pull/3823

- Added a new SharedRoleRepresenter and TeacherStudentRepresenter
- Replaced arrays of teachers and students in the course with single objects
- Updated tests

I have not removed any of the old fields yet because we'll want to update the FE and release that before we take them out.

### DIFF
--- a/app/controllers/api/v1/periods_controller.rb
+++ b/app/controllers/api/v1/periods_controller.rb
@@ -98,8 +98,8 @@ class Api::V1::PeriodsController < Api::V1::ApiController
     result = CreateOrResetTeacherStudent.call(user: current_human_user, period: @period)
 
     render_api_errors(result.errors) || respond_with(
-      result.outputs.role,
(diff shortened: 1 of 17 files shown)


Dantemss:
There's one small incompatible change here which is that the create teacher student API returns a teacher student record instead of a role, meaning after the update teachers won't be able to create new teacher students until they reload the FE. This only happens once when they click "view as student" for a period they haven't ever viewed before, so I think it's acceptable, but what do you think?

TomWoodward (approved):

TomWoodward:
> until they reload the FE. This only happens once when they click "view as student" for a period they haven't ever viewed before, so I think it's acceptable, but what do you think?

the limitations you describe here seem to make the issue pretty minor, i think we can go ahead with it.

PR #2404, opened 2021-11-03
use uuids for practice questions

jivey:
Goes with https://github.com/openstax/tutor-js/pull/3825

- Add`exercise_uuid` in the representer, which allows the FE to match PracticeQuestions by uuid to avoid issues with ecosystem updates. (Maybe we should remove `exercise_id` in the future, but some FE things broke and need to be updated.)
- Stop accepting exercise ids from the FE, we can get them off the tasked, which makes the endpoint and routine simpler.
- Dedup if multiple matching versions were picked when creating the practice task (thanks @Dantemss for the query)

I left `content_exercise_id` in there for now to minimize the scope of this change, but I wonder if we should change it to `content_exercise_uuid` over a couple of deploys and background migrations (or remove it and get it from tasked?)

### DIFF
--- a/app/controllers/api/v1/ecosystems_controller.rb
+++ b/app/controllers/api/v1/ecosystems_controller.rb
@@ -82,8 +82,7 @@ class Api::V1::EcosystemsController < Api::V1::ApiController
   def practice_exercises
     exercises = GetPracticeQuestionExercises[
       role: @role,
-      course: @course,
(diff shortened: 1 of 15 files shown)


Dantemss (approved):
We can change the way we save practice questions and the API at a later time to use the exercise uuids as well.

PR #2405, opened 2021-11-11
Updated spec cassette to contain the new URL for accounts dev

Dantemss:
Shouldn't really affect anything, but just in case...


jivey (approved):

Dantemss:
I'm closing this one because we'll actually use dev.openstax.org/accounts

PR #2406, opened 2021-11-15
Shuffle choices

jivey:
The main pieces to this are:
- Add `shuffle_answer_choices` field on the grading template
- Use `answer_id_order` param on the first attempt only to sort the tasked's `answer_ids` field so we can use that on the front end to preserve the submitted order
- Allow teacher-created exercises to set `is_answer_order_important` field if there are more than 2 answers.

### DIFF
--- a/app/subsystems/tasks/models/grading_template.rb
+++ b/app/subsystems/tasks/models/grading_template.rb
@@ -53,7 +53,8 @@ class Tasks::Models::GradingTemplate < ApplicationRecord
       default_due_time: '21:00',
       default_due_date_offset_days: 7,
       default_close_date_offset_days: 7,
-      allow_auto_graded_multiple_attempts: false
+      allow_auto_graded_multiple_attempts: false,
+      shuffle_answer_choices: true
     },
     {
       task_plan_type: :reading,
@@ -68,7 +69,8 @@ class Tasks::Models::GradingTemplate < ApplicationRecord
       default_due_time: '07:00',
       default_due_date_offset_days: 7,
       default_close_date_offset_days: 7,
-      allow_auto_graded_multiple_attempts: false
+      allow_auto_graded_multiple_attempts: false,
+      shuffle_answer_choices: true
     }
   ]
 
--- a/app/subsystems/tasks/models/task.rb
+++ b/app/subsystems/tasks/models/task.rb
@@ -224,6 +224,10 @@ class Tasks::Models::Task < ApplicationRecord
     !!grading_template&.allow_auto_graded_multiple_attempts
   end
 
+  def shuffle_answer_choices
+    !!grading_template&.shuffle_answer_choices
+  end
+
   def late_work_penalty_applied
     grading_template&.late_work_penalty_applied || 'not_accepted'
   end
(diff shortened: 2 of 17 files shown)


app/subsystems/tasks/models/task.rb
```diff
@@ -224,6 +224,10 @@ def allow_auto_graded_multiple_attempts
     !!grading_template&.allow_auto_graded_multiple_attempts
   end
 
+  def shuffle_answer_choices
+    !!grading_template&.shuffle_answer_choices
```
Dantemss, line 228: I think this will default to false for tasks without a grading_template (practice). Is that ok?

app/subsystems/tasks/models/tasked_exercise.rb
```diff
@@ -38,6 +38,15 @@ class Tasks::Models::TaskedExercise < IndestructibleRecord
 
   delegate :uuid, to: :exercise, prefix: :exercise
 
+  def answer_id_order
+    answer_ids
+  end
+
+  def answer_id_order=(order)
+    return unless attempt_number == 0
```
Dantemss, line 46: attempt_number has some oddness where it is updated as soon as you set other fields. Maybe you want attempt_number_was?

db/migrate/20211103150618_add_shuffle_answer_choices_fields.rb
```diff
@@ -0,0 +1,5 @@
+class AddShuffleAnswerChoicesFields < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tasks_grading_templates, :shuffle_answer_choices, :boolean, default: false, null: false
```
Dantemss, line 3: Current courses get false

app/subsystems/tasks/models/grading_template.rb
```diff
@@ -68,7 +69,8 @@ class Tasks::Models::GradingTemplate < ApplicationRecord
       default_due_time: '07:00',
       default_due_date_offset_days: 7,
       default_close_date_offset_days: 7,
-      allow_auto_graded_multiple_attempts: false
+      allow_auto_graded_multiple_attempts: false,
+      shuffle_answer_choices: true
```
Dantemss, line 73: Future courses get true

Dantemss (approved):
Looks good. Check my comment on answer_id_order= but if it works as expected then it's fine.

app/subsystems/tasks/models/tasked_exercise.rb
jivey: I was thinking it was okay because the order assignment happens in the consume before the attempt number is bumped, but I updated it to use that just in case: https://github.com/openstax/tutor-server/pull/2406/commits/af21ab4babdf32ac9e908499bd5cec58b2e58d64

app/subsystems/tasks/models/task.rb
jivey: Added a commit to default this to true, I'm glad you noticed that! Seems like it'll be a nice practice experience https://github.com/openstax/tutor-server/pull/2406/commits/2a9bdf52216fbb25de0e18150751bfee2b5dfeee

Dantemss (approved):

Dantemss (approved):

PR #2411, opened 2021-12-06
Updated the webpacker gem to remove the node-gyp dependency

Dantemss:
Should fix failing PRs due to node-gyp failing to build node-sass or something


Dantemss:
Seems like there's another issue with sass-loader...

Dantemss:
The second issue turned out to be a typo in Tutor's stylesheets that the old version of sass did not care about.

Dantemss:
There's no code to be covered here so codecov can be ignored

jivey (approved):

PR #2424, opened 2022-03-03
fix homepage css

jivey:
A miss-targeted css rule was causing some rendering issues: https://github.com/openstax/tutor-js/issues/3855

Also remove a couple of rules that were interfering with the placement of the left arrow button.


Dantemss (approved):
Seems like this is only used in the homepage so if the fix is good there it seems good to me

PR #2445, opened 2022-10-24
upgrade nokogiri to 1.13.9

jivey:


jivey:
I don't think we need https://github.com/openstax/tutor-server/pull/2444/files, the lockfile versions are still satisfied, I'm not sure why dependabot would bump all the gems that require it.

PR #2450, opened 2022-10-25
upgrade rails

jivey:


jivey:
Upgrading Rails manually just like the last one instead of several PRs that bump things we don't want bumped.

Dantemss (approved):

PR #2472, opened 2023-08-08
Updated gems and fixed tests

Dantemss:
Just do we can deploy something that is not failing tests


jivey:
Looks like a selenium upgrade will let us remove webdriver: https://github.com/titusfortner/webdrivers/issues/247#issuecomment-1641879165

Dantemss:
Removing webdrivers requires updating selenium-webdrivers which requires ruby 3. I tried to update it but ran into other issues so I think for now I'll try to just fix the tests in ruby 2.7. The tests already work locally but something is wrong in GH actions.

Dantemss:
I'm going to ignore the migrations failure since it's a gem issue (it cannot install the gems from the base commit due to the Chrome changes).

